PR #5, opened 2011-07-06
allowed line terminators to be included in strings when escaped

rapha:


tuchida:
What ECMA262 specification does this comply with?
```js
var lf = '\u000A', cr = '\u000D', ls = '\u2028', ps = '\u2029';
console.log(
  eval('"\\'+lf+'"') === lf,
  eval('"\\'+cr+'"') === lf,
  eval('"\\'+ls+'"') === lf,
  eval('"\\'+ps+'"') === lf
);
```
SpiderMonkey
```
false false false false
```
V8
```
false false false false
```

p-bakker:
Given that this is a very old PR, as it doesn't seem standard compliant behavior and since we'll be up to date with the latest spec in this area soon (see https://github.com/mozilla/rhino/pull/978), I'm closing this PR

PR #7, opened 2011-08-12
Bugfix for parsing xml that starts with whitespace

rasmuserik:
Rhino has problems when parsing xml that starts with whitespace (and `XML.ignoreWhitespace === true`), ie:

```
js> (new XML(" <a/>")).toSource()
&lt;a/&gt;
```

This patch fixes this issues such that:

```
js> (new XML(" <a/>")).toSource()
<a/>
```

Unit test also added.

This is implemented by removing the special case for strings not starting with `<`. According to ECMA-357 10.3, numbers and booleans are passed to the same xml-conversion as strings, so as far as I can see the special case is not needed.

### DIFF
--- a/src/org/mozilla/javascript/optimizer/Codegen.java
+++ b/src/org/mozilla/javascript/optimizer/Codegen.java
@@ -1148,7 +1148,7 @@ public class Codegen implements Evaluator
 
     String getBodyMethodName(ScriptNode n)
     {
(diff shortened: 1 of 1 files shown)


p-bakker:
Converted to draft as this has been lying around since forever and needs someone to go through to determine whether we want/needs this PR not

PR #39, opened 2012-03-21
ES5-Test262 progress

anba:
This is a bit of my work towards ECMAScript 5 compliance. For now I've mainly focused on test262 to find out what is currently missing for ES5 conformance in Rhino, so this is by far not a complete. (Out of the full test262 test suite, only seven test cases still fail. Yeah!)

As the change set itself actually has grown quite big, I don't think it's possible (nor reasonable) to just apply it as a whole (at least not into the master branch). And I deliberately did not squash all commits into a single one in order to reveal somewhat the thought process behind the changes. 

I'm open for suggestions how to get this into Rhino in a sane way.

Cheers,
André

### DIFF
--- a/src/org/mozilla/javascript/IRFactory.java
+++ b/src/org/mozilla/javascript/IRFactory.java
@@ -1921,11 +1921,6 @@ public final class IRFactory extends Parser
             } else if (name.equals("With")) {
                 type = Node.SPECIALCALL_WITH;
             }
-        } else if (child.getType() == Token.GETPROP) {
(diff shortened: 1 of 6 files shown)


p-bakker:
Is there any plan to merge the remainder of this PR, the part with the breaking changes to the Scriptable and Callable interfaces?

At least I assume they are still needed to progress in ES Latest compliance?

p-bakker:
Converted to draft as this one is very outdated. Just not choosing it because there's some stuff in here that is still somewhat useful

PR #57, opened 2012-07-04
Maven dependency and build management

anba:
I've created some Maven configuration files to build Rhino through Maven (in addition to Ant). The changes are meant to give users a drop-in replacement for Ant without requiring to apply any other Maven specific changes, such as, for example, switching to the Maven standard directory layout.


gbrail:
The current version of Rhino in master now supports building using Gradle. That has worked better than Maven because we are better able to support our legacy stuff.

PR #70, opened 2012-07-28
Patch for Bug 757410

anba:
Generators save and later restore the current stack when processing the 'yield' operation. Our current implementation for restoring the stack unfortunately confuses the Java classfile verifier, so at class load time a VerifierError is thrown. This happens because the verifier can no longer ensure that the proper types are placed on the stack, since the stack-state is saved in a simple Object[]. Re-ordering a few operations is necessary so the verifier will accept the generated class again. But this is only done for generators because it creates slightly less efficient code compared to the standard case.


gbrail:
Merged (and rebased) to master for 1.8.0.

PR #76, opened 2012-08-01
Fall back to interpreter when codegen hits 64k limit and fails.

headius:
Currently, if optimize level is set to >= 0, it is possible for code to fail to run at all when it exceeds the JVM's 64k bytecode size limit. This patch (a work in progress) modifies Codegen to fall back to Interpreter when any errors occur during code generation.

I recognize that this catch-all is not ideal, but I'm posting this now to start a discussion. It is not clear to me why having code fail to execute is better than falling back to interpretation, but the hard break at the 64k limit means most applications of Rhino set optimize level to -1. I believe this has harmed the world's impression of Rhino and its performance.

This could also be an optional setting, enabled by default but possible to disable for those who really want code >64k to blow up. I'm not sure who those people are.

### DIFF
--- a/src/org/mozilla/javascript/optimizer/Codegen.java
+++ b/src/org/mozilla/javascript/optimizer/Codegen.java
@@ -72,44 +72,59 @@ public class Codegen implements Evaluator
 
         String mainClassName = "org.mozilla.javascript.gen." + baseName + "_" + serial;
 
-        byte[] mainClassBytes = compileToClassFile(compilerEnv, mainClassName,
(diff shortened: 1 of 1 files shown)


rbri:
I think we can close this.

gbrail:
I think that we have covered this int he product.

headius:
@rbri @gbrail I'm not sure I can find where this has been addressed. my patch was never merged, obviously, and as far as I can tell neither was @hns's version. I would like to know that this has been addressed, but I'm failing at github 😄

PR #78, opened 2012-08-05
Patch for Bug 780458:

anba:
Use the fast ToInt32 conversion code from vm/NumericConversions.h to avoid calling Math.IEEEremainder(). The new method is called "toInt32_Slow" (although it's really fast compared to Math.IEEEremainder()) so people don't start calling it directly. Direct calls should be avoided because the method size is too big for inlining by the JVM.


gbrail:
I elected to use the V8 version of this instead.

PR #80, opened 2012-08-14
Patch for Bug 782363 ("Increment/Decrement alters const variables")

anba:
- Update Interpreter#doVarIncDec() and BodyCodegen#visitIncDec() in a similar way to handle constants
- Update optimizer/Block and optimizer/Optimizer to process the Token.SETCONSTVAR opcode, also make sure not to re-assign the type of constants
- Replace the various getNewWordLocal() implementations with a single shared implementation to reduce duplicated code
- While I was at it, I also fixed two bugs in that code:
  - getNewWordLocal() wasted variable space because it started to search for free entries at `firstFreeLocal + 2` even if `isConst=false`
  - getConsecutiveSlots() could throw ArrayIndexOutOfBounds exceptions when `count=2 & isConst=true`


gbrail:
Merged and rebased to master for 1.8.0.

PR #81, opened 2012-08-17
Patch for Bug 783507 ("Implement ES.next quasi-literals (Rhino)")

anba:
This is a first stab at implementing quasi-literals (resp. the new official name "string templates"), it's still kind of rough and could need some clean-up, but should be sufficient as a starting point. I've used the current ES6 draft (July 8, 2012 Draft) for this patch.

There are four individual patches:
[Patch 1] fixes two bugs in the current code base which prevented the test cases to succeed
[Patch 2] contains the majority of the changes (frontend as well as backend)
[Patch 3] adds the new "String.raw" function
[Patch 4] adds some test cases for quasi-literals and String.raw

### DIFF
--- a/src/org/mozilla/javascript/CodeGenerator.java
+++ b/src/org/mozilla/javascript/CodeGenerator.java
@@ -6,7 +6,10 @@
 
 package org.mozilla.javascript;
 
+import java.util.List;
+
 import org.mozilla.javascript.ast.AstRoot;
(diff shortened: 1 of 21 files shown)


tonygermano:
I think this can finally be closed since https://github.com/mozilla/rhino/pull/879 and https://github.com/mozilla/rhino/pull/904 have been merged.

p-bakker:
Merged via #904

PR #87, opened 2012-08-28
Patch for Bug 780458 ("Math.IEEEremainder makes ToInt32 slow for non-integer values") (V8)

anba:
Alternative patch for Bug 780458, this time using V8's implementation for ToInt32.

I've slightly changed `doubleToInt32()` and removed the fast-paths for positive/negative infinitiy and NaN in `ScriptRuntime.to[Uint16, Uint32, Int32]()`. Both changes resulted in a noticeable performance improvement in a microbenchmark, I guess the JVM was able to inline the methods when those changes were present.


gbrail:
Merged and pushed to master. Added a NOTICE.txt to cover the V8 license.

PR #88, opened 2012-09-10
Patch for Bug 789277 ("JSC: "missing ; after statement" message prints out for the line after the problem one")

anba:
Change strict-mode warning for missing semi-colons to point to the actual statement which misses the semi-colon instead of the next token.

TokenStream:
- add helper methods charAt(int) and substring(int, int) to reduce string/reader input clutter
- change getLine() to use the new helper methods
- handle "\r\n" newline sequences in getLine()
- add getLine(int, int[]) to retrieve a line based on an absolute input source position
- getLine(int, int[]) may return `null` to signal the requested line is no longer available, which can happen in the reader input case

Parser:
- change warnMissingSemi(int, int) to use TokenStream#getLine(int, int[]), so the line information of the actual statement which misses the semi-colon can be retrieved

### DIFF
--- a/src/org/mozilla/javascript/Parser.java
+++ b/src/org/mozilla/javascript/Parser.java
@@ -189,6 +189,39 @@ public class Parser
         }
     }
 
+    private void addStrictWarning(String messageId, String messageArg,
+                                  int position, int length,
+                                  int line, String lineSource, int lineOffset) {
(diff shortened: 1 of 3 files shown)


gbrail:
Merged and rebased in master.

PR #91, opened 2012-09-18
applied fixes for bug 542479

pa314159:


tuchida:
The bug is AstNode#toSource of number literal.
https://bugzilla.mozilla.org/show_bug.cgi?id=542479

ref. https://github.com/mozilla/rhino/commit/a01ab971ad65d1c7ba9e748c52734bc1b27226c5
It had been fix already.

PR #97, opened 2012-11-11
(Potential) fix for parsing of keywords (e.g. "null") at EOF to return correct length.

mikelehen:
[This should be code-reviewed by somebody familiar with the rhino token parsing as I don't know if this is the correct way to fix this.]

Prior to this change, TokenStream would return an incorrect length for keywords ("null", "true", etc.) encountered immediately prior to the end-of-file.  I added a test to ParserTest to reproduce the issue and made a potential fix.

My fix is to not call ungetChar() if getChar() returned EOF_CHAR, since getChar() does not increment "cursor" when encountering EOF, but ungetChar() always decrements "cursor" thus leading to the incorrect length.

### DIFF
--- a/src/org/mozilla/javascript/TokenStream.java
+++ b/src/org/mozilla/javascript/TokenStream.java
@@ -369,7 +369,8 @@ class TokenStream
                         }
                     }
                 }
-                ungetChar(c);
+                if (c != EOF_CHAR)
+                    ungetChar(c);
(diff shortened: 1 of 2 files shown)


mikelehen:
No idea if this is still relevant and it's been over 5 years, so I'm going to go ahead and just close this. :-)

PR #100, opened 2012-12-23
Implement ES.next for-of

tuchida:
This patch is implementing for-of syntax which iterates Array and String.
https://developer.mozilla.org/en-US/docs/JavaScript/Reference/Statements/for...of
http://wiki.ecmascript.org/doku.php?id=harmony:iterators

I've used the current ES6 draft (December 21, 2012 Draft) for this patch.
Conflict with for each-in, referred to behavior of Firefox19(Aurora 19.0a2 (2012-12-19)).

### DIFF
--- a/src/org/mozilla/javascript/CodeGenerator.java
+++ b/src/org/mozilla/javascript/CodeGenerator.java
@@ -475,6 +475,7 @@ class CodeGenerator extends Icode {
           case Token.ENUM_INIT_KEYS:
           case Token.ENUM_INIT_VALUES:
           case Token.ENUM_INIT_ARRAY:
+          case Token.ENUM_INIT_VALUES_IN_ORDER:
             visitExpression(child, 0);
             addIndexOp(type, getLocalBlockRef(node));
             stackChange(-1);
--- a/src/org/mozilla/javascript/ES6Iterator.java
+++ b/src/org/mozilla/javascript/ES6Iterator.java
@@ -6,21 +6,9 @@
 
 package org.mozilla.javascript;
 
-import java.util.Iterator;
+public abstract class ES6Iterator extends IdScriptableObject {
 
-/**
- * This class implements ArrayIterator and StringIterator objects. See
- * http://wiki.ecmascript.org/doku.php?id=harmony:iterators
- */
-public final class NativeElementIterator extends IdScriptableObject {
-    private static final long serialVersionUID = 1L;
-    private static final Object ELEMENT_ITERATOR_TAG = "ElementIterator";
-
-    static void init(ScriptableObject scope, boolean sealed) {
-        // NativeElementIterator
-        // Can't use "NativeElementIterator().exportAsJSClass" since we don't want
-        // to define "NativeElementIterator" as a constructor in the top-level scope.
-        NativeElementIterator prototype = new NativeElementIterator();
+    static void init(ScriptableObject scope, boolean sealed, IdScriptableObject prototype, String tag) {
         if (scope != null) {
             prototype.setParentScope(scope);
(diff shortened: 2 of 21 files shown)


gbrail:
@tuchida -- same question here -- the code has changed quite a bit. 

I had to make a few merges, which I made in the branch "tuchida-for-of_loop". The test passes except for the test on "for (var n of null) {}".

Can you take a look at the branch and see if the merge looks OK to you? I'd also like to see if we can find an external test for this, although I can't find one easily in the V8 test suite.

I think I just wrote this same comment in your own repo -- sorry.

tuchida:
> I'd also like to see if we can find an external test for this, although I can't find one easily in the V8 test suite.

It seems that test262 have begun support of ES6.
https://github.com/tc39/test262/pull/308

As far as I know, my branch does not meet the specification of the following.

 * The specification is `Array.prototype[Symbol.iterator]`, but `Array.prototype.iterator`.
 * surrogate pair
```js
var c = 0;
for (var c of '𠮷野家') {
  c++;
}
c;   // Expect 3, but 4.
```

eshepelyuk:
Hello @tuchida 
1. You could use my simple `Symbol` implementation from PR #203 to implement more spec compatible `for-of` loop.
2. Second thing I would like to ask if it possible to improve this PR by implementing latest version of `Iterator Protocol` as described in [ES 2015 standard](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols#iterator)
3. I would also like to use this work for finalizing #159 where `Iterator and Iterable` protocol is widely used. 

So, after those changes we could use more of `test262` suites to prove `for-of` implementation.
What do you think ?

tuchida:
Hello @eshepelyuk

I use your `Symbol`. Does it this will be your help?

eshepelyuk:
Hello

You can use current simple `Symbol` implementation to just make more `test262` tests pass.
But since you've implemented `NativeElementIterator` - you may notice that it works not according to `Iterator protocol`. 
So I asked if you can make this iterator to implement ES6 iterator protocol, so `NativeElementIterator` can be later reused for a variety of ES6 code.

eshepelyuk:
Btw you maybe interested in taking a look at PR #205 that implements sample framework for selectively running test262 as a part of `Rhino` unit tests.

tuchida:
Thanks! `Test262SuiteTest` is cool.

I used the test case and read specification. I want to these iterable objects. I will send PR in future as possible.

## Arguments

> 9.4.4.6 CreateUnmappedArgumentsObject(argumentsList)
>  7. Perform DefinePropertyOrThrow(obj, @@iterator, PropertyDescriptor {[[Value]]:%ArrayProto_values%, [[Writable]]: true, [[Enumerable]]: false, [[Configurable]]: true}).

> 9.4.4.7 CreateMappedArgumentsObject ( func, formals, argumentsList, env )
>  22. Perform DefinePropertyOrThrow(obj, @@iterator, PropertyDescriptor {[[Value]]:%ArrayProto_values%, [[Writable]]: true, [[Enumerable]]: false, [[Configurable]]: true}).

## %TypedArray%

> 22.2.3.30 %TypedArray%.prototype [ @@iterator ] ( )

## java.util.Iterator, java.lang.Iterable

eshepelyuk:
Hello @gbrail 
Is there an chance this PR will be merged soon to master ? 
I'd like to use it for completing implementation of #159. The `iterator protocol` implemented by @tuchida is the last remaining thing for implementing Maps and Sets support for `Rhino`.

gbrail:
Rebased and merged into master. Thanks!

PR #102, opened 2013-01-17
test and fix for crash parsing regexp /0{0/

alagopus:
This admittedly strange regexp is a minimized version of a test case from the coffee-script project. Judging from a comment in NativeRegExp.java, incomplete quantifiers are supposed to be ignored. Instead, an java.lang.ArrayIndexOutOfBoundsException is thrown, causing the REPL to terminate:

``` sh
$ java -jar build/rhino1_7R5pre/js.jar
Rhino 1.7 release 5 PRERELEASE 2013 01 17
js> /0{0/
Exception in thread "main" java.lang.ArrayIndexOutOfBoundsException: 3
    at org.mozilla.javascript.regexp.NativeRegExp.parseTerm(NativeRegExp.java:1124)
[ . . . more stack trace elided]
```

The change set in this pull request includes a unit test and fix for the open coded parsing of {} quantifiers.


gbrail:
This very old PR is actually quite reasonable, so I created https://github.com/mozilla/rhino/pull/342 to cherry-pick it.

PR #106, opened 2013-01-27
rhino exits the JVM even when run as a subshell of another java shell - ...

rotty3000:
...bug-835147


gbrail:
This was actually merged long ago in another PR.

PR #108, opened 2013-02-21
Attach jsdoc nodes to function params

nicks:
Hopefully, this should be a straightforward and non-controversial parser change. Thanks!


concavelenz:
Closure Compiler no longer uses rhino as a parser.

---

Review status: 0 of 2 files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/mozilla/rhino/108)

gbrail:
This looks like a very old but reasonable fix.

nicks:
hahahaha, i'm sure you're setting some sort of record with these merges :)

Issue #94, opened 2012-11-03
Git project


asashour:
Hi all,

I noticed few points in the current Git project:

- There are some confusing tags, like "head" and "done", I believe they should be removed or renamed to something more meaningful
- "build" and "lib" folders should be added to .gitignore
- .csvignore still exists in three places: root, in "toolsrc/org/mozilla/javascript/tools/debugger" and its "downloaded" subfolder.

Thanks,
Ahmed

d10:
This issue can be closed.

PR #113, opened 2013-03-21
removing old .cvsignore files

sghill:
Agreeing with issue #94 that the cvs files probably aren't necessary anymore


gbrail:
Merged to master.

PR #114, opened 2013-03-21
adding gitignore for build directories and artifacts

sghill:
Running `ant all` no longer generates files that git thinks are new. Issue #94 talks about this a bit.


gbrail:
Manually added some of these paths to the existing .gitignore.

PR #115, opened 2013-04-07
Created readme for Github project

shirishp:
Created README taking reference from https://developer.mozilla.org/en-US/docs/Rhino


gbrail:
Copied some of this into the README that we had a while back.

shirishp:
Thank you. :)

With regards,
Shirish (http://twitter.com/_Garbage_)

PR #120, opened 2013-06-01
Implement ES6 Arrow Function

tuchida:
This patch is implementing Arrow Function syntax.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/arrow_functions
http://wiki.ecmascript.org/doku.php?id=harmony:arrow_function_syntax

I've used the current ES6 draft (May 14, 2013 Draft) for this patch.

 * No support Rest parameters.
 * I'm not sure about the specification of sctrict mode.
 * When a Arrow Function have called as constructor, error message is unkind.

```js
var fn = () => 10;
new fn();
/*
js: "<stdin>", line 3: uncaught JavaScript runtime exception: TypeError: "
() => 10
" is not a constructor.
	at <stdin>:3
*/
```

### DIFF
--- a/src/org/mozilla/javascript/ArrowFunction.java
+++ b/src/org/mozilla/javascript/ArrowFunction.java
@@ -1,0 +1,75 @@
+/* -*- Mode: java; tab-width: 8; indent-tabs-mode: nil; c-basic-offset: 4 -*-
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.javascript;
+
+/**
+ * The class for  Arrow Function Definitions
+ * EcmaScript 6 Rev 14, March 8, 2013 Draft spec , 13.2
+ */
+public class ArrowFunction extends BaseFunction {
+    
+    static final long serialVersionUID = -7377989503697220633L;
+    
+    private final Callable targetFunction;
+    private final Scriptable boundThis;
+
+    public ArrowFunction(Context cx, Scriptable scope, Callable targetFunction, Scriptable boundThis)
+    {
+        this.targetFunction = targetFunction;
+        this.boundThis = boundThis;
+
+        ScriptRuntime.setFunctionProtoAndParent(this, scope);
+
+        Function thrower = ScriptRuntime.typeErrorThrower();
+        NativeObject throwing = new NativeObject();
+        throwing.put("get", throwing, thrower);
+        throwing.put("set", throwing, thrower);
+        throwing.put("enumerable", throwing, false);
+        throwing.put("configurable", throwing, false);
+        throwing.preventExtensions();
+
+        this.defineOwnProperty(cx, "caller", throwing, false);
+        this.defineOwnProperty(cx, "arguments", throwing, false);
+    }
+
+    @Override
+    public Object call(Context cx, Scriptable scope, Scriptable thisObj, Object[] args)
+    {
+        Scriptable callThis = boundThis != null ? boundThis : ScriptRuntime.getTopCallScope(cx);
+        return targetFunction.call(cx, scope, callThis, args);
+    }
+
+    @Override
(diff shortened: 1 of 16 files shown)


gbrail:
I would love to get this in but it's been a while and there is a lot of stuff changed here.

@tuchida -- would you have time to rebase this PR with the latest master so that we can try and merge it cleanly? (And since you know this bit of the code better, you would be better qualified to produce a proper merge.)

tuchida:
I rebased the branch.

I have created ArrowFunction.java based on BoundFunction.java. A arrow function creates a two function (unbind this and bound) at run time. I think that ArrowFunction.java is needlessness, but I could not remove it at that time.

gbrail:
Finally merged -- thanks!

I'm going to open an issue to add more tests, however. For instance, perhaps we could add arrow function tests from the V8 test suite.

tntim96:
This merge seems to parse `=>` OK, but the `toSource` is not yet working.

```JavaScript
var a3 = a.map(s => s.length);
```
will `toSource` to
```JavaScript
var a3 = a.map(functions => s.length);
```

> I'm going to open an issue to add more tests

It might me good to add a common place for AST `toSource` tests. I'm not sure if there's currently a specific place for them. The ones I've found are in classes like `Bug491621Test` which is far from ideal. Would be good to eventually move these to a common area, that hopefully reflects the relevant specification, with the bug number noted as a comment if that's useful - at least the new tests can go to this area. I'm happy to help write the tests (can use Bug491621Test as a prototype) but don't want to create something similar without a bit more thought.

gbrail:
Yes -- more tests for "toSource" would be great! We've been using either
test262 or tests from the V8 test suite to test new features, but the
parser itself isn't something that we have good unit tests for -- if you
could think of a framework for that and contribute some tests then we'll be
better prepared next time.

On Fri, Jul 24, 2015 at 3:49 PM, tntim96 <notifications@github.com> wrote:

> This merge seems to parse => OK, but the toSource is not yet working.
>
> var a3 = a.map(s => s.length);
>
> will toSource to
>
> var a3 = a.map(functions => s.length);
>
>  I'm going to open an issue to add more tests
>
> It might me good to add a common place for AST toSource tests. I'm not
> sure if there's currently a specific place for them. The ones I've found
> are in classes like Bug491621Test which is far from ideal. Would be good
> to eventually move these to a common area, that hopefully reflects the
> relevant specification, with the bug number noted as a comment if that's
> useful - at least the new tests can go to this area. I'm happy to help
> write the tests (can use Bug491621Test as a prototype) but don't want to
> create something similar without a bit more thought.
>
> —
> <https://github.com/mozilla/rhino/pull/120#issuecomment-124756001>.
>



-- 
*Greg Brail* | *apigee <https://apigee.com/>* | twitter @gbrail
<http://twitter.com/gbrail> @apigee <https://twitter.com/apigee>
<http://iloveapis.com/>

tntim96:
I wonder could this be done at two levels:

1. Modify or subclass `Test262SuiteTest`, parse each harnessFile and then run `toSource` before executing the test - it should all still pass. This should catch everything
2. Another area to test formatting. I guess a few variations would be needed for different language levels, etc... Any idea how this should be organized?

tntim96:
I've created https://github.com/mozilla/rhino/pull/221 - please double check.

> Make a modified or subclassed version of Test262SuiteTest to parse each harnessFile and then run toSource before executing the test - it should all still pass. This should catch everything

That seems to work fine already (so mustn't cover this particular case). Do you want to add it to the test suite?

tuchida:
@tntim96 Sorry, and thanks.

AstNode#toSource has almost the same thing as Decompiler.java.
https://github.com/tuchida/rhino/commit/318da76bceaeadc0dd0319892fe7200640fbfae1#diff-2 https://github.com/tuchida/rhino/commit/318da76bceaeadc0dd0319892fe7200640fbfae1#diff-3

I think Rhino is better to remove decompiler like SpiderMonkey.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/toString#Gecko-specific_notes
> Function.prototype.toString() has been implemented by saving the function's source. The decompiler was removed, so that the indentation parameter is not needed any more. See bug 761723 for more details.

PR #122, opened 2013-06-17
Correctly attach jsdoc comments to AST. Replace == with isNaN calls. Remove executable permissions from some files.

dimvar:


p-bakker:
Had a brief look at whether there might be still something usefull in this PR, by running the UnitTests contained in https://github.com/mozilla/rhino/blob/9ccf1c7e574b455227b85519012c45b2ce1c3009/testsrc/org/mozilla/javascript/tests/AttachJsDocsTest.java against the current master. Result was about 50% of the tests passing and the other 50% either failing or erroring out.

So there might to something usefull in this PR still, so leaving it open for now

p-bakker:
Converted to draft as not mergable in it's current state, but may contain some useful changes

PR #133, opened 2014-04-05
Don't swallow empty lines in doctest; split lines on Mac/Windows/Unix.

cscott:
Use a line-splitting regexp that works on Mac, Windows, and Unix, without
swallowing blank lines.


gbrail:
Rebased and merged to master.

PR #140, opened 2014-05-22
'undefined' pattern should be treated as empty string in RegExp constructor

tntim96:
http://www.ecma-international.org/ecma-262/5.1/#sec-15.10.4.1
https://sourceforge.net/p/htmlunit/bugs/1599/

### DIFF
--- a/src/org/mozilla/javascript/regexp/NativeRegExp.java
+++ b/src/org/mozilla/javascript/regexp/NativeRegExp.java
@@ -183,7 +183,7 @@ public class NativeRegExp extends IdScriptableObject implements Function
             this.lastIndex = thatObj.lastIndex;
             return this;
         }
-        String s = args.length == 0 ? "" : escapeRegExp(args[0]);
+        String s = args.length == 0  || args[0] instanceof Undefined ? "" : escapeRegExp(args[0]);
         String global = args.length > 1 && args[1] != Undefined.instance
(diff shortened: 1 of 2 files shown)


gbrail:
This is already fixed, in another patch, from André Bargull, in the apigee/rhino fork.

Can you verify if that build fixes this problem for you?

tntim96:
It's the same fix, but my PR also has test cases you might want to include.

I've got another PR for you to include if that's OK. As I already have a
Rhino fork, it seems difficult to create a second fork from yours (any idea
how?). I tried following
https://adrianshort.org/create-multiple-forks-of-a-github-repo/, but my PR
ended up here, https://github.com/tntim96/rhino-apigee/pull/1. Is that
usable for you? If it's easier for me to push that PR to mozilla/rhino,
I'll do that. I'll probably end up trashing my mozilla/rhino fork, and
create a proper one from yours.

Also, I wonder would you consider using this Maven PR,
https://github.com/mozilla/rhino/pull/57, that seems to work well. It
allows the project to open up immediately in IntelliJ with the correct
source and test paths, and upload to Maven with 'mvn deploy' which is nice.

Finally, if you're merging and new language features, can you encourage
developers to look at implementing the AST? You could also notify me and
I'll help if I can. Do you have a labeling strategy that would assist with
this?

Cheers,
Tim

PR #167, opened 2015-03-11
Extract default method when interface be implemented by a single function

tuchida:
I want to use the Stream API in JDK8.
```js
java.nio.file.Files.readAllLines(java.nio.file.Paths.get('test.js'),
                                 java.nio.charset.Charset.defaultCharset()).stream()
    .forEach(function(line) {
      print(line);
    });
```
But error.
```
js: Cannot convert function to interface java.util.function.Consumer since it contains methods with different names
```

Rhino can convert JavaScript functions to Java interfaces if interface has only one method.
https://developer.mozilla.org/ja/docs/Rhino/Scripting_Java#JavaScript_Functions_as_Java_Interfaces

`Consumer` has two method, but `andThen` is default method.
http://docs.oracle.com/javase/8/docs/api/java/util/function/Consumer.html

I think it is acceptable to convert in this case.

### DIFF
--- a/src/org/mozilla/javascript/InterfaceAdapter.java
+++ b/src/org/mozilla/javascript/InterfaceAdapter.java
@@ -7,6 +7,7 @@
 package org.mozilla.javascript;
 
 import java.lang.reflect.Method;
+import java.util.ArrayList;
 
 /**
  * Adapter to use JS function as implementation of Java interfaces with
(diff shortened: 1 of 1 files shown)


gbrail:
Sorry I didn't look at this sooner...

This seems like a reasonable change, but I don't think we can incorporate it into Rhino now as-is. "Method.isDefault()" is a new function that was introduced in Java 8 (I believe that default methods are a Java 8 feature). Today we compile Rhino with Java 6 (we just upgraded from Java 5 actually) and it's important that we keep Java 6 compatibility for the time being.

tuchida:
What do you think of this?
https://github.com/tuchida/rhino/commit/e87fca95fd89cda1010d72858df6728a0ddc30eb

I compiled it with jdk1.7.0_65.

tuchida:
Sorry, this PR is incomplete.
```java
public interface Test {
    void foo();
    default void bar() {
        System.out.println("bar");
    }

    static void print(Test t) {
        t.foo();
        t.bar();
    }
}
```
```
js> Packages.Test.print(function() { print('foo'); });
foo
foo
```
I will request again if it can fix.

gbrail:
Thanks.

I'm about ready to call what's in "master" now the 1.7.6 release, and push
it out to Maven. Then we might be able to start to tackle some of the ES6
features.

I don't see people clamoring for the release, but I'll be out next week, so
I'll probably wait until later in April to push the release out.

In the meantime, if you come up with a fix that works well soon, then we
can get it in to 1.7.6, and otherwise we can wait.

On Tue, Mar 31, 2015 at 10:34 PM, tuchida <notifications@github.com> wrote:

> Sorry, this PR is incomplete.
>
> public interface Test {
>     void foo();
>     default void bar() {
>         System.out.println("bar");
>     }
>
>     static void print(Test t) {
>         t.foo();
>         t.bar();
>     }
> }
>
> js> Packages.Test.print(function() { print('foo'); });
> foo
> foo
>
> I will request again if it can fix.
>
> —
> <https://github.com/mozilla/rhino/pull/167#issuecomment-88356396>.
>



-- 
<http://twitter.com/gbrail>

tuchida:
Where I tried, could not call default method. #169

PR #169, opened 2015-04-07
Enable to call default method on java8

tuchida:
When call `Predicate#negate` (default method) in java8.
https://docs.oracle.com/javase/8/docs/api/java/util/function/Predicate.html#negate--

``` js
var p = java.util.function.Predicate({
  test: function(t) {
    return true;
  }
}).negate();
p; // null
```

By right, the method return `Predicate`'s instance, but returned null.
The `negate` method is generated from JavaAdapter in Rhino, and that body is empty.

So I made this:
- Do not generate default method in JavaAdapter.
- Add default method to JavaMember.

### DIFF
--- a/src/build.xml
+++ b/src/build.xml
@@ -14,8 +14,10 @@ Requires Ant version 1.2
 
   <available property="jdk15"
              classname="java.lang.reflect.ParameterizedType" />
+  <available property="jdk18"
+             classname="java.util.stream.Stream" />
 
-  <target name="compile" depends="compile-most,compile-jdk15">
+  <target name="compile" depends="compile-most,compile-jdk15,compile-jdk18">
   </target>
 
   <target name="shell" depends="compile">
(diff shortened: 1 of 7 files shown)


gbrail:
Thanks -- I get this, and I think that's a good way to handle it. 

However, I'd like to publish 1.7.6 soon, and I think that changing the build process so that 1.8 is required to build it is a bit of a big deal -- I'd rather push that out until 1.8.0. (I'd also like to change the build process so that we can use InvokeDynamic optionally at runtime, so that's a good time to do that as well.)

tuchida:
Thanks.

> we can use InvokeDynamic optionally at runtime

Sounds good!

eshepelyuk:
@gbrail could you share usage scenarios for `InvokeDynamic` ?

gbrail:
I merged this in a different PR -- things have changed but it's still valid. Take a look at https://github.com/mozilla/rhino/pull/340.

gbrail:
Fixed this in https://github.com/mozilla/rhino/pull/343.

PR #180, opened 2015-04-27
Implementing possibility of writing  JS tests code inside JUnit Test.

eshepelyuk:
Now one can write JS test like this

````
@RhinoTest(
    inline = "assertEquals(\"abc\".startsWith(\"\", +Infinity), true);"
)
public class StringStartsWithTest extends ScriptTestsBase {
}
````

And code snippet from `inline` property will be executed as it was read from file.
Additionaly `assert.js` will be implicitely included.

Feature is intended for debugging and developing purposes to speead up debugiing huge JS tests suites.


gbrail:
Rebased and merged into master.

PR #181, opened 2015-04-28
Issue 176, Second attempt

raimi:
Trivial new Exception-Creation method with (hopefully) correct and understandable javadoc. Solves at least my problem and destroys nothing else.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,5 +12,4 @@ lib/
 .idea/
 out/
 target/
-
-
+*~
--- a/src/org/mozilla/javascript/ScriptRuntime.java
+++ b/src/org/mozilla/javascript/ScriptRuntime.java
@@ -4243,7 +4243,23 @@ public class ScriptRuntime {
         return new JavaScriptException(error, filename, linep[0]);
     }
 
+
+    /**
+     * Equivalent to executing "new Error(message)" from JavaScript.
+     * @param cx the current context
+     * @param scope the current scope
+     * @param message the message
+     * @return a JavaScriptException you should throw
+     */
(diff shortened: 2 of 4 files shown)


gbrail:
This looks pretty simple -- if no one objects I'll try it out.

gbrail:
I sorry I didn't look at this sooner. The change looks quite reasonable.

However, we don't have a test for it, and I'd like to increase test coverage of new features by making sure that we have them. Is there any possibility that you could write a quick test? 

Since it's something that you need a combination of JS and Java code to test, it'd probably be easiest to add a test to testsrc/org/mozilla/javascript/tests and follow the pattern of the other tests there. Doesn't need to be a super-complex test, but we just want to ensure that there are no regressions in the future.

raimi:
@gbrail ok, i hacked a test, that actually tests something :) it shows that my original method works.

BUT: I left two FIXMEs: I believe I found bugs:
* The internal throwError() loses the sourceName while the lineNumber is correct
* It still allows overwriting built-in error types ! this should not work at all, see the "Step 3" test.

Thanks for considering anything.

gbrail:
So, do you want to try and fix the bugs that you found in the tests? I may be able to look at this in more detail later but haven't yet. But if you want to try and fix them then I will hold off...

raimi:
To be honest, I'd rather have my actual problem fixed and released as 1.7.7. It's a trivial change that breaks nothing and there are now tests in place that verify the behaviour. My day-job-project is waiting for this change quite urgently.

The issues I found have been in there for ages. I'd give it a shot but I can't promise anything. I'll gladly provide a pull request against 1.7.7 :)

Thank you very much!

PR #190, opened 2015-05-11
Complete project migration to Gradle

### LINKED ISSUE #178: Implement build system with Gradle
Hello @gbrail 
A you mentioned in mailing list - it would be good to migrate build system to either `Maven` or `Gradle`.
I'd like to tackle the task of implementing project build with `Gradle`. 
Although not promising that it could be really fast but if you see agree I'd like to start it now.
It also could solve Issue #168 

I am also interested to know your opinion about directory structure. Can I change it to be Maven / Gradle compatible, i.e. looking like src/main/java, src/main/resources etc.

What do you think ?

eshepelyuk:
Closes #178 

1. Removed all `Ant` scripts leftovers,
1. Removed `CVS` leftovers.
1. Changed file structure to comply with standard Maven / Gradle directory structure   
    i.e `src/main/java`, `src/main/resources`, etc
1. Code coverage with `JaCoCo`
    - run it with `./gradlew jacocoTestReport` 
    - find report at `build/reports/jacoco`
1. ZIP archive creation with `Gradle` 
    - run it with `./gradlew distZip`
    - find ZIP at `build/distributions`
1. More tests can be run directly from IDE, e.g. benchmarks tests


README.md
```diff
@@ -55,7 +55,10 @@ More resources if you get stuck:
 
 ### How to Build
 
-Rhino builds with `Gradle`. Here are some useful tasks:
+Rhino builds with `Gradle`. You don't need to install `Gradle` locally to build `Rhino` thanks to [Gradle wrapper](http://gradle.org/docs/current/userguide/gradle_wrapper.html)
+Just clone project from `GitHub` and use provided `gradlew.sh` or `gradlew.bat` for `Linux` / `Mac OS` or `Windows` accordingly.
```
sainaen: Minor correction: even though file for *nix is a bash script, it's called just `gradlew`, not `gradlew.sh`.

RELEASE-NOTES.md
```diff
@@ -64,7 +64,7 @@ André Bargull (60):
 *     Add missing @Override annotations
 *     Add missing generic type info to deprecatedsrc/
 *     Add missing generic type info to toolsrc/
-*     Add missing generic type info to testsrc/
+*     Add missing generic type info to src/test/resources/
```
sainaen: Another minor thing: I don't think a changelog should be updated -- even if folder is no longer there, it's certainly existed when it was written.

gbrail:
OK, I will look, but I am going to take the time to make sure this is really stable first since ant has been with the project for so long. Thanks for doing this!

gbrail:
Thank you very much for all the Gradle work, but here's why I have been hesitant to merge this:

I understand rationally that Git will do lots of smart things when we re-name the source directories, but I guess until I try it I won't be comfortable with that. There is a queue of pull requests, plus lots of people who have other versions of the project, plus some other forks on GitHub, and I fear that if we re-name those directories we are going to waste a lot of time manually merging things.

Either I'm not yet convinced of the power of git, or this is a real concern.

Is there a way that I can cherry-pick the changes here that remove ant and add the other improvements without re-naming everything?

eshepelyuk:
Unfortunately I don't know how to do it using cherry picks.

eshepelyuk:
Hello @gbrail 
FYI I've merged current `master` to this branch and also update Gradle to latest version 2.6.
So this branch reflect all the latest changes from master and completely built with `Gradle` including all the improvements described above.

gbrail:
I've cherry-picked most of these changes, so that the gradle code that we have in master does everything that we need it to do. (It does not actually publish the final build to Maven Central -- that requires signing things and the Gradle recipe for that was complex -- but ant didn't do it right either.)

Thanks for leading the way with Gradle!

I'm ready to get rid of ant...

However, I don't think that we should re-name the project directories. There are always a few PRs in the queue and if we re-name the classes, I fear that we are going to end up with a big mess when we try to merge.

PR #191, opened 2015-05-14
Added readline and write functions to Global

ecornely:
I played with d8 from v8 chrome javascript engine, played as well with jaegermonkey from mozilla, and they both offer an easy way to interact with the command line.

I was missing a way to print without a newline so added the write function.

I was missing an easy way to ask questions in a command line and added the readline function.

I'm not sure the names are perfect but I chose write and readline so that my scripts are compatible with d8.

### DIFF
--- a/toolsrc/org/mozilla/javascript/tools/shell/Global.java
+++ b/toolsrc/org/mozilla/javascript/tools/shell/Global.java
@@ -97,9 +97,11 @@ public class Global extends ImporterTopLevel
             "load",
             "loadClass",
             "print",
+            "write",
             "quit",
             "readFile",
             "readUrl",
+            "readline",
             "runCommand",
             "seal",
             "serialize",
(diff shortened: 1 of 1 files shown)


gbrail:
Look like reasonable changes, except for the last one (need to check again).

I'd love to see some tests. I realize that console tests are hard to write and we have very few, but there's never a bad time to start. Any chance that you can take a crack?

ecornely:
An other solution would be to automatically load a javascript like a ~/.rhino/rinorc.js or something like that where I can create usefull functions like readline by using the available Java API

gbrail:
I agree that "readline" and "write" are reasonable things to have, and the names are fine. I added them to master, and you're welcome to try them out. "readline" also adds an optional "prompt" argument.

It was easy enough to do this without your PR, but thanks for it because it made it very easy to see what to do. Instead of the PR, however, what I did was different in two ways:

* used the existing "ShellConsole" object to properly "readline" without stepping on other console input
* added help

thanks!

PR #193, opened 2015-05-27
Support ES6/ES2015 MethodDefinition

tuchida:
This patch is not implemented this specifications. (I've see ecma-262_6th_edition_final_draft_-04-14-15.pdf).

# non-constructor

```js
assertThrows('new (({ a() {} }).a)', TypeError);
```

> 14.3.8 Runtime Semantics: DefineMethod
>  5. If functionPrototype was passed as a parameter, let kind be Normal; otherwise let kind be Method.

> 9.2.5 FunctionCreate (kind, ParameterList, Body, Scope, Strict, prototype)
>  2. If kind is not Normal, let allocKind be "non-constructor".

# function name property

```js
assertEquals("a", { a() {} }.a.name);
```

> 12.2.6.9 Runtime Semantics: PropertyDefinitionEvaluation
>  6. If IsAnonymousFunctionDefinition(AssignmentExpression) is true, then
>   c. If hasNameProperty is false, perform SetFunctionName(propValue, propKey).

# computed property

```js
assertEquals(123, {
  ['abc' + 'efg']() {
    return 123;
  }
}.abcefg());
```

### DIFF
--- a/src/org/mozilla/javascript/CodeGenerator.java
+++ b/src/org/mozilla/javascript/CodeGenerator.java
@@ -1086,6 +1086,9 @@ class CodeGenerator extends Icode {
             } else if (childType == Token.SET) {
                 visitExpression(child.getFirstChild(), 0);
                 addIcode(Icode_LITERAL_SETTER);
+            } else if (childType == Token.METHOD) {
+                visitExpression(child.getFirstChild(), 0);
+                addIcode(Icode_LITERAL_SET);
             } else {
                 visitExpression(child, 0);
                 addIcode(Icode_LITERAL_SET);
--- a/src/org/mozilla/javascript/Decompiler.java
+++ b/src/org/mozilla/javascript/Decompiler.java
@@ -323,7 +323,12 @@ public class Decompiler
             switch(source.charAt(i)) {
             case Token.GET:
             case Token.SET:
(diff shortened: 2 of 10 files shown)


gbrail:
Hi -- sorry I didn't look sooner, but I'm trying to understand your comments:

When you say:

"This patch is not implemented this specifications."

Do you mean:

"The patch is not implemented according to the specification"

or:

"The patch is now implemented according to the specification."

Thanks!

tuchida:
Thanks!

It does not yet implemented the following specification.

> non-constructor

ArrowFunction and getter/setter and native method is also non-constructor.
```js
var arrow = () => {};
new arrow();  // TypeError

var a = { get b() {} };
var getter = Object.getOwnPropertyDescriptor(a, 'b').get;
new getter();   // TypeError

new String.prototype.match();  // TypeError
```
I think `Function` should have `functionKind`.

> function name property

The name is not the only identifier. There is a need to change `ast/FunctionNode.java`.

> computed property

I think that this should be implemented in another pull request.

gbrail:
Merged this into master. I will close it when we do a release.

gbrail:
Actually, will close now. Release soon.

tuchida:
Great!
I am pleased that Rhino has started the implementation of syntax from ES6.
Beating the Nashorn!

PR #196, opened 2015-06-02
Update .gitignore for eclipse generated files.

asashour:
Eclipse automatically generates .classpath and .project files, which should not be committed by default, this change adds them to .gitignore.


gbrail:
Merged into master. Thanks!

PR #204, opened 2015-06-18
ES5 handling of null and undefined passed as this to Function.prototype.call / Function.prototype.apply

eshepelyuk:
@gbrail @sainaen 

1. Before ES5 null or undefined passed as thisArg were replaced by Global scope object. This PR 
introduces new flag that controls if we have to keep old behaviour or ES5 compatible
2. Additionally this PR intorduces control of ES6 method using Rhino language version. So now if `find` / `findIndex` / `String.repeat` etc are used with version < ES6 - exception will be thrown

This PR should help to properly resolve #174  

##### Note 
Due to current `Rhino` method signatures `undefined` can not be forwarded to function as spec say.
Rhino mostly accepts `Scriptable` instances and `Undefined` doesn't implement this interface.


gbrail:
OK -- now for a brief point on compatibility:

I think it's OK to add new stuff for all versions of Rhino if it doesn't break backward compatibility.

For instance, the new String methods introduced by ES6, IMHO, can safely be introduced for all versions -- there's no need to check if ES6 is specified. The Rhino project has done this in the past, and I think that in general it is perfectly safe to continue. The only things that we need to check the version flag are are things that could break backward compatibility, like the "this" check that we've been debating.

In other words, I think that there's no need to add the check for ">= ES6" before implementing the new String methods.

We have also used the version flag to detect whether completely new classes, like Symbol, should be added to the global namespace. I think that fine -- it doesn't complicated the implementation much, it seems to be in line with what other JavaScript implementations do, and since in theory the new classes might conflict with existing classes in code, it helps with backward compatibility.

However, putting version checks on new methods does complicate the code, and furthermore the world of JavaScript today is so fluid that coders are used to testing for "new" functions anyway, and don't rely on a language level.

So this is a long way of saying that I don't think this patch should explicitly disable the new String and Array class methods based on the version.

eshepelyuk:
Np, I will fix this

eshepelyuk:
The one big issue I see, that implementations of string and array methods from ES6 rely on this new feature from ES5 abiut handling `thisArg`. And they can not be implemented at all with old bahaviour of handling `thisArg`.
So a customer could have an issue setting version to 1.5 and using findIndex for example.
Because in this case thisArg will be handled old way, so findIndex fill behave impredictably.

eshepelyuk:
Update PR.

eshepelyuk:
Please do not merge. Adding some more features, so tests may fail.

eshepelyuk:
Pushed all changed, now ready to review / merge

gbrail:
Reading this through, it looks like we are trying to make sure that the new String methods (startsWith, endsWith, normalize) are not supported if the version < ES6. I think that would be OK except for two things:

* We already pushed a release that supports them, so we might break someone -- although I think that we can work around that, although I have to think about how
* If we are going to exclude methods from the prototype by language level, it's not enough to simply throw a Java exception -- they need to not be on the prototype in the first place. The only way to do that is on the "initPrototypeId" method, and that's not set up to get the Context passed to it, and there are probably other things that could go there as well. Same goes for "findPrototypeId."

Basically, Rhino isn't set up from the ground up to support objects that have different prototypes depending on language version without making a bunch of global changes.

I think that that only realistic ways of handling this would be:

1) Support the new methods for all language levels, and live with the fact that older releases will have inconsistent handling of certain values compared to ES6.
2) Create an "ES6NativeString" class that extends NativeString and adds all the new methods to the existing class.

If you think that the second one makes sense, then I can take a crack at it, maybe next week.

eshepelyuk:
As for me - 1st option is easier and it seems the way Rhino` did before.
So, let me fix this ES6 restrictions for those particular methods if you agree to proceed with Option 1

eshepelyuk:
Another question, should we stick to new Undefined constant implementing `Scriptable` ?
Or instead try to replace current `Undefined.instance` ?

eshepelyuk:
Well, pushed ES6 restriction removal when version < `ES6`. 
I.e. option N1

Issue #159, opened 2015-03-10
Support maps and sets


gbrail:
ES6 adds:

Map
Set
WeakMap
WeakSet

these are supported by recent versions of ES6.

eshepelyuk:
Hello @gbrail 
I am working on this issue. Could you please add corresponding labels ?
During implementation I had to implement naive support for `Symbol` as well to conform to `test262` suite.
I would like to create separate PR with simple `Symbol` implementation to unblock other `ES6` tasks like #164 and maybe #162 before implementing this task.
What do you think ?

gbrail:
Absolutely -- but as you said, a lot of stuff depends on Symbol. So it'd make sense to get Symbol built first, in a separate PR.

eshepelyuk:
I plan to submit PR after 1.7.7 release. Is it OK ?

gbrail:
Yep -- thanks!

On Wed, Jun 17, 2015 at 9:51 AM, Evgeny Shepelyuk <notifications@github.com>
wrote:

> I plan to submit PR after 1.7.7 release. Is it OK ?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/mozilla/rhino/issues/159#issuecomment-112874841>.
>



-- 
*Greg Brail* | *apigee <https://apigee.com/>* | twitter @gbrail
<http://twitter.com/gbrail> @apigee <https://twitter.com/apigee>
<http://iloveapis.com/>

eshepelyuk:
Hello @gbrail 
What do you think about integrating [Apache Commons Collections 4.0](https://commons.apache.org/proper/commons-collections/) into `Rhino` project ?
I would like to use it classes for implementing ES6 iteration, where, according to spec, changes applied to map or set should be visible to iterator even after iterator creation. 
Such behaviour is not supported in standard JDK classes and implementing it looks really ugly and verbose.

gbrail:
I think that we should stay away from including a dependency. Rhino is used
in a lot of different contexts and if we end up with a dependency on this
library, then it is likely (guaranteed, actually) that we will end up with
a conflict with some other project that depends on a different version.
Rhino is a pretty low-level component.

If the requirements are different, I don't see a problem with Rhino having
its own Map and Set implementations designed expressly for the needs of
JavaScript.

FWIW, a lot of the semantics of a map are already implemented in
ScriptableObject, via the slot array.

[...]

PR #205, opened 2015-06-20
Simple framework for running Test262 ES6 test suite

eshepelyuk:
During implementation of #159 I've started to use [Test262](https://github.com/tc39/test262) and discovered usage issues like

1. need to install `Python` and deal with running scripts manually
1. need to assembe jar each time before running test
1. inability to debug test262 tests
1. no integration with project test phase

So to overcome it I've implemented this simple test262 micro framework. It uses [Git Submodule](http://git-scm.com/docs/git-submodule) to fetch test262 sources and then uses properties file to detect which tests to run.

To start using the suite

1. cd test262
1. git submodule init
1. git submodule update

After this you should get test262 sources locally and then you can run `org.mozilla.javascript.tests.Test262SuiteTest`.
The framework uses `testsrc/test262.properties` for configuring tests that should be run from `test262` suite.


gbrail:
OK, that looks promising -- I'll try it out.

In addition, we run test262 (using the JAR and Python) regularly on the CI server:

http://ci.apigee.io/job/Rhino%20test262/

PR #207, opened 2015-06-27
TypedArray and arguments are Iterable.

tuchida:
And fixed some specification violation.


tuchida:
@eshepelyuk Sorry. Do I make a mistake in how to use your test?

sainaen:
@tuchida 
I think what @eshepelyuk meant is that you can replace several ignores that share common prefix, with just one. For example,

     ! map-contract-expand.js
     ! map-contract.js
     ! map-expand-contract.js
     ! map-expand.js
     ! map.js

can be replaced with single

     ! for-of/map

Note: `! map` would also exclude files that are not excluded right now and have 'map' in the name, like  `arguments-mapped.js`.

tuchida:
@sainaen
Thanks! I understood.

sainaen:
Wait, but why did you add a `! map`? Isn't it ignores more than needed?

tuchida:
Thanks!

partial matching
https://github.com/mozilla/rhino/blob/ce6771e502339c9a17809d8742b3d2af705464c7/testsrc/org/mozilla/javascript/tests/Test262SuiteTest.java#L149

gbrail:
Merged to master and rebased. Thanks!

PR #209, opened 2015-06-28
Make array literal construction always set own properties of instance

sainaen:
When new array created with array literal, previously we would create an array instance and call `ScriptableObject.putProperty()` for each defined element of array. That contradicts the spec that requires implementation to always set own property of newly created array instance, even when [[Set]], because of its recursive traversal nature, would update property on `Array.prototype`.

Consider the case, when `Array.prototype` has "0" property defined, with `writable` property set to `false` in its descriptor:
```js
    Object.defineProperty(Array.prototype, "0", {
        value: 10,
        configurable: true,
        enumerable: false,
        writable: false
    });
```
Now, all arrays that do not have "0" property defined, in [[Get]] would find a prototype property and return its value:

```js
    [][0]
    //> 10
```

But what's worse, when new array is created and it 1st element set to provided value, because `writable` property of its descriptor is `false`, `putProperty()` will not update the value:

```js
    [123][0]
    //> 10
```

This fix, ensures that when new array instance created from literal, only its own property will be set.

Also test262's [`Array/prototype/concat/15.4.4.4-5-b-iii-3-b-1.js`](https://github.com/tc39/test262/blob/master/test/built-ins/Array/prototype/concat/15.4.4.4-5-b-iii-3-b-1.js) now passes.

### DIFF
--- a/src/org/mozilla/javascript/ScriptRuntime.java
+++ b/src/org/mozilla/javascript/ScriptRuntime.java
@@ -3804,7 +3804,7 @@ public class ScriptRuntime {
                 ++skip;
                 continue;
             }
-            ScriptableObject.putProperty(array, i, objects[j]);
+            array.put(i, array, objects[j]);
             ++j;
         }
         return array;
--- a/testsrc/test262.properties
+++ b/testsrc/test262.properties
@@ -12,3 +12,5 @@ built-ins/String
     ! prototype/trim/15.5.4.20-1-2.js
     ! raw/special-characters.js
     ! raw/zero-literal-segments.js
+
+built-ins/Array/prototype/concat/15.4.4.4-5-b-iii-3-b-1.js


gbrail:
Thanks!

I want to make sure that we maintain test coverage for all fixes in the regular code, so that every developer and CI job will cover fixes like this. Evgeny's work to run test262 tests in the regular build (via a git submodule) works great, but only if we explicitly enable tests that we know are passing.

Can you make the corresponding change in this fix so that we have a test for it? I THINK that's a matter of adding Array/prototype/concat/15.4.4.4-5-b-iii-3-b-1.js to "testsrc/test262.properties."

Can you please do that?

@eshepelyuk is that the right way to add 262-based test coverage to fixes like this?

eshepelyuk:
Just a small change - `built-ins` should be added in the beginning. 
So to enable single test, e.g. `15.4.4.4-5-b-iii-3-b-1.js` we should add single line to `testsrc/test262.properties`
```
 built-ins/Array/prototype/concat/15.4.4.4-5-b-iii-3-b-1.js
```

gbrail:
Thanks!

So in the spirit of learning to fish and all that, can you add this @sainaen? Thanks!

sainaen:
Added.

I'm working on unignoring and adding the rest of Array tests in separate PR. In particular, I think I have a fix for handling array-like objects with lengths bigger than 2^32 passed to Array's method (according to ES6, methods that don't create arrays should handle those cases).

gbrail:
Rebased and merged. Thanks!

PR #210, opened 2015-06-29
Apply a argument for error message.

tuchida:
```
js> 'use strict';(function(eval) {})();
js: "<stdin>", line 6: "{0}" is not a valid identifier for this use in strict mode.
```

### DIFF
--- a/src/org/mozilla/javascript/Parser.java
+++ b/src/org/mozilla/javascript/Parser.java
@@ -253,7 +253,7 @@ public class Parser
     void reportError(String messageId, String messageArg, int position,
                      int length)
     {
-        addError(messageId, position, length);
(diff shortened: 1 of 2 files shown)


gbrail:
Similar to my comment on the last PR -- we need to have a test for this. If there is a test262 test that previously failed and now passes, then can you add that test to testsrc/test262.properties so that a test gets run as part of the regular build? Thanks!

tuchida:
Sorry. I did not explain enough.

The problem is that the error message is unkind.
```
"{0}" is not a valid identifier for this use in strict mode.
```
expect
```
"eval" is not a valid identifier for this use in strict mode.
```
It is not related to ECMA-262 specification.

gbrail:
I get it, and thanks for the test!

PR #216, opened 2015-07-07
negative test in test262

tuchida:
https://github.com/tc39/test262/blob/6f22dad1529fc1b8a0f263a357569bea8eb20176/test/language/statements/for-of/let-bound-names-fordecl-tdz-for-of.js#L7

> negative: ReferenceError

The test case will throw a ReferenceError. But passed the test it did not become so.

### DIFF
--- a/testsrc/org/mozilla/javascript/tests/Test262SuiteTest.java
+++ b/testsrc/org/mozilla/javascript/tests/Test262SuiteTest.java
@@ -11,7 +11,8 @@ import org.junit.runner.RunWith;
 import org.junit.runners.Parameterized;
 import org.junit.runners.Parameterized.Parameters;
 import org.mozilla.javascript.Context;
-import org.mozilla.javascript.JavaScriptException;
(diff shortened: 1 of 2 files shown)


gbrail:
I am getting back in the swing of things now and just merged the fix from @sainaen on array prototype methods. Since I merged that one first, I see a bunch for new test failures for Array.prototype.splice when I merge this fix.

Plus, this fix changes the way that we handle results from test262, so I was hoping a few others could look at that code as well and make sure that we don't introduce any false positives.

Can you rebase this and see if you can figure out why the new tests are failing with this fix? Thanks!

tuchida:
except S15.4.4.12_A6.1_T2.js

gbrail:
OK -- there was as previous commit that was blocking this one, which we had to revert because it was incomplete. So this one works now and there are are a lot more test262 tests running, so thanks!

tuchida:
Thanks!

PR #222, opened 2015-07-27
Conversion of JSObject gives priority to Map than Interface implementation.

tuchida:
```java
public class Overload {
    public static String x(Collection<String> x) {
        return "collection";
    }
    public static String x(Map<String, String> x) {
        return "map";
    }
}
```
Rhino 1.7R3
```
js> Packages.Overload.x([])
collection
js> Packages.Overload.x({})
map
```
1.7R4 or later
```
js> Packages.Overload.x([])
collection
js> Packages.Overload.x({})
js: The choice of Java method Overload.x matching JavaScript argument types (object) is ambiguous; candidate methods are: 
    class java.lang.String x(java.util.Collection)
    class java.lang.String x(java.util.Map)
```
The cause is this commit.
https://github.com/mozilla/rhino/commit/3e83f18588cda4fe6701ba388afd3d99b5849d53#diff-b08a68e60375aeef7bb378b20d93ea93L400

Conversion to Map is same priority as Interface implementation.

### DIFF
--- a/src/org/mozilla/javascript/NativeJavaObject.java
+++ b/src/org/mozilla/javascript/NativeJavaObject.java
@@ -362,9 +362,13 @@ public class NativeJavaObject implements Scriptable, Wrapper, Serializable
                 }
             }
             else if (to.isInterface()) {
-                if (fromObj instanceof NativeObject || fromObj instanceof NativeFunction) {
+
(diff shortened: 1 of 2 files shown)


gbrail:
Closed this on July 27.

PR #229, opened 2015-08-19
Force UTF-8 encoding for test JS files

eshepelyuk:
When default system encoding is used - some tests can fail because they contain UTF-8 characters.
This PR forces UTF-8 encoding for test input files.


gbrail:
Closed and merged. thanks!

PR #231, opened 2015-08-31
ES6 improvements: iterator and function

eshepelyuk:
* Added ES6 `@@toStringTag` to iterator objects
* Generic reusable iteration `next` algorithm
* more `test262` tests
* Very basic implementation of ES6 compliant function object
  - name is configurable
  - length is configurable
  - create new function object by default for version >= ES6
* updated test262 to latest version
* Enable more test262 tests

### DIFF
--- a/src/org/mozilla/javascript/ES6Iterator.java
+++ b/src/org/mozilla/javascript/ES6Iterator.java
@@ -92,7 +92,7 @@ public abstract class ES6Iterator extends IdScriptableObject {
             return Id_next;
         } else if (ITERATOR_PROPERTY.equals(s)) {
             return Id_iterator;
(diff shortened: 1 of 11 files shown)


gbrail:
Looks OK to me -- is it ready to go for you? If so I can test and merge it.

Also, I see you might have your IDE set to replace multiple imports from the same package with "*". (This happened in the native array class I think). I'm old-fashioned and try to avoid using import with a "*" in the theory that it might cause compilation problems later (and it's an old Java style I adopted a long time ago). 

So next time, can you set up your IDE so that it doesn't do that?

Other random note -- at some point we'll need to re-write the code that generates the "ID switch" stuff to take into consideration the "@" syntax used by iterators...

eshepelyuk:
Fixed imports. Please proceed with testing / merging.

Issue #233, opened 2015-09-05
Object.prototype.toString.call() does not work correctly on null or undefined


swannodette:
In the 1.7.7 release `Object.prototype.toString.call(null)` returns `"[object Object]"` rather than the correct `"[object Null]"` value. The `undefined` case is similarly wrong. See the ECMAScript 262 Edition 5.1 [Section 15.2.4.2](http://www.ecma-international.org/ecma-262/5.1/#sec-15.2.4.2).

I note that on master these cases actually crash Rhino.

PR #234, opened 2015-09-07
Make Object.prototype.toString behavior to conform to ECMA-262 15.2.4.2

swannodette:
Object.prototype.toString.call(null) should return [object Null]
Object.prototype.toString.call(undefined) should return [object Undefined]

Fixes https://github.com/mozilla/rhino/issues/233

### DIFF
--- a/src/org/mozilla/javascript/ScriptRuntime.java
+++ b/src/org/mozilla/javascript/ScriptRuntime.java
@@ -842,6 +842,10 @@ public class ScriptRuntime {
 
     static String defaultObjectToString(Scriptable obj)
     {
+        if (obj == null)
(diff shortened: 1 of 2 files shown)


eshepelyuk:
You could use `Undefined.isUndefined()` method to check if certain object is `undefined`

swannodette:
@eshepelyuk thanks for the feedback. Will do that and submit a fresh PR.

gbrail:
Thanks to you both -- will merge when I get the new PR. 

In addition, are there any es262 tests (in test262.properties) that we can enable to test this fix using the 262 test suite?

swannodette:
Closing this one I've created a new on off a branch.

PR #236, opened 2015-09-08
Make Object.prototype.toString behavior to conform to ECMA-262 15.2.4.2

### LINKED ISSUE #233: Object.prototype.toString.call() does not work correctly on null or undefined

swannodette:
Object.prototype.toString.call(null) should return [object Null]
Object.prototype.toString.call(undefined) should return [object Undefined]

Fixes #233

### DIFF
--- a/src/org/mozilla/javascript/ScriptRuntime.java
+++ b/src/org/mozilla/javascript/ScriptRuntime.java
@@ -842,6 +842,10 @@ public class ScriptRuntime {
 
     static String defaultObjectToString(Scriptable obj)
     {
+        if (obj == null)
+            return "[object Null]";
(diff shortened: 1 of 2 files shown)


gbrail:
This looks quite reasonable and didn't break anything, so I merged it. Thanks!

Issue #238, opened 2015-09-14
Delayed execution of BoundFunction


AlexTrotsenko:
We have JS code, which creates Java (android) UI (dialog).
Also this JS code passes some other JS Functions to this UI component in order to be called later: when user do some interaction with UI (clicks dialog button).

Currently it works fine for all other interpreted functions, except BoundFunction, which is `bind` to null.

According to specification this should lead to using global object as `this`. But beacuase this function is invoked after the end of the ScriptRuntime.doTopCall of the 1st script ScriptRuntime.getTopCallScope(cx) results in the IllegalStateException.

Since spec says, that global scope should be used, BoundFunction should store refernce to this global scope at time of creation and use it later.

szegedi:
## Short story

I think this issue is invalid and the current behavior is compliant with ES5.1 spec.

## Detailed explanation

[ES5 spec 15.3.4.5 "Function.prototype.bind"](http://es5.github.io/#x15.3.4.5) says _"8. Set the [[BoundThis]] internal property of F to the value of thisArg."_ and _"12. Set the [[Call]] internal property of F as described in 15.3.4.5.1."_.

[15.3.4.5.1. in turn says](http://es5.github.io/#x15.3.4.5.1) _"5. Return the result of calling the [[Call]] internal method of target providing boundThis as the this value and providing args as the arguments._

It says nothing about substituting null for global scope anywhere in this part of the spec. Now granted, if we follow the spec further, the ordinary `[[Call]]` internal method of functions as specified in [13.2.1](http://es5.github.io/#x13.2.1) says "1. Let funcCtx be the result of establishing a new execution context for function code using the value of F's [[FormalParameters]] internal property, the passed arguments List args, and the this value as described in 10.4.3."

[10.4.3](http://es5.github.io/#x10.4.3) finally says: _"2. Else if thisArg is null or undefined, set the ThisBinding to the global object."_

So there you have it. According to spec:
* bound functions need not treat _this_ being null in any special way on either bind call, or when they're invoked, and
* when the target function is invoked through bound function, the target function's [[Call]] needs to check for null and replace it with global object (unless function is strict), but this happens on invocation, and for the global that's current on invocation.

This means the code for Rhino `BoundFunction` is up to spec. I'll close this issue and the associated PR, but feel free to reopen it if you disagree.

PR #239, opened 2015-09-14
fixes mozilla/rhino#238. Fixed issue, preventing delayed execution of BoundFunction

AlexTrotsenko:
Added storing of topCallScope at time of BoundFunction creation according to spec: ScriptRuntime.getTopCallScope(cx) results in the IllegalStateException if function was bound to null and executed some times later (after the end of the ScriptRuntime.doTopCall).


gbrail:
Can you point to or track down a test that we can add to this? Just want to make sure it doesn't regress some day. Thanks!

szegedi:
Closing as per explanation in #238

PR #240, opened 2015-09-16
Fixed bug when repeated an empty string MAX_INT times

tuchida:
Fixed bug `empty-string-returns-empty.js` that is test case on test262.

```js
''.repeat(MAX_INT);  // expected results is empty string, but different.
```

### DIFF
--- a/src/org/mozilla/javascript/NativeString.java
+++ b/src/org/mozilla/javascript/NativeString.java
@@ -475,7 +475,7 @@ final class NativeString extends IdScriptableObject
 
                     long size = str.length() * (long) cnt;
                     // Check for overflow
-                    if (cnt >= Integer.MAX_VALUE || size >= Integer.MAX_VALUE) {
+                    if (cnt > Integer.MAX_VALUE || size > Integer.MAX_VALUE) {
                         size = Integer.MAX_VALUE;
                     }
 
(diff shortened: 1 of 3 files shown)


gbrail:
This is better -- and thanks! 

I'm still not sure that we have it right, though. What happens if "cnt" is greater than MAX_INT? It looks like we adjust the size of the StringBuilder, but won't we still try to append to the string until we have exceeded MAX_INT? I don't see anything in the spec about what to do if "cnt" exceeds MAX_INT but can we throw a RangeError in that case?

tuchida:
I could not find the spec when it exceeds the MAX_INT.

http://www.ecma-international.org/ecma-262/6.0/#sec-string.prototype.repeat
http://www.ecma-international.org/ecma-262/6.0/#sec-stringcreate

But other engines check size.

https://hg.mozilla.org/mozilla-central/file/f1dffc8682fb/js/src/builtin/String.js#l172
http://trac.webkit.org/browser/trunk/Source/JavaScriptCore/runtime/StringPrototype.cpp?rev=186037#L734
https://chromium.googlesource.com/v8/v8.git/+/7a7b692b30d50daacc1d054281f723d6c2255ccf/src/string.js#940

I think that should throw RangeError.

gbrail:
OK, Thanks.

I made a few changes, including adding the range check as we discussed.

I also changed the code style a bit -- moved "repeat" into a separate function, and tried to do a bit more of the work with ints instead of doubles.

Can you take a quick look at my commit and make sure I didn't miss anything? Then I'll merge. Thanks!

https://github.com/mozilla/rhino/commit/8c1d3a696dd59ff0869f2c16ce254bf9f1f7acac

tuchida:
LGTM
Thanks!

PR #241, opened 2015-09-16
Fixed some specifications violation for arrow function.

tuchida:
These test cases have come to pass.

## Can not use reserved word for identifier
```js
// SyntaxError
var af = enum => 1;
```
## Transform to STRICT_SETNAME when outer function is strict mode
```js
'use strict';
assert.throws(ReferenceError, function() {
  var af = _ => {
    foo = 1;
  };

  af();
});
```
## Line terminator not allowed between arguments and arrow
```js
// SyntaxError
var af = ()
=> {};
```


gbrail:
This one looks good to me. Rebased and merged. Thanks!

Issue #242, opened 2015-09-19
Implement octal and binary literals


gbrail:
The Node.js 4.0 code base, in particular, makes pretty heavy use of these.

gbrail:
Open question on this:

Should we add a switch to only work in ES6 mode? If we don't, will it break existing code?

gbrail:
I'm thinking that there is little risk since these kinds of literals just wouldn't compile previously. But we've been a little lose (in the name of performance and code simplicity) about whether certain features should check for the "ES6" version in the Context, so I wanted to think about it first.

PR #246, opened 2015-09-28
Implement octal and binary literals

tuchida:
https://github.com/mozilla/rhino/issues/242

This test case is wrong.
https://github.com/tc39/test262/blob/0d9733e93be9e6783f56a3fd4163a4ca083b159c/test/language/literals/numeric/octal-invalid-truncated.js
https://github.com/tc39/test262/issues/433

### DIFF
--- a/src/org/mozilla/javascript/Parser.java
+++ b/src/org/mozilla/javascript/Parser.java
@@ -2957,11 +2957,14 @@ public class Parser
           case Token.NUMBER: {
               consumeToken();
               String s = ts.getString();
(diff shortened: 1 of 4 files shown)


gbrail:
Awesome!

My only question on this (and for others who may be watching). Should we add a switch to only work in ES6 mode? If we don't, will it break existing code?

tuchida:
Thank you for your feedback.
I made those valid only in ES6.

gbrail:
Sorry I didn't close this -- I merged it quite a while ago! Thanks!

PR #262, opened 2016-02-01
Cache ClassLoading misses

WinZib:


WinZib:
probably need to create a map of set (for different classloader maintain own set of classloading misses) More than that such implementation won't work under OSGi

WinZib:
see the change for supporting classloading from different classloaders

gbrail:
Thank you for contributing this!

Do you think it's possible that an app might create a large number of context classloaders? If that's the case and they use this feature then they run the risk of generating a big memory leak. 

My other concern is that previously, "classOrNull" just called "Class.forName()" rather than explicitly using the context classloader. I know that there is a lot that I don't know about classloading. Will this cause any apps to behave differently?

Finally, I don't think that I can commit this without some tests, preferably a new test or two that will somehow try to verify that the caching is working. It does look like we test those functions in the existing test suite, so you can also run "./gradlew jacocoTestReport" to see the coverage data and verify that your new code paths are being tested.

WinZib:
>>Do you think it's possible that an app might create a large number of context classloaders? If that's the >>case and they use this feature then they run the risk of generating a big memory leak.

yes, it's a good point. Should I restrict it by some custom value or get it from System.getProperty?

>>Will this cause any apps to behave differently?
yes, it strongly depends on ClassLoader implementation but base behaviour should be common

WinZib:
update pull request 
@gbrail pls have a look

WinZib:
@gbrail pls review this pull request

gbrail:
Sorry that this fell off my radar. It sounds cool and there's a test ;-)

It does sound to me like we need some docs for this, since it sounds like a new feature. Would you consider writing something up? Even an article on the Mozilla Wiki (which I think everyone should contribute to).

Either that, or at the minimum can you please put some detailed instructions into a commit so that people know how to use it? Thanks!

emukidid:
With multiple concurrent engines running this would help alleviate blocking in this scenario:

`java.lang.Thread.State: BLOCKED
        at sun.misc.URLClassPath.getNextLoader(URLClassPath.java:479)
        - waiting to lock <75dbfc79> (a sun.misc.URLClassPath) owned by "Thread: 13361" t@26902
        at sun.misc.URLClassPath.getResource(URLClassPath.java:248)
        at java.net.URLClassLoader$1.run(URLClassLoader.java:366)
        at java.net.URLClassLoader$1.run(URLClassLoader.java:363)
        at java.security.AccessController.doPrivileged(Native Method)
        at java.net.URLClassLoader.findClass(URLClassLoader.java:362)
        at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
        - locked <39f29ec8> (a java.lang.Object)
        at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:349)
        at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
        at org.mozilla.javascript.Kit.classOrNull(Kit.java:60)
        at org.mozilla.javascript.NativeJavaPackage.getPkgProperty(NativeJavaPackage.java:129)
        - locked <16b6cd8f> (a org.mozilla.javascript.NativeJavaPackage)
        at org.mozilla.javascript.ImporterTopLevel.getPackageProperty(ImporterTopLevel.java:111)
        at org.mozilla.javascript.ImporterTopLevel.has(ImporterTopLevel.java:91)
        at org.mozilla.javascript.ScriptableObject.getBase(ScriptableObject.java:2695)
        at org.mozilla.javascript.ScriptableObject.hasProperty(ScriptableObject.java:2429)
        at org.mozilla.javascript.ScriptRuntime.initScript(ScriptRuntime.java:3581)
        at org.mozilla.javascript.gen._Unknown_source__109714._c_script_0(<Unknown source>)
        at org.mozilla.javascript.gen._Unknown_source__109714.call(<Unknown source>)
        at org.mozilla.javascript.gen._Unknown_source__109714.exec(<Unknown source>)
        at org.mozilla.javascript.Context.evaluateReader(Context.java:1293)`

WinZib:
@gbrail pls provide me a link to appropriate  Wiki's space? I'll try to prepare some article

gbrail:
I'm again sorry I didn't answer sooner.

If you could put a few sentences description into the change comment, or even in the PR, then I can put it in the release notes.

I have one more concern about the thread-safety of this implementation. I'll add comments to the PR.

There is also a Rhino Wiki but I'm not sure how you get access -- I might have to upload that for you:

https://developer.mozilla.org/en-US/docs/Mozilla/Projects/Rhino/Documentation

src/org/mozilla/javascript/Kit.java
```diff
@@ -56,11 +97,23 @@
      */
     public static Class<?> classOrNull(ClassLoader loader, String className)
     {
+
+        LRUCache lruCache = null;
+        if (enabledNonClassCaching) {
+            lruCache = getOrCreateCache(loader);
+
+            if (lruCache.containsKey(className))
```
gbrail: I don't know that we can assume that LRUcache is only used by one thread, so I think this needs to be synchronized.

src/org/mozilla/javascript/Kit.java
```diff
@@ -56,11 +97,23 @@
...
+        }
+
         try {
             return loader.loadClass(className);
-        } catch (ClassNotFoundException ex) {
-        } catch (SecurityException ex) {
-        } catch (LinkageError ex) {
+        } catch  (ClassNotFoundException ex) {
+            if (enabledNonClassCaching && lruCache != null) {
+                lruCache.add(className);
```
gbrail: Again, "add" is not thread-safe and I think it should be.

src/org/mozilla/javascript/Kit.java
```diff
@@ -144,6 +197,52 @@ public static int xDigitToInt(int c, int accumulator)
         return (accumulator << 4) | c;
     }
 
+    public static void reinitNonClassCache() {
+        notClassMap = Collections.synchronizedMap(new WeakHashMap<ClassLoader, LRUCache>());
+    }
+
+    private static synchronized LRUCache getOrCreateCache(ClassLoader loader) {
```
gbrail: It is good that this is synchronized. However, we also have to synchronize usage of the cache once it's been created.

Also, if you're doing this for efficiency, then "notClassMap" could be a ConcurrentHashMap, and then you see if this is cleaner and uses less locking if you use "putIfAbsent". That leads to other complexity though, and it's probably a toss-up whether it buys you anything.

src/org/mozilla/javascript/Kit.java
```diff
@@ -144,6 +197,52 @@ public static int xDigitToInt(int c, int accumulator)
...
+    public static int getNonClassCacheSize() {
+        int result = 0;
+
+        for (Map.Entry<ClassLoader, LRUCache> entry : notClassMap.entrySet()) {
+            if (entry instanceof WeakReference) {
+                WeakReference<ClassLoader> weak = (WeakReference<ClassLoader>)entry;
+                if (!weak.isEnqueued())
+                    result += entry.getValue().size();
+            } else {
+                System.err.println("strange entry type");
```
gbrail: I think that if this is really something that should never happen then you should throw an AssertionError here and be done with it.

OTOH if it's something that we need to recover from then you should return a reasonable result.

Otherwise, having Rhino print to stderr doesn't really help anyone and will just cause us problems later.

Also this isn't thread-safe ;-)

emukidid:
Bump! I still have a fork specifically for this issue so it'd be nice if it could be merged soonish :P

testsrc/org/mozilla/javascript/KitTest.java
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    @AfterClass
+    public static void disableCaching() throws Exception {
+        enabledFld.set(null, enabledNonClassCaching);
+    }
+
+    private static class TestClassLoader extends URLClassLoader {
+        public TestClassLoader() {
+            super(((URLClassLoader)getSystemClassLoader()).getURLs());
```
tonygermano, line 67: In java 9+ getSystemClassLoader does not return an instance of URLClassLoader.

src/org/mozilla/javascript/Kit.java
```diff
@@ -34,20 +73,22 @@
         } catch (Exception ex) {
             // Assume any exceptions means the method does not exist.
         }
+
+        reinitNonClassCache();
+
+        //identify enabling of non-class cache
+        try {
+            enabledNonClassCaching = CACHE_USE_PROPERTY_NAME_YES.equals(System.getProperty(CACHE_USE_PROPERTY_NAME));
```
p-bakker, line 81: Config settings like this should be exposed through a feature flag in the Context class (see https://github.com/mozilla/rhino/pull/619#discussion_r344956743 for example

p-bakker:
For this PR to be considered for merging, the pending comments would first have to be resolved

p-bakker:
@emukidid dunno if you still maintain a branch with this and want to have it merged, but since @WinZib seems mia, maybe you want to create a new PR from your branch and help it over the finish?

emukidid:
I don't have the time to help finish it to a level that would be acceptable sorry.

src/org/mozilla/javascript/Kit.java
tonygermano: I don't believe the JRE has a concurrent version of a WeakHashMap. Google Guava does, but we probably don't want to pull in that dependency.

p-bakker:
Converted to draft as it needs work to be considered for merging

PR #269, opened 2016-02-10
Fixed Stackoverflow with string concatenation

szarnekow:
Also improved performance of ConsString.toString.

If the script was used in a deep call or along with other intermediate nodes in the script's AST, the flattening was not sufficient to avoid stack overflows.


gbrail:
OK -- looks like a clever improvement on the algorithm. Thanks!

gbrail:
I rebased this and merged it a while back but forgot to close this. Thanks!

Issue #266, opened 2016-02-08
Caching ContinuationPending exception or allow subclassing it


spielzeugland:
In a project we are using continuations quite heavily. There are multiple interceptions per script call. For performance reasons we would like to avoid the cost of creating a ContinuationPending exception every time by overriding org.mozilla.javascript.Context#captureContinuation(). We would care about setting the right continuation and application state. Unfortunately the constructor of ContinuationPending exception is package visible only. Making it protected would be one option to solve this for us.

gbrail:
That's a simple thing to fix and I don't see what it would hurt (pending tests). 

Any chance you'd be able to try it out yourself and create a pull request?

gbrail:
Thanks for the PR!

PR #275, opened 2016-03-10
Allow subclassing of ContinuationPending

spielzeugland:
Proposal for #266 "Caching ContinuationPending exception or allow subclassing it"


gbrail:
Not much controversial here -- I'd say to add a test but I'm not sure what it'd test. Thanks!

PR #277, opened 2016-04-29
fix undefined script value become a string value after pass java call…

bruce007lee:
….below to reproduce:

js code:
TestObj.testValue(notExist);

java code:
public void testValue(Object val){
   System.out.println(val); //become "undefined" string
}

### DIFF
--- a/src/org/mozilla/javascript/NativeJavaObject.java
+++ b/src/org/mozilla/javascript/NativeJavaObject.java
@@ -489,7 +489,7 @@ public class NativeJavaObject implements Scriptable, Wrapper, Serializable
         case JSTYPE_UNDEFINED:
             if (type == ScriptRuntime.StringClass ||
                 type == ScriptRuntime.ObjectClass) {
-                return "undefined";
+            	return null;
             }
             else {
                 reportConversionError("undefined", type);


gbrail:
Thanks for submitting this and I'm sorry that I didn't comment way back when you submitted it.

I fear that if we change the string value of "undefined" in Java code, which has been built in to Rhino for a very long time, that we will break too much existing code.

p-bakker:
The LiveConnect (3) spec that governs how values are converted between Java and JavaScript specifies both JavaScript null and undefined ought to be converted to a Java null value when passed to Java: https://www.oracle.com/java/technologies/javase/liveconnect-docs.html#JS_NULL

So Rhino converting undefined  to a String doesn't align

If we want to fix this at some point, we can add some Context flag to switch between different behaviours:
- leave the current behaviour that returns converts to a 'undefined' string
- Follow LC3 spec and convert undefined to null
- return the undefined value as is

With the latter option, the integrator can use `Undefined.isUndefined(Object obj)` to check whether dealing with an undefined value

A Context flag could allow an integrator to choose the behaviour they want

Some relevant links: 
- https://www.oracle.com/java/technologies/javase/liveconnect-docs.html#CONFORMANCE
- https://www.askapache.com/pub/mozilla/js/src/liveconnect/README.html
- https://www.codejava.net/java-se/applet/liveconnect-the-api-for-communication-between-java-applet-and-javascript

p-bakker:
Am closing this PR  as this PR in is current state won't be merged, as it breaks backwards compatibility. 

Case #760 remains to work on this, but requires a different/larger PR than this one too be considered

PR #281, opened 2016-07-19
fixed typo

qwertyppp11:


sainaen:
Hi! Thanks for your PR, but I'm not sure it's a typo. _iff_ is typically used as a shortend version of _if and only if_ (see [Wikipedia](https://en.wikipedia.org/wiki/If_and_only_if)) which fits in this case.

qwertyppp11:
Found 12 "iff" in source. Seems it is not typo.

PR #285, opened 2016-09-19
	Check before getting "toJSON" property because some built-in objects…

svella:
… (NativeJavaArray and XML) throw an exception if you try to get a non-existent property. This allows for a JSON.stringify replacer function to actually get called and have a chance to do something sane with these objects instead of just barfing.

### DIFF
--- a/src/org/mozilla/javascript/NativeJSON.java
+++ b/src/org/mozilla/javascript/NativeJSON.java
@@ -276,7 +276,7 @@ public final class NativeJSON extends IdScriptableObject
             value = getProperty(holder, ((Number) key).intValue());
         }
 
-        if (value instanceof Scriptable) {
+        if (value instanceof Scriptable && hasProperty((Scriptable) value, "toJSON")) {
             Object toJSON = getProperty((Scriptable) value, "toJSON");
             if (toJSON instanceof Callable) {
                 value = callMethod(state.cx, (Scriptable) value, "toJSON",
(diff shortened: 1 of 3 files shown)


gbrail:
Sure, this change makes sense.

I know it is a small change but what is the chance that you can write a test case for it?

svella:
Will do, though may not get to it for a few days.

svella:
Just as a sanity check - build succeeds and all tests pass when building using JDK7. The tests fail to build using JDK6 with error: 
```
com/google/caliper/model/Measurement.class(com/google/caliper/model:Measurement.class): warning: Cannot find annotation method 'value()' in type 'javax.persistence.Access': class file for javax.persistence.Access not found
An exception has occurred in the compiler (1.6.0_65). Please file a bug at the Java Developer Connection (http://java.sun.com/webapps/bugreport)  after checking the Bug Parade for duplicates. Include your program and the following diagnostic in your report.  Thank you.
com.sun.tools.javac.code.Symbol$CompletionFailure: class file for javax.persistence.AccessType not found
:compileTestJava FAILED
```
And when building with JDK 8 tests regress-226507.js and regress-98901.js both fail with stack overflows.
As far as I can tell these failures have nothing to do with my changes. Are these known issues? I JDK7 what is used to build the releases?

gbrail:
Today we build on Java 7, although I used to have a CI job that ran regularly on Java 8.

I think that at this point (as opposed to a year ago) it's time to drop support for Java 6. I know that this will make a lot of people happy ;-)

There are a few tests that seem to be measuring some sort of timing and seem to behave differently on Java 8. I'm not sure that they're valid tests any more. It's probably time to disable them or fix them (if we can figure out if anything needs fixing).

Other than that I will merge this later in the week -- next few days are busy though.

gbrail:
Thanks for submitting this!

PR #288, opened 2016-10-29
Param to omit "Missed Context.javaToJS() conversion" warning 

wildloop:
Rhino is used by Alfresco as JavaScript engine, they patched this library to omit "Missed Context.javaToJS() conversion" warning.
Please add parameter that does this work.


TheDiamondYT1:
Any updates on this?

gbrail:
It would be great to get rid of this silly System.err.println even though it's been in the codebase forever.

I was thinking that it would make more sense to make this a parameter on the Context object, but on reflection I doubt that many people would want to turn it back on.

PR #292, opened 2016-12-19
Update NativeDate.java

Jugolo:
I has calculate it and sometime this will fail and this should fix it

### DIFF
--- a/src/org/mozilla/javascript/NativeDate.java
+++ b/src/org/mozilla/javascript/NativeDate.java
@@ -539,7 +539,7 @@ final class NativeDate extends IdScriptableObject
 
         // d: date count from 1 March
         int mdays, mstart;
-        switch (d / 30) { // approx number of month since March
+        switch (Math.round(d / 30)) { // approx number of month since March
             case 0: return d + 1;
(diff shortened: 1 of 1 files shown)


gbrail:
Thanks for submitting this. It's a pretty critical piece of code, however,
and it's hard to accept this PR without more information, such as:

   - A test that failed previously and is fixed with this change
   - Some context as to why you uncovered this issue
   - Some assurance that the change doesn't break anything else.

Thanks -- if you can help with some of this info, then we can see about
merging it.

Jugolo:
Okay i will came with a math exempel later :)

Jugolo:
Okay the time is 1482314077000
wee method Day will return: 17156
Year will be 2016

First wee will calculate Day-Year
So
 17156
-2016
=15140

Wee do anthoer calculate
 15140
-59
=15081

And finaly
 15081
/30
=9.8666666666667

So it need to be round up so it is 10

Jugolo:
Hi sorry i ask but any news?

gbrail:
OK -- this looks reasonable and the tests are good. Trying to get back into a few regular Rhino updates.

PR #297, opened 2017-03-21
Support for `arguments` object as TypedArray constructor argument

anton-bannykh:
Before the change the following would cause `org.mozilla.javascript.EcmaError: Error: invalid argument`
```
function foo() {
  return new Int32Array(arguments);
};
```

The proposed change is completely untested and should be considered more of a bug report.


gbrail:
Sorry I have not caught up on all of these.

This looks like a very reasonable change but I'd really like to see a test case.

rbri:
Have created a new PR #457 with the same code change but including an unit test.

PR #299, opened 2017-04-01
Attempt to invoke method on an undefined object

zlaski:
Handle this condition more gracefully, producing a SEVERE warning in the
log and also throwing a proper JavaScript exception.  The behavior is
now consistent with, e.g., invalid arguments to querySelector().

### DIFF
--- a/src/org/mozilla/javascript/ScriptRuntime.java
+++ b/src/org/mozilla/javascript/ScriptRuntime.java
@@ -2474,7 +2474,9 @@ public class ScriptRuntime {
           String property, Context cx, Scriptable thisObj)
     {
         if (thisObj == null) {
-            throw undefCallError(obj, property);
+            throw Context.throwAsScriptRuntimeEx(Context.reportRuntimeError(
+                    "Attempt to invoke method '" + property +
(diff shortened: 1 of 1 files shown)


gbrail:
I'm not sure what this makes better, and it also breaks some tests. If you can explain how this makes Rhino be closer to the JavaScript specs, and include some tests, then we could look at it.

zlaski:
I should have written this up previously, but my proposed patch makes (or made) Rhino's behavior consistent with Chrome's JavaScript engine.  I'm no JS expert and do not know if Chrome is compliant in this respect.  Hence, I'm not sure whether or not I broke JS's intended semantics; if I have, then obviously this patch is inappropriate.

I stumbled upon this during my work (and have since decided to use jsoup), so unfortunately I cannot spend any more time on it.

p-bakker:
Closing this case as it's been a while, its unclear if the change is actually an improvement and it doesn't come with a reproducible case demonstrating the issue its trying to fix

The simple code below doesn't reveal any issues:
```
null.test() // throws TypeError
undefined.test() // throws TypeError
```

PR #319, opened 2017-08-27
some jvmbridge cleanup

rbri:
; jdk 1.3 support is no longer needed, all implementations are protected, some methods removed because of simplification


gbrail:
Thanks for cleaning this up!

rbri:
Anything i can do to get this also merged?
My plan was to do this as a first step and remove some of the methods from the interface. For a next step we can discuss the removal of more methods from the interface (at least for performance reasons this might make sense)

gbrail:
It seems OK to me but it's bigger so I wanted to read it more carefully.

gbrail:
Rebased and merged.

PR #322, opened 2017-09-02
fix static array functions and add test cases

rbri:
Looks like the last refactoring of the array functions had introduces a bug. Hope you like the fix and i did the test stuff right.

### DIFF
--- a/src/org/mozilla/javascript/NativeArray.java
+++ b/src/org/mozilla/javascript/NativeArray.java
@@ -257,6 +257,11 @@ public class NativeArray extends IdScriptableObject implements List
               case ConstructorId_findIndex:
               case ConstructorId_reduce:
               case ConstructorId_reduceRight: {
+                // this is a small trick; we will handle all the ConstructorId_xxx calls
+                // the same way the object calls are processed
+                // so we adjust the args, inverting the id and
+                // restarting the method selection
+                // Attention: the implementations have to be aware of this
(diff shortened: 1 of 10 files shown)


rbri:
Any feedback / any hope to get this merged?

src/org/mozilla/javascript/NativeArray.java
```diff
@@ -1570,7 +1570,7 @@ else if (start < 0)
     private static Object iterativeMethod(Context cx, IdFunctionObject idFunctionObject, Scriptable scope,
                                           Scriptable thisObj, Object[] args)
     {
-        int id = idFunctionObject.methodId();
```
gbrail: I get what you're fixing here.

This case is already convoluted enough and the use of "Math.abs" is not going to be obvious to someone else using the code.

Can you add a comment to the effect that this works because at the very bottom of the file, we use negative numbers to represent the non-prototype versions of these functions and we want them all to work? Otherwise this looks good.

rbri: Ok, you are right. Was not really obvious to me also. Will update this later today.

rbri: Have added some inline comments, hope this makes this a bit more clear.

rbri:
>Merged #322.
>
Thanks Greg, next pull request is on the way

	RBRi

Issue #290, opened 2016-12-18
Hashes for Slots in ScriptableObject should be seeded


botic:
As far as I can interpret the slot allocation in `ScriptableObject` it uses `String.hashCode()` to generate a hash number, which then is used to determine a slot for the value. This makes it easy to exploit with an algorithmically attack.

Take the following example in RingoJS:

```javascript
function loadJSON(file) {
   const start = Date.now();
   const data = require("./" + file);
   console.log("File:", file, "Time: " + (Date.now() - start) + " ms");
}

loadJSON(require("system").args[1]);
```

For a 20 MB huge JSON with randomized keys it takes under a second to load. An attacker could use colliding hash codes since it's easy to generate them with the knowledge that slots are allocated with `String.hashCode()`. Just an example sequence of colliding hashes: `AaAaAaAaAaAaAaAaAaAaAaAaAaAaAaAaAaAa`, `AaAaAaAaAaAaAaAaAaAaAaAaAaAaAaAaAaBB`, `AaAaAaAaAaAaAaAaAaAaAaAaAaAaAaAaBBAa`, ……… Then the example above takes 20 minutes on my machine. And just imagine an application calling `JSON.parse()` without any bounds for the request's body.

To fix this problem Java changed its hash-based data structure's implementation with Java 7: https://vaskoz.wordpress.com/2013/04/06/java-7-hashing-drastically-better-than-java-6/ 

You can find the full code here: https://gist.github.com/botic/d409564dca727eea8086fb6a1cb00623

Further infos / links:

 * [28C3 - Effective DoS on web application platforms](https://events.ccc.de/congress/2011/Fahrplan/attachments/2007_28C3_Effective_DoS_on_web_application_platforms.pdf)
* [V8 / Chromium Patch](https://codereview.chromium.org/9083001/#ps1)
* [V8 CVE-2011-5037](https://www.cvedetails.com/cve/CVE-2011-5037/)
* [V8 Mailing List Discussion](https://groups.google.com/d/topic/nodejs/qZieqGfoZbM/discussion)

davidmann4:
maybe useful lib if you want to find some collisions; (I guess you need to modify the hashing algorithm)
https://github.com/hastebrot/V8-Hash-Collision-Generator

gbrail:
Thanks -- that's an important issue. There seem to be very few people able to work actively on Rhino right now. This looks like a great issue for a new author who might be looking for a way to contribute to a new project.

botic:
[...]

PR #328, opened 2017-09-22
Improve hashing logic in ScriptableObject for collision resistance.

gbrail:
See the commit log and related issue #290 for more information.

This is a big change with big implications, but it seems to be holding up well. V8 benchmarks as well as microbenchmarks are very close to the way that they were before and any differences seem to be generally lost in the noise.


gbrail:
Replacing with a more comprehensive version.

PR #329, opened 2017-09-22
remove methods find and findIndex from the prototoye

rbri:
the methods find and findIndex are not available on the prototoye in real browsers (tested with chrome, ff and ie). Because of this i think we have to remove this methods.
Have added simple test for the method calls on array objects.

### DIFF
--- a/src/org/mozilla/javascript/NativeArray.java
+++ b/src/org/mozilla/javascript/NativeArray.java
@@ -170,10 +170,6 @@ public class NativeArray extends IdScriptableObject implements List
                 "map", 1);
         addIdFunctionProperty(ctor, ARRAY_TAG, ConstructorId_some,
                 "some", 1);
-        addIdFunctionProperty(ctor, ARRAY_TAG, ConstructorId_find,
-                "find", 1);
-        addIdFunctionProperty(ctor, ARRAY_TAG, ConstructorId_findIndex,
-                "findIndex", 1);
         addIdFunctionProperty(ctor, ARRAY_TAG, ConstructorId_reduce,
                 "reduce", 1);
         addIdFunctionProperty(ctor, ARRAY_TAG, ConstructorId_reduceRight,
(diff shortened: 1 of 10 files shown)


gbrail:
For some historical reason (perhaps I was careless in a previous year), "Array" in Rhino has a huge number of functions defined on its constructor. In fact, ECMA script only defines a few: from(), isArray(), and of():

https://www.ecma-international.org/ecma-262/8.0/index.html#sec-properties-of-the-array-constructor

So this would seem to be the right thing to do. The question is whether we should remove some of those other functions, or if we're going to break someone if we do.

rbri:
All the other functions are still available in FF.

rbri:
What should we do here? Do you like to stay with all the constructor functions? If yes i will update this request in a way we only get the additional test cases and the HtmlUnit adaption will remove the two.
I personally think we can safely remove the two functions and mark the change in the release notes; but it is your decisions.
My target is to bring the HtmlUnit Rhino branch a bit closer to the current Rhino. Hope we can benefit from you 'Symbol' work and Rhino can benefit from some fixes / updates we did.

gbrail:
Given that we already shipped this, and that we do it for a lot of Array methods, I think we should leave these methods on the constructor.

rbri:
Ok, will send the changed test as separate pull request.

sainaen:
> For some historical reason (perhaps I was careless in a previous year), "Array" in Rhino has a huge number of functions defined on its constructor.

Just to clarify: this is a non-spec-ed feature that was [introduced back in 2005](https://developer.mozilla.org/en-US/docs/Web/JavaScript/New_in_JavaScript/1.6#New_features_in_JavaScript_1.6) called “generics” or “generic methods”. Both [Array](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array#Array_generic_methods) and [String](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String#String_generic_methods) constructors do this (see the first link for some rationale.) It's still present in Firefox, though String generics show warnings now.

PR #330, opened 2017-09-22
ensureScriptable must supports delegator instances

rbri:


gbrail:
I don't understand this one. Why should we check if these objects are instances of classes from HTMLUnit? It's not something that Rhino depends on today.

rbri:
Sorry, my fault. Will prepare a new one.

rbri:
Hi Greg,
sorry for the stupid commit. Maybe it was a bit too late yesterday ;-)
The fix is for an error we found while using Rhino for HtmlUnit (i guess you are aware that htmlunit uses a slightliy customized version of Rhino). Rhino offers the option to use a Delegator instance instead of a Scriptable(Object). At least at this two changed places Rhino has to take care of Delegator instances also.

src/org/mozilla/javascript/ScriptableObject.java
```diff
@@ -2153,9 +2154,13 @@ protected boolean isGenericDescriptor(ScriptableObject desc) {
     }
 
     protected static Scriptable ensureScriptable(Object arg) {
-        if ( !(arg instanceof Scriptable) )
-            throw ScriptRuntime.typeError1("msg.arg.not.object", ScriptRuntime.typeof(arg));
-        return (Scriptable) arg;
+        if (arg instanceof Scriptable) {
+            return (Scriptable) arg;
+        }
+        if (arg instanceof Delegator) {
```
gbrail: OK -- this makes more sense now. Still, Delegator is a subclass of Scriptable so this particular test will never be "true." If the intention is to get the "delegee" then do we want to have this test first?

rbri:
What a mess. Looks like we did the same stupid error in our HtmlUnit version.Two persons had a look at the code - maybe this was not enough.
Have created a new request with the hopefully correct code; please close this one.
Thanks for your patience ;-)

rbri:
Ok, looks like i can close this also.

PR #332, opened 2017-09-23
remove this constant because SubString is not immutable

rbri:
remove this constant because SubString is not immutable and using/manipulation this instance has side effects
In the past there was already a bunch of fixes because of this. To avoid to get trapped again i think we have to remove this in general. As far as i can see there are no negative side effects for performance and/or memory (we use this patch in HtmlUnit already for a while)


gbrail:
Makes sense to me. Thanks!

PR #334, opened 2017-09-26
Fix a looping parser case

gbrail:
When error recovery is enabled, this prevents an infinite loop in parsing.

### DIFF
--- a/src/org/mozilla/javascript/Parser.java
+++ b/src/org/mozilla/javascript/Parser.java
@@ -3064,9 +3064,6 @@ public class Parser
               pos = ts.tokenBeg; end = ts.tokenEnd;
               return new KeywordLiteral(pos, end - pos, tt);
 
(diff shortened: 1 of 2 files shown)


sainaen:
I've checked a couple of cases that could be affected by this change and everything seems to be ok. 👍 

Also, if I'm reading it correctly, Nashorn's parser does a kind of [similar thing](http://hg.openjdk.java.net/jdk10/jdk10/nashorn/file/3397ed166912/src/jdk.scripting.nashorn/share/classes/jdk/nashorn/internal/parser/Parser.java#l2786): it consumes both parenthesis and then asserts that the next token is `ARROW`. So, at least this way isn't _that_ hack-ish.

gbrail:
Thanks for looking at this! I'm going to merge it since it at least gets us closer to fixing the underlying problem.

PR #335, opened 2017-09-26
ensureScriptable(..) and ensureScriptableObject(..) have to take care of Delegator instances

rbri:
Next try hope this is correct now.

### DIFF
--- a/src/org/mozilla/javascript/FunctionObject.java
+++ b/src/org/mozilla/javascript/FunctionObject.java
@@ -398,6 +398,9 @@ public class FunctionObject extends BaseFunction
         } else {
             if (!isStatic) {
                 Class<?> clazz = member.getDeclaringClass();
+                if (thisObj instanceof Delegator) {
+                    thisObj = ((Delegator) thisObj).getDelegee();
+                }
                 if (!clazz.isInstance(thisObj)) {
                     boolean compatible = false;
                     if (thisObj == scope) {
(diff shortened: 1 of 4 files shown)


gbrail:
I'm not super excited about the way that this is going. It seems that we'd have to end up calling "getDelegee" every place in the codebase where we might have a Delegate, and then that kind of defeats the purpose of this feature. I thought that the idea of Delegate was just to have some way to put another Scriptable in front of another, and in turn that would solve a large number of use cases.

Perhaps with some examples of what you're trying to do we can figure out how to best solve the problem.

rbri:
He he, making you super excited was not the idea of this pull request.
And yes i understand your point. While looking at the source a found many more places where i have the feeling there is a need to handle this Delegator instances also.

But: All changes in this pull request are changes we did already in the HtmlUnit for; all this changes where done on demand - means we had some problem at this special places.
As far as is remember we use this Delegator instance only for a single class. We made several attempts in past to get rid of this but there was some good reasons to stay with the delegator. I will not start looking the whole code and add Delegator instance checks all over. But it will be great if we can add this patch because we already know that we have to handle the Delegators at this places. There is some risk that we will find more in the future but this is a different story.

gbrail:
OK -- just looking at the benchmarks since this also introduces a bunch more instanceof checks on every function invocation.

My other concern is that we don't have a test for this. I think we really need a test case that failed before and passes with this change. Otherwise there is nothing stopping future developers from breaking this again.

p-bakker:
Not exactly sure how the Delegator class would be used, but it looks similar to JavaScript Proxies

Wondering if Proxies could replace Delegator or help in any way, by maybe Delegator extending Proxy

Just thinking out loud

rbri:
Yes the Delegator is some kind of a proxy but i fear this is in use not only by HtmlUnit.
Will try to find/create a test case for this.

rbri:
Will close this - because i have prepared a new one (less changes and including tests)

PR #336, opened 2017-09-26
simple tests for the method calls on array objects added

rbri:


gbrail:
More testing makes sense to me. Thanks!

PR #337, opened 2017-09-26
replace tab with 4 blanks

rbri:
there are a lot of code file with mixed tab/blank. At least for the code we should use always blanks like the majority of files does


gbrail:
This project has a very old and diverse codebase. While it'd be neat to impose a coding standard on all the code that's been contributed over the years, it's a little late to do anything other than try and make sure that new code follows a more modern standard.

I don't think that we should be proactively changing formatting in existing code, especially since at any time there are several PRs in flight and big reformats make it harder to keep them all straight.

rbri:
Ok, your decision ;-). No problem with that from my side. Will create a new pull request for minor cleanup if i find something.

PR #344, opened 2017-10-04
Support collision-resistant objects; make thread-safety configurable

gbrail:
Replace the hand-rolled hash code in ScriptableObject with code that works the same way, but falls back to java.util.HashMap when an object has many properties.

Make synchronization of object properties configurable via a feature flag in Context. By default objects are not thread-safe. If the feature flag is set, then they are.


gbrail:
This PR is a pretty big one, in that it changes the way that object properties are stored in a pretty fundamental way. I'd like to merge it soon, as I've tested it quite a bit. It'd be great if a few people out in the community could look at this code as well.

Issue #348, opened 2017-10-16
getOwnPropertyDescriptor of Java wrapped objects getter/setter functions can not be invoked using "call"/"apply"


tmallery:
I'm getting an error when I try to "call" the get/set properties of the result of Object.getOwnPropertyDescriptor() when a wrapped Java object is the target.

org.mozilla.v17.javascript.EvaluatorException: Can't find method org.mozilla.v17.javascript.Callable.call(object,string). (exp#69)
	at org.mozilla.v17.javascript.DefaultErrorReporter.runtimeError(DefaultErrorReporter.java:77)
	at org.mozilla.v17.javascript.Context.reportRuntimeError(Context.java:971)
	at org.mozilla.v17.javascript.Context.reportRuntimeError(Context.java:1026)
	at org.mozilla.v17.javascript.Context.reportRuntimeError1(Context.java:989)
	at org.mozilla.v17.javascript.NativeJavaMethod.call(NativeJavaMethod.java:145)

In Chromes JS engine, the getter/setter properties of the object returned from Object.getOwnPropertyDescriptor() are functions which can then be executed using "call" and "apply". This is not true for Rhino because the get and set functions are MemberBox objects and not FunctionObjects.

p-bakker:
@tmallery am closing this as I assume https://github.com/mozilla/rhino/pull/353 fixed it. If not, please reopen

PR #353, opened 2017-10-29
384 - member box call error

tmallery:
This fixes both the issue of not being able to call "call" on the returned value from getOwnPropertyDescriptor().get, it also stops the JS conversion error message from showing up when the get/set properties are referenced in a boolean condition. #348

### DIFF
--- a/src/org/mozilla/javascript/ScriptableObject.java
+++ b/src/org/mozilla/javascript/ScriptableObject.java
@@ -253,17 +253,24 @@ public abstract class ScriptableObject implements Scriptable,
             if (getter == null && setter == null) {
                 desc.defineProperty("writable", (attr & READONLY) == 0, EMPTY);
             }
-            if (getter != null)
-			{
-				if( getter instanceof MemberBox ) desc.defineProperty("get", new FunctionObject("f", ((MemberBox)getter).member(),scope), EMPTY);
-				else if( getter instanceof Member ) desc.defineProperty("get", new FunctionObject("f",(Member)getter,scope), EMPTY);
-				else desc.defineProperty("get", getter, EMPTY);
-			}
-			if (setter != null) {
-				if( setter instanceof MemberBox ) desc.defineProperty("set", new FunctionObject("f", ((MemberBox)setter).member(),scope), EMPTY);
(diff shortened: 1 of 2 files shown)


src/org/mozilla/javascript/ScriptableObject.java
```diff
@@ -253,8 +253,17 @@ ScriptableObject getPropertyDescriptor(Context cx, Scriptable scope) {
             if (getter == null && setter == null) {
                 desc.defineProperty("writable", (attr & READONLY) == 0, EMPTY);
             }
-            if (getter != null) desc.defineProperty("get", getter, EMPTY);
-            if (setter != null) desc.defineProperty("set", setter, EMPTY);
+            if (getter != null)
+			{
+				if( getter instanceof MemberBox ) desc.defineProperty("get", new FunctionObject("f", ((MemberBox)getter).member(),scope), EMPTY);
```
gbrail: Sadly there is not a single code style for this huge, old code base, but generally we follow the style of the code "around us" when making a change. You can see that this change is pretty different from other Rhino code. Specifically:

1) Please use spaces and not tabs
2) Put the spaces on the other side of the parentheses (like you see elsewhere in the file)
3) It's not a hard and fast rule yet, but using brackets for every "if" and "else" statement, even if there is only a single statement, avoids lots of errors later.

(I also think this is why the Travis CI build is failing for this PR.)

testsrc/org/mozilla/javascript/tests/MemberBoxCallTest.java
```diff
@@ -0,0 +1,108 @@
...
+
+public class MemberBoxCallTest {
+
+    Scriptable scope;
+
+    @Test
+    public void testPrototypeProperty() {
+		Context cx = Context.enter();
+		try {
+			assertEquals(evaluate(cx,
```
gbrail: I see a few of properties and functions on "AnnotatedHostObject," but it looks like this test is only checking for "foo." Is there a reason why we can't test the other properties? Similarly, are you sure that all the code paths in ScriptableObject are being exercised here for both Member and MemberBox?

gbrail:
Thanks! I'm just going to squash this myself and merge it.

PR #354, opened 2017-10-30
Add a setting to CompilerEnvirons to allow the 'use strict' directive…

peterj99a:
… to be ignored.

The directive implementation changed between 1.7.1 and 1.7.2, for backward compatibility the user may want to disable it.

### DIFF
--- a/src/org/mozilla/javascript/CompilerEnvirons.java
+++ b/src/org/mozilla/javascript/CompilerEnvirons.java
@@ -23,6 +23,7 @@ public class CompilerEnvirons
         optimizationLevel = 0;
         generatingSource = true;
         strictMode = false;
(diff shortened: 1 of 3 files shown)


gbrail:
We also talked off-line about addressing the thing that changed in 1.7.7.2 (supporting use strict inside a function body) and making that dependent on the language level, so that it won't break compatibility for anyone on an older language level, as it does now. Do you think that fixing it that way would be more appropriate?

Also, we often fix these sorts of things by adding a feature flag to Context. If we do want to support this feature, do you think that makes more sense?

p-bakker:
@gbrail as this PR is 7 years old, is related to a backwards compatibility issue in quite an old Rhino version and there's been no activity on this, shall we close it?

Issue #355, opened 2017-10-30
Function.prototype.apply fails with "Array like" second argument


tmallery:
TypeError: second argument to Function.prototype.apply must be an array

With this test code:
function test() { numArgs = arguments.length; }
test.apply(this, {'length':2, '0':'turkey', '1':'gravey'});

Per Mozillas MDN this should be allowed: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/apply

p-bakker:
Seems long fixed

PR #358, opened 2017-11-01
355 function prototype apply argument array like

tmallery:
#355
This fix matches ECMA5 behavior where an object passed in to apply as the second argument allows the call to proceed without exception.

### DIFF
--- a/src/org/mozilla/javascript/ScriptRuntime.java
+++ b/src/org/mozilla/javascript/ScriptRuntime.java
@@ -2696,13 +2696,27 @@ public class ScriptRuntime {
 
         return function.call(cx, scope, callThis, callArgs);
     }
+    
+    /**
+      * @return true if the passed in Scriptable looks like an array
+      */
+    private static boolean isArrayLike(Scriptable obj)
+    {
+        return obj != null && (
+            obj instanceof NativeArray ||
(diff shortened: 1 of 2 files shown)


src/org/mozilla/javascript/ScriptRuntime.java
```diff
@@ -2696,13 +2696,27 @@ public static Object applyOrCall(boolean isApply,
...
+    
+    /**
+      * @return true if the passed in Scriptable looks like an array
+      */
+    private static boolean isArrayLike(Scriptable obj)
+    {
+        return obj != null && (
+            obj instanceof NativeArray ||
+            obj instanceof Arguments ||
+            ScriptableObject.getProperty(obj, "length") != Scriptable.NOT_FOUND
```
gbrail: Can we use "hasProperty" here instead? It could actually make a performance difference.

gbrail:
Thanks! That looks great. Now can you please squash this into one commit (right now there are 10)? Then I can take it from there and rebase and merge or whatever we need.

tmallery:
Thanks for the review and I believe I've finally squashed the changes into one commit. Local 'check' checks out.

gbrail:
Thanks -- looks good!

PR #360, opened 2017-11-10
Make "hybrid" sort algorithm more robust.

gbrail:
Add bounds checking so that we don't make so many assumptions about
how the comparator behaves.
Make the "median of three" algorithm actually take the median
of three.

### DIFF
--- a/src/org/mozilla/javascript/Sorting.java
+++ b/src/org/mozilla/javascript/Sorting.java
@@ -98,10 +98,10 @@ public final class Sorting {
     }
 
     /*
-    Return the index of the smallest of three elements in the specified array range -- the
+    Return the index of the median of three elements in the specified array range -- the
     first, the last, and the one in the middle.
     */
-    private static int median(final Object[] a, int start, int end,  Comparator<Object> cmp)
+    public static int median(final Object[] a, int start, int end, Comparator<Object> cmp)
     {
         final int m = start + ((end - start) / 2);
         int smallest = start;
--- a/src/org/mozilla/javascript/Sorting.java
+++ b/src/org/mozilla/javascript/Sorting.java
@@ -1,6 +1,5 @@
 package org.mozilla.javascript;
 
-import java.util.Arrays;
 import java.util.Comparator;
 
 public final class Sorting {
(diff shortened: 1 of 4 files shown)


src/org/mozilla/javascript/Sorting.java
```diff
@@ -85,10 +97,28 @@ private static int log2(int n)
         return (int)(Math.log10(n) / Math.log10(2.0));
     }
 
-    private static int median(int n1, int n2, int n3)
+    /*
+    Return the index of the smallest of three elements in the specified array range -- the
+    first, the last, and the one in the middle.
```
sainaen: Is this correct? From my reading `median()` returns the index of the median of three elements (start, end and in the middle), not the smallest in range.

testsrc/jstests/sorting-comparators.js
```diff
@@ -0,0 +1,99 @@
...
+    if (x < y) {
+      return 1;
+    }
+    if (x > y) {
+      return -1;
+    }
+    return 0;
+  });
+
+// Torture tests devised by @sainen and @bensummers
```
sainaen: Thanks, but I'd say that attribution isn't in necessary/useful here. It's just a test, plus comparator itself comes from underscore. (And it's sain*a*en. 🙂 )

testsrc/jstests/sorting-comparators.js
```diff
@@ -0,0 +1,99 @@
...
+      .map(function(value, index) { return {index: index, criteria: value}; }),
+      function(left, right) {
+          var a = left.criteria;
+          var b = right.criteria;
+          if (a !== b) {
+              if (a > b || a === void 0) return 1;
+              if (a < b || b === void 0) return -1;
+          }
+          return left.index < right.index ? -1 : 1;
+      });
```
sainaen: Since both these ‘torture’ cases are basically the same, how about replacing one of them with something different that also throws currently?

Like, 
```js
// should work ok, comparator is similar to the underscore's one
sortAndCompare([1, 1, 1, 1, 1, 1, 1, 1, 1, 0, 1, 1, 1, 1, 1, 1, 1, 1, 1], function (a, b) { return a >= b ? 1 : -1; });
// weird comparator, array may not be sorted, but shouldn't fail/throw
assertDoesNotThrow(function () {
  [1, 1, 1, 1, 1, 1, 1, 1, 1, 0, 1, 1, 1, 1, 1, 1, 1, 1, 1].sort(function (a, b) { return -1; });
});
// yet another weird comparator, again shouldn't fail/throw
assertDoesNotThrow(function () {
  [1, 1, 1, 1, 1, 1, 1, 1, 1, 0, 1, 1, 1, 1, 1, 1, 1, 1, 1].sort(function (a, b) { return 1; });
});
```

gbrail:
Thanks for the review and suggestions! I kind of think that if someone sends us code and they're not going to be part of the Git commit then I should credit them ;-) I'll merge this soon.

PR #371, opened 2017-12-22
Improve UX around Test262

sainaen:
This addresses several small annoyances with how working with test262 cases goes:

1. adds an explanation of the format to the file itself, since it's not very obvious
1. allows leading and trailing whitespace for all the lines
1. `Test262SuiteTest` now throws exception with an informative error message when path from `test262.properties` doesn't exist, before it was just ignored and a couple of paths were wrong
1. `Test262SuiteTest` now prints out warning to the console when an exclusion pattern didn't match anything
1. streamlines code around the `test262.properties` parsing (less duplication) and adds comments, I hope it's now easier to follow

As a bonus, fixing the third point brought the number of test262 cases from 25752 to 26229 (+477) 🎉

### DIFF
--- a/testsrc/org/mozilla/javascript/tests/Test262SuiteTest.java
+++ b/testsrc/org/mozilla/javascript/tests/Test262SuiteTest.java
@@ -19,12 +19,12 @@ import org.mozilla.javascript.drivers.TestUtils;
 import org.mozilla.javascript.tools.SourceReader;
 import org.mozilla.javascript.tools.shell.ShellContextFactory;
 import org.yaml.snakeyaml.Yaml;
+import org.yaml.snakeyaml.error.YAMLException;
 
 import java.io.*;
 import java.util.*;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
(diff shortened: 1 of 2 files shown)


gbrail:
OK! I was just about to push a "1.7.8-RC" before going out for a little while so that we can do a proper 1.7.8 release in the new year. However, this would be an easy thing to include. I'm going to "squash" it and merge it!

sainaen:
Cool! Thanks for the quick response.

Next thing I'd like to work on is bringing up to date test262 executor as there were changes in the format of the file headers since it was first introduced here. It's still only a test update, so shouldn't interfere with the release process, but I could wait until 1.7.8 is ready.

By the way, do you want me to squash commits in my future PRs?

gbrail:
No problem. I only get to work on Rhino in short bursts these days but
today's one of those days.

In general, we don't have a hard and fast rule about squashing commits, but
in general if I see a lot of little commits I think it's easier on future
maintainers if we squash them, yes.

On Fri, Dec 22, 2017 at 12:04 PM, Ivan Vyshnevskyi <notifications@github.com

Issue #377, opened 2018-01-05
1.7.8 RC1: Dynamic Scopes vs. initStandardObjects (NativeSymbol) 


raimi:
As discussed in https://groups.google.com/forum/#!topic/mozilla-rhino/ez3kr8-o0oI : there is an issue in my specific way of using dynamic scopes initStandardObjects, namely NativeSymbol.

I prepared a testcase.

raimi:
To be clear: passing sealed = false to initStandardObjects(ScriptableObject scope, boolean sealed) makes the problem go away.

gbrail:
Here's a PR for this with a test case:

https://github.com/mozilla/rhino/pull/381

Can you try it out? I'd like to merge it before we do Rhino 1.7.8 later this week.

raimi:
Good news: This fixes both the test-case and our application. Perhaps you can accept both my and your PR? Doesn't hurt to have more tests :)

gbrail:
Fixed this one in https://github.com/mozilla/rhino/pull/378.

PR #378, opened 2018-01-05
https://github.com/mozilla/rhino/issues/377 now OK test case

raimi:
This reproduces my problem and also suggests a way to the solution: cx.initStandardObjects(new TopLevel(), false) WOULD work.

So the issue is the "isSealed" bit being set to early for NativeSymbol.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ Rhino builds with `Gradle`. Here are some useful tasks:
 ```
 ./gradlew jar
 ```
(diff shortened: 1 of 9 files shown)


gbrail:
Sorry -- missed this. I'm going to replace my test with this one.

raimi:
Merged Greg's changes and his testcase. Cleaned up my testcase. Hope he just accepts this change...

gbrail:
OK -- looks good thanks!

PR #381, opened 2018-01-16
Fix error when creating context in "sealed" mode at ES6 level.

gbrail:
Fix Symbol initialization so that "sealed" initialization works with ES6 language level.


gbrail:
Fixed this one in a different PR.

PR #383, opened 2018-01-19
Make ToNumber(String) conversion more spec-compliant

### LINKED ISSUE #368: Unexpected result with isNaN method
I'm using Mozilla Rhino engine version 1.7.7.2 with OpenRASP project.

The following statement returns expected result `false`

```
isNaN('0x7f.0x0.0x0.0x1')
```

At lease in Chrome console, it returns `true`

Can you verify that?

sainaen:
Update ToNumber conversion applied to the String type:
* plus (+) or minus (-) signs are not allowed before the hex literals,
  return NaN for them
* also return NaN for HexIntegralLiteral followed by anything other than
  whitespace (in contrast to how 'parseInt()' works)
* add support for binary ('0b101010') and octal ('0o52') literals

New behavior is applied only if the language version is >=ES6, so it
shouldn’t brake any existing code.

This also enables built-ins/Number test262 cases.

-----
Fixes #368.

I’ve also verified that the new `ToNumberLegacyConversionsTest` works in the master branch without this patch, so it in fact represents the ToNumber's current behavior.

### DIFF
--- a/src/org/mozilla/javascript/NativeGlobal.java
+++ b/src/org/mozilla/javascript/NativeGlobal.java
@@ -249,7 +249,7 @@ public class NativeGlobal implements Serializable, IdFunctionCall
             }
         }
 
-        double d = ScriptRuntime.stringToNumber(s, start, radix);
+        double d = ScriptRuntime.stringPrefixToNumber(s, start, radix);
         return ScriptRuntime.wrapNumber(negative ? -d : d);
     }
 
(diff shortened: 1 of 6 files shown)


sainaen:
I have one question related to this.

I decided to not change the tests that cover some non-spec’ed functionality in MozSuite, which leaves it on the language version < ES6. Is it ok?
I think in SpiderMonkey they slowly remove parts of their own tests that are covered by test262, maybe that's something we should consider too?

sainaen:
Thanks for merging!

> I do think it's safest to leave this to ">=ES6" for now since backward compatibility can show up at the worst time!

Sure, let’s leave it for now.

It’s just that I’m a bit concerned it may become harder to update in future if we drift away too much. Maybe create an issue to keep track of it?

PR #389, opened 2018-02-10
Report parsing error for default values in destructuring assignments

### LINKED ISSUE #385: IllegalStateException
When evaluating the attached file , an `IllegalStateException` is thrown.
[rhino_ise.zip](https://github.com/mozilla/rhino/files/1691899/rhino_ise.zip)
Rhino version: `1.7.7.2`
Kotlin code to reproduce the issue if `path` points to the attached file:
```
File(path).reader().use {
            val context = Context.enter()
            val scope = context.initStandardObjects()
            context.evaluateReader(scope, it, path, 1, null)
        }
```

sainaen:
Closes #385. (See my comments there for even more context.)

### DIFF
--- a/src/org/mozilla/javascript/Parser.java
+++ b/src/org/mozilla/javascript/Parser.java
@@ -2253,6 +2253,13 @@ public class Parser
             tt = peekToken();
         }
         if (Token.FIRST_ASSIGN <= tt && tt <= Token.LAST_ASSIGN) {
+            if (inDestructuringAssignment) {
+                // default values inside destructuring assignments,
+                // like 'var [a = 10] = b' or 'var {a: b = 10} = c',
+                // are not supported
+                reportError("msg.destruct.default.vals");
+            }
+
             consumeToken();
 
             // Pull out JSDoc info and reset it before recursing.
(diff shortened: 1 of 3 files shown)


sainaen:
Travis-CI check failed, because
```
Global vars should not be more than 2.5 times slower than local, Ratio: 2.8160377358490565
```
but I think it’s an intermittent perf problem on the CI servers. The same test passes locally.

sainaen:
Here we go again
```
In "testsrc/tests/ecma_3/RegExp/regress-289669.js":
  FAILED! O(N^2) behavior on String.replace(/RegExp/, ...) BigO 2 < 2 : Expected value 'true', Actual value 'false'
```
😞 

Something weird is going on at Travis-CI. I’ll force rebuild tomorrow morning (UTC+0).

sainaen:
OK, tests did pass in two last runs on Travis CI ([315](https://travis-ci.org/mozilla/rhino/builds/340103135), [319](https://travis-ci.org/mozilla/rhino/builds/341512668)).

gbrail:
Those tests are flaky. I think that they made sense back in the old days
but on a modern CPU and a modern JVM I think that those kinds of benchmarks
are not helpful. I'm happy to disable more of those 'BigO' tests in favor
of actual benchmarks like the ones we currently use (and new ones!)

gbrail:
Looks good -- thanks!

PR #392, opened 2018-02-18
Add addError(String messageId, int c) method

sideshowbarker:
This change adds an `addError(String messageId, int c)` method. The
intent of it is for cases like the “illegal character” error, to cite
the offending character in the error message.

So this change also updates the `illegal.character` message property to
add a message argument to it, and updates the `illegal.character` error
call sites to use the `addError(String messageId, int c)` method.

Thus the method expects to be called with a character value for the `c`
int arg, and makes a string from it by first casting that int to a char.

### DIFF
--- a/src/org/mozilla/javascript/Parser.java
+++ b/src/org/mozilla/javascript/Parser.java
@@ -237,6 +237,12 @@ public class Parser
                  ts.tokenEnd - ts.tokenBeg);
     }
 
+    void addError(String messageId, int c) {
+        String messageArg = Character.toString ((char) c);
+        addError(messageId, messageArg, ts.tokenBeg,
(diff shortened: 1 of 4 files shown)


gbrail:
I would love to see more tests for the parser. (There are some in ParserTest.java.) Would it be possible to add a simple test case for this?

sideshowbarker:
> Would it be possible to add a simple test case for this?

Yup — will try to write one up asap and get it added on this PR branch

sideshowbarker:
>  Would it be possible to add a simple test case for this?

Added in 715eaaa and passing when running `./gradlew -Dtest.single=ParserTest test`

gbrail:
Thanks for doing this!

PR #393, opened 2018-02-18
Show word in “identifier is a reserved word” error

sideshowbarker:
This change updates the `reserved.id` message property to add a message
argument for indicating the offending reserved word. The change also
updates the `reserved.id` error call site to add the message argument to
the call (with the serialization of the current token as the value).

### DIFF
--- a/src/org/mozilla/javascript/Parser.java
+++ b/src/org/mozilla/javascript/Parser.java
@@ -3084,7 +3084,7 @@ public class Parser
 
           case Token.RESERVED:
               consumeToken();
-              reportError("msg.reserved.id");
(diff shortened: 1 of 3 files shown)


gbrail:
Same as the last question ... would it be possible to add a test for this, perhaps in ParserTest.java?

sideshowbarker:
> Same as the last question ... would it be possible to add a test for this, perhaps in ParserTest.java?

Yup, will do soon

sideshowbarker:
> would it be possible to add a test for this, perhaps in ParserTest.java?

Added in 1154942

gbrail:
Thanks!

I rebased and manually merged this into master.

Issue #386, opened 2018-02-03
NullPointerException


havrikov:
When evaluating the attached file , a `NullPointerException` is thrown.
[rhino_npe.zip](https://github.com/mozilla/rhino/files/1691901/rhino_npe.zip)
Rhino version: `1.7.7.2`
Kotlin code to reproduce the issue if `path` points to the attached file:
```
File(path).reader().use {
            val context = Context.enter()
            val scope = context.initStandardObjects()
            context.evaluateReader(scope, it, path, 1, null)
        }
```

sainaen:
Thanks for reporting this one too!

Here’s simplified reproducer:
```js
const [y,y] = [];
```

`IRFactory` uses `Parser`'s [`createDestructuringAssignment()`](https://github.com/mozilla/rhino/blob/master/src/org/mozilla/javascript/Parser.java#L3887-L3907) to transform destructuring assignment into a series of regular assignments. When it realizes that we’re trying to declare constant `y` twice, it tries to report an error, but uses normal [`addError()`](https://github.com/mozilla/rhino/blob/master/src/org/mozilla/javascript/Parser.java#L235) that expects `TokenStream` to be present (`ts`) to construct an error message with position of the offending element.

I see that [`addWarning()`](https://github.com/mozilla/rhino/blob/master/src/org/mozilla/javascript/Parser.java#L200) and [`addStrictWarning()`](https://github.com/mozilla/rhino/blob/master/src/org/mozilla/javascript/Parser.java#L185), for example, check if `ts` is `null` and set token’s start position to `-1` and length to `0` when it is. The fix could be as simple as adding the same check to `addError()`, if we don’t care about printing an exact position of the re-declaration in the error message.

sainaen:
@havrikov 
I wonder, did you use some fuzzer to find #385 and this one? The code looks autogenerated.

havrikov:
Yes, I am currently exploring grammar-based fuzzing.

sainaen:
Oh, sounds awesome! Test case reduction still requires work though. 😉

Bamagik:
Just submitted a PR, seems to have fixed the issue, however, I am unsure whether it outputs the expected result as desired.

p-bakker:
The nullPointer error reported here got fixed in https://github.com/mozilla/rhino/pull/987

Created a followup case (#989), as the error thrown now for `const [y,y] = []` is a TypeError, while it should be a SyntaxError

PR #394, opened 2018-02-21
Added checking in Error function in Parser

Bamagik:
In cases of TokenStream not being defined. Like in issue#386 https://github.com/mozilla/rhino/issues/386

### DIFF
--- a/src/org/mozilla/javascript/NativeDate.java
+++ b/src/org/mozilla/javascript/NativeDate.java
@@ -438,7 +438,6 @@ final class NativeDate extends IdScriptableObject
 
         double y = Math.floor(t / (msPerDay * 365.2425)) + 1970;
         double t2 = TimeFromYear(y);
-
         /*
          * Adjust the year if the approximation was wrong.  Since the year was
          * computed using the average number of ms per year, it will usually
--- a/src/org/mozilla/javascript/Parser.java
+++ b/src/org/mozilla/javascript/Parser.java
@@ -233,8 +233,12 @@ public class Parser
     }
 
     void addError(String messageId, String messageArg) {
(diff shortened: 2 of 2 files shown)


gbrail:
Thanks! 

I'd love to see a test! 

I know that we don't have that many tests for the parser, but it seems that people are using Rhino to parse and help interpret JavaScript in lots of different projects and improving the parsers's quality via better code coverage would be a big help.

src/org/mozilla/javascript/NativeDate.java
```diff
@@ -438,7 +438,6 @@ private static int YearFromTime(double t)
 
         double y = Math.floor(t / (msPerDay * 365.2425)) + 1970;
         double t2 = TimeFromYear(y);
-
```
szegedi: This is irrelevant and should be removed.

szegedi: This is irrelevant and should be removed.

src/org/mozilla/javascript/Parser.java
```diff
@@ -233,8 +233,12 @@ void addError(String messageId, int position, int length) {
     }
 
     void addError(String messageId, String messageArg) {
-        addError(messageId, messageArg, ts.tokenBeg,
-                 ts.tokenEnd - ts.tokenBeg);
+        int beg = -1, end = -1;
```
szegedi: I'd really like us to adopt a single-assignment style for local variables in Rhino whenever possible, so
```java
final int beg, end;
if (ts != null) {
    ...
} else {
    beg = end = -1;
}
```
would be much preferred.

p-bakker:
The issue this PR fixes is already covered by #987, so closing this PR

Issue #390, opened 2018-02-14
ArrayIndexOutOfBoundsException - EmbeddedSlotMap?


tmallery:
I'm seeing an ArrayIndexOutOfBoundsException in ScriptableObject.getIds(boolean, boolean) because the EmbeddedSlotMap iterator is returning one more result than its count, "map.size()", is returning.
Has anyone else seen this? I haven't found the cause yet so I can't make a test case, I only see the error after a lot of processing has happened so I don't yet know the sequence of events that causes this problem. However I tried removing EmbeddedSlotMap by having SlopMapContainer just make HashSlotMap and that "fixed" this issue.

This also might be caused by a fix I'm currently doing to allow Object.getOwnPropertyDescriptor("z", 0) and "z".propertyIsEnumerable(0) to work. I'll provide more information as I find it.

gbrail:
Can you share the stack trace? I hope we didn't miss an edge case --
certainly across lots of product usage and tests I haven't -- but anything
is possible and a stack trace would help a lot!

tmallery:
ArrayIndexOutOfBoundsException: 231 (evaluateExpression_39102403#1)
org.mozilla.v17.javascript.WrappedException: Wrapped java.lang.ArrayIndexOutOfBoundsException: 231 (evaluateExpression_39102403#1)
	at org.mozilla.v17.javascript.Context.throwAsScriptRuntimeEx(Context.java:1924)
	at org.mozilla.v17.javascript.MemberBox.invoke(MemberBox.java:148)
	at org.mozilla.v17.javascript.FunctionObject.call(FunctionObject.java:452)
	at org.mozilla.v17.javascript.Interpreter.interpretLoop(Interpreter.java:1492)
	at script(evaluateExpression_39102403:1)
	at org.mozilla.v17.javascript.Interpreter.interpret(Interpreter.java:823)
	at org.mozilla.v17.javascript.InterpretedFunction.call(InterpretedFunction.java:109)
	at org.mozilla.v17.javascript.ContextFactory.doTopCall(ContextFactory.java:402)
	at org.mozilla.v17.javascript.ScriptRuntime.doTopCall(ScriptRuntime.java:3514)
	at org.mozilla.v17.javascript.InterpretedFunction.exec(InterpretedFunction.java:120)
	at org.mozilla.v17.javascript.Context.evaluateString(Context.java:1254)
Caused by: java.lang.ArrayIndexOutOfBoundsException: 231
[...]

PR #398, opened 2018-02-27
Changes to process integer object as integer and long as long, not as…

alexandrmaximenko:
Here is the problem: rhino is used inside handlebars library. When handlebars have to process js, it calls rhino for help. Assume there is a simple piece of js code: "data.lenght". When handlebars ask rhino to process it, it is expected a number without a decimal place (for instance 4) to be returned. In fact 4.0 is returned.
Small changes in this pull request change this behavior.

### DIFF
--- a/src/org/mozilla/javascript/Context.java
+++ b/src/org/mozilla/javascript/Context.java
@@ -327,6 +327,14 @@ public class Context
      */
     public static final int FEATURE_THREAD_SAFE_OBJECTS = 17;
 
+    /**
+     * If set, then all integer numbers will be returned without decimal place. For instance
+     * assume there is a function like this:
+     * <code>function foo() {return 5;}</code>
+     * 5 will be returned if feature is set, 5.0 otherwise.
+     */
+    public static final int FEATURE_INTEGER_WITHOUT_DECIMAL_PLACE = 18;
+
     public static final String languageVersionProperty = "language version";
     public static final String errorReporterProperty   = "error reporter";
 
--- a/src/org/mozilla/javascript/ContextFactory.java
+++ b/src/org/mozilla/javascript/ContextFactory.java
@@ -291,6 +291,9 @@ public class ContextFactory
 
           case Context.FEATURE_THREAD_SAFE_OBJECTS:
               return false;
+
+          case Context.FEATURE_INTEGER_WITHOUT_DECIMAL_PLACE:
+              return false;
         }
         // It is a bug to call the method with unknown featureIndex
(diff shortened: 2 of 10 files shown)


gbrail:
This seems like it's going to break someone, somewhere, who is expecting those properties to return numbers with decimal points in them. If this is important then we should consider whether it should be enabled by a feature flag.

Also, do you have an example of a bit of JavaScript that doesn't work due to the existing behavior?

alexandrmaximenko:
It would be great to enable this on demand by feature flag. What is the way to put it under toggle?

Using handlebars I'm adding some js helper, something like this:

Handlebars.registerHelper('bannerLength', function(bannerTotal, options) {
   return bannerTotal.length;
});

and to process it Rhino is used inside handlebars. It returns number with decimal place.

gbrail:
Yes, perhaps others don't agree, but I feel that this change has a real chance to break existing code, and to even cause a performance regression. At the same time, I understand that it's important to your project and others may feel the same.

The way to make something optional in Rhino is to use the Context class, and by following the pattern used there.

For instance, you would:

1) Add a constant to Context.java like "FEATURE_CONVERT_NUMBER_TO_INTEGER" with comments that describes it.
2) Add a case in ContextFactory.java to set the default value (which I think should be false)
3) Use it in your code by calling Context.hasFeature when you want to select the behavior. (And if you don't have a Context object in scope, you can call Context.getCurrentContext to get it.)

And again, you can follow the example of existing flags...

And once you do that, I'd like to also see some tests specifically for this behavior that test the behavior with the flag both on and off.

Thanks!

alexandrmaximenko:
Thank you for clear explanation, will update pull request soon.

gbrail:
Thanks for doing this -- it looks good, and if others find it useful or doesn't break their code then we can always look at making it the default.

alexandrmaximenko:
Thank you for your time.
I'd like to ask a question - what is the approximate time for next release (where this functionality can be used)?

gbrail:
There's no formal schedule -- it has to do with need.

It turns out that I'd like to get the release out anyway to some of my own
customers, so it might happen soon, because of:

https://github.com/mozilla/rhino/issues/390

Do you have a deadline you're working to?

In general, releases don't take long. However, there is some upstream
effect as they end up in YUM repos, Homebrew, etc.

On Tue, Mar 13, 2018 at 6:38 AM, alexandrmaximenko <notifications@github.com

alexandrmaximenko:
It really didn't take long and exceeded our expectations, thank you very much!

PR #416, opened 2018-04-05
Missing properties are not enumerable when checking enumerability

### LINKED ISSUE #415: Property 0 not found. for Object('z').propertyIsEnumerable(0); Since 1.7.7.2
Since 1.7.7.2 I believe this commit https://github.com/mozilla/rhino/commit/6f0f1f669ed3da8a476af736c97ab41ba32873cc
the following code is throwing a `Property X not found.` error.

```js
Object('z').propertyIsEnumerable(0);
```

Happens to be used in Babel/Core-js here: https://github.com/zloirock/core-js/blob/master/modules/_iobject.js

whitlockjc:
When checking property enumerability _(`Object.prototype.propertyIsEnumerable`)_ on a `String`, missing properties should return `false` instead of throwing a `Property {0} not found.` exception.

Fixes #415 _(regression in `1.7.7.2`)_

### DIFF
--- a/src/org/mozilla/javascript/NativeObject.java
+++ b/src/org/mozilla/javascript/NativeObject.java
@@ -181,20 +181,31 @@ public class NativeObject extends IdScriptableObject implements Map
                 }
             } else {
                 String s = ScriptRuntime.toStringIdOrIndex(cx, arg);
-                if (s == null) {
-                    int index = ScriptRuntime.lastIndexResult(cx);
-                    result = thisObj.has(index, thisObj);
-                    if (result && thisObj instanceof ScriptableObject) {
-                        ScriptableObject so = (ScriptableObject) thisObj;
-                        int attrs = so.getAttributes(index);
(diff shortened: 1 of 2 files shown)


gbrail:
Thank you for looking at this and fixing it -- it's a real regression. 

I'd like to take a little time and figure out if there's a way to fix this without catching the exception and checking the error message on it. Is it just the String class that's throwing an exception from "has"? Perhaps we should fix that, or even fix the implementation of "has" in ScriptableObject so it can't leak out an exception.

gbrail:
Given how many things out there extend ScriptableObject and may have the same undesirable behavior for "has," I think that this is a fine fix.

However I think we have more to fix, as I see that V8 and Rhino differ on the enumerability of string properties and other stuff. So I may open an issue to fix the underlying problem, which requires that classes that override "has" should also override "getAttributes" to return the proper attributes for these properties, as NativeArray already does.

whitlockjc:
Agreed.

PR #417, opened 2018-04-08
fixes for version handling

rbri:
method version(xxx) should return the new version identifier if the version was updated

 make the default version more explicit because:

* there was no chance to get an idea about the current meaning of the default version
* we had to do always a double check for the version and the default version (i fear we had some places in the code where the check for the default version is missing)

Hope this makes the code more maintainable...

### DIFF
--- a/src/org/mozilla/javascript/Arguments.java
+++ b/src/org/mozilla/javascript/Arguments.java
@@ -35,8 +35,7 @@ final class Arguments extends IdScriptableObject
         calleeObj = f;
 
         int version = f.getLanguageVersion();
-        if (version <= Context.VERSION_1_3
-            && version != Context.VERSION_DEFAULT)
+        if (version <= Context.VERSION_1_3)
         {
             callerObj = null;
         } else {
--- a/src/org/mozilla/javascript/Context.java
+++ b/src/org/mozilla/javascript/Context.java
@@ -71,6 +71,15 @@ public class Context
     public static final int VERSION_UNKNOWN =   -1;
 
     /**
+     * The default version.
+     * <p>Be aware, this version will not support many of the newer
+     * language features and will not change in the future.</p>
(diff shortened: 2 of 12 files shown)


gbrail:
I like the idea here and if we were making Rhino again from scratch I'm sure that we'd do it just this way.

However I know that there is a lot of code out there that doesn't set a language version at all, which means that it is effectively using version "0". This change will probably break that code since version "0" and version "180" don't behave the same way. That means that projects that upgrade their Rhino version will probably break because they see those subtle differences. 

We've already caused some regressions in production code that I know about and this feels to me like another one...

rbri:
Ok, from your point of view what is the meaning of the default version?
* we will not change the default version in the future
* we will not change any of the checks to include the default version also?
So the default version will be useless (more or less).

If this is your idea of the default version i will try to document this a bit. My code (HtmlUnit) used the default version so far, that was the reason why i start fighting with this. Will use a specific version in the future.

rbri:
What do you think about the first commit of the PR. Do you think the version() implementation is correct?

gbrail:
Yes -- unfortunately I think that we should leave the default version alone. There is so much code out there that depends on Rhino I'd be careful about breaking a lot of it. I realize that this makes it useless and worse yet it makes people think that Rhino is not as capable as it is, as many of the more recent language features aren't available with that version. We can alleviate this to some degree by periodically updating the default version in the "Shell" class since people are presumably more flexible there...

The rest of the change makes sense to me -- I have not used the "version" function, however, and it makes sense to me that it should return something useful.

Can you also add a test for the "200" version, and also when you're ready, "squash" these commits into one? I don't think that we should always do that, but when we go back and forth a few times it makes everything cleaner!

rbri:
Ok, agree, will provide a new pr and close this.

PR #419, opened 2018-04-10
method Global#version(xxx) should return the new version identifier

rbri:
if the version was updated


rbri:
yes the tests are working locally

gbrail:
Works for me too. Thanks!

PR #423, opened 2018-04-15
two minor improvements from HtmlUnit code

rbri:
* window list is sorted
* Command 'Go to line' added

These changes are in use by HtmlUnit for a long time now (since 2013), i think we can merge this to improve the debugger a bit.


gbrail:
Any improvement to the GUI is great. I don't use it much, however, so I might be missing some ideas....

If I just run the debugger, and am in the console, "Go to line" results in a NullPointerException. Is that something that we should handle better? (Or can we disable "Go to line" when the console window is active?) Otherwise, if I open a file it works fine.

rbri:
have done some fixes for this; hope this is stable now in all cases

gbrail:
Great -- thanks!

PR #427, opened 2018-04-22
Delegator: improve the design for flexible subclassing

rbri:
For HtmlUnit we have so subclasses of Delegator. This patch makes it possible (or simple) to implement different ways of delegee resolution by simply overwrite the getDelegee() method.
Benefits of this:
* simpler design, you can implement lazy resolution
* you are more flexible regarding the object you have to store for delegee access, no need to store the delegee itself it is sufficient to store an object that knows the delegee

```
    @Override
    public Window getDelegee() {
        return webWindow_.getScriptableObject();
    }
```


gbrail:
Makes sense -- thanks!

PR #432, opened 2018-04-27
431 update expiry of cached commonjs entity if no change on filesystem

dhogan:
Upon commonjs entity cache expiration, subsequent filesystem resolution needs to update the expiry of the cache entry if the module has not changed on the filesystem. The isResourceChanged method should return true if the lastModified value of the cached entry does NOT equal the lastModified value of the UrlConnection pointing to the file.
With this change, the cached value will once again be used to resolve a require reference; without this change, the paths will be consulted to load the corresponding filesystem artifact with every require reference following initial cache period expiration.

### DIFF
--- a/src/org/mozilla/javascript/commonjs/module/provider/UrlModuleSourceProvider.java
+++ b/src/org/mozilla/javascript/commonjs/module/provider/UrlModuleSourceProvider.java
@@ -278,7 +278,7 @@ public class UrlModuleSourceProvider extends ModuleSourceProviderBase
                 return ((HttpURLConnection)urlConnection).getResponseCode() ==
                     HttpURLConnection.HTTP_NOT_MODIFIED;
             }
-            return lastModified == urlConnection.getLastModified();
+            return lastModified != urlConnection.getLastModified();
         }
 
         private long calculateExpiry(URLConnection urlConnection,


gbrail:
Thanks!

Can you provide a use case that ensures the proper behavior and prevents future regressions?

dhogan:
I don't see the unit-test context which would allow this to be easily tested. But our use-case involved our application throwing ~150,000 FileNotFoundExceptions per minute. Tracking down the code from the SoftCachingModuleScriptProvider#getModuleScript-> CachingModuleScriptProviderBase#getModuleScript-> ModuleSourceProviderBase#loadSource and ModuleSourceProviderBase#entityNeedsRevalidation and ModuleSourceProviderBase#loadFromPrivilegedLocations led to the realization that, upon cache expiry, the url paths (in our case file urls) would be consulted to reload the module. If the cached timestamp matched that of the url timestamp, then the expiry of the cached entity should be updated so that subsequent references would not lead to filesystem resolution. The bug I fixed would prevent this from happening, because the UrlModuleSourceProvider#URLValidator#isResourceChanged incorrectly reported that a resource changed if the cached timestamp was equal to the file urlConnection timestamp, rather than reporting that the resource changed if the comparison is NOT equal. The absence of resource change is necessary to update cache expiry - see URLValidator#updateValidator.

gbrail:
OK -- you've tested it, I can see the code, and commonjs does have regression tests. So let's go for it. Thanks!

PR #436, opened 2018-05-10
some fixes/enhancements for the typed array support

rbri:


rbri:
Hi Greg,
i'm still in the progress to reduce the differences between the HtmlUnit Rhino fork and Rhino itself. So far we are really close. Currently i'm working on using the typed array impl from Rhino instead our one. This PR contains some updates to fix some issues and add some features. Hope you are agree that this makes rhino a bit better.

gbrail:
Looks good to me. Would you mind if I squashed these into one commit (or would you mind doing it?)

gbrail:
Well there's the hard way and the easy way and the easy way is to press the button in GitHub when I merge the PR ;-)

PR #438, opened 2018-05-15
Improvements to MemberBox

### LINKED ISSUE #429: Serialization of MemberBox with a delegate is broken
Serialization of `MemberBox` with a non-null `delegateTo` field is broken: the field is declared transient, but never restored upon deserialization. This results in incorrectly deserialized objects.

It can be fixed in two ways: 
1. fail early: throw an exception from `writeObject` if serialization is attempted with a non-null `delegateTo`.
2. make it work: make the field non-transient.

Now, it's only `NativeError` that uses the delegate mechanism, and is typically setting it to otherwise serializable values. It seems to me that making the field non-transient is harmless. It was originally made transient in d357f56 11 years ago by @norrisboyd; not sure why – possibly due to a mistake?

Note that making it non-transient will change the serialization format of MemberBox, though.

szegedi:
Fixes #429 by allowing serialization of `MemberBox.delegateTo`. 

It also adds some more improvements enabled by the fact that in modern Java, we now have `java.lang.reflect.Executable` as a unifying type for both `Method` and `Constructor`. Using it reduces the need to typecheck and cast and it also allows unified access to parameter types and vararg flag, so they no longer need to be duplicated in the `MemberBox`.

### DIFF
--- a/src/org/mozilla/javascript/MemberBox.java
+++ b/src/org/mozilla/javascript/MemberBox.java
@@ -21,49 +21,32 @@ final class MemberBox implements Serializable
 {
     static final long serialVersionUID = 6358550398665688245L;
 
-    private transient Member memberObject;
-    transient Class<?>[] argTypes;
-    transient Object delegateTo;
-    transient boolean vararg;
+    private transient Executable memberObject;
+    Object delegateTo;
 
-
-    MemberBox(Method method)
+    MemberBox(Executable executable)
     {
-        init(method);
+        this.memberObject = executable;
     }
 
-    MemberBox(Constructor<?> constructor)
-    {
-        init(constructor);
-    }
-
-    private void init(Method method)
-    {
-        this.memberObject = method;
-        this.argTypes = method.getParameterTypes();
-        this.vararg = method.isVarArgs();
-    }
-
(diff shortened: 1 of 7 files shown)


src/org/mozilla/javascript/MemberBox.java
```diff
@@ -21,49 +21,32 @@
 {
     static final long serialVersionUID = 6358550398665688245L;
 
-    private transient Member memberObject;
-    transient Class<?>[] argTypes;
-    transient Object delegateTo;
-    transient boolean vararg;
+    private transient Executable memberObject;
```
gbrail: Can we make this final? I'm asking because some of the blogs talk about JVM optimizations for reflection (and MethodHandles, which we don't use yet) that only happen when they're final.

szegedi: Unfortunately, it can't be both transient and final as then it couldn't be assigned upon deserialization.

gbrail:
Of course -- thanks.

I often wish that Rhino didn't have serialization. But it does, so we need to leave it like this.

gbrail:
OK -- thanks!

PR #439, opened 2018-05-15
Modernizing the ContextAction

szegedi:
Parametrized return type of ContextAction, to reduce casts in client code. 
Replaced Rhino usages of ContextAction with lambdas.

This PR probably makes sense to be reviewed with `?w=1` at the end of the URL.

### DIFF
--- a/src/org/mozilla/javascript/Context.java
+++ b/src/org/mozilla/javascript/Context.java
@@ -512,7 +512,7 @@ public class Context
      * @return The result of {@link ContextAction#run(Context)}.
      */
     @Deprecated
-    public static Object call(ContextAction action)
+    public static <T> T call(ContextAction<T> action)
     {
         return call(ContextFactory.getGlobal(), action);
     }
@@ -545,7 +545,7 @@ public class Context
     /**
      * The method implements {@link ContextFactory#call(ContextAction)} logic.
      */
-    static Object call(ContextFactory factory, ContextAction action) {
+    static <T> T call(ContextFactory factory, ContextAction<T> action) {
         Context cx = enter(null, factory);
         try {
             return action.run(cx);
--- a/src/org/mozilla/javascript/ContextAction.java
+++ b/src/org/mozilla/javascript/ContextAction.java
@@ -11,8 +11,9 @@ package org.mozilla.javascript;
 /**
  * Interface to represent arbitrary action that requires to have Context
  * object associated with the current thread for its execution.
+ * @param T the type of the return value of action execution
  */
-public interface ContextAction
+public interface ContextAction<T>
 {
     /**
      * Execute action using the supplied Context instance.
@@ -21,6 +22,6 @@ public interface ContextAction
      *
      * @see ContextFactory#call(ContextAction)
      */
-    public Object run(Context cx);
+    public T run(Context cx);
 }
 
--- a/src/org/mozilla/javascript/ContextFactory.java
+++ b/src/org/mozilla/javascript/ContextFactory.java
@@ -512,7 +512,7 @@ public class ContextFactory
      *                   Scriptable scope, Scriptable thisObj,
      *                   Object[] args)
      */
-    public final Object call(ContextAction action)
+    public final <T> T call(ContextAction<T> action)
     {
         return Context.call(this, action);
     }
--- a/src/org/mozilla/javascript/JavaAdapter.java
+++ b/src/org/mozilla/javascript/JavaAdapter.java
@@ -604,7 +604,7 @@ public final class JavaAdapter implements IdFunctionCall
 
     public static Scriptable runScript(final Script script)
     {
-        return (Scriptable)ContextFactory.getGlobal().call(cx -> {
+        return ContextFactory.getGlobal().call(cx -> {
             ScriptableObject global = ScriptRuntime.getGlobal(cx);
             script.exec(cx, global);
             return global;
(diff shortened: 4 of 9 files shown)


src/org/mozilla/javascript/Context.java
```diff
@@ -505,7 +505,7 @@ public static void exit()
      * @return The result of {@link ContextAction#run(Context)}.
      */
     @Deprecated
-    public static Object call(ContextAction action)
+    public static <T> T call(ContextAction<T> action)
```
gbrail: This is a public interface of Rhino that's probably used by a lot of projects. I don't think that this actually changes the public interface (it should just introduce a compiler warning) but what did you observe? I see that you had to change a lot of tests... I am in general trying to be very careful about the impact of changes to the public interface like this.

szegedi: This type signature just gets erased to Object - it's the same as when Java collection classes got retrofitted for types with Java 1.5, so it's all binary compatible. The benefit of this is that in most cases users of the call invocation no longer have to be casting returns, like [here](https://github.com/mozilla/rhino/pull/439/files#diff-7a7863311e0e96498e06c552114aed15L612) and [here](https://github.com/mozilla/rhino/pull/439/files#diff-ca6960f128d5cd5d051c4fba4daae2f3L37).

Recompiling against this change might cause warnings (mostly in IDEs) that some casts are no longer necessary. In general, the less explicit casts you need in your code, the better.

As for tests, I did not have to change them; they would've passed unchanged. I just lambdafied the ContextActions in them. Ability to express ContextActions as lambdas is independent of having  a type parameter on ContextAction; on the other hand, lambdas save us from having to specify `ContextAction<Object>` instead of `ContextAction` to avoid raw type warnings. So I mostly lambdafied them to avoid having to have `ContextAction<Object>` everywhere.

szegedi:
I also checked today with `javap -p -c -v -cp buildGradle/classes/java/main org.mozilla.javascript.Context` to see how do type signatures look, and I see:

```
  public static <T extends java.lang.Object> T call(org.mozilla.javascript.ContextAction<T>);
    descriptor: (Lorg/mozilla/javascript/ContextAction;)Ljava/lang/Object;
```
So as you can see, the binary signature to the method didn't change at all… It's still `Object` as all type information is compile-only in Java and gets erased in actual signatures and has no runtime impact. I even did a full diff of javap outputs for Context between master and this branch (after masking irrelevant differences, like indices of constant pool entries and bytecode offsets) and it looks like this:

```diff
325c323
<   public static java.lang.Object call(org.mozilla.javascript.ContextAction);
---
>   public static <T> T call(org.mozilla.javascript.ContextAction<T>);
339,340d336
<         : new           #xx                 // class org/mozilla/javascript/Context$1
<         : dup
345c341
<         : invokespecial #xx                 // Method org/mozilla/javascript/Context$1."<init>":(Lorg/mozilla/javascript/Callable;Lorg/mozilla/javascript/Scriptable;Lorg/mozilla/javascript/Scriptable;[Ljava/lang/Object;)V
---
>         : invokedynamic #xx,  0             // InvokeDynamic #xx:run:(Lorg/mozilla/javascript/Callable;Lorg/mozilla/javascript/Scriptable;Lorg/mozilla/javascript/Scriptable;[Ljava/lang/Object;)Lorg/mozilla/javascript/ContextAction;
349c345
<   static java.lang.Object call(org.mozilla.javascript.ContextFactory, org.mozilla.javascript.ContextAction);
---
>   static <T> T call(org.mozilla.javascript.ContextFactory, org.mozilla.javascript.ContextAction<T>);
2579a2576,2585
>   private static java.lang.Object lambda$call$0(org.mozilla.javascript.Callable, org.mozilla.javascript.Scriptable, org.mozilla.javascript.Scriptable, java.lang.Object[], org.mozilla.javascript.Context);
>     Code:
>         : aload_0
>         : aload         4
>         : aload_1
>         : aload_2
>         : aload_3
>         : invokeinterface #xx,  5          // InterfaceMethod org/mozilla/javascript/Callable.call:(Lorg/mozilla/javascript/Context;Lorg/mozilla/javascript/Scriptable;Lorg/mozilla/javascript/Scriptable;[Ljava/lang/Object;)Ljava/lang/Object;
>         : areturn
> 
```

As you can see, `call(ContextAction)` on line 323 and `call(ContextFactory, ContextAction)` on line 345 only have a parametric type change (again, 100% binary compatible with Object) and the only other change (`invokedynamic` instead of `new/dup/invokespecial` and the new `lambda$call$0` method) is the reflection of [this](https://github.com/mozilla/rhino/pull/439/files#diff-d6ca902c16f796b2593d249bdfddab0dL535).

I think Rhino is overdue for API modernization; generics are basically Java 1.5 stuff. Admittedly, Rhino being a runtime for a dynamic language we don't have a lot of places where it makes sense to propagate type parameters, but `ContextAction` being from-Java-to-Java API is one of them; I don't by all means want to go overboard and "parametrize all the things". E.g. there's also `Callable`, but I wouldn't parametrize it as it's meant to express something that's callable in a script, so having it be non-parametrized and always returning `Object` makes sense.

If it makes you more comfortable, I can split out the ContextAction parametrization into a separate change to be submitted here, and move all the lambda-replacements into their own separate change.

gbrail:
Oh -- I merged this into another branch, not into master. Is there a way that you can create a new PR with these changes that we can merge to master? Or if you're not sure I can look and see what I can do...

At any rate, I'm convinced now that this is OK -- thanks!

szegedi:
Yeah, sorry, I set you up for this as I based the branch on top of the other one as a way to deal with diffs more easily… I made sure the change is merged into master now.

PR #443, opened 2018-05-20
Bound functions must not leak global object into strict functions

### LINKED ISSUE #442: Bound functions leak global scope into strict functions
While investigating #238 I noticed that our BoundFunction is replacing null "this" with global too early; it shouldn't be doing that at all (see #238 for relevant ES5 specification citations).

This expression needs to evaluate to true as strict functions aren't allowed to see global scope as their "this" (you can verify with either Nashorn or V8):
```
((function() { "use strict"; return this == null }).bind(null))()
```
Unfortunately, in Rhino it evaluates to false.

szegedi:
Fixes #442

### DIFF
--- a/src/org/mozilla/javascript/BoundFunction.java
+++ b/src/org/mozilla/javascript/BoundFunction.java
@@ -49,8 +49,7 @@ public class BoundFunction extends BaseFunction {
   @Override
   public Object call(Context cx, Scriptable scope, Scriptable thisObj, Object[] extraArgs)
   {
-    Scriptable callThis = boundThis != null ? boundThis : ScriptRuntime.getTopCallScope(cx);
-    return targetFunction.call(cx, scope, callThis, concat(boundArgs, extraArgs));
+    return targetFunction.call(cx, scope, boundThis, concat(boundArgs, extraArgs));
   }
 
   @Override
--- a/testsrc/org/mozilla/javascript/tests/StrictBindThisTest.java
+++ b/testsrc/org/mozilla/javascript/tests/StrictBindThisTest.java
@@ -1,0 +1,14 @@
+package org.mozilla.javascript.tests;
+
+import junit.framework.TestCase;
+
+public class StrictBindThisTest extends TestCase {
+    public void testStrictBindThis() {
+        Utils.runWithAllOptimizationLevels(cx -> {
+            assertTrue((Boolean)cx.evaluateString(cx.initSafeStandardObjects(), 
+                "((function() { 'use strict'; return this == null }).bind(null))()", 
+                "testBindScope.js", 1, null));
+            return null;
+        });
+    }
+}


szegedi:
[`language/function-code/10.4.3-1-77gs.js`](https://github.com/tc39/test262/blob/master/test/language/function-code/10.4.3-1-77gs.js) tests exactly this. The test doesn't seem excluded to me, so I'm not sure if this is indicative of the test suite failing to detect failures? Surely if I run this from command line with:
```
java -jar ./buildGradle/libs/rhino-1.7.11-SNAPSHOT.jar test262/test/language/function-code/10.4.3-1-77gs.js
```
on master it fails with:
```
js: "test262/test/language/function-code/10.4.3-1-77gs.js", line 13: exception from uncaught JavaScript throw: 'this' had incorrect value!
```
While in the branch of this PR it silently passes.

gbrail:
Sorry to open a can of worms but I think it's important to understand deep stuff like this...

All of the "language/function-object" tests in test262 are disabled, so I added a line to test262.properties to enable them. When I did that, we had 423 test failures, all in that suite (we run each test three times so that's 141 test cases), and with this change there are 429 (143 failing test cases).

I mucked around in the XML for a while and I might have misinterpreted some of the output, but it looks like this change fixes:

* 10.4.3-1-77-s
* 10.4.3-1-77gs

and it introduces new failures in these:

* 10.4.3-1-97gs
* 10.4.3-1-95s
* 10.4.3-1-96gs
* 10.4.3-1-97-s
* 10.4.3-1-95gs
* 10.4.3-1-96-s

Since you're in the middle of this, perhaps you'll be able to see what's going on?

szegedi:
I completely agree. I misunderstood the `test262.properties` file at first, now I understand that's it an include/exclude file, so if `language/function-code` wasn't specified, it wasn't included at all. You're right that those new failures should be looked into; I will do so. The gist of those failures is that the substitution of null for global (for non-strict functions) should be handled by the target function invocation; it seems like for some reason in these cases it isn't. So it's possible that that early substitution in `BoundFunction` was masking the lack of a necessary substitution somewhere downstream; I'll investigate and add it to this PR.

szegedi:
Just a heads up that I've spent some more time on this, and have a version that fixes more of the cases. It turned out to be a harder issue than I expected, so still needs some more work.

rbri:
@szegedi maybe you can push you current version of this, so others can try to improve this further

szegedi:
Pushed my additional pending changes on request from @rbri; with the understanding that even with this second commit it still doesn't fix all the cases.

p-bakker:
Converted to draft as this isn't mergable in it's current state 

Might be a useful basis to work on better strict-mode support in the future or be replaced by another PR autoverkeer

PR #451, opened 2018-06-26
Fix #448: Correctly wrap the result of Math.imul as an Int32.

### LINKED ISSUE #448: Math.imul sometimes returns incorrectly wrapped negative numbers
Expected, with Node.js v10.0.0:
```
$ /localhome/doeraene/opt/node-v10.0.0-linux-x64/bin/node -v
v10.0.0
$ /localhome/doeraene/opt/node-v10.0.0-linux-x64/bin/node 
> Math.imul(2147483611, -14)
518
```
Actual, with Rhino 1.7.9 (also 1.7.7+, since the introduction of `Math.imul` in 79e11acac27a41bbabbeb0014fd0da5a6c2d8339):
```
$ java -jar rhino-1.7.9.jar 
Rhino 1.7.9 2018 03 15
js> Math.imul(2147483611, -14)
-4294966778
js> Math.imul(2147483611, -14) | 0
518
```
Negative numbers smaller than -2**31 should be wrapped around to the positive side, basically using `ToInt32`.

sjrd:
The test https://github.com/tc39/test262/blob/041da54c02ae7d17edb8c134ab7691c4f643bafe/test/built-ins/Math/imul/results.js did not pass before this change, and passes after. I assume enabling the corresponding test262 directory is enough testing for the fix.

### DIFF
--- a/src/org/mozilla/javascript/NativeMath.java
+++ b/src/org/mozilla/javascript/NativeMath.java
@@ -6,8 +6,6 @@
 
 package org.mozilla.javascript;
 
-import org.mozilla.javascript.typedarrays.Conversions;
-
 /**
  * This class implements the Math native object.
  * See ECMA 15.8.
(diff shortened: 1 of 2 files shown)


sjrd:
The problem with the old implementation is that, in
```java
long product = (x * y) % Conversions.THIRTYTWO_BIT;
```
the product `x * y` of two unsigned 32-bit integers produces 64 bits worth of an *unsigned* integer. So it is possible to get a number whose bit 63 is 1, but that's the sign bit, so as a signed `long` it becomes negative. The subsequent `%` is broken since it interprets its arguments as signed, but you wanted them unsigned.

There are other ways to fix the issue. One that would stay closer to the spec algorithm would be to force an unsigned modulus operation, using `& 0xffffffffL` instead of `% THIRTYTWO_BIT` (or `Long.remainderUnsigned(x * y, THIRTYTWO_BIT)` if you prefer). Another is to directly use `(int) (x * y)` since the conversion from `long` to `int` applies the correct modulus *and* fixes the sign according to step 4 of the algorithm.

My implementation is closer to the intended meaning of `imul`, which is to perform a signed 32-bit multiplication, and is therefore also (probably) more efficient.

gbrail:
I agree that your implementation is what the spec intends, that it's more efficient than what's there now, and that it passes the tests ;-) So I think that we should adopt it.

gbrail:
However, since you have two commits in here that change the same function, I'd prefer that you "squash" them into one. (You can usually do this in git using "git rebase -i"). If you need help I can do it -- but in general it's your PR so I'd rather than you did. Thanks!

sjrd:
Don't worry, I perform interactive rebases in my sleep :p

Squashed.

gbrail:
Great. Thanks!

PR #452, opened 2018-06-30
Fix the conversions in typedarrays.Conversions.

sjrd:
Their implementation did not account for

a) the wrapping of number values larger than Integer.MAX_VALUE  or lower than Integer.MIN_VALUE.
b) the behavior of `%` for negative values.

The complex problem of a) is already dealt with by `ScriptRuntime.toInt32`, which we now leverage. Instead of `%` to implement the `modulo` operation from the ECMAScript specification (which is incorrect), we use sign-extending conversions such as `(int)(byte)x` and bitwise ands such as `& 0xff`.

This fixes the test262 test in `built-ins/TypedArrays/internals/Set/conversion-operation.js`. It does not affect any other test in `built-ins/TypedArrays/` (neither progression nor regression).

I stumbled upon the implementations of `Conversions` while trying to fix `Math.imul`, and noticed they were buggy for pretty much the same reason `js_imul` was buggy. So I figured I'd send a fix.

### DIFF
--- a/src/org/mozilla/javascript/typedarrays/Conversions.java
+++ b/src/org/mozilla/javascript/typedarrays/Conversions.java
@@ -14,33 +14,14 @@ import org.mozilla.javascript.ScriptRuntime;
 
 public class Conversions
 {
-    public static final int EIGHT_BIT = 1 << 8;
-    public static final int SIXTEEN_BIT = 1 << 16;
-    public static final long THIRTYTWO_BIT = 1L << 32L;
-
     public static int toInt8(Object arg)
     {
-        int iv;
-        if (arg instanceof Integer) {
-            iv = (Integer)arg;
-        } else {
-            iv = ScriptRuntime.toInt32(arg);
-        }
-
-        int int8Bit = iv % EIGHT_BIT;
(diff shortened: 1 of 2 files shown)


src/org/mozilla/javascript/typedarrays/Conversions.java
```diff
@@ -14,33 +14,14 @@
...
 {
-    public static final int EIGHT_BIT = 1 << 8;
-    public static final int SIXTEEN_BIT = 1 << 16;
-    public static final long THIRTYTWO_BIT = 1L << 32L;
-
     public static int toInt8(Object arg)
     {
-        int iv;
-        if (arg instanceof Integer) {
-            iv = (Integer)arg;
```
sjrd: Note that this fast-path was completely redundant, because [`ScriptRuntime.toInt32`](https://github.com/mozilla/rhino/blob/f192525fa96ad2c1211ec984e300e475bf7e4bb6/src/org/mozilla/javascript/ScriptRuntime.java#L1236) already takes care of this optimization.

gbrail:
Great -- thanks!

Issue #401, opened 2018-03-03
Allow comparing and hashing `NativeContinutation` objects


michbarsinai:
At the moment, `NativeContinuation`s are either identical (same object) or not equal. To implement formal verification, we need a way of comparing the state of the program in these continuations. That is, the stack, program counter, and heap.

The goal is that iff to `NativeContinuation` objects are equal, then continuing them would result in the same run. Corollary: continuations of two programs in the same state will be equal, while continuations of programs in different states (stack/heap values or program counter) will not be equal.

The hashcode should follow the usual contract: If two continuations are equal, they have the same hashcode.

BTW, this could be implemented by adding methods to `NativeContinuation`, or by creating an object that represent their internal state and equaling it.

PR #453, opened 2018-07-01
Allow comparing continuations for equality.

szegedi:
Implements #401.

This PR is separated into 4 commits for ease of review:
- First one simply adds a convenience static method to `NativeContinuation` to compare two continuations' implementations using vanilla `Objects.equal`. This is done instead of overriding `NativeContinuation.equals` in order to avoid backwards compatibility concerns. (`NativeContinuation` is a `ScriptableObject` and Rhino doesn't override `equals` for majority of them, except, curiously, for `NativeJavaPackage`,  but I digress…) 
- Second one introduces an algorithm for structural equality comparison of object graphs, specifically tailored to Rhino objects as well as Java collections. It also provides package-private static methods that act as hooks into bespoke structural comparison of some Rhino-specific objects (notably, `ArrowFunction`, `BoundFunction`, and `IdFunctionObject`). This comparison is "structural" in the sense that different object layouts of even semantically equivalent objects compare as false, e.g.
```javascript
var a1 = { 'foo': 42 }
var a2 = { 'foo': 42 }
var b = { 'a1': a1, 'a2': a2 }
var c = { 'a1': a1, 'a2': a1 }
```
`b` and `c` would not be considered structurally equivalent, since `b.a1 !== b.a2` but `c.a1 === c.a2`. This is important as we compare continuations, because if `b` occurs in one continuation being compared and `c` occurs in another continuation being compared, they must compare unequal otherwise if the continuations contain such strict comparison expressions as per above we'd fail to recognize their inequality.
- Third commit finally uses the previously defined structural comparison as the implementation for `Interpreter$CallFrame.equals` – the class that is used as the actual underlying state of Rhino interpreter continuations. It also adds an equals-consistent `hashCode` implementation.
- Finally, the fourth commit serves to put in place workarounds for two issues in Rhino I have identified while working on this that would otherwise block it. The workarounds are thus in a separate, marked commit so they can be easily found once the issues were fixed to remove them.

A reasonable amount of tests has also been provided.

### DIFF
--- a/src/org/mozilla/javascript/ArrowFunction.java
+++ b/src/org/mozilla/javascript/ArrowFunction.java
@@ -72,4 +72,8 @@ public class ArrowFunction extends BaseFunction {
         }
         return super.decompile(indent, flags);
     }
+    
+    static boolean equalObjectGraphs(ArrowFunction f1, ArrowFunction f2, EqualObjectGraphs eq) {
+        return  eq.equalGraphs(f1.boundThis, f2.boundThis) && eq.equalGraphs(f1.targetFunction, f2.targetFunction);
+    }
 }
--- a/src/org/mozilla/javascript/BoundFunction.java
+++ b/src/org/mozilla/javascript/BoundFunction.java
@@ -80,4 +80,8 @@ public class BoundFunction extends BaseFunction {
     System.arraycopy(second, 0, args, first.length, second.length);
     return args;
   }
+  
+  static boolean equalObjectGraphs(BoundFunction f1, BoundFunction f2, EqualObjectGraphs eq) {
+      return  eq.equalGraphs(f1.boundThis, f2.boundThis) && eq.equalGraphs(f1.targetFunction, f2.targetFunction) && eq.equalGraphs(f1.boundArgs, f2.boundArgs);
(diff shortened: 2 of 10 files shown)


gbrail:
This looks good to me but it does a lot and I'd love another pair of eyes on this. Is there anyone out there who's sufficiently familiar with continuations to give it a try and do a quick review?

gbrail:
I haven't had any feedback on this from anyone else but I'll test it one more time myself and unless there are other comments I'll merge it.

szegedi:
Hi Greg,

sorry… I literally just arrived home from a 7-day holiday so couldn't comment earlier.

These changes don't really affect any existing functionality; it's all new functionality. The way continuations work also doesn't change at all. The only thing that's somewhat different is that `Interpreter.CallFrame` class now has a semantical `equals` and `hashCode`.

@michbarsinai has been using a fork with this functionality for a while now in his [BPjs project's "modified-rhino" branch](https://github.com/bThink-BGU/BPjs/tree/modified-rhino).

gbrail:
Thanks -- I'm glad that someone has been using this and I understand it's all new.

PR #457, opened 2018-07-08
Support for `arguments` object as TypedArray constructor argument

rbri:
This is the same as #297 but includes a simple test


gbrail:
Looks good to me. Thanks!

Issue #449, opened 2018-06-17
getAllIds on Float32Array fails


szegedi:
Trying to query all IDs on a `Float32Array` throws an `IllegalArgumentException`. The issue seems to be that there is a gap of undefined instance IDs between `Float32Array` and its superclass `NativeTypedArrayView`.

Here's a little test that illustrates the issue:

```java
package org.mozilla.javascript.tests;

import junit.framework.TestCase;
import org.mozilla.javascript.*;

public class Float32ArrayTest extends TestCase {
    public void test() throws Exception {
        Context cx = Context.enter();
        cx.setLanguageVersion(Context.VERSION_ES6);
        Scriptable global = cx.initStandardObjects();
        ScriptableObject obj = (ScriptableObject)cx.evaluateString(global, "new Float32Array(5)", "", 1, null);
        obj.getAllIds();
        Context.exit();
    }
}
```
It results in:
```
java.lang.IllegalArgumentException: 9
	at org.mozilla.javascript.IdScriptableObject.getInstanceIdName(IdScriptableObject.java:699)
	at org.mozilla.javascript.typedarrays.NativeArrayBufferView.getInstanceIdName(NativeArrayBufferView.java:100)
	at org.mozilla.javascript.typedarrays.NativeTypedArrayView.getInstanceIdName(NativeTypedArrayView.java:400)
	at org.mozilla.javascript.IdScriptableObject.getIds(IdScriptableObject.java:629)
	at org.mozilla.javascript.ScriptableObject.getAllIds(ScriptableObject.java:995)
	at org.mozilla.javascript.tests.Float32ArrayTest.test(Float32ArrayTest.java:20)
```

The issue might be similar to #437.

szegedi:
It seems to me that the below change fixes it. Not sure why these values were 10 and 11 in the first place, though. A less brittle way of specifying them, counting from superclass `MAX_INSTANCE_ID` might be desirable.

```diff
--- a/src/org/mozilla/javascript/typedarrays/NativeTypedArrayView.java
+++ b/src/org/mozilla/javascript/typedarrays/NativeTypedArrayView.java
@@ -439,8 +439,8 @@ public abstract class NativeTypedArrayView<T>
      * These must not conflict with ids in the parent since we delegate there for property dispatching.
      */
     private static final int
-        Id_length               = 10,
-        Id_BYTES_PER_ELEMENT    = 11,
+        Id_length               = 4,
+        Id_BYTES_PER_ELEMENT    = 5,
[...]

Issue #437, opened 2018-05-15
NativeArrayBuffer.initPrototypeId throws IllegalArgumentException


szegedi:
Trying to access property "isView" on a `NativeArrayBuffer` throws an `IllegalArgumentException`. The issue seems to be that `PrototypeValues.maxId` for `NativeArrayBuffer` is 3 (to account for `Id_isView` property) but `initPrototypeId` only recognizes `Id_constructor` and `Id_slice`. 

`Id_isView ` seems to have some special handling as it's separately initialized; there's this whole business with it and `ConstructorId_isView` that I don't fully understand how's it supposed to work.

Here's a little test that illustrates the issue:

```java
package org.mozilla.javascript.tests;

import junit.framework.TestCase;
import org.mozilla.javascript.*;

public class ArrayBufferInitTest extends TestCase {
    public void test() throws Exception {
        Context cx = Context.enter();
        cx.setLanguageVersion(Context.VERSION_ES6);
        Scriptable global = cx.initStandardObjects();
        cx.evaluateString(global, "(new ArrayBuffer(5)).isView", "", 1, null);
        Context.exit();
    }
}
```
It results in:
```
java.lang.IllegalArgumentException: 3
	at org.mozilla.javascript.typedarrays.NativeArrayBuffer.initPrototypeId(NativeArrayBuffer.java:156)
	at org.mozilla.javascript.IdScriptableObject$PrototypeValues.ensureId(IdScriptableObject.java:312)
	at org.mozilla.javascript.IdScriptableObject$PrototypeValues.get(IdScriptableObject.java:190)
	at org.mozilla.javascript.IdScriptableObject.get(IdScriptableObject.java:410)
	at org.mozilla.javascript.ScriptableObject.getProperty(ScriptableObject.java:2331)
	at org.mozilla.javascript.ScriptRuntime.getObjectProp(ScriptRuntime.java:1579)
	at org.mozilla.javascript.ScriptRuntime.getObjectProp(ScriptRuntime.java:1572)
	at org.mozilla.javascript.gen.c_1._c_script_0(:1)
	at org.mozilla.javascript.gen.c_1.call()
	at org.mozilla.javascript.ContextFactory.doTopCall(ContextFactory.java:404)
	at org.mozilla.javascript.ScriptRuntime.doTopCall(ScriptRuntime.java:3508)
	at org.mozilla.javascript.gen.c_1.call()
	at org.mozilla.javascript.gen.c_1.exec()
[...]

PR #464, opened 2018-07-20
fixes for #437 and #449

rbri:
Should be fixed now, have not added many more unit tests but org.mozilla.javascript.tests.harmony.TypedArraysTest already includes tests for the isView function already.

Side note: the call in the issue is 'invalid'; the isView function is a static one. Calling isView on an object returns undefined (checked with ff)

### DIFF
--- a/src/org/mozilla/javascript/EqualObjectGraphs.java
+++ b/src/org/mozilla/javascript/EqualObjectGraphs.java
@@ -15,8 +15,8 @@ import java.util.Objects;
 import java.util.Set;
 import java.util.SortedMap;
 import java.util.TreeMap;
+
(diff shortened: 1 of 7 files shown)


rbri:
changing the line separator is not really related to this issue but the old impl hurts me every time i regenerate id classes on my win machine. The git checkout seems to translate the line separator; i think we have to reflect this in this tool also

rbri:
Hi Greg,
any hope for this one?

gbrail:
I’ve been travelling a bit. Tomorrow!

rbri:
Ah, enjoy you trip.

gbrail:
Looks good to me. Thanks!

PR #465, opened 2018-07-23
Retaining of comments and their position in the actual code after parsing.

ravikishore1979:
I wanted to use rhino parser to generate a new js file from existing js file after some modifications to some methods or variable names etc. for example as below

old JS file:
`Application.run(function ($rootScope) {
    "use strict";
    /* perform any action on the variables within this block(on-page-load) */
    $rootScope.onAppVariablesReady = function () {
        /*
         * variables can be accessed through '$rootScope.Variables' property here
         * e.g. $rootScope.Variables.staticVariable1.getData()
         */
    };

});`

Post Refactoring:
`/* perform any action on the variables within this block(on-page-load) */
    App.onAppVariablesReady = function () {
        /*
         * variables can be accessed through '$rootScope.Variables' property here
         * e.g. $rootScope.Variables.staticVariable1.getData()
         */
    };`
Using rhino parser I was able to achieve this but comments are lost. So providing a fix such that if "CompilerEnvirons.setRecordingComments(true)" then comments are added to the AST.


ravikishore1979:
Thanks for the comments.

This changes will not effect backward Compatibility as this will be executed only if "recordingComments" of "CompilerEnvirons" is set.

I'll make below changes and then reply.
1. Rebase and squash with fewer commits.
2. Write extensive test cases for this PR.

gbrail:
Thank you!

I'd be happy with "more extensive" test cases ;-) I saw that you were
supporting comments in a number of different places in the source and I
just want to make sure that we have a test for each one.

ravikishore1979:
Hi Greg Brail,
I couldn't find the reason for the failure of build. Can you help out to fix this issue?

gbrail:
I think we still have some flaky tests in the suite -- I re-ran the Travis job and it passed. So I think you're good to go from a test perspective.

ravikishore1979:
Thank you Greg Brail.
I am done with writing test cases for the inline comments used in loops, conditional blocks, try-catch etc.

gbrail:
This looks good. I'm going to squash it because it's got a bunch of "remove pom.xml" kinds of commits.

ravikishore1979:
Thanks Greg Brail.

PR #467, opened 2018-08-05
Array fixes

rbri:
some more fixes for our array implementations


gbrail:
Thanks -- I like to see it when more tests are passing!

Do you think that any non-Rhino code would be depending on the signature of "getIndexIds"? I've never seen it but I thought I'd ask.

Finally, can you squash these commits? Otherwise I can do it.

PR #468, opened 2018-08-11
Add @Override and some more cleanup

rbri:
Hi Greg,

hope you like this a bit, i think after switching to java 8 it is time to add the @Override annotation and streamline the code a bit.
I plan to provide some more patches to get more tests running; because of this i have done the cleanup before.


gbrail:
These kinds of changes are scary when the code base is in a lot of flux, but right now seems like a good time. I know that these kinds of changes are also a lot of work!

Anyway, I think these are good -- let me do one more test and merge.

At some point, it'd be interesting to update a few of the "checkstyle" rules so that more things can be enforced. Right now, since the codebase is so old and inconsistent, checkstyle runs but only a few basic things are enforced.

PR #469, opened 2018-08-12
Date fixes

rbri:
some more date tests are passing now

### DIFF
--- a/src/org/mozilla/javascript/ScriptableObject.java
+++ b/src/org/mozilla/javascript/ScriptableObject.java
@@ -1026,7 +1026,6 @@ public abstract class ScriptableObject implements Scriptable,
             }
 
             String methodName;
-            Object[] args = ScriptRuntime.emptyArgs;
             if (tryToString) {
                 methodName = "toString";
             } else {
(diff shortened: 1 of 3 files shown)


src/org/mozilla/javascript/ScriptableObject.java
```diff
@@ -1026,44 +1026,11 @@ public static Object getDefaultValue(Scriptable object, Class<?> typeHint)
...
 
             String methodName;
-            Object[] args;
+            Object[] args = ScriptRuntime.emptyArgs;
             if (tryToString) {
                 methodName = "toString";
-                args = ScriptRuntime.emptyArgs;
             } else {
                 methodName = "valueOf";
-                args = new Object[1];
```
gbrail: This is a lot of code and it must have been doing something -- or at least it was in some previous version of Rhino. Can you give us some context as to what's changing, what tests it fixes, and whether we think it'll break people used to older versions of Rhino?

rbri: Yes i was surprised also about this huge amount of code used for nothing. I think this was something left over from the really beginning, but did not any real history searching.

If you like to have a test case for this you can have a look at arg-to-number.js (and all the other args-to... for dates. There is an assert that proofs the valueOf method is called without any parameters.

If you have a look at all the core classes offering valueOf methods, all these methods are defined using an arity 0 and none of them uses any argument.

Did also some search on the internet and was not able to find some spec (even old one) that has any hint about this parameter.

And finally there was no failing test after this removal.

rbri:
Hope the details in my comment for the source code are sufficient. The 262 test suite seems to be a treasure chest full of test cases that might improve Rhinos js support quality. Hope to find some more time to work on this a bit more. But there are far too many issues on my list.

Thanks for your work and all the questions.....

PR #472, opened 2018-08-17
Implement the built-in Set and Map classes for ES6.

gbrail:
Implement these two new classes from the spec and enable relevant
test262 tests for them.

These new classes are only available when the version is
VERSION_ES6 or higher.

WeakSet and WeakMap will come next once we have these squared away.


gbrail:
I'd love a review of this from Attilla, @rbri, or whoever else has time. It's a big change that won't affect existing code but there's a bit of tricky code in here. Thanks!

testsrc/test262.properties
```diff
@@ -567,6 +568,23 @@ built-ins/Object
     ! values/tamper-with-global-object.js
     ! values/tamper-with-object-keys.js
 
+built-ins/Set
+    ! Symbol.species
+    # Expects Function.toString to work just like V8
+    ! prototype/Symbol.iterator.js
+    ! prototype/keys/keys.js
+    # Lots of work to do on strict mode
+    !prototype/forEach/callback-this-non-strict.js
```
rbri: inconsistent format, missing blank after the !

testsrc/test262.properties
```diff
@@ -340,15 +340,9 @@ built-ins/Date
 
 # this is really; slow disable for now
 # built-ins/decodeURI
-    ! name.js
-    ! S15.1.3.1_A2.4_T1.js
-    ! S15.1.3.1_A5.2.js
```
rbri: the idea of leaving the excludes in was to be able to re enable this without breaking, i suggest to leave it as is was before

src/org/mozilla/javascript/Hashtable.java
```diff
@@ -0,0 +1,224 @@
...
+import java.util.NoSuchElementException;
+
+/**
+ * This generic hash table class is used by Set and Map. It uses
+ * a standard HashMap for storing keys and values so that we can handle
+ * lots of hash collisions if necessary, and a list to support the iterator
+ * capability. This second one is important because JavaScript handling of
+ * the iterator is completely different from the way that Java does it.
+ */
+
```
rbri: remove this empty line

src/org/mozilla/javascript/Hashtable.java
```diff
@@ -0,0 +1,224 @@
...
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.NoSuchElementException;
+
+/**
+ * This generic hash table class is used by Set and Map. It uses
+ * a standard HashMap for storing keys and values so that we can handle
+ * lots of hash collisions if necessary, and a list to support the iterator
+ * capability. This second one is important because JavaScript handling of
+ * the iterator is completely different from the way that Java does it.
```
rbri: Nice documentation including the motivation for this impl. HtmlUnit has since a long time it's own impl of Map and Set. Both are using LinkedHashSet for the elements; can you please explain where the difference is that makes this impl necessary.

src/org/mozilla/javascript/Hashtable.java
```diff
@@ -0,0 +1,224 @@
...
+        if (v == null) {
+            return null;
+        }
+
+        // To keep existing iterators moving forward as specified in EC262,
+        // we will remove the "prev" pointers from the list but leave the "next"
+        // pointers intact. Once we do that, then the only things pointing to
+        // the deleted notes are existing iterators. Once those are gone, then
+        // these objects will be GCed.
+        if (v == first) {
```
rbri: Sounds strange, maybe it is worth to have some unit tests for this class.

rbri:
var kvArray = [['key1', 'value1'], ['key2', 'value2']];
var testMap = new Map(kvArray);
var myMap = new Map(testMap);
alert(myMap.size);

fails with something like
org.openqa.selenium.WebDriverException: com.gargoylesoftware.htmlunit.ScriptException: TypeError: Expected argument of type object, but instead had type string (script in http://localhost:12345/ from (1, 39) to (8, 10)#5)

rbri:
```
function logElement(value, key, m) {
  alert(value);
  alert(key);
  alert(m);
  alert(this);
}
function test() {
try {
  var myMap = new Map([[, 'value1'], ['key2', {}]]);
  myMap.forEach(logElement);
}catch(e){alert(e)}
}
```
produces 
[value1, ReferenceError: "key" is not defined.]
instead of 
[value1, undefined, [object Map], [object Window], [object Object], key2, [object Map], [object Window]]

rbri:
`alert(myMap[Symbol.iterator]);`

produces 
function [Symbol.iterator]() {...
instead of
function entries() {....

rbri:
Your helper class Hashtable has to be serializable.

rbri:
And finally to sum up the firsts tests - great works, seems to work in most cases. Great to see Rhino is still alive.

gbrail:
Thanks -- those tests exposed a real problem with iteration on Map that the test262 and other tests I had failed to catch -- thanks!

Making Map.prototype[Symbol.iterator] === Map.prototype.entries and so forth is harder -- there are tests for it, but the Rhino machinery around IdScriptableObject and stuff like that doesn't make it easy. I'll take a look but it might have to be after the rest of this feature gets merged.

I'll finish this up and look at your other PR in the next day or two.

rbri:
Take your time, i'm currently working on moving htmlunit to github, this requires time also.


    RBRi
    --------------------------
    WETATOR
    Smart web application testing
    http://www.wetator.org

gbrail:
Thanks for the review!

I believe I addressed your issues in the second commit in this PR. It'd be great to run another test if you can!

rbri:
Have done a check again with HtmlUnit. Looks better.

From a quick look it might be possible that the handling of the this arg of the forEach function is still not correct.
https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Global_Objects/Map/forEach

But this might also be a problem of HtmlUnit.

From my point of view you can merge this into the trunk. If there are still issues we can fix them later. Will have no time for this until next week.

PR #473, opened 2018-08-18
fix native array constructor called with null and same for the setter

rbri:
these are the remaining fixes required to have the native array impl from rhino to pass all the tests currently passing for the HtmlUnit impl.
If this is merged we can remove the classes from HtmlUnit.
[Next stop will be to use the new Map/Set stuff]


gbrail:
This looks good. Thanks!

PR #475, opened 2018-08-27
Add WeakMap and WeakSet

gbrail:
This are pretty simple now that Map and Set are implemented.


rbri:
I like to see this on the head, if done i will try to use it from HtmlUnit.

PR #478, opened 2018-09-24
Override getArity

raphj:
This override fixes getArity unconditionaly returning 0 instead of returning what getLength returns on ArrowFunction objects.

### DIFF
--- a/src/org/mozilla/javascript/ArrowFunction.java
+++ b/src/org/mozilla/javascript/ArrowFunction.java
@@ -65,6 +65,11 @@ public class ArrowFunction extends BaseFunction {
     }
 
     @Override
+    public int getArity() {
+        return getLength();
+    }
+
+    @Override
     String decompile(int indent, int flags)
     {
         if (targetFunction instanceof BaseFunction) {
--- a/testsrc/jstests/harmony/function-arity.js
+++ b/testsrc/jstests/harmony/function-arity.js
@@ -1,0 +1,26 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+load("testsrc/assert.js");
+
+// Check correctness of function properties length and arity
+// NOTE: arity is non-standard
+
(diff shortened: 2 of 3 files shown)


raphj:
So, I am a bit ashamed to admit this actually made me download and compile Rhino and try my patch. I did work around this bug in my code by using `getLength()` instead of `getArity()` so it works with current versions of Rhino.

I can see four Javascript tests:

    assertEquals(((a,b) => {}).arity, (function (a,b) {}).arity)
    assertEquals(((a,b) => {}).arity, ((a,b) => {}).length)
    assertEquals(((a,b) => {}).length, (function (a,b) {}).length)
    assertEquals(((a,b) => {}).length, 2)

The third one does not actually test this patch, but is probably interesting to have anyway.

`getArity` isn't currently tested anywhere if my grep-fu does not fail me. Testing for `.arity` in Javascript might not be desirable since this is not a standard property of Javascript functions if I am not mistaken. So writing the first two tests in Java might make sense. Or deprecating and removing `getArity`, since `getLength` is here?

What do you think? In which file(s) should those tests be put?

gbrail:
Oh yeah, this is weirder than I thought and is likely a historical thing. ECMAScript says nothing about an "arity" property, so even though we add it to function objects it's not part of current versions of the language. (It's probably historical and if we take it out we'll definitely break someone somewhere.) Furthermore, most subclasses of BaseFunction return the same value for getArity and getLength but not all of them.

So I think it'd be fine to do what you did and I don't see it breaking everything. The fact that your PR passed the automated tests helps.

I would love to see you add the tests that you posted above just for regression purposes, with a comment as to what you're testing.

An easy way to test this would be to add your tests as a new file in testsrc/jstests/harmony. Look at the other classes in that directory to see how they to asserting and how they all return "success" at the end.

Then, add a class to org.mozilla.javascript.tests.harmony -- otherwise your JS will never be executed. Just do what the other classes in there do ;-)

raphj:
Your last sentence probably saved me a few hours. Thanks!
Tests related to property `arity` fail without my commit and pass with it.
Let me know if things look good to you.

gbrail:
Thanks for doing this!

PR #480, opened 2018-10-06
Codegen 64k limit

rbri:
Here comes the next try to implement an fallback when the compiler reaches the usual java bytecode limit. (see https://sourceforge.net/p/htmlunit/bugs/1989/)

My starting point was hannes branch hns@207bb86, but the patch here is different

1. have not added a feature for this, because the discussion in the past shows no use case where someone likes to disable this
2. regarding the exception handling i stay with ClassFileFormatException; the name is not the best one but this was always triggered when an bytecode limit was reached.
3. Hannes impl contains a bug (found as side effect from my test cases); as stated in the source comments we have to rebuild the tree because the compile step might have done some manipulation

Outside of this i have done the usual cleanup and simplifications at some places. Sorry, but this for this code base i have to do this stuff while analyzing.
Hope you like it (at least we have tests for this) and as always squash if you like.


gbrail:
This looks good. I wish that it wasn't so much reformatting so that we can go back later if we have to.

At the same time, is there any way that you can squash this down to fewer commits? I'm also willing to give it a try later in the week.

And just to be clear -- what this mainly does is fall back to interpreter, right? From the description people might think that we "fixed" the 64K limit

rbri:
> This looks good. I wish that it wasn't so much reformatting so that we can go back later if we have to.

Angsthase

Good point, will provide a complete new PR as soon as you have merged the refactoring PR; i think this is a good deal; OK?

rbri:
ok the proposed new PR is there (#486). Will close this.

Issue #462, opened 2018-07-14
Illegal Reflective Operation


ArnaudBuchholz:
I have a project based on Rhino that throws this new error:

WARNING: Illegal reflective access by org.mozilla.javascript.MemberBox (file:/E:/Nano%20et%20Nono/Arnaud/dev/GitHub/gpf-js/node_modules/rhino-1_7r5-bin/rhino1_7R5/js.jar) to method sun.net.www.protocol.http.HttpURLConnection.setRequestMethod(java.lang.String)
WARNING: Please consider reporting this to the maintainers of org.mozilla.javascript.MemberBox
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release

Is there anything we can do to prevent that?

gbrail:
A few people have seen this with Java 9. The fix doesn't seem simple but
perhaps someone can try...

This happens because, via reflection, Rhino is trying to call a method that
is on a class that's not supposed to be visible through the Java 9 module
system.

My theory is that some code using Rhino is trying to call
java.net.HttpURLConnection.setRequestMethod(). However since this is
implemented by the sun.net.www.protocol... class, and Rhino is calling it
via reflection, the JVM doesn't know that Rhino is trying to use the
interface, and thinks that it's illegally calling a method on a class
that's not supposed to be visible.

I can think of two ways to "fix" this...

1) In MemberBox, we can add some code that walks the class tree upwards via
reflection to try and discover the highest-level interface that implements
a method, and try to call that method rather than the method on the
implementing class. I don't know exactly how to do this but it seems
plausible.

2) We can change Rhino's bytecode generation to directly insert calls to
Java methods into the generated bytecode rather than using reflection via
MemberBox. This would be a lot harder and not help us when in interpreted
mode. I'm pretty sure that this is what Nashorn did.

2 and a half) perhaps this works differently when InvokeDynamic and its
MethodHandles are used...

If anyone has time, it'd be great to write a little bit of experimental
code to test this hypothesis and possible solutions!

ArnaudBuchholz:
First of all, thank you for taking the time to answer.

I am clearly not a JAVA guy but I understand the reflection concept.

_"This happens because, via reflection, Rhino is trying to call a method that
is on a class that's not supposed to be visible through the Java 9 module
system."_

I am not sure what this Java 9 module system is but if the method is available on the class (provided it is public) I don't see why we wouldn't be able to invoke it.

[...]

PR #481, opened 2018-10-06
Memberbox cleanup

rbri:
Another cleanup pr; have unified (and sync with the name of the delegate method) the naming of the methods in member box and added some more methods. Hope this will make the code more clean ans straightforward.


src/org/mozilla/javascript/FunctionObject.java
```diff
@@ -91,7 +91,7 @@ public FunctionObject(String name, Member methodOrConstructor,
         }
         String methodName = member.getName();
         this.functionName = name;
-        Class<?>[] types = member.argTypes();
```
gbrail: Not on this line but...

If we're doing a "cleanup" PR, can we also clean up "import *" stuff? I see that this file and others have it. I'd love to update the checkstyle rule to disallow it (as we do in most Java codebases I've worked on) but there is too much of it in Rhino!

src/org/mozilla/javascript/NativeCallSite.java
```diff
@@ -133,22 +133,7 @@ private Object js_toString(Scriptable obj)
         return sb.toString();
     }
 
-    private Object getUndefined()
```
gbrail: Thanks! That old code is kind of like something from one of those "don't write code like this" web sites ;-)

gbrail:
Here's another idea for ANOTHER PR... take a look at this:

https://github.com/mozilla/rhino/issues/462

Since you are already digging in to MemberBox.

The problem is that the user might try to instantiate, say, java.net.URLConnection and call a method on it. As far as the user is concerned they are only using an abstract interface. But the class that actually gets instantiated is a private class deep inside some sort of "sun" package, so the Java 9 module system logs a warning.

But once we put an object in MemberBox (or wherever) if we could introspect the method being called and walk the inheritance tree up to the "top" somehow, then we won't get that error. 

I don't know how else to explain that other than with a diagram but I think that's the idea!

src/org/mozilla/javascript/FunctionObject.java
rbri: As you wish master ;-)

rbri:
Sorry but i see no chance to work on this java9/10 stuff for the next weeks

gbrail:
Thanks! I manually rebased and merged this.

PR #484, opened 2018-10-10
XmlMemberGet toSource implementation

### LINKED ISSUE #483: XmlMemberGet toSource() implementation
Rhino has support for E4X extensions, such as attribute get operator `a.@b`. It is parsed correctly as `XmlMemberGet` node, but `toSource()` call causes `IllegalArgumentException`, thrown from `AstNode.operatorToString()`, which is given '.' or '..' token.

Following snippet resembles this issue (default CompilerEnvirons is used):
```java
String script = "a.@b;";
Parser parser = new Parser(environment);
AstRoot root = parser.parse(script, null, 1);
/* up to 1.7.9 following will throw IllegalArgumentException */
root.toSource();
```

I have prepared a pull request, it will be opened right after this issue is created.

kuzjka:
This fixes #483.

`XmlMemberGet` (E4X support) node is parsed correctly, but calling `toSource()` causes `IllegalArgumentException`

As `Token.DOT` and `Token.DOTDOT` are not included to operators map in `AstNode` class, I've created helper method in `XmlMemberGet` to get string representation of the dots.

Another option is to add these tokens to the _operatorNames_ map in `AstNode`, so `operatorToString` could deal with them. Please, feel free to correct me, I'll update the code.

### DIFF
--- a/src/org/mozilla/javascript/ast/XmlMemberGet.java
+++ b/src/org/mozilla/javascript/ast/XmlMemberGet.java
@@ -87,8 +87,24 @@ public class XmlMemberGet extends InfixExpression {
         StringBuilder sb = new StringBuilder();
         sb.append(makeIndent(depth));
         sb.append(getLeft().toSource(0));
-        sb.append(operatorToString(getType()));
+        sb.append(dotsToString());
         sb.append(getRight().toSource(0));
         return sb.toString();
+    }
+
+    /**
+     * Gives string representation of inner dots token.
+     * @return  String representation of inner dots token (e.g. '.' or '..')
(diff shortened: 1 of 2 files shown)


kuzjka:
Whooops!

Forgot to remove copyrights headers, auto-added by IDE :)

On my way home, will fix it soon

gbrail:
Just to keep anyone from being surprised by that header message, can you squash these commits into one (I usually do it using "git rebase -i"). Thanks!

kuzjka:
Squashed

gbrail:
Thanks!

PR #485, opened 2018-10-10
fix serialization for the new Map and Set classes

rbri:
Serialization of the context was broken because NativeCollectionIterator was not serializable


rbri:
Sorry, there was not enougth time today will fix this later

rbri:
Ok, this one needs more time because also WeakHashMap is not serializable. Will create a new PR.

PR #486, opened 2018-10-11
Fall back to the interpreter mode if the byte code generation fails because of the 64k limit

rbri:
Here comes the next try to implement an fallback when the compiler reaches the usual java bytecode limit. (see https://sourceforge.net/p/htmlunit/bugs/1989/)

My starting point was hannes branch hns/rhino@207bb86, but the patch here is different

1. have not added a feature for this, because the discussion in the past shows no use case where someone likes to disable this
1. regarding the exception handling i stay with  ClassFileFormatException; the name is not the best one but this was  always triggered when an bytecode limit was reached
3. Hannes impl contains a bug (found as side effect from my test  cases); as stated in the source comments we have to rebuild the tree  because the compile step might have done some manipulation  

Outside of this i have done some cleanup and simplifications


rbri:
cool

On Thu, 18 Oct 2018 11:25:26 -0700 Greg Brail wrote:
>
>I tried this out with Trireme because it has some tests for large source files, and it works 
fine. Thanks!
>

PR #487, opened 2018-10-11
fix serialization for NativeMap, NativeSet, NativeWeakMap, and NativeWeakSet

rbri:


rbri:
And one more thing
Map and Set foreach have to take care of the special null value you are using. I'm not in front of my PC for the weekend but I'm sure you can find/fix this.

PR #488, opened 2018-10-21
more map fixes

rbri:
* another map fix found while using the latest code with HtmlUnit (hope this is the last one)
* again some work regarding the Test262Suite; more test included and have done a small hack to dump tests already passing but excluded

Hope you like both changes and merge both soon. Seems to be the last fix needed to switch HtmlUnit to the new Map/Set impl.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,4 +10,4 @@ cache:
   directories:
     - $HOME/.gradle/caches
     - $HOME/.gradle/wrapper
-
+install: travis_wait mvn install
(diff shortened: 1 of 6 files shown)


gbrail:
The main change here looks good -- what do you want to do? I can test this locally and merge it, or we can keep trying to make Travis more reliable...

rbri:
Thanks for taking the time....
I tested changes tested locally and they working fine here but i looks like we have a timing problem with the additional tests. The test suite runs mow longer.. If we like to have more tests we have to solve this isuue (and i like to have more tests to be able to change things without fear)
Will be great if you can
* merge the map fix because i need this change to be able to use the Map/Set stuff in HtmlUnit
* then merge the updated test suite and make it running on travis

rbri:
Knock, knock....

gbrail:
Sorry...

I created https://github.com/mozilla/rhino/pull/492 that cherry-picks your map fix. I'll merge it as soon as the tests pass.

Then can you make a PR for the others that shows the test changes? I'd like @szegedi to look too since both of you have been working on the test infrastructure.

rbri:
Thanks, will create a new PR over the weekend and then i will close close this one.

rbri:
as proposed #493 is available, will close this

PR #493, opened 2018-11-10
Improved Test262SuiteTest; many more tests

rbri:
hack to dump already passing tests
exclude one more class of tests
more tests are passing

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,9 @@ language: java
 dist: trusty
 jdk: 
   - openjdk8
+script:
+  - travis_wait 20 ./gradlew check
+
 # Took this stuff from the travis docs:
 before_cache:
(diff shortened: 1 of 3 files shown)


rbri:
Hi Greg,
after reading the doc a bit more careful a some coffee if made the right change to our travis build setup.
Hope you agree and we can have this larger tests suite as part of our build.

gbrail:
I got some OOM errors on the Regexp tests just now on my workstation. I have to run but will try again and see if it's repeatable... (built-ins/RegExp/S15.10.2...)

rbri:
But travis was green and i can't remember to see an oom while testing. JDK 1.8?

gbrail:
it works better if I increase the heap size in build.gradle from 256 MB to 1 GB, which is still reasonable IMHO given that we are now running more than 80,000 tests. I'm going to check that in and then merge this.

PR #509, opened 2019-01-06
[build] bump to gradle 5.1

grimreaper:
This allows users with Java 11 to build Rhino.
In doing so do the following
- regenerate the wrapper (ignoring default jvm changes)
- update wrapper jar
- remove custom wrapper task, which is duplicative of just keeping the version in the `.properties` file.


gbrail:
Interestingly, as I was writing that the tests ended on my machine after 14 minutes, with the last 10 minutes or so just sitting there after having run all tests. So it doesn't hang "forever".

Still, I think that we should figure this out before merging this.

gbrail:
The test problems I saw with older gradle 5.x versions don't seem to happen with the very latest, 5.4.1. I updated that in a separate PR.

PR #510, opened 2019-01-08
Fixes Parser.throwStatement()

stijnkliemesch:
Fixes Parser.throwStatement() creating ThrowStatement with incorect length

### DIFF
--- a/src/org/mozilla/javascript/Parser.java
+++ b/src/org/mozilla/javascript/Parser.java
@@ -1754,7 +1754,7 @@ public class Parser
             reportError("msg.bad.throw.eol");
         }
         AstNode expr = expr();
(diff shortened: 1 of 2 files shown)


stijnkliemesch:
The created ThrowStatement had its length property initialized with not a length value but the ASTNode's endposition value.

gbrail:
Very straightforward and it doesn't break anything. However in the service of making this codebase better over time I'd still love to see a new test case. Thanks!

stijnkliemesch:
The testcase triggers a specific use of the ThrowStatement instance that, without the fix, ends up using the incorrect value.

This case is how I encountered the bug, I suppose the bug mentioned in the testcase commit message should have been a seperately reported issue before initiating this pull request.

gbrail:
Thanks for the test case!

Issue #511, opened 2019-01-10
Cannot convert function to interface java.util.function.Consumer since it contains methods with different names


GuillermoBlasco:
Given an arrow function passed to a `forEach` the intepreter crashes with the message:
 ```
org.mozilla.javascript.EvaluatorException: Cannot convert function to interface java.util.function.Consumer since it contains methods with different names (<script name>#<line number>)
```

The point is that the `InterfaceAdapter:27` expects an interface with at least one method, and if many all called the same. Which makes sense prior to `default` keyword for interfaces. But given that Consumer interface now has a `andThen` default method `InterfaceAdapter:27` finds that there are two methods called differently.

Rhino version: 1.7.10

JS code interpreted:
```
tc.getValues().forEach(function (value) {
  if (value.get('XX').get() === 'xx') {
    if (value.get('YY') === undefined) {
        value.set('YY', new Date());   
    }
  }  
});
``` 
Where `tc.getValues()` returns a NativeArray. 

Stack trace:
```
rg.mozilla.javascript.EvaluatorException: Cannot convert function to interface java.util.function.Consumer since it contains methods with different names (<script_name>#<line number>)
	at org.mozilla.javascript.DefaultErrorReporter.runtimeError(DefaultErrorReporter.java:77) ~[rhino-1.7.10.jar:1.7.10]
	at org.mozilla.javascript.Context.reportRuntimeError(Context.java:998) ~[rhino-1.7.10.jar:1.7.10]
	at org.mozilla.javascript.Context.reportRuntimeError(Context.java:1053) ~[rhino-1.7.10.jar:1.7.10]
	at org.mozilla.javascript.Context.reportRuntimeError1(Context.java:1016) ~[rhino-1.7.10.jar:1.7.10]
	at org.mozilla.javascript.InterfaceAdapter.create(InterfaceAdapter.java:52) ~[rhino-1.7.10.jar:1.7.10]
	at org.mozilla.javascript.NativeJavaObject.createInterfaceAdapter(NativeJavaObject.java:681) ~[rhino-1.7.10.jar:1.7.10]
	at org.mozilla.javascript.NativeJavaObject.coerceTypeImpl(NativeJavaObject.java:657) ~[rhino-1.7.10.jar:1.7.10]
	at org.mozilla.javascript.Context.jsToJava(Context.java:1874) ~[rhino-1.7.10.jar:1.7.10]
[...]

PR #512, opened 2019-01-10
Issue #511, fixing InterfaceAdapter abstract name lookup.

GuillermoBlasco:
Solving issue #511. Now default methods are ignored, so they do not conflict with
interface (abstract) methods to implement.

### DIFF
--- a/src/org/mozilla/javascript/InterfaceAdapter.java
+++ b/src/org/mozilla/javascript/InterfaceAdapter.java
@@ -7,6 +7,7 @@
 package org.mozilla.javascript;
 
 import java.lang.reflect.Method;
+import java.lang.reflect.Modifier;
 
 /**
  * Adapter to use JS function as implementation of Java interfaces with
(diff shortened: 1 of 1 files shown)


gbrail:
Would it be possible to add a test case so that this doesn't regress? For instance, it looks like we have some default method tests already in testsrc/jstests/java8/call-default-method.js

Jeffrey-P-McAteer:
Hi has there been anything blocking this fix from being merged in? I'm in a position at work where having this engine accept callback functions as `Consumer<T>` objects would be useful.

GuillermoBlasco:
Hi @Jeffrey-P-McAteer 
Nothing but the missing tests :D anyhow I've been using rhino with that fix during almost a year and worked like a charm.

rPraml:
I ran into the same issue

tonygermano:
Suggest closing this as superseded by https://github.com/mozilla/rhino/pull/631

p-bakker:
Closing as superceded by #631

PR #513, opened 2019-01-14
Fix bug that caused modified JavaScript files to never be reloaded

travishaagen:
Fix bug related to reloading JavaScript files that have been modified.

The function,

```
        boolean updateValidator(URLConnection urlConnection, long request_time,
                UrlConnectionExpiryCalculator urlConnectionExpiryCalculator)
        throws IOException
        {
            boolean isResourceChanged = isResourceChanged(urlConnection);
            if(!isResourceChanged) {
                expiry = calculateExpiry(urlConnection, request_time,
                        urlConnectionExpiryCalculator);
            }
            return isResourceChanged;
        }
```

checks if a resource has changed, which means that it was modified. The bug was to consider this as not-modified in the conditional statement.

### DIFF
--- a/src/org/mozilla/javascript/commonjs/module/provider/UrlModuleSourceProvider.java
+++ b/src/org/mozilla/javascript/commonjs/module/provider/UrlModuleSourceProvider.java
@@ -152,7 +152,7 @@ public class UrlModuleSourceProvider extends ModuleSourceProviderBase
         try {
             urlConnection.connect();
             if(applicableValidator != null &&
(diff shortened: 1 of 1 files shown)


gbrail:
Sounds like a very reasonable fix. Can you add a test case?

travishaagen:
@gbrail : Added modified/not-modified unit tests.

```
./gradlew -Dtest.single=*UrlModuleSourceProviderTest test
```

PR #363, opened 2017-11-16
Fix positions recording around arrow functions

sainaen:
Two actual fixes:
* record position of the parameterless arrow function (fix by @nabice, fixes #303)
* record position and length of the 'fake' return statement inside function's expression body

Tests are passing locally; doesn't introduce new checkstyle violations.

Issue #303, opened 2017-04-24
ARROW_FUNCTION position set error


nabice:
JS Code:
`setTimeout(() => {}, 0);`

In org.mozilla.javascript.Parser.java, line 887,
params.position is -1, this will make a wrong FunctionNode position.

hyb1996:
Rhino supports arrow function? I try your code but only get a syntax error.

nabice:
支持的。可能你没有用到最新的版本？

hyb1996:
I am using the latest release [1.7.7.1](https://github.com/mozilla/rhino/releases/tag/Rhino1_7_7_1_RELEASE) but cannot run any code with arrow function. How to enable this feature? 
怎么支持呀 我设置语言版本为1.8和ES6也不行耶

nabice:
不太清楚你怎么弄的，直接编译最新的是可以的。没有进行什么设置。

hyb1996:
I finally found it was supported in the latest [SNAPSHOT](http://oss.jfrog.org/repo/org/mozilla/rhino/1.7.8-SNAPSHOT/).

Yidadaa:
@hyb1996 活捉autoJS作者2333

gbrail:
Since this seems to be fixed in master I'm closing this.

nabice:
@gbrail, I found it still not fixed. FunctionNode has wrong position.

gbrail:
OK. I must be misunderstanding the problem then. Perhaps if you can give us
some code to reproduce it and describe what's incorrect then we can take a
look.

nabice:
file: a.js
`setTimeout(() => {}, 0);`

When a.js is parsed, setTimeout's first argument is a FucntionNode instance. Property `position` of the instance should be 12, but 0 we get.
It can be fixed like this, probably not correct:

```
diff --git a/src/org/mozilla/javascript/Parser.java b/src/org/mozilla/javascript/Parser.java
index da03f0c7..787e8bc3 100644
--- a/src/org/mozilla/javascript/Parser.java
+++ b/src/org/mozilla/javascript/Parser.java
@@ -3106,7 +3106,7 @@ public class Parser
             Comment jsdocNode = getAndResetJsDoc();
             int lineno = ts.lineno;
             int begin = ts.tokenBeg;
-            AstNode e = (peekToken() == Token.RP ? new EmptyExpression() : expr());
[...]

PR #517, opened 2019-01-24
fixes position for ParenthesizedExpression nodes (#129)

### LINKED ISSUE #129: Position incorrect for Parenthesized expression.
Given a file containing:

(a);

The absolute position of the AstNode will be 1 (the char index of the 'a'). 
Expected absolute position to be 0 (the index for the '(').

Contrast this with an array literal which has absolute position set correctly:

[a];

kuzjka:
Fixes #129 by setting position explicitly for _ParenthesizedExpression_ nodes while parsing.
It avoids mapping of ParenthesizedExpression to source with only right paren included, which seems to be unwanted.

One test has broken: `ArrowFnPositionBugTest#testArrowFnWithArgsPosition()`. But there was a notice about inconsistent position evaluation for ParenthesizedExpression nodes, so as far as this inconsistency is eliminated, we can update the test. Please, make a retrospective look at #303 and PR #363 and confirm that it's ok.

### DIFF
--- a/src/org/mozilla/javascript/Parser.java
+++ b/src/org/mozilla/javascript/Parser.java
@@ -3212,20 +3212,20 @@ public class Parser
             if (peekToken() == Token.FOR) {
                 return generatorExpression(e, begin);
             }
-            ParenthesizedExpression pn = new ParenthesizedExpression(e);
+            mustMatchToken(Token.RP, "msg.no.paren", true);
+            if (e.getType() == Token.EMPTY && peekToken() != Token.ARROW) {
+              reportError("msg.syntax");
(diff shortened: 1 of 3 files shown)


kuzjka:
Thank you for the review, Greg.
Also, could you please tell us if there is any release plan for Rhino for this year.

gbrail:
There's no fixed schedule. Given that a lot of things have been fixed recently, it's sounding like a good time for a release in the next month or so.

gbrail:
This looks good to me, passes the tests, and no one who depends on the AST a lot has had a question. Thanks for submitting it!

Issue #521, opened 2019-02-10
Feature request: implement  String.prototype.padStart(), String.prototype.padEnd(), and Array.prototype.includes()


manishmotwani3:
Hi, the implementations of the following functions is missing in the latest version of Rhino.
Can we add these?

String.prototype.padEnd( maxLength [ , fillString ] )
https://www.ecma-international.org/ecma-262/8.0/#sec-string.prototype.padend

String.prototype.padStart( maxLength [ , fillString ] )
https://www.ecma-international.org/ecma-262/8.0/#sec-string.prototype.padstart

Array.prototype.includes ( searchElement [ , fromIndex ] )
http://www.ecma-international.org/ecma-262/8.0/index.html#sec-array.prototype.includes

rbri:
Have created a pull request for  padStart and padEnd. Includes is already there.

rbri:
I think we can close this

PR #523, opened 2019-02-16
implement #521 (padStart and padEnd)

rbri:
and some more cleanup and minor fixes.

### DIFF
--- a/src/org/mozilla/javascript/Arguments.java
+++ b/src/org/mozilla/javascript/Arguments.java
@@ -18,7 +18,7 @@ import org.mozilla.javascript.NativeArrayIterator.ARRAY_ITERATOR_TYPE;
  */
 final class Arguments extends IdScriptableObject
 {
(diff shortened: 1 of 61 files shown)


rbri:
Any hope for this one?

gbrail:
Absolutely. Sorry for the delay. I'll look.

gbrail:
In general... I think it's a good practice to separate changes that add or modify functionality from formatting changes. I know I'm always the one asking for commits to be squashed, but in this case I'd feel a lot better if the formatting changes and the String changes were separate PRs or at least separate commits.

Why? If we end up having to revert something or isolate a problem later it'll be easier if it's not all merged together into the same commit.

Is that a hard change to make? If it's a big burden we can go ahead anyway, but I think it's a good principle to follow.

gbrail:
Ah -- they're already separate commits! Then never mind! I'll keep reading.

rbri:
Thanks.

PR #525, opened 2019-03-10
add first array includes impl

rbri:
add first array includes impl;
fix null/undefined handling


gbrail:
This basically looks good to me. Let me test it out and read it again and then I think that we're good to go.

rbri:
At least no old test is broken and some more are passing :-)

PR #530, opened 2019-03-31
functions are valid keys for WeakMap/WeakSet

rbri:


rbri:
knock knock

src/org/mozilla/javascript/ScriptRuntime.java
```diff
@@ -2892,7 +2892,8 @@ public static boolean isObject(Object value)
             return false;
         }
         if (value instanceof ScriptableObject) {
-            return "object".equals(((ScriptableObject)value).getTypeOf());
+            String type = ((ScriptableObject)value).getTypeOf();
+            return "object".equals(type) || "function".equals(type);
```
gbrail: I'm a bit concerned about fixing this problem right here, as opposed to fixing it at the source. "isObject" is in theory a generic thing that's called from lots of places and I'm worried that changing it here might have other effects even if it doesn't break any tests.

Furthermore, I don't really love using "getTypeOf" here since it's kind of expensive versus a pair of instanceof checks.

rbri:
>I'd also love to re-check the ECMAscript standard for the exact definition of this "isObject" functionality and see where else it's used in the code.

Did this already, you have introduced this method as part of the initial Map implementation and this method is used so far only in this situation (im my memories are correct ;-)

>Unfortunately I'm travelling now and won't be able to sit down and look at this more carefully today or tomorrow.

Enjoy

	RBRi

src/org/mozilla/javascript/ScriptRuntime.java
gbrail: Yes, I now re-read the code and I remember. Thanks for fixing this!

gbrail:
Merged. Thanks!

rbri:
Am 11. April 2019 00:50:26 MESZ schrieb Greg Brail <notifications@github.com>:
>gbrail commented on this pull request.
>
>
>
>> @@ -2892,7 +2892,8 @@ public static boolean isObject(Object value)
>             return false;
>         }
>         if (value instanceof ScriptableObject) {
>-            return
>"object".equals(((ScriptableObject)value).getTypeOf());
>+            String type = ((ScriptableObject)value).getTypeOf();
>+            return "object".equals(type) || "function".equals(type);
>
>Yes, I now re-read the code and I remember. Thanks for fixing this!
>
>-- 
>You are receiving this because you authored the thread.
>https://github.com/mozilla/rhino/pull/530#discussion_r274196506

We are all getting older....

Issue #531, opened 2019-04-02
Using a symbol as array index crashes with a class cast exception


rbri:
Hi Greg,
found this simple example that fails.

  var arr = ['Apple', 'Banana'];
  arr[Symbol.iterator] = function () { return i; };

Not sure how to fix this but i hope you have an idea because you have done the symbol support so far.

gbrail:
The fix looks good -- merged!

PR #532, opened 2019-04-03
fix for #531

rbri:
using a symbol as array index crashes with a class cast exception

please review

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -204,7 +204,7 @@ publishing {
 }
 
 jacoco {
-    toolVersion = "0.7.9"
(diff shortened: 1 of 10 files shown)


rbri:
knock knock

gbrail:
Looks good! I'm going to squash the Jacoco commits and push in a second.

gbrail:
I should have read this one more carefully! Adding a method to the Scriptable interface means that existing code that embeds Rhino will break. I have a new PR coming that keeps this test working but doesn't add to the Scriptable interface...

PR #543, opened 2019-05-09
another jdk switch no longer required

rbri:


gbrail:
This makes sense to me. Thanks!

Issue #219, opened 2015-07-21
Array.concat not working on Java Arrays


tangrui:
In version 1.7.6 and 1.7.7:

``` javascript
a = java.lang.reflect.Array.newInstance(java.lang.Integer.TYPE, 1);
a[0] = 1;
b = java.lang.reflect.Array.newInstance(java.lang.Integer.TYPE, 1);
b[0] = 2;
c = a.concat(b); // this prints out something like [I@2f410acf,[I@424c0bc4
c.toSource() // this prints out [[1], [2]]
```

But the same code works correctly in version 1.7R5, and prints `1,2` and `[1, 2]`.

gbrail:
One thing I am learning is the wide variety of ways that people have used Rhino in the past to integrate Java code with JavaScript, and we clearly are missing test cases for many of them.

For backward compatibility purposes, we need to figure out how to make this work the way that it did before. Not sure what the fix is right now but perhaps some of the others who worked on the NativeArray code would have an idea.

tuchida:
The cause is this commit.
https://github.com/mozilla/rhino/commit/82d9a936

In ES5's specification,
http://www.ecma-international.org/ecma-262/5.1/#sec-15.4.4.4
> 5.b. If the value of the [[Class]] internal property of E is "Array", then

I think this commit is reasonable.

In ES6, add symbol `@@isConcatSpreadable`.
http://www.ecma-international.org/ecma-262/6.0/index.html#sec-array.prototype.concat
> 7.b. Let spreadable be IsConcatSpreadable(E).

You could add `@@isConcatSpreadable` to `NativeJavaArray` after symbol implementation.

PR #546, opened 2019-05-10
Implement @@isConcatSpreadable and make Java arrays spreadable

gbrail:
Array.prototype.concat will now honor the @@isConcatSpreadable symbol,
by treating arguments with the symbol set to a truthy value as
arrays that should be spread apart, and everything other than
a native array as a single object.

In addition, mark native Java arrays (the "JavaArray") class as
"spreadable" to preserve backward compatibility with older releases.

This fixes https://github.com/mozilla/rhino/issues/219


gbrail:
@rbri You did a lot of recent work on arrays -- can you look at this change?

rbri:
Outside of the minor comments another step forward into a bright rhino future ;-). Like to see this improvement as soon as possible in the head.

PR #547, opened 2019-05-11
Array of and copyWithin

rbri:
first simple impl

### DIFF
--- a/src/org/mozilla/javascript/NativeArray.java
+++ b/src/org/mozilla/javascript/NativeArray.java
@@ -181,6 +181,8 @@ public class NativeArray extends IdScriptableObject implements List
                 "reduceRight", 1);
         addIdFunctionProperty(ctor, ARRAY_TAG, ConstructorId_isArray,
                 "isArray", 1);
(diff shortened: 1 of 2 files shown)


src/org/mozilla/javascript/NativeArray.java
```diff
@@ -641,6 +651,17 @@ private static Object jsConstructor(Context cx, Scriptable scope,
         return new NativeArray(len);
     }
 
+    private static Object jsOf(Context cx, Scriptable scope, Scriptable thisObj, Object[] args)
```
gbrail: Is there any reason why this can't be js_of like other stuff in this class? A minor thing I know.

rbri: no real reason, i have used the pattern that was used for the static methods, because this is more constructor like.

Not really work in progress because my time is limited and a had this around for some time. Made the pull request because this fixes some more cases and breaks no existings.

gbrail:
I used your commits as the basis of https://github.com/mozilla/rhino/pull/555. I made some minor test fixes and added an implementation of Array.from, so I'm going to close this one.

PR #549, opened 2019-05-14
Un-do recent addition to the Scriptable interface.

gbrail:
More of the classes inside Rhino need the new methods that handle
Symbol properties. However, for backward compatibility we cannot
add new methods to the Scriptable interface, or existing code
will break.

### DIFF
--- a/examples/Matrix.java
+++ b/examples/Matrix.java
@@ -159,16 +159,6 @@ public class Matrix implements Scriptable {
     public void put(String name, Scriptable start, Object value) {
     }
 
-
-    /**
-     * Set a named property.
-     *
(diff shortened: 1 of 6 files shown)


gbrail:
@rbri I think this fixes the problem that you were trying to fix in issue 531 without causing a new method to be added to the Scriptable interface. Can you check? Thanks!

rbri:
Makes sense for me

gbrail:
Yep -- that PR added a method to Scriptable, and that breaks existing code that embeds Rhino because tons of code out there implements Scriptable. This un-does that, but keeps the fix for the issue. The test for that issue remains in the codebase so it's still fixed.

PR #551, opened 2019-05-14
More compatibility fixes for Array.prototype.concat.

gbrail:
Restore old behavior that would check the constructor and treat
objects with the Array constructor as "spreadable" for the purposes
of concat. This is the default for language versions < ES6, and
for ES6 we follow the spec. This makes older language versions
follow older ECMAscript specs and be compatible with older Rhino
releases.

This also fixes some bugs that caused us to treat arguments
to concat differently if the first argument was a native array.
Rather than restore that bug, we fixed it and removed the test.


gbrail:
@rbri more work on this -- backward compatibility is also important and I wanted to support the new Symbol and ECMAspec without it. The code was getting unwieldy so I broke it up.

rbri:
Please merge it into the head, can't wait ;-)

Issue #550, opened 2019-05-14
NullPointException while retrieving a number from an object


guygoldwasser:
B"H
Hi,
while fetching an attribute from an object, in case the content of the attribute is a number, code is checking if some Context attributes is set in order to decide which type to return e.g. integer or double.
the problem is that not always context is entered and null pointer exception raise.
seems that check if context!=null is missing while coerce a type.
at NativeJavaObject.java line 520. Rhino 1.7.9 release
if(context.hasFeature(Context.FEATURE_INTEGER_WITHOUT_DECIMAL_PLACE))

while context might be null at this point and there is no check for that.

Usually the only usecase I do enter context is for compilation. after compilation ends, I do exit because performance is much better then. I never had problems working this way.
js code example will be any getter e.g. model['intContent'] or model.intContent etc.

PR #553, opened 2019-05-15
Fix flag tests that assume Context is available.

gbrail:
This fixes https://github.com/mozilla/rhino/issues/550.

I found another place where we are doing the same thing. It makes sense to fix both.

There are other places where we assume that Context.getCurrentContext() returns non-null, but those are more deeply embedded.


guygoldwasser:
Thanks a lot gbrail!
will test that soon

Issue #164, opened 2015-03-10
Support ES6 additions to Array


gbrail:
These include:

fill
keys
values
entries

(find and findIndex are already implemented)

rbri:
Seems to be done, i suggest to close this.

gbrail:
Those are done. But although I didn't list them all those years ago, we still don't have:

from
of
and copyWithin

p-bakker:
Seems all taken care of, except for new Array methods in ES2019 (flat & flatMap), for which we have https://github.com/mozilla/rhino/issues/949 and https://github.com/mozilla/rhino/issues/948, so closing this issue

PR #555, opened 2019-05-17
Complete work on Array.of, from, and copyWithin.

gbrail:
Fix a few tests, and add an implementation of from.


gbrail:
This will let us close https://github.com/mozilla/rhino/issues/164.

rbri:
Great progress, will merge this into the HtmlUnit fork (there is also a from impl) as soon this is on the head and provide feedback.
Many thanks for all the improvements you did the last days.

PR #561, opened 2019-05-25
Java List and array access improvements

makusuko:
Add a new wrapper calss NativeJavaList that provides natural JS array access for Java lists. Improve NativeJavaArray for better compatability with JS via the addition of indexOf and includes functions. Add a couple of tests for the list and array access.

### DIFF
--- a/src/org/mozilla/javascript/NativeJavaList.java
+++ b/src/org/mozilla/javascript/NativeJavaList.java
@@ -7,9 +7,7 @@
 package org.mozilla.javascript;
 
 
-import java.lang.reflect.Array;
 import java.util.List;
-import java.util.Objects;
 
 /**
  * This class reflects Java Lists into the JavaScript environment.
--- a/src/org/mozilla/javascript/NativeJavaList.java
+++ b/src/org/mozilla/javascript/NativeJavaList.java
@@ -6,62 +6,34 @@
 
 package org.mozilla.javascript;
 
+
 import java.lang.reflect.Array;
+import java.util.List;
 import java.util.Objects;
 
 /**
- * This class reflects Java arrays into the JavaScript environment.
+ * This class reflects Java Lists into the JavaScript environment.
  *
- * @author Mike Shaver
+ * @author Markus Sunela
  * @see NativeJavaClass
  * @see NativeJavaObject
  * @see NativeJavaPackage
  */
 
-public class NativeJavaArray
-    extends NativeJavaObject
-    implements SymbolScriptable
-{
+public class NativeJavaList
+        extends NativeJavaObject
+        implements SymbolScriptable {
+
     private static final long serialVersionUID = -924022554283675333L;
-
-    /**
-     * Implementation of the Array.indexOf method.
-     */
-    private static class IndexOfMethod extends BaseFunction {
-        private Object array;
-
-        IndexOfMethod(Object array) {
-            this.array = array;
-        }
-
-        @Override
-        public Object call(Context cx, Scriptable scope, Scriptable thisObj, Object[] args) {
-            Object arg = args[0];
-            if (arg instanceof Wrapper)
-                arg = ((Wrapper)arg).unwrap();
-
-            int start = 0;
-            if (args.length > 1)
-                start = ((Number)args[1]).intValue();
-            if (start < 0)
-                start = args.length + start;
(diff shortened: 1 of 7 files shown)


gbrail:
Thanks. I'll take a look at this next week after we finish the 1.7.11 release. One difference between this and other built-in classes is in the way that it adds new methods -- other native classes inherit from IdScriptableObject whereas this one uses a different method. One thing that we might want to do is to make this class work that way too.

Also, a number of the prototype classes on regular arrays should work on Java arrays too -- we should look and see why not.

makusuko:
Yeah, I tried to use a couple of ways to implement the JavaScript functions. The thing is, that the wrapped object should expose all the methods and properties from the object implementing the List interface. For example, in our projects we tend to have business objects, which implement List interface, but have quite a lot of other business logic in them too.

src/org/mozilla/javascript/NativeJavaArray.java
```diff
@@ -23,6 +24,64 @@
...
+            this.array = array;
+        }
+
+        @Override
+        public Object call(Context cx, Scriptable scope, Scriptable thisObj, Object[] args) {
+            Object arg = args[0];
+            if (arg instanceof Wrapper)
+                arg = ((Wrapper)arg).unwrap();
+
+            int start = 0;
```
tuchida, line 43: Is there a test when `start` is specified as an argument?
```js
var a = java.util.Arrays.asList(1, 2, 3).toArray();
a.indexOf(1, 5);  // -1?
a.indexOf(1, -5); // throw error?
```

src/org/mozilla/javascript/NativeJavaArray.java
```diff
@@ -23,6 +24,64 @@
...
+            if (args.length > 1)
+                start = ((Number)args[1]).intValue();
+            if (start < 0)
+                start = args.length + start;
+
+            int length = Array.getLength(array);
+            for (int index = start; index < length; index++)
+                if (Objects.deepEquals(arg, Array.get(array, index)))
```
tuchida, line 51: Is there a test to make sure `deepEquals` is called?

testsrc/org/mozilla/javascript/tests/AccessingJavaList.java
```diff
@@ -0,0 +1,83 @@
...
+import junit.framework.TestCase;
+import org.mozilla.javascript.Context;
+import org.mozilla.javascript.ContextFactory;
+import org.mozilla.javascript.Scriptable;
+import org.mozilla.javascript.tools.shell.Global;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class AccessingJavaList extends TestCase {
```
rPraml, line 18: This test is passing in the current codebase.

src/org/mozilla/javascript/NativeJavaList.java
```diff
@@ -0,0 +1,196 @@
...
+    @Override
+    public boolean has(Symbol key, Scriptable start) {
+        return SymbolKey.IS_CONCAT_SPREADABLE.equals(key);
+    }
+
+    @Override
+    public Object get(String id, Scriptable start) {
+        if (id.equals("length"))
+            return Integer.valueOf(list.size());
+        else if (id.equals("includes"))
```
rPraml, line 97: This is only one extension of the list.
I think we were both trying to reach the same goal: See https://github.com/mozilla/rhino/pull/830
A `JavaList` should be treated like an `Array` and provide functions like `join`, `includes` etc...

rPraml (commented):
This is a very old PR and goes into the same direction as #830
Personally I can take up the topic again after #1031 is merged

rPraml:
Most of the test cases SHOULD¹ pass in the current codebase of #1083.  (¹ plugins.gradle.com is currently down...)
I've adjusted only one code https://github.com/mozilla/rhino/pull/1083/commits/faeada0e575b81a32142433baafe623789d95b4a which is correct in my mind. (There is no `includes` in java.util.List, so you have to `use the Array.prototype.includes`)
@makusuko If you agree, this PR can be closed.

p-bakker:
Closing this one for lack of activity and the idea behind this PR likely already been solved in other PRs

If there's still merit to this PR, please reopen after a rebase

PR #562, opened 2019-05-26
Add manual OSGi manifest

makusuko:
As Rhino project already manually extends the MANIFEST.MF, it is very easy (and completely trouble-free) to add the OSGi headers so that the library can be readily utilized in OSGi environment. Only thing worth noting is, that OSGi version numbers only support <major>.<minor>[.<patch>] format, so I added a small regex to remove the possible -RCx, -BETA, -SNAPSHOT etc. endings from the version.


gbrail:
Thanks! This is very simple. I merged it manually and changed the tabs to spaces... It'll be part of 1.7.11.

PR #579, opened 2019-06-22
Fix Array.includes returns a wrapped Boolean and align with specs

nename0:
I have noticed a bug when executing this line in the latest version (master and 1.7.11):
```
js> [1,2,3,4].filter(t => [1,2].includes(t));
1,2,3,4
```
Now of course this should be `[1,2]`. So I started to dig: both `filter` and `includes` work fine for themselves. However when combining them something goes wrong.
`includes` returns a `NativeBoolean` and the `filter` function checks the result of the callback using `ScriptRuntime.toBoolean`. 
However in JS this is true: `!(new Boolean(false)) == false`. So NativeBoolean objects are always true. Therefore `includes` should return an java `Boolean`.

### DIFF
--- a/src/org/mozilla/javascript/NativeArray.java
+++ b/src/org/mozilla/javascript/NativeArray.java
@@ -351,7 +351,7 @@ public class NativeArray extends IdScriptableObject implements List
                 return js_lastIndexOf(cx, scope, thisObj, args);
 
               case Id_includes:
-                return Boolean.valueOf(((long) js_indexOf(cx, scope, thisObj, args)) > -1);
+                return js_includes(cx, scope, thisObj, args);
 
               case Id_fill:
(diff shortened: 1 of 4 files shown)


rbri:
Great catch. This fix seems to fix a bunch of 262 include test. See the Travis output for details.

nename0:
Ok, so while digging for the test262 tests, I found out that the current `Array.includes` implementation does not respect the [specs](https://tc39.es/ecma262/#sec-array.prototype.includes). The specs explicitly state that `includes` differes from `indexOf` in some parts.
So I reimplemented `includes` as a new method and now all test262 tests pass.

rbri:
Great

gbrail:
Looks good to me too. Thanks!

PR #586, opened 2019-07-20
redefinition of permanent properties should throw a type error

### LINKED ISSUE #574: Rhino allows redefine the length property and even changes the array object
Rhino version: 1.7.11, 1.7.10, 1.7.9
**testcase:**
````
var NISLFuzzingFunc = function () {
        var arr = [1,2,3,4];
        print("Before redefine, arr is: " + arr)
        print("Before redefine, arr.length = " + arr.length);
        Object.defineProperty(arr, "length", {
                get: function(){
                        return 11;
                }
        });
        print("After redefine, arr.length = " + arr.length);
        print("After redefine, arr is: " + arr)
};

NISLFuzzingFunc();
````
**command:**
``java -jar rhino/rhino-1.7.11.jar -debug -version 200 testcase.js``
**result:**
````
Before redefine, arr is: 1,2,3,4
Before redefine, arr.length = 4
After redefine, arr.length = 11
After redefine, arr is: 1,,3,4
````
------
Rhino not only allows redefine length of array, but also changes the array object. While others JS engines are not allowed to redefine property length, such as spiderMoneky, v8, javascriptCore, and chakra.

rbri:
closes #574


gbrail:
This looks good but it feels to me like one of those things that will break legacy code. I wonder if we should put it behind the "ES6" language version by checking the version before throwing the error.

The other PRs look good, I'll test and merge soon!

p-bakker:
@rbri I know it's been ages, but any idea what happened to the changes in all the commits in this PR?

For example the [first one](https://github.com/mozilla/rhino/pull/586/commits/af92448061744fdf692dd84def4101fada306be9) was about #574 and it seems that commit never made it into the codebase, as the issue reported in 574 still exists

rbri:
@p-bakker will try to start my time machine ....

rbri:
ok, looks like this really got lost in the mud. Will prepare a new PR based on the first commit, the others are looking like faulty usage of git. But will check that later.

rbri:
step 1 - spend some time on the first commit ([af92448](https://github.com/mozilla/rhino/pull/586/commits/af92448061744fdf692dd84def4101fada306be9)) but this is no longer working. There are several problems with the code, at least the arguments property of functions gets overwritten for arrow functions. This overwrite fails after applying this because the arguments property from basefunction is declared as permanent. There are several other problems also.

Think we have to leave the issue https://github.com/mozilla/rhino/issues/574 open and we have to start a new attemp from scratch.

rbri:
the other commits are not related to the issue and i guess they are all applied.

p-bakker:
Ah, to bad, thnx for taking a look!

PR #593, opened 2019-08-16
fix interpreter fallback when using streams (fixes #592)

### LINKED ISSUE #592: Fallback to interpreter fails when using a reader to compile
When Context.compileImpl is called with a Reader rather than a String to supply the source, if the compilation fails due to ClassFileFormatException and we try to use the interpreter for a second attempt, the second parse on [this line](https://github.com/mozilla/rhino/blob/master/src/org/mozilla/javascript/Context.java#L2538) will fail as the Reader would've already run through all available data.

One possible workaround is to comment out [this line](https://github.com/mozilla/rhino/blob/master/src/org/mozilla/javascript/Context.java#L2516) so that we always have the source in the String rather than the Reader.

rbri:


gbrail:
This is fine and helpful, but... How often will the stream be read from a reader that supports "reset"? I believe that FileReader does not, which means that we should use a BufferedReader instead, and then we end up with size issues...

I don't necessarily think that reading it directly into a String is the best solution here, but perhaps we either need to read into a BufferedReader or accept the idea that the user of this function is going to have to do some extra work.

I know that there are some comments on the attached issue too -- what do you think?

Sorry I didn't look at this earlier in the week.

rbri:
Sorry folks, but emukidid was correct (and i was not aware of this even after this long time providing patches to rhino). The Parser always converts the given reader into a string before parsing. Made a major change of my first fix by doing the reader to string conversion in the context. This simplifies the code at some places and ends the discussion about resetting the reader.
Hope everybody likes this one :-)

rbri:
I think the failing test is a false positive - at least on my machine the suite passes.

rbri:
Thanks

PR #595, opened 2019-08-22
Fix parseInt for numbers with a leading 0

### LINKED ISSUE #594: ParseInt() returns NAN,0 for some numbers
https://github.com/mdukehall/javascript-algorithms/blob/master/reverseint.js
Running this in rhino works fine for most numbers.
'008', '009' return as 0 and '-900' returns as NaN.

moreheadm:
Fixes #594. As discussed in the issue ES5 and later say that integers starting with a leading zero should be assumed have a radix of 10.

### DIFF
--- a/src/org/mozilla/javascript/NativeGlobal.java
+++ b/src/org/mozilla/javascript/NativeGlobal.java
@@ -244,9 +244,6 @@ public class NativeGlobal implements Serializable, IdFunctionCall
                 if (c == 'x' || c == 'X') {
                     radix = 16;
                     start += 2;
-                } else if ('0' <= c && c <= '9') {
-                    radix = 8;
-                    start++;
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


rbri:
Looks like we have to add a version chech

gbrail:
Thanks for submitting this. However, Rhino is used in a ton of places and we have to be careful about backward compatibility. 

I'd really rather see us pull together a more comprehensive fix that:

1) Re-enables some of the test262 tests for "parseInt"
2) Adds a version check (VERSION_ES6 is always safe, but perhaps a newer one) so that the new behavior is only triggered by users who are OK with a more recent language version.'
3) Checks to see if there are additional requirements for parseInt in ECMA 262 that we should be implementing too!

gbrail:
@rbrbi submitted another change that fixes this same problem but adds a commit check.

https://github.com/mozilla/rhino/pull/623

Thanks to both of you for working on this!

PR #599, opened 2019-09-16
Eased validation in ScriptRuntime.enumInitOrder for host objects

stijnkliemesch:
This changeset eases the validation slightly, to allow host objects implementing Scriptable and SymbolScriptable, but not extending ScriptableObject specifically, to serve as an iterable in a for(... of ...) loop.

I'm not sure what unit test file I'd have to add a testcase to for this matter, any suggestions?

P.S. You'll have to pardon my commit message, I wasn't aware of the formatting that github assumes.

### DIFF
--- a/src/org/mozilla/javascript/ScriptRuntime.java
+++ b/src/org/mozilla/javascript/ScriptRuntime.java
@@ -2271,15 +2271,11 @@ public class ScriptRuntime {
     }
 
     private static Object enumInitInOrder(Context cx, IdEnumeration x) {
-        if (!(x.obj instanceof ScriptableObject)) {
+        if (!(x.obj instanceof SymbolScriptable) || !ScriptableObject.hasProperty(x.obj, SymbolKey.ITERATOR)) {
             throw typeError1("msg.not.iterable", toString(x.obj));
         }
 
-        ScriptableObject xo = (ScriptableObject)x.obj;
-        if (!ScriptableObject.hasProperty(xo, SymbolKey.ITERATOR)) {
-            throw typeError1("msg.not.iterable", toString(x.obj));
-        }
-        Object iterator = ScriptableObject.getProperty(xo, SymbolKey.ITERATOR);
+        Object iterator = ScriptableObject.getProperty(x.obj, SymbolKey.ITERATOR);
         if (!(iterator instanceof Callable)) {
             throw typeError1("msg.not.iterable", toString(x.obj));
         }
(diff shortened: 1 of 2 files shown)


gbrail:
This seems quite reasonable to me, and since Symbols are fairly recent, I don't THINK it'll break any old code.

I do think that we should write SOME tests for this, showing that a dummy object that implements the right interface and has the iterator could work, and that a dummy object that doesn't won't, and possibly even make sure that one of those tests fails before your change.

The easiest way to do that, I think, is to add some unit tests by adding a new class in testsrc/org/mozilla/javascript/tests. Most of the actual JS code you'd need to write for this test is very small (basically try to create an iterator and see if you get an exception) and most of it would be in the various Java classes.

I think that's important because otherwise some day in the future someone will try and reverse this test, or something, and then we will indeed break something!

stijnkliemesch:
Would commit 190d955 suffice? 

I executed the test locally without the changes from b02144d and it failed as expected, with the following result:

```
org.mozilla.javascript.EcmaError: TypeError: [object Object] is not iterable
	at org.mozilla.javascript.ScriptRuntime.constructError(ScriptRuntime.java:4256)
	at org.mozilla.javascript.ScriptRuntime.constructError(ScriptRuntime.java:4234)
	at org.mozilla.javascript.ScriptRuntime.typeError(ScriptRuntime.java:4267)
	at org.mozilla.javascript.ScriptRuntime.typeError1(ScriptRuntime.java:4279)
	at org.mozilla.javascript.ScriptRuntime.enumInitInOrder(ScriptRuntime.java:2275)
	at org.mozilla.javascript.ScriptRuntime.enumInit(ScriptRuntime.java:2248)
	at org.mozilla.javascript.gen._eval__2._c_anonymous_1(<eval>)
	at org.mozilla.javascript.gen._eval__2.call(<eval>)
	at org.mozilla.javascript.optimizer.OptRuntime.call0(OptRuntime.java:36)
	at org.mozilla.javascript.gen._eval__2._c_script_0(<eval>)
	at org.mozilla.javascript.gen._eval__2.call(<eval>)
	at org.mozilla.javascript.ContextFactory.doTopCall(ContextFactory.java:409)
	at org.mozilla.javascript.ScriptRuntime.doTopCall(ScriptRuntime.java:3566)
	at org.mozilla.javascript.gen._eval__2.call(<eval>)
	at org.mozilla.javascript.gen._eval__2.exec(<eval>)
	at org.mozilla.javascript.Context.evaluateString(Context.java:1266)
	at org.mozilla.javascript.tests.IterableTest.testForOfUsingArrayIterator(IterableTest.java:175)
```

I don't know if you want the order of these commits the other way around so first the test fails and then the fix comes afterwards.

stijnkliemesch:
I noticed you also wanted a test for a dummy object that doesn't implement the right interface, I added two variants of this in a68bccf.

gbrail:
This makes sense to me now. Thanks!

Issue #479, opened 2018-09-28
A potential XXE vulnerability found on rhino


brianwrf:
Howdy,

Just found a potential XXE vulnerability on rhino as show below, it seems function `toXml` didn't add any protection from XXE vulnerability when parsing XML document.

<https://github.com/mozilla/rhino/blob/master/xmlimplsrc/org/mozilla/javascript/xmlimpl/XmlProcessor.java#L225>

You may need to follow the OWASP guide below which provides concise information to prevent this vulnerability. 
<https://www.owasp.org/index.php/XML_External_Entity_(XXE)_Prevention_Cheat_Sheet#Java>

Thanks,
Brian

nightwatchcyber:
@gbrail / @rbri - did someone from the Rhino project team have a chance to look at this issue?

git4dipu:
@brianwrf Can you help us understand the vulnerability with some attack vector? 

The reason i am asking is:

-  I wrote a main method in XmlProcessor.java to verify toXml vulnerability and tried few xmls with below entity attack. 
     <?xml version="1.0" encoding="UTF-8"?>
     <!DOCTYPE foo [<!ELEMENT foo ANY >
     <!ENTITY xxe SYSTEM "/" >]>

- But it failed with "Scanner State 24 not Recognized". I believe Java Parser is not allowing DOCTYPE injection (i.e. state 24).

TypeError: Cannot parse XML: Scanner State 24 not Recognized  (_2B0601040181900D819C200A819C200100#94)
Execution time: 446.0 ms
org.xml.sax.SAXException: Scanner State 24 not Recognized 
                at com.sun.org.apache.xerces.internal.parsers.DOMParser.parse(DOMParser.java:271)
                at com.sun.org.apache.xerces.internal.jaxp.DocumentBuilderImpl.parse(DocumentBuilderImpl.java:347)
                at org.mozilla.javascript.xmlimpl.XmlProcessor.toXml(XmlProcessor.java:274)
                at org.mozilla.javascript.xmlimpl.XmlProcessor.main(XmlProcessor.java:401)

So not sure if this is truly vulnerable. Please correct me if i am wrong.

FYI    - @nightwatchcyber @gbrail  @rbri

gbrail:
Looking at the code, and sticking in a handy "printf," I see that although
there are standard APIs now to disable external entity fetching, we're not
explicitly disabling that feature, so in theory we could be vulnerable.

It'd be a simple change to put in the code, if someone is looking for a
[...]

PR #600, opened 2019-09-18
Adding secure configuration for XML parsers

tophersmith:
This PR properly configures both the DocumentBuilderFactory and TransformerFactory instances in the XmlProcessor class to prevent XML eXternal Entity (XXE) attacks as recommended by OWASP.

A side note for anybody reading this in the future: The original implementation was not actually vulnerable to this attack using default built-in DBF or TF instances as by a set of slightly unique occurrences the code dodged all of the known attack types.

For example, XXE attacks are typically run by adding DOCTYPE and ENTITY declarations to the XML to trick the Java parsers into making protocol requests for more information to resolve malicious entities - file:// pulls in file and directory information into the XML doc, and http(s):// (s)ftp:// can trick the target server into making requests on behalf of an attacker (this is known as SSRF). Because the XmlProcessor wraps all incoming xml in a synthetic parent, This attack is nullified as DOCTYPEs and ENTITY defs can only be made before the first data tag.

The other attack vectors usually rely on XML implementations (XSLT and XSD) that require gathering more information from external sources, but the TransformerFactory's usage of output=XML and DOMSource negate that as well.

The good news about this is that this code fix should not have any behavior changes, as this will further enforce a case that can't happen out of the box.

All this being said, the PR should be included as this codefix protects against cases where an application may be using a non-default XML parser (most Java XML parsers are loaded via Service Loading or system properties at runtime with a fallback of one of the defaults). The PR sets all available best practice features or properties to catch these additional implementations.

### DIFF
--- a/src/org/mozilla/javascript/Context.java
+++ b/src/org/mozilla/javascript/Context.java
@@ -347,6 +347,13 @@ public class Context
      */
     public static final int FEATURE_LITTLE_ENDIAN = 19;
 
+    /**
+     * Configure the XMLProcessor to parse XML with security features or not.
+     * Security features include not fetching remote entity references and disabling XIncludes
+     * @since 1.7 Release 12
+     */
+    public static final int FEATURE_ENABLE_XML_SECURE_PARSING = 20;
+    
     public static final String languageVersionProperty = "language version";
     public static final String errorReporterProperty   = "error reporter";
 
--- a/src/org/mozilla/javascript/ContextFactory.java
+++ b/src/org/mozilla/javascript/ContextFactory.java
@@ -298,6 +298,9 @@ public class ContextFactory
 
           case Context.FEATURE_LITTLE_ENDIAN:
               return false;
+
+          case Context.FEATURE_ENABLE_XML_SECURE_PARSING:
+              return true;
         }
         // It is a bug to call the method with unknown featureIndex
         throw new IllegalArgumentException(String.valueOf(featureIndex));
--- a/testsrc/org/mozilla/javascript/tests/CustomTestDBF.java
+++ b/testsrc/org/mozilla/javascript/tests/CustomTestDBF.java
@@ -1,0 +1,48 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.javascript.tests;
+
+import javax.xml.parsers.DocumentBuilder;
+import javax.xml.parsers.DocumentBuilderFactory;
+import javax.xml.parsers.ParserConfigurationException;
+
+public class CustomTestDBF extends DocumentBuilderFactory {
+    public static final String INTENTIONAL_CONFIG_EXCEPTION = "Intentionally thrown";
+    
+    @Override
+    public Object getAttribute(String arg0) throws IllegalArgumentException {
+        // TODO Auto-generated method stub
+        return null;
+    }
+
+    @Override
+    public boolean getFeature(String name) throws ParserConfigurationException {
+        // TODO Auto-generated method stub
+        return false;
+    }
+
+    @Override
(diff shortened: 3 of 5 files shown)


tophersmith:
This resolves https://github.com/mozilla/rhino/issues/479

tophersmith:
Looks like a test failure in regress-363258.js which is some kind of timer test? I will lay odds that a re-build will not have this come up.

gbrail:
Thank you. I agree that this is something we should fix! I do think that we need to look in to three things, however:

First, Rhino is an old codebase that's used in lots of places, and in theory this will break backward compatibility. Therefore, I think that we need to add a flag to determine whether we configure the parser in "secure" mode. You can look at the Context and ContextFactory classes to see how feature flags are implemented, and then at how those flags are used to see how to check them. It's a common pattern in Rhino.

I would also suggest a flag with a name like "FEATURE_DISABLE_XML_EXTERNAL_ENTITIES" or something similar that will make it clear what it does. 

However, since this is a potential security vulnerability, I think that the flag should be on by default. (That is, "disable XML external entities" should default to "true" in ContextFactory.hasFeature()). But by adding this flag, if someone does indeed depend on Rhino to process external entities, they still can.

Second, especially since we have a flag, I'd really like to see tests that check with the flag set both ways. (Adding a new file to testsrc/org/mozilla/rhino/javascript/tests is a fine way to do it.) I'm not sure how you'd test pulling down an external entity, however, but perhaps there is some other XML feature we can test for?

Third, I'm not 100% sure what we need to do if we cut and paste code from the OWASP GitHub repository, since it uses a different license from Rhino. I think I know a few people to ask.

tophersmith:
@gbrail Thanks for the review. I've incorporated your suggestions. Feature switch is in place and I've added two tests for the PR. Unfortunately, the only way I could figure out a test case is not thread-safe, so we may want to look into how much we want this or if we can add some advice to the test runs. We might have a conflict if two tests that use XML happen simultaneously.

I don't have much to add on the direct copy (#3) point you raised, but OWASP Cheatsheets are under CC3, with allowance on copying and modifications, though I'm not sure if that's on the data sheet or the code as well.

gbrail:
Unfortunately, the license on that GitHub code has the "ShareAlike" flag, set which puts some restrictions on the project if we "build upon" the code. A lawyer I informally consulted with suggested not incorporating that code, and that's the best advice I can get for now.

Sorry to be so literal about this, but these license things are difficult.

I see that Java documents two products that turn off external identification:

https://docs.oracle.com/en/java/javase/12/docs/api/java.xml/javax/xml/parsers/DocumentBuilderFactory.html#setAttribute(java.lang.String,java.lang.Object)

What if we just wrote our own code to set those two?

tophersmith:
Hi Greg, I looked into this a bunch more. I'm committing a new version after reading through a bunch of confusing sun an javax code... The two attributes mentioned aren't quite enough to completely lock down the potential attacks (they don't handle ENTITY values or XIncludes) 

So I've instead rewritten my code from my own research and it's significantly more concise (OWASP's recommendations are semi-tiered it seems, or just exhaustive and overlapping). 

- All Doctype declarations are disabled (which is already kind of true from my first comment) 
- XIncludes and external dtds are disabled in case the underlying parser is configured to allow them. 

I've also extracted the configuration for the TransformerFactory to match the DBF setup.

tophersmith:
Test failure is the timer resolution again

gbrail:
This is very thoroughly done -- thanks! Just to keep things clean, I'm going to squash your commits and merge them. It also looks like we have another flaky test on Travis or it's time to find another CI setup...

tirsoh:
Great to see this fix has been merged! Any idea when we can get a release with this change?

tirsoh:
@gbrail Any update on when we can get a release of this?

gbrail:
It's true that it doesn't take very long, but at the same time there are a
few other things in flight. I'd like to do it around the end of the year.
If that's a big problem then I can try to find more time.

tirsoh:
@gbrail just checked to see if a release was available since it is the new year now but I don't see anything yet. Got any idea when you might be able to get a release cut?

PR #601, opened 2019-09-25
fix unnecssary synchronized issue for NativeJavaPackage.getPkgProperty:

qxo:
not synchronized for get , only synchronized for put


gbrail:
Thanks for looking at this!

I'm not sure I agree that this is unnecessary synchronization. It looks to me that after the change, we are testing the contents of "negativeCache" outside the "synchronized" block, and in that case I don't believe we can guarantee that that object will behave correctly.

qxo:
move "negativeCache"  check code to synchronized block ?

gbrail:
Yes -- I think that the check of "negativeCache" still needs to be inside the "synchronized" block if we need this class to be thread-safe.

On the other hand, regular Rhino objects are no longer thread-safe by default unless configured that way, so we could make an argument that we should do the same thing here.

Can you give us a little more detail on what problem you're trying to fix?

PR #607, opened 2019-10-08
Fix syntax error for comments in array

### LINKED ISSUE #602: Syntax error with comments in array declaration
The following javascript generates a syntax error when recording comments is enabled: var array = [/* */];

It can be solved by removing the "break" at the end of the "if (compilerEnv.isRecordingComments())" block in the method peekToken() in Parser.java .

PalAditya:
Fixes #602 

The error came from the fact that token type comment was not expected within array literal, so added that case to handle that error. Also added multiple test cases in ```testsrc/org/mozilla/javascript/tests/ParserTest.java``` under the function ```testCommentInArray()``` to check all possible cases (single/multiple/no comments, and comments between literals)

Please mention if something else needs to be fixed for this PR 😄

### DIFF
--- a/src/org/mozilla/javascript/Parser.java
+++ b/src/org/mozilla/javascript/Parser.java
@@ -3272,10 +3272,9 @@ public class Parser
                     elements.add(new EmptyExpression(ts.tokenBeg, 1));
                     skipCount++;
                 }
-            }else if(tt == Token.COMMENT) {
+            } else if(tt == Token.COMMENT) {
                 consumeToken();
-            }
-            else if (tt == Token.RB) {
+            } else if (tt == Token.RB) {
                 consumeToken();
                 // for ([a,] in obj) is legal, but for ([a] in obj) is
                 // not since we have both key and value supplied. The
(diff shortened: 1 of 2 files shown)


PalAditya:
Capture
Just adding the local Jacoco coverage for easy reference :)

src/org/mozilla/javascript/Parser.java
```diff
@@ -3272,7 +3272,10 @@ private AstNode arrayLiteral()
                     elements.add(new EmptyExpression(ts.tokenBeg, 1));
                     skipCount++;
                 }
-            } else if (tt == Token.RB) {
+            }else if(tt == Token.COMMENT) {
```
gbrail: If I'm going to suggest only one thing, I'd prefer if changes kept the same coding style (spaces, newlines, etc) as the code around it. Unfortunately applying any one particular style to this ancient codebase is a big job, but if we can at least keep things consistent with the lines around them then we'll all be happier I think.

src/org/mozilla/javascript/Parser.java
```diff
@@ -3272,7 +3272,10 @@ private AstNode arrayLiteral()
...
+                consumeToken();
+            }
+            else if (tt == Token.RB) {
```
gbrail: Same here -- It looks like other code in this file puts "else" on the same line as the preceding brace, with a space between the brace and else...

PalAditya: Noted. Will fix the formatting and push the changes.

gbrail:
Thanks! I'm going to squash these two commits into one and merge them.

PR #608, opened 2019-10-10
Make XML processor configuration more robust.

gbrail:
When the default XML processor is replaced, not all implementations
support the same attributes and features. Make the code robust.
This means not failing if attempts to set certain XML processing
features fail.

### DIFF
--- a/xmlimplsrc/org/mozilla/javascript/xmlimpl/XmlProcessor.java
+++ b/xmlimplsrc/org/mozilla/javascript/xmlimpl/XmlProcessor.java
@@ -67,20 +67,33 @@ class XmlProcessor implements Serializable {
      */
     private void configureSecureDBF(DocumentBuilderFactory dbf){
         try {
+            // This feature is required to be supported by all DocumentBuilderFactories.
+            dbf.setFeature(XMLConstants.FEATURE_SECURE_PROCESSING, true);
+            // Disallow XIncludeAware as it is an SSRF target using xi:include.
+            // This should also be supported on all XML processors.
+            dbf.setXIncludeAware(false);
+        } catch (ParserConfigurationException e) {
+            throw new RuntimeException("XML parser (DocumentBuilderFactory) cannot be securely configured.", e);
+        }
+
+        // The rest of these features should be set for the best security by default.
+        // However, not all XML processing implementations support them.
+        // So we will attempt to set each one but continue if we can't.
+        
+        try {
             //Prevent File attacks in DBF
             //Disallow all doctypes, removing all ENTITY-type tags as a vector
             dbf.setFeature("http://apache.org/xml/features/disallow-doctype-decl", true);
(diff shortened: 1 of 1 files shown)


gbrail:
Some of the XML changes we added didn't work in one of the environments I was able to test this in because some of the XML attributes were not supported. This change reworks the exception handling logic so that not all attempts to set XML parser configuration are fatal errors.

@tophersmith : Can you look and try this out? Thanks!

tophersmith:
I think this is fine. It's a best-effort approach, which is good. 

The only comment I'd make is that I had spent some time looking to add logging statements to talk about what configuration happened/didn't. I know we have that RuntimeException in there, but perhaps you know a way to add logging statements where the parser couldn't be configured properly. Is that a worthwhile addition?

Can you share any info on the DBF/TF implementation that failed in your testing?

gbrail:
Here's the stack trace of what failed:

Caused by: java.lang.IllegalArgumentException: Not supported: http://javax.xml.XMLConstants/property/accessExternalDTD
	at org.apache.xalan.processor.TransformerFactoryImpl.setAttribute(TransformerFactoryImpl.java:571)
	at org.mozilla.javascript.xmlimpl.XmlProcessor.configureSecureTF(XmlProcessor.java:94)
	at org.mozilla.javascript.xmlimpl.XmlProcessor.<init>(XmlProcessor.java:134)
	at org.mozilla.javascript.xmlimpl.XMLLibImpl.<init>(XMLLibImpl.java:110)
	at org.mozilla.javascript.xmlimpl.XMLLibImpl.init(XMLLibImpl.java:45)

This is in some legacy code (everyone has some!) but it's a good example of what we might encounter in the field.

Logging is a more difficult thing. Rhino doesn't log today. (There used to be a few places where it wrote stuff to System.err, and that caused problems, as Rhino gets embedded in a lot of places.) Rhino also doesn't use SLF4j or anything like that for logging, and doesn't depend on anything. So I'm not sure what to do about logging other than to perhaps add some option that would allow users to register a logging callback.

Adding something like that might be a good enhancement request for someone!

tophersmith:
Ok I see what's going on. In general, TransformerFactories allow you to set all the features that they will use when creating their Transformers. Xalan, for whatever reason, only allows the SECURE_PROCESSING and has conditional logic to manually set the correct features on underlying Transformers and other XML factories. So their code is correct, but not in line with other impls and your changes are absolutely necessary as they will honor the SECURE_PROCESSING to actually set the correct flags on the underlying parsers.

I'm approving. Thank you for catching this and the correction.

See: http://svn.apache.org/repos/asf/xalan/java/trunk/src/org/apache/xalan/processor/TransformerFactoryImpl.java
and
http://svn.apache.org/repos/asf/xalan/java/trunk/src/org/apache/xalan/transformer/TransformerIdentityImpl.java

PR #610, opened 2019-10-15
Fix limitations found in static analysis

gbrail:
Fix some of the results found by SpotBugs.


src/org/mozilla/javascript/ScriptRuntime.java
```diff
@@ -137,7 +137,7 @@ public Object call(Context cx, Scriptable scope, Scriptable thisObj,
         Scriptable.class;
 
     // Locale object used to request locale-neutral operations.
-    public static Locale ROOT_LOCALE = new Locale("");
+    public static final Locale ROOT_LOCALE = new Locale("");
```
rbri: Because we are at Java8 this is no longer needed - we can use Locale.ROOT

PR #615, opened 2019-11-08
Fix bug about argment of Number.prototype.toFixed

### LINKED ISSUE #587: RangeError should be throw if the argument of Number.prototype.toFixed is less than 0
### Testcase:
````
1. var foo = function(interval){
2. 	var p = interval.toFixed(-2);
3. 	print(p);
4. };
5. var parameter = -634619;
6. foo(parameter);
````

### Command:
`java -jar rhino/rhino-1.7.11.jar -debug -version 200 testcase.js`

### Output:
`-634600`

### Expected output:
Throw a RangeError exception on Line 2

### Description:
According to the ECMAScript-262 standard, if the argument of Number.prototype.toFixed is less than 0, the engine should throw a RangeError exception. When running the testcase above, rhino does not throw a RangeError exception, but instead outputs -634600. The references of ECMAScript-262 are as follows: 
http://www.ecma-international.org/ecma-262/5.1/index.html#sec-15.7.4.5
http://www.ecma-international.org/ecma-262/9.0/index.html#sec-number.prototype.tofixed

 Is this a bug of Rhino? Looking forward to your reply.

YaoHouyou:
Fixed #587

For the definition of argument of  `Number.prototype.toFixed` in ECMAScript-262, if the argument is less than 0, RangeError should be thrown in ES5, ES5.1, ES9 and the newer version, but it is not a necessary operation in ES6-ES8 because the implementation is permitted to extend. However, rhino does not throw RangeError if version < 200(ES6), and  it throws RangeError if version is 200. They are just opposite.

If rhino originally planned to extend it in ES6,  this is a solution. 

The references of ECMAScript-262 are as follows:
https://tc39.es/ecma262/#sec-number.prototype.tofixed
http://www.ecma-international.org/ecma-262/10.0/index.html#sec-number.prototype.tofixed
http://www.ecma-international.org/ecma-262/9.0/index.html#sec-number.prototype.tofixed
http://www.ecma-international.org/ecma-262/8.0/index.html#sec-number.prototype.tofixed
http://www.ecma-international.org/ecma-262/7.0/index.html#sec-number.prototype.tofixed
http://www.ecma-international.org/ecma-262/6.0/index.html#sec-number.prototype.tofixed
http://www.ecma-international.org/ecma-262/5.1/index.html#sec-11.9.1
[ES5](http://www.ecma-international.org/publications/files/ECMA-ST-ARCH/ECMA-262%205th%20edition%20December%202009.pdf)

### DIFF
--- a/src/org/mozilla/javascript/NativeNumber.java
+++ b/src/org/mozilla/javascript/NativeNumber.java
@@ -146,7 +146,7 @@ final class NativeNumber extends IdScriptableObject
             return ScriptRuntime.wrapNumber(value);
 
           case Id_toFixed:
-            int precisionMin = cx.version < Context.VERSION_ES6 ? -20 : 0;
+            int precisionMin = cx.version < Context.VERSION_ES6 ? 0 : -20;
             return num_to(value, args, DToA.DTOSTR_FIXED, DToA.DTOSTR_FIXED, precisionMin, 0);
 
           case Id_toExponential: {


YaoHouyou:
@gbrail @rbri 
After the bug was fixed #587, I re-queried the ES standard for confirmation. I think the bug is still not fixed. A detailed description can be seen in this issue. I tried to fix the bug, but It failed when run the command `./gradlew check` (because I didn't modify the test case). I hope to help you.

gbrail:
I'm not sure that we need to change anything here.

No other JavaScript engine that I know even tries to offer compliance with all of the various intermediate sets of JavaScript standards. However, Rhino has always had a "version" built in so that programs can maintain backward compatibility. 

Right now, "ES6" is the highest version number. If new language features come along that might break compatibility, then we might need to introduce another one.

But unless that happens, I think that ">= ES6" should be a flag that tells us that Rhino should be compatible with current versions of the language, and pass as much of test262 as we can.

So given all that, while you understand the various changes that have been made in the language over time, I think that the current behavior of having this function return RangeError for invalid inputs is correct because it's what "modern" JS engines do, even though it wasn't in the standard at the exact time that "ES6" was first published.

(I agree that we should write this all up somewhere!)

p-bakker:
Closing this as we're focusing on the latest version of EcmaScript

Issue #616, opened 2019-11-08
In strict mode, ScriptableObject.putImpl(Object,int,Scriptable,Object) prematurely breaks on !isExtensible, setters don't get a chance to be called


stijnkliemesch:
Executing the following code:

    public static void main(String[] cmdlineArgs) {
        final Context context = ContextFactory.getGlobal().enterContext();
        try {
            final Script compiledScript = context.compileString(""
                    + "\"use strict\";\n"
                    + "var a = function(){this._x = 2;};\n"
                    + "Object.defineProperty(a.prototype, \"x\", {get:(function x() {return this._x;}), set:(function x(x) {this._x = x;})});\n"
                    + "Object.freeze(a.prototype);\n"
                    + "var b = new a();\n"
                    + "var c = b.x;\n"
                    + "b.x = 1;\n"
                    + "b.x+c;",
                    "eval", 1, null
            );
            ScriptableObject global = context.initSafeStandardObjects();
            final Object obj
                    = ScriptRuntime.doTopCall(
                            (Context cx, Scriptable scope, Scriptable thisObj, Object[] args) -> compiledScript.exec(context, scope),
                            context, global, null, new Object[]{}, true);
            System.out.println(obj);
            int i = 0;
        } finally {
            Context.exit();
        }
    }

Expected output is:

     3.0

instead, this results in a thrown Exception

    Exception in thread "main" org.mozilla.javascript.EcmaError: TypeError: Cannot add properties to this object because extensible is false. (eval#7)
        at org.mozilla.javascript.ScriptRuntime.constructError(ScriptRuntime.java:4251)
        at org.mozilla.javascript.ScriptRuntime.constructError(ScriptRuntime.java:4229)
        at org.mozilla.javascript.ScriptRuntime.typeError(ScriptRuntime.java:4262)
        at org.mozilla.javascript.ScriptRuntime.typeError0(ScriptRuntime.java:4268)
        at org.mozilla.javascript.ScriptableObject.putImpl(ScriptableObject.java:2813)
        at org.mozilla.javascript.ScriptableObject.put(ScriptableObject.java:521)
        at org.mozilla.javascript.IdScriptableObject.put(IdScriptableObject.java:479)
        at org.mozilla.javascript.ScriptableObject.putProperty(ScriptableObject.java:2507)
        at org.mozilla.javascript.ScriptRuntime.setObjectProp(ScriptRuntime.java:1743)
[...]

PR #617, opened 2019-11-08
Fix for #616

stijnkliemesch:
Affects strict mode only.

No longer prematurely throws a TypeError in ScriptableObject.putImpl(Object,int,Scriptable,Object) when instance member isExtensible is false.

Instead, now it first queries the slotmap, so that if a GetterSlot instance is returned, any present setter may be called. All other outcomes of this method should be unaffected.

### DIFF
--- a/src/org/mozilla/javascript/ScriptableObject.java
+++ b/src/org/mozilla/javascript/ScriptableObject.java
@@ -2804,20 +2804,18 @@ public abstract class ScriptableObject implements Scriptable,
     {
         // This method is very hot (basically called on each assignment)
         // so we inline the extensible/sealed checks below.
-        if (!isExtensible) {
-            Context cx = Context.getContext();
-            if (cx.isStrictMode()) {
-                throw ScriptRuntime.typeError0("msg.not.extensible");
-            }
-        }
         Slot slot;
         if (this != start) {
             slot = slotMap.query(key, index);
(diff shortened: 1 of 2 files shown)


stijnkliemesch:
Do I add the testcase as a new .js file somewhere in test262/test/built-ins/Object/ like freeze/ or preventExtensions/ ?

stijnkliemesch:
Also, after this commit, the tests show the following output:

    org.mozilla.javascript.tests.Test262SuiteTest STANDARD_OUT
        Test is marked as failing but it does not: test262/test/built-ins/Object/freeze/15.2.3.9-2-c-3.js
        Test is marked as failing but it does not: test262/test/built-ins/Object/freeze/15.2.3.9-2-c-2.js
        Test is marked as failing but it does not: test262/test/built-ins/Object/freeze/15.2.3.9-2-c-4.js

Yet the tests seem to "succeed" despite this. Is this a blocking issue?

gbrail:
Those messages are from the Test262 driver. Basically, some tests are disabled in the file test262.properties, but for some reason (I don't recall why) they were run anyway and passed.

That's a good thing -- it means that we pass more of the test262 suite now -- and we want to preserve our progress by re-enabling those tests so that we don't regress.

Can you take a look at test262 properties and see if modifying that file makes the warnings go away? If you can't figure out I can probably do it later.

stijnkliemesch:
Do these 3 extra activated Test262 testcases already cover the issue well enough, or is there still a testcase required for showing #616 is fixed ?

PR #620, opened 2019-11-13
Fixes for calling several Object.prototype members.

stijnkliemesch:
Fixes for calling several Object.prototype members with an owner value of either undefined or null, now throws a TypeError. Affected methods: valueOf, hasOwnProperty, propertyIsEnumerable, isPrototypeOf.

### DIFF
--- a/src/org/mozilla/javascript/NativeObject.java
+++ b/src/org/mozilla/javascript/NativeObject.java
@@ -160,9 +160,15 @@ public class NativeObject extends IdScriptableObject implements Map
           }
 
           case Id_valueOf:
+              if (cx.getLanguageVersion() >= Context.VERSION_1_8 && (thisObj == null || Undefined.isUndefined(thisObj))) {
+                  throw ScriptRuntime.typeError0("msg." + (thisObj == null ? "null" : "undef") + ".to.object");
+              }
             return thisObj;
 
           case Id_hasOwnProperty: {
+              if (cx.getLanguageVersion() >= Context.VERSION_1_8 && (thisObj == null || Undefined.isUndefined(thisObj))) {
+                  throw ScriptRuntime.typeError0("msg." + (thisObj == null ? "null" : "undef") + ".to.object");
+              }
               boolean result;
               Object arg = args.length < 1 ? Undefined.instance : args[0];
               if (arg instanceof Symbol) {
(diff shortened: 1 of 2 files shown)


gbrail:
Thanks!

ECMAscript 3, at least, didn't say anything about throwing TypeError for these calls, at least not in my very brief reading.

https://www.ecma-international.org/publications/files/ECMA-ST-ARCH/ECMA-262,%203rd%20edition,%20December%201999.pdf

I fear that by introducing this change we will break existing code unless we add a "> VERSION_ES6" check to these changes.

stijnkliemesch:
I'm not sure which spec version this was introduced in, but in 5.1 there is this for example:

https://www.ecma-international.org/ecma-262/5.1/

    15.2.4.5 Object.prototype.hasOwnProperty (V)
    
    When the hasOwnProperty method is called with argument V, the following steps are taken:
    
        1. Let P be ToString(V).
        2. Let O be the result of calling ToObject passing the this value as the argument.
        3. Let desc be the result of calling the [[GetOwnProperty]] internal method of O passing P as the argument.
        4. If desc is undefined, return false.
        5. Return true.

The second step there refers to ToObject which is in this spec described as follows:

    9.9 ToObject

    The abstract operation ToObject converts its argument to a value of type Object according to Table 14:

Table 14 — ToObject

Argument Type | Result
-- | --
Undefined | Throw a TypeError exception.
Null | Throw a TypeError exception.
Boolean | Create a new Boolean object whose [[PrimitiveValue]] internal property is set to the value of the argument. See 15.6 for a description of Boolean objects.
Number | Create a new Number object whose [[PrimitiveValue]] internal property is set to the value of the argument. See 15.7 for a description of Number objects.
String | Create a new String object whose [[PrimitiveValue]] internal property is set to the value of the argument. See 15.5 for a description of String objects.
Object | The result is the input argument (no conversion).

So in effect, the values "undefined" and "null" each result in a thrown TypeError because of the ToObject conversion that's called as part of the sequence.

The spec I linked specifies this for the methods affected by dbcdfea.

> I fear that by introducing this change we will break existing code unless we add a "> VERSION_ES6" check to these changes.

Considering the spec, can we then use ">= VERSION_1_8" or perhaps a new ">= VERSION_1_8_5"?

rbri:
> Considering the spec, can we then use ">= VERSION_1_8" or perhaps a new ">= VERSION_1_8_5"?

Not another version please - the version switches are already a nightmare.

gbrail:
I think that adding a check for ">= VERSION_1_8" would be the best way to
implement this without breaking existing code that uses older "versions" in
Rhino.

Overall I think that the strategy should be that we keep things up to date
with the latest specs at version "200" (aka "ES6") and that we should use
the version flags when those specs require changes that would break
existing code.

stijnkliemesch:
Sorry for the messy force-pushes, I'm still learning how to handle git, please ignore 9ae29e4.

gbrail:
In your PR branch I only see one commit and it looks clean to me. I'll run some tests and merge it.

PR #623, opened 2019-11-23
fix for #595;

rbri:
backward compatibility test;
some code cleanup


rbri:
I know there is already a similar PR.
Have done more or less the same but this time with version check and backward compatibility test.

PR #624, opened 2019-11-28
fix Boolean(document.all)

rbri:
found this while trying to merge all the fixes done in htmlunit/core-js back to rhino (if not HtmlUnit specific).

### DIFF
--- a/src/org/mozilla/javascript/NativeBoolean.java
+++ b/src/org/mozilla/javascript/NativeBoolean.java
@@ -72,9 +72,13 @@ final class NativeBoolean extends IdScriptableObject
             if (args.length == 0) {
                 b = false;
             } else {
+                // see special handling in ScriptRuntime.toBoolean(Object)
+                // avoidObjectDetection() is used to implement document.all
+                // see Note on page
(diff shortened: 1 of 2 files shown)


gbrail:
"avoidObjectDetection" is a legacy part of Rhino intended for compatibility with very very old web browsers. But it seems to be used for a few things inside Rhino, including how Boolean is implemented. 

You're the one who understands Rhino's integration with HTML best, but I am a little surprised that this has been broken for so long. I hesitate to do it but I really feel like I have to ask, "are you sure about this?"

rbri:
> with very very old web browsers

But all the current browsers are still returning false for Boolean(document.all).
See the note at https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Boolean

> I am a little surprised that this has been broken for so long.

That is more or less the case for all the fixes/improvements i like to reintegrate from HtmlUnits fork. Looks like Marc and Ahmed are stopping to report pr's (or patches at that time) against rhino somewhere in 2012 because of the slow development at the rhino side.

gbrail:
You're right -- I shouldn't be surprised how far Rhino is behind other things ;-)

PR #625, opened 2019-12-01
Functions as var

rbri:
functions declared as var f = function f() {...} within a function should not impact higher scope variable with the same name


gbrail:
This makes sense. Thanks!

PR #626, opened 2019-12-03
use int instead of short like in many other places, add a test case t…

### LINKED ISSUE #529: ArrayIndexOutOfBoundsException thrown when compile long js
When we upgrade rhino from 1.7.R4 to 1.7.7.1, compilation error occurs for an exists long js which is ok in 1.7.R4.

the js has 2445 lines and 82277 characters, contains java importer, inner function and many comments. Before we upgrde the rhino, it's ok, we try to debug the source code and it's too complex to find out the root cause.

```java
java.lang.ArrayIndexOutOfBoundsException: -32748
        at org.mozilla.classfile.ClassFileWriter$StackMapTable.executeBlock:1731
        at org.mozilla.classfile.ClassFileWriter$StackMapTable.executeWorkList:1712
        at org.mozilla.classfile.ClassFileWriter$StackMapTable.verify:1626
        at org.mozilla.classfile.ClassFileWriter$StackMapTable.generate:1472
        at org.mozilla.classfile.ClassFileWriter.stopMethod:290
        at org.mozilla...optimizer.BodyCodegen.generateBodyCode:1292
        at org.mozilla...optimizer.Codegen.generateCode:306
        at org.mozilla...optimizer.Codegen.compileToClassFile:166
        at org.mozilla...optimizer.ClassCompiler.compileToClassFiles:155
        at com.huawei...script.JsCompiler.processSource:314
        at com.huawei...script.JsCompiler.compile:74
```
the js file structure is like this
```java
var javaImporter = JavaImporter();   

javaImporter.importPackage(Packages.java.lang);
javaImporter.importPackage(Packages.java.util); 
javaImporter.importClass(Packages.com.xxx.xxx.XXX);

with(javaImporter)
{
	function run(message)
	{
               function isEmpty(paramField)
               { 
   	               if("" ==paramField && null == paramField && "null" == String(paramField))
	              {
		              return true;
	              }
	             return false;
                }
                //...
         }

        function _isEmpty(value)
       {
            return value == "";
       }
      //...
}
```

JDK 1.8
Version:1.7.7.1

rbri:
…hat passes after this change

fixes #529

### DIFF
--- a/src/org/mozilla/classfile/ClassFileWriter.java
+++ b/src/org/mozilla/classfile/ClassFileWriter.java
@@ -325,9 +325,9 @@ public class ClassFileWriter {
             index = putInt16(itsExceptionTableTop, codeAttribute, index);
             for (int i = 0; i < itsExceptionTableTop; i++) {
                 ExceptionTableEntry ete = itsExceptionTable[i];
-                short startPC = (short) getLabelPC(ete.itsStartLabel);
-                short endPC = (short) getLabelPC(ete.itsEndLabel);
-                short handlerPC = (short) getLabelPC(ete.itsHandlerLabel);
+                int startPC = getLabelPC(ete.itsStartLabel);
+                int endPC = getLabelPC(ete.itsEndLabel);
+                int handlerPC = getLabelPC(ete.itsHandlerLabel);
                 short catchType = ete.itsCatchType;
                 if (startPC == -1)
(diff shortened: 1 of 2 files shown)


rbri:
the new test runs for nearly one minute - maybe we have to adjust the travis timeout

rbri:
I fear this is too simple, will close this for now.

PR #631, opened 2019-12-10
#511 fixing InterfaceAdapter abstract name lookup.

rPraml:
this is a follow up to #511 and #512 

It fixes the lookup:
- take only abstract methods (ignore default methods)
- skip "equals", "hashCode" and "toString"

There are also tests now.

### DIFF
--- a/src/org/mozilla/javascript/NativeJavaObject.java
+++ b/src/org/mozilla/javascript/NativeJavaObject.java
@@ -692,7 +692,8 @@ public class NativeJavaObject
                 reportConversionError(value, type);
             }
             else if (type.isInterface() && (value instanceof NativeObject
-                    || value instanceof NativeFunction)) {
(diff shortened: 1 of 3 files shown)


src/org/mozilla/javascript/NativeJavaObject.java
```diff
@@ -692,7 +692,8 @@ else if (value instanceof Wrapper) {
                 reportConversionError(value, type);
             }
             else if (type.isInterface() && (value instanceof NativeObject
-                    || value instanceof NativeFunction)) {
+                    || value instanceof NativeFunction
+                    || value instanceof ArrowFunction)) {
```
rPraml: we should also allow to create adapters from arrowFunctions

gbrail:
Looks good. Thanks!

Issue #482, opened 2018-10-09
Does rhino supports yield statements ?


eugenhoro:
I am trying to build AST of the following javascript (es6) file using the code below and I am getting an empty AST. 

function  * foo(x) {
    while (true) {
        x = x * 2;
        yield x;
    }
}

var g = foo(2);
console.log('start')
console.log(g.next())
console.log(g.next())
console.log(g.next())

---------------------
The code for buiding the AST : 

        String fileContent = String.join(System.lineSeparator(), getFileLines(file));
        AstRoot top = null;
        Reporter reporter = new Reporter();
        CompilerEnvirons compilerEnv = new CompilerEnvirons();
        compilerEnv.setErrorReporter(reporter);
        compilerEnv.setReservedKeywordAsIdentifier(true);
        compilerEnv.setIdeMode(true);
        top = new Parser(compilerEnv, compilerEnv.getErrorReporter()).parse(fileContent, file, 1);

lfmunoz:
val script= """
           function execute() {
                var c = 'a'.charCodeAt(0);
                while (c <= 'z'.charCodeAt(0)) {
                    yield String.fromCharCode(c++);
                }
            }
            for (let h in execute()) out.println(h)
        """.trimIndent()

Don't use * to define the generator function.

Issue #411, opened 2018-03-19
Finish ES6 generators work


gbrail:
Some time back I tried to finish ES6 generators by adapting the generator code that is already in there to the new syntax. Some basic things were working but there is some sort of bug in try/catch processing that's causing a code generation problem, and I don't understand the code generation stuff enough to know what it is.

This might be a good issue for someone who is already familiar with the code generation stuff or just someone who wants to learn Rhino -- perhaps you can help figure out what's going wrong!

To reproduce:

1) Check out the branch "greg-generator-2"
2) ./gradlew jar
3) java -jar ./buildGradle/libs/rhino-1.8.0-SNAPSHOT.jar -version 200 testsrc/jstests/harmony/generators-basic.js

You should see an IllegalArgumentException in ClassFileWriter.

Issue #396, opened 2018-02-23
Can not use generator


XsJIONG:
I runned follow javascript code but got a exception:
function* counter() {
  var c=0;
  while (true)
    yield c++;
}
var gen=counter();

After reading the source code(Parser.java), I found that if the parser find a "function" keyword, it requires that the follow symbol must be "(" or it will throw an exception

Plz fix it!

sainaen:
Rhino implements older version of iterators and generators, not the ones introduced in ES6/ES2015. In this version, you don’t need an asterisk:
```js
function counter() {
  var c = 'a'.charCodeAt(0);
  while (c <= 'z'.charCodeAt(0))
    yield String.fromCharCode(c++);
}
for (let h in counter()) print(h);
// prints out letters from 'a' to 'z'
```
For more examples you can check out older versions of [MDN page on Iterators and generators](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Iterators_and_Generators$revision/109528).

I agree, it would be great to have this updated, but there’s just not many people working on Rhino right now.

Xs-JIONG:
I tried your code but also got an exception
org.mozilla.javascript.EvaluatorException: missing ; before statement (Main.js#4)
How to fix it?

sainaen:
Weird, I’m not getting any errors. How do you run it?

PR #633, opened 2019-12-18
Add support for ES6 generators.

gbrail:
This includes "function *", "yield", and "yield *" support.
It is only activated if the language level is <= "LANGUAGE_ES6",
or "200". At lower language levels, the "yield" keyword still
triggers the pre-ES6 generator support that Rhino supported in
previous releases.

This addresses a few issues:

https://github.com/mozilla/rhino/issues/158
https://github.com/mozilla/rhino/issues/396
https://github.com/mozilla/rhino/issues/411
https://github.com/mozilla/rhino/issues/482


gbrail:
I'd appreciate it if anyone has time to try this out. @rbri you commented, perhaps you can see what we missed -- it's a pretty complicated thing that's been sitting in my Git client for years now!

src/org/mozilla/javascript/ScriptRuntime.java
```diff
@@ -2627,6 +2627,19 @@ public static Object callIterator(Object obj, Context cx, Scriptable scope)
...
+     * Given an iterator result, return true if and only if there is a "done"
+     * property that's true.
+     */
+    public static boolean isIteratorDone(Context cx, Object result)
+    {
+        if (!(result instanceof Scriptable)) {
+            return false;
+        }
+        final Object prop = getObjectProp((Scriptable)result, "done", cx);
+        return toBoolean(prop);
```
rbri: Maybe a constant for "done" somewhere

src/org/mozilla/javascript/Parser.java
```diff
@@ -1913,11 +1925,19 @@ private AstNode returnOrYield(int tt, boolean exprContext)
...
+            (peekToken() == Token.MUL)) {
+          yieldStar = true;
+          consumeToken();
+        }
+
         AstNode e = null;
         // This is ugly, but we don't want to require a semicolon.
         switch (peekTokenOrEOL()) {
           case Token.SEMI: case Token.RC:  case Token.RB:    case Token.RP:
-          case Token.EOF:  case Token.EOL: case Token.ERROR: case Token.YIELD:
```
rbri: Is this Lang version dependant?

src/org/mozilla/javascript/ScriptRuntime.java
gbrail: I was being contrarian there since it's not like we're going to have to change the value of "done" some day. But I agree that it probably makes sense to constant-ize it!

src/org/mozilla/javascript/Parser.java
gbrail: I don't think so because "yield" has been a supported keyword for many Rhino releases now. Let me run a quick test and check.

src/org/mozilla/javascript/ScriptRuntime.java
rbri: But the string is used at several places with the same semantics.

rbri:
@gbrail any plans for this to go to the head?

gbrail:
Although all this works, the "GeneratorFunction" business and the generator
function prototype chain in general is wrong in the case of compiled code,
and it'll actually take a pretty big change to the compiled bytecode to
make it work properly. (It also might make the bytecode less
convoluted which is not a bad thing anyway.) So, I didn't want to hold back
the release.

I'll push it soon and work on fixing the bytecode but I don't get too much
time for this so it might be a few weeks if I'm being realistic.

rbri:
Ok, I'm planning a new Htmlunit release for this weekend. Will do it without this, there are already enough rhino improvements. Thanks for working on this.

gbrail:
The commit history has gone pretty off on this one so I'm going to replace it with a cleaner one.

PR #635, opened 2019-12-24
Dynamic scope fix

stijnkliemesch:
A given scope was ignored by the implementations of Object.create and Object.defineProperties.

The important part was in "case ConstructorId_create:" that newObject.setParentScope(scope); happens, instead of "inheriting" the scope from the given prototype object (first parameter of Object.create).

The rest is only for completion sake. Implementation of Object.defineProperties and handling of the second argument of Object.create both end up ignoring the given scope anyway. Even if Context.toObject(propsObj, scope) returns a new object that has the given scope, the call to obj.defineOwnproperties(cx, ensureScriptableObject(props)) does nothing with it. Only in some fringe edge case scenario with a custom WrapperFactory impl would that matter.

### DIFF
--- a/src/org/mozilla/javascript/NativeObject.java
+++ b/src/org/mozilla/javascript/NativeObject.java
@@ -433,7 +433,7 @@ public class NativeObject extends IdScriptableObject implements Map
                 Object arg = args.length < 1 ? Undefined.instance : args[0];
                 ScriptableObject obj = ensureScriptableObject(arg);
                 Object propsObj = args.length < 2 ? Undefined.instance : args[1];
-                Scriptable props = Context.toObject(propsObj, getParentScope());
+                Scriptable props = Context.toObject(propsObj, scope);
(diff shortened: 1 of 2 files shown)


stijnkliemesch:
I'm not sure how to adhere to the DYNAMIC_SCOPE context feature in the code, should there instead be an if(cx.hasFeature(DYNAMIC_SCOPE)) construct to choose between either getParentScope() or the given scope?

gbrail:
Dynamic scope is one of those obscure features that I wasn't around for and don't know the details of. The only place it's referenced is in ScriptRuntime.doTopCall, and I would imagine that we don't need to worry about it elsewhere in the code. You have added a test case and not broken anything else so I'm inclined to go ahead with this unless you think we need more validation.

stijnkliemesch:
In that case I think the feature is fine as-is, no need to complicate things.

gbrail:
I agree -- that makes sense. Thanks!

PR #636, opened 2019-12-28
many more tests are passing already and some cleanup

rbri:


gbrail:
Love to have more tests. Thanks!

PR #640, opened 2020-01-08
Fix handling of "return" in iterators.

gbrail:
Clean up iterator handling for Array, Map, and Set so that it
properly handles "undefined" values for "return".
Add tests for negative cases.


gbrail:
@rbri you will be interested in this since you work a lot with Arrays and it makes Array.from work more properly.

rbri:
Thanks, have seen it already. Will check if I can use the impl from htmlunit also.
As soon as I c find some time I will make another pr for __proto__ handling.

rbri:
Looks like typedarray-arg-other-ctor-buffer-ctor-species-null.js is passing.

PR #666, opened 2020-05-09
fix string.split with limit 0

rbri:


thysce:
Lgtm

gbrail:
Agreed. Thanks for fixing this!

Issue #665, opened 2020-05-09
A strange testcase about RegExp Object of rhino.


ty5491003:
### Testcase & Output
testcase1:
```js
var a = new RegExp("1");
print(a(1));
```
output1:
```js
null
```

testcase2:
```js
var a = new RegExp("1");
print(a(2));
```

output2:
```js
1
```

### Other JS Engines Output
```js
TypeError: a is not a function.
```

### Rhino Version
1.7.11

### Description
In my opinion, `a` is an Object, not a function. I don't know why rhino can execute these two testcases without throwing an exception and get two different result? Maybe it's a bug? or a feature?

So I need your help. Looking forward to your reply, thanks!

ty5491003:
I see this merge, thanks for your work.
Closed.

PR #668, opened 2020-05-11
fix for issue #665 (maybe we have to adjust the version switch to ver…

rbri:
…sion 1_6)


rbri:
fixes #665

gbrail:
LGTM. Thanks!

PR #669, opened 2020-05-12
fix for the recursion detection when converting an array into a string

### LINKED ISSUE #659: An issue about Array.prototype.map
###### Revision
rhino-1.7.11

###### Test case
```
var f = function() {
    var obj = [0,1];
    var a = obj.map(function() {return obj;});
    print(a);
};
f();
```

###### Execution steps
    java -jar rhino-1.7.11.jar -debug -version 200 testcase.js

###### Output
 ```
0,1,
```
###### Expected behavior
 ```
0,1,0,1
```

###### Description
According to ES Standard, each element in the array `obj` should execute the callback function of `Array.prototype.map`. Therefore, when running this testcase blow, the output of rhino should be `0,1,0,1`.
The references are as follows:
ecma-international.org/ecma-262/5.1/index.html#sec-15.4.4.19

rbri:
fixes #659

### DIFF
--- a/src/org/mozilla/javascript/NativeArray.java
+++ b/src/org/mozilla/javascript/NativeArray.java
@@ -979,7 +979,9 @@ public class NativeArray extends IdScriptableObject implements List
         // so we don't leak memory
         try {
             if (!iterating) {
-                cx.iterating.put(thisObj, 0); // stop recursion.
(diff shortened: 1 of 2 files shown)


gbrail:
Cool fix. Thanks!

Issue #656, opened 2020-04-04
An issue about Object.prototype.isPrototypeOf


YiWen-y:
###### Revision
rhino-1.7.11

###### Test case
```
var f = function (){
    print(typeof Object.prototype.isPrototypeOf);
    var a = Object.prototype.isPrototypeOf.call(undefined, []);
    print("a:", a);
};
f();
```

###### Execution steps
    java -jar rhino-1.7.11.jar -debug -version 200 testcase.js

###### Output
 ```
function
a: false
```
###### Expected behavior
Throw a Type Error in line 3.

###### Description
According to ES standard, the `this` value should be converted via `ToObject` in case of the `isPrototypeOf` call . The `ToObject` should throw and error for `undefined` values.

rbri:
Have added a pull request but i can't reproduce your finding. What is wrong with the test?

rbri:
This is fixed now. Please close.

PR #672, opened 2020-05-19
add testcase for issue #656

rbri:


gbrail:
I'm in favor of adding more tests even for things that already work!

PR #675, opened 2020-05-19
Symbol.length is 0

### LINKED ISSUE #648: A bug about Symbol.length
###### Revision
rhino-1.7.11

###### Test case
```
var NISLFuzzingFunc = function() {
    print(typeof Symbol);
    print(Symbol.length);
};
NISLFuzzingFunc();
```

###### Execution steps
    java -jar rhino-1.7.11.jar -debug -version 200 testcase.js

###### Output
```
function
1
```

###### Expected behavior
```
function
0
```

###### Description
`Symbol` was first defined in ES6, and according to ES standard, the length property value of Symbol Constructor is 0. However, when running this test case , the output of rhino is 1.
The references are as follows:
http://www.ecma-international.org/ecma-262/6.0/index.html#sec-properties-of-the-symbol-constructor

rbri:
fixes #648

### DIFF
--- a/src/org/mozilla/javascript/NativeSymbol.java
+++ b/src/org/mozilla/javascript/NativeSymbol.java
@@ -159,7 +159,7 @@ public class NativeSymbol
     {
         switch (id) {
         case Id_constructor:
-            initPrototypeMethod(CLASS_NAME, id, "constructor", 1);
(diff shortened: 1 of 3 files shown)


gbrail:
Thanks for all the help!

PR #677, opened 2020-05-20
improve seal() and freeze() processing;

### LINKED ISSUE #676: Object.seal() behavior is not correct
According to the specification, the values of present properties of a sealed object can be changed as long as they are writable. The code below runs correctly in real browsers but fails in rhino with EcmaError: TypeError: Cannot add properties to this object because extensible is false.
```javascript
fun=function() 
{ 
        "use strict"; 
        var r = {}; 
        Object.defineProperties(r, 
        { 
                a: { writable: true, value: 'aaa' } 
        }); 
        r.a='abefore'; 
        Object.seal(r); 
        r.a='aafter'; 
}; 
fun();
```

rbri:
fixes #676


gbrail:
Not sure what happened on Travis, but this works fine. Thanks!

PR #681, opened 2020-05-22
An error should be thrown when defining a property for a read-only va…

### LINKED ISSUE #573: An error should be thrown when defining a property for a read-only variable in strict mode.
As the title suggests, the testcase is:
```
var func = function (number) {
    "use strict";
    number.snippetText = "";
};
func(-334918463);
```
The defination and assignment at the third line leads to an exception when I run this testcase in V8 and SpiderMonkey. But Rhino doesn't throws any exception.

rbri:
…riable in strict mode

fixes #573


gbrail:
Thanks!

PR #682, opened 2020-05-23
Side effect fix

rbri:
@gbrail - see the comment on the second commit (it was a nightmare to hunt for this)


gbrail:
I understand this -- thanks for figuring it out. I can see why it'd be a problem, not only for tests, but I can imagine that it would cause incorrect script behavior as well.

gbrail:
This is a nice fix and it makes some flaky tests less flaky. Thanks!

PR #683, opened 2020-05-24
various fixes for array calls using this pattern

### LINKED ISSUE #673: A rhino's implementation does not meet ES standards about 'Array.prototype.concat'
### Testcase
```js
var a = Array.prototype.concat.call(null, [1]);
```

### Expect Output
A TypeError should be thrown.

### Rhino Actual Output
No exception is thrown.

### Rhino Version
1.7.12 (latest release) &
34cf08c (lastest commit version)

### ES Standard
According to ES standard(5.1, 6, and 10), The first step of `Array.prototype.concat` is:

>1. Let O be ? ToObject(this value).

In this testcase, `ToObject(null)` should throw a TypeError, but rhino didn't do it. So I suspect it is a bug of rhino.

Looking forward to your reply, thanks.

rbri:
Array.prototype.foo.call(null, ....);

fixes #673

### DIFF
--- a/src/org/mozilla/javascript/NativeArray.java
+++ b/src/org/mozilla/javascript/NativeArray.java
@@ -316,25 +316,25 @@ public class NativeArray extends IdScriptableObject implements List
                 return toStringHelper(cx, scope, thisObj, true, false);
 
               case Id_join:
-                return js_join(cx, thisObj, args);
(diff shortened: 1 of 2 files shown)


gbrail:
Thanks!

PR #693, opened 2020-05-31
fix handling of undefined parameter value in String.normalize()

### LINKED ISSUE #692: A rhino's implementation does not meet ES standards about 'String.prototype.normalize'
#### Version

> 1.7.12 (latest release)

#### Test case

```javascript
print('123'.normalize('NFC')); 
print('123'.normalize(undefined)); 
```

#### Execution steps

#### Rhino Output

 ```bash
123
js: "/export/xxx/temp/xxxx.js", line 1: uncaught JavaScript runtime exception: RangeError: The normalization form should be one of NFC, NFD, NFKC, NFKD
	at /export/xxx/temp/xxxx.js:1
 ```

#### Expected Output

 ```bash
123
123
 ```

#### Description

According to ES Standard(6, and 10), The steps of `String.prototype.normalize ([form])`  are:

> 1. Let O be ? [RequireObjectCoercible](http://ecma-international.org/ecma-262/10.0/index.html#sec-requireobjectcoercible)(this value).
> 2. Let S be ? [ToString](http://ecma-international.org/ecma-262/10.0/index.html#sec-tostring)(O).
> 3. **If form is not present or *form* is undefined, set form to `"NFC"`.**
> 4. Let f be ? [ToString](http://ecma-international.org/ecma-262/10.0/index.html#sec-tostring)(form).
> 5. If f is not one of `"NFC"`, `"NFD"`, `"NFKC"`, or `"NFKD"`, throw a RangeError exception.

I think the implementation of Rhino is missing step3. Maybe you can fix it.

rbri:
fixes #692

### DIFF
--- a/src/org/mozilla/javascript/NativeString.java
+++ b/src/org/mozilla/javascript/NativeString.java
@@ -472,14 +472,18 @@ final class NativeString extends IdScriptableObject
                 }
                 case Id_normalize:
                 {
+                    final Normalizer.Form form;
+                    if (args.length > 0 && Undefined.isUndefined(args[0])) {
+                        return Normalizer.normalize(ScriptRuntime.toString(requireObjectCoercible(cx, thisObj, f)), Normalizer.Form.NFD);
+                    }
+
                     String formStr = ScriptRuntime.toString(args, 0);
 
-                    Normalizer.Form form;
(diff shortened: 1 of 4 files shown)


gbrail:
This looks good but it has a merge conflict with what's now in master. Can you fix it? Thanks!

PR #694, opened 2020-06-01
it should not be possible to change the [[Prototype]]

rbri:
of a non-extensible object;
some cleanup

based on commit 6267a13de8c2123604f1b7ff5b9b6d9e516f805c from anba


gbrail:
Is this going to break apps that previously worked with language versions < ES6? I don't know how prevalent this sort of thing is but it often seems that if Rhino allows something then someone has already done it.

PR #696, opened 2020-06-02
fix all this-value-not-obj-coercible.js tests for string

rbri:


gbrail:
This works fine -- looks like we need to figure out what's up with Travis once again. Thanks!

PR #703, opened 2020-06-06
Implement a ScriptEngine interface for Rhino

gbrail:
This is the beginning of a ScriptEngine implementation for Rhino. It's not based on the old Sun code, which we can't use for license reasons.

It's a bit opinionated -- by default, it uses the "ES6" language version, and has debug and optimizations enabled. 

It has one bulit-in function so far -- "print." I'm reluctant to include all of the built-in functions from the shell's "Global" class because some of them do dangerous things that I think we should not enable by default.

There is work to do if we want more built-in functions. There are also unimplemented methods on the ScriptEngineFactory, namely the ones that turn generic bits of code into JavaScript.


build.gradle
```diff
@@ -132,9 +131,30 @@ task runtimeJar(type: Jar) {
...
+        include 'META-INF/services/**'
+    }
+    manifest {
+        attributes(
+                "Manifest-Version": "1.0",
+                "Implementation-Version": project.version,
+                "Implementation-Title":  "Mozilla Rhino ScriptEngine",
+                "Implementation-Vendor": "Mozilla Foundation",
+                "Implementation-URL": "http://www.mozilla.org/rhino",
+                "Built-Date": new Date().format("yyyy-MM-dd"),
```
ebourg, line 148: Adding a timestamp to the manifest is a bad idea, it makes the jar non reproducible.

PR #712, opened 2020-06-11
Support making JavaScript functions out of Java lambdas

gbrail:
This is a cleaner way to make idiomatic JavaScript objects out of Java


gbrail:
FWIW, this whole thing makes native classes a lot easier to write and they perform just barely worse than IdScriptableObject and a lot better than anything that uses reflection.

However, it makes serialization even more complicated, so I'll get back to this after we figure out what to do about that.

PR #718, opened 2020-06-20
travis ci: switch from trusty to xenial + set explicit -Xss in tests

syjer:
Hi, this PR fix the travis CI build. See https://travis-ci.org/github/syjer/rhino/builds/700382792 .

Currently (https://travis-ci.org/github/mozilla/rhino/builds/700365658#L530) it fails due to reaching the limit of the stack size by most likely a small amount.

Setting it explicitly in the test task (with a value that is in the normal range) fix the build.

I've also updated the distribution.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: java
-dist: trusty
+dist: xenial
 jdk: 
   - openjdk8
 script:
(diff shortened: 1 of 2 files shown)


gbrail:
Makes sense to me. Thanks!

syjer:
uh, oops, sorry, it seems that it has broken on circle ci, I guess I need to adjust the value with a bigger value.

I'll provide another PR that cover there too.

Issue #717, opened 2020-06-20
Proposal: Reduce short living allocations


A-Herzog:
The implicit memory management of Java allows to implement algorithms very straight forward but relying too much on this concept can result in very heavy garbage collector loads.

In my JS workloads I can see (using Flight Recorder) there are two lines of code in Rhino which result in 90 GB of short living allocations per minute. For both cases I would like to suggest improvements which will reduce these allocations to zero (and therefore reduce the time needed to collect them in gc to zero). The disadvantage is, that in both cases some additional code is necessary, so the clearness of the methods is reduced.

### org.mozilla.javascript.ScriptRuntime

When executing an "i++" in JS the previously maybe as Integer stored "i" becomes a Double. (The increment method tests if i is a Number and in this case `Number.doubleValue()` is used.) Boxing small integer values does not need any allocation because `Integer.valueOf(int)` is using a cache. `Double.valueOf(double)` always needs to allocate a Double object, so a simple `for (var i=0;i<10;i++)` results in 10 Double allocations. Even more: When creating a JS variable by "var i=3;" it will be stored as an Integer. But the "int i=0" in the for loop generates a Double - for some unknown reasons. So "i++" will operate on Double all the time.

By using the following code in ScriptRuntime the unnecessary use of Double variables can be avoided. (Please note the "Added code starts/ends here" markers. Everything else is not changed.)

```javascript
private static Object doScriptableIncrDecr(Scriptable target,
		String id,
		Scriptable protoChainStart,
		Object value,
		int incrDecrMask)
{
	boolean post = ((incrDecrMask & Node.POST_FLAG) != 0);

	// Added code starts here
	if (value instanceof Integer) {
		int intNumber = ((Integer)value).intValue();
		if ((incrDecrMask & Node.DECR_FLAG) == 0) {
			++intNumber;
		} else {
			--intNumber;
		}
		final Integer intResult = wrapInt(intNumber);
		target.put(id, protoChainStart, intResult);
		return post?value:intResult;
	}

	if (value instanceof Double) {
		final double d=((Double)value).doubleValue();
		if (Math.floor(d)==d && Math.abs(d)<NativeNumber.MAX_SAFE_INTEGER) {
			int intNumber = (int)d;
			if ((incrDecrMask & Node.DECR_FLAG) == 0) {
				++intNumber;
			} else {
				--intNumber;
			}
			final Integer intResult = wrapInt(intNumber);
			target.put(id, protoChainStart, intResult);
[...]

PR #720, opened 2020-06-21
implement first proposal from issue #717

rbri:


rbri:
Have closed this because it looks like the test suite fails and is slower. Have to do a more detailed analysis of the impact.

PR #721, opened 2020-06-21
avoid some auto boxing

rbri:
use Double.valueOf() instead of new Double()
some cleanup
try to optimize the code a bit to avoid unnecessary conversations and Double object creation
make some methods static


gbrail:
This all makes sense to me -- lots less garbage! Let me test it locally and merge.

PR #724, opened 2020-06-22
try to fix circle, increase Xss

syjer:
Hi, this PR increase a little bit more the Xss for circle ci.

Currently it fail https://app.circleci.com/pipelines/github/mozilla/rhino/39/workflows/97f3c138-2e1e-4f09-9b7e-e43580af8d2f/jobs/44/steps , with the increase, it seems to pass there too https://app.circleci.com/pipelines/github/syjer/rhino/4/workflows/b7280c33-48da-4587-af30-61096c5307ba/jobs/4/parallel-runs/0/steps/0-107 (I hope it's public).

Looks like there is a difference in the version of the openjdk 8 in use on both ci environment, I would guess this can cause this difference in behavior.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -67,7 +67,7 @@ test {
     useJUnit()
     exclude "**/benchmarks/**"
     
-    jvmArgs += '-Xss1152k'
+    jvmArgs += '-Xss1280k'
 
     jacoco.excludes = ['**/testsrc_tests_ecma_3_RegExp_perlstress*']
 


gbrail:
It doesn't hurt much to try this, although I suspect either that test is no longer a good one or we need to fix something deeper.

syjer:
I would guess the [test](https://github.com/mozilla/rhino/blob/master/testsrc/tests/js1_5/Regress/regress-98901.js) make sense, as it highlight the limit of using recursion in the `IRFactory` for traversing/transforming the nodes (in this specific case https://github.com/mozilla/rhino/blob/master/src/org/mozilla/javascript/IRFactory.java#L812 + https://github.com/mozilla/rhino/blob/master/src/org/mozilla/javascript/IRFactory.java#L221 ) in long expressions.

Issue #704, opened 2020-06-06
Proposal: Get rid of all serialization in Rhino


gbrail:
Rhino tries to make all JavaScript objects and other objects Serializable, because it's always done that.

I think that we should consider getting rid of Serialization support for a future release for a few reasons:

1. There is a pretty big category of security holes that are created when things are Serializable.
2. Keeping things Serializable is a maintenance burden and makes it harder for us to change internal structures.
3. Others in the Java community are considering removing it from Java some day.
4. Most other JavaScript engines handle this using JSON.

Can people who actually Serialize and de-serialize Rhino objects please comment here on whether  you are using Serialization, and how?

rbri:
HtmlUnit does not use serialization of JavaScript objects.
Everything that gives us more freedom to change the code base is very welcome.

gbrail:
I'd actually still like to do this -- serialization is being more and more generally acceptable as being terrible for interoperability and security. Unless someone has code that will break I am going to create a new PR soon.

michbarsinai:
Combined with continuations, serialization is a super-powerful tool, as it allows a single machine to run virtually unlimited amount of threads. This is by capturing and serializing continuations, and storing them in a DB. Think of a server that runs multiple processes based on external events. I've heard of at least one system that's doing this using Rhino. I know there was also an experimental project that used serialized continuations to move a computation between servers, but I'm not sure if it ever reached production.

In BPjs we depend on serialized continuations for re-trying computations.

I of course understand the complications serializations add, but given the power they give to Rhino, I'd be happy to see them stay.

tonygermano:
The Continuations API announcement specifically mentioned serialization https://web.archive.org/web/20210226081112/https://developer.mozilla.org/en-US/docs/Mozilla/Projects/Rhino/New_in_Rhino_1.7R2

Serializing Continuations has also appeared in a few SO questions that I came across:
https://stackoverflow.com/questions/8616512/interpreting-javascript-in-java-with-rhino-pausing-resuming-scripts
https://stackoverflow.com/questions/5221175/is-rhino-the-only-interpreter-with-support-for-sandboxing-and-serializable-conti

p-bakker:
@gbrail based on the discussion here and in the [Next steps for Rhino after the v1.7.14 release](https://github.com/mozilla/rhino/discussions/972) discussion, it looks like serialization, at least in the context of Continuations is something that is still desirable.

[...]

PR #727, opened 2020-06-24
Remove serialization.

gbrail:
Get rid of every class that implements "Serializable" and associated
machinery for readObject/writeObject and the like. Get rid of tests
that assume it's there (not very many of those actually).

Remove the "serialize" and "deserialize" commands from the shell.

Remove SerializableInputStream and SerializableOutputStream.


gbrail:
@rbri and others who are regularly running the latest Rhino -- any chance you can try this and see if it causes big problems? It removes some classes and support that were previously supported although I think that they were generally not used.

rbri:
@gbrail did a first quick test with you PR applied to the HtmlUnit branch. Looks correct so far - no regressions. I only had failing tests because HtmlUnit supports serialization of dom objects (Scriptables). This was expected and is something i have to solve inside HtmlUnit.
For me it is fine to go ahead with this.

rbri:
As this is a major change, maybe we can announce the plan and doing the next release without this. Then we will merge this directly after the release to have this for some time in the snapshots.

gbrail:
Yes, I was thinking that it would be something for 1.8.0. And if it's
really useful, we can work on a better way to do it -- but, for instance, I
noticed that when you serialize an object it just keeps going and going
until you are doing things like serializing the internal state of the
interpreter, and I'm not sure that doing that really makes a lot of sense.
Perhaps we can come up with a better way to serialize objects that supports
your use case but doesn't serialize a bunch of stuff that I imagine you
can't use.

tonygermano:
See https://github.com/mozilla/rhino/issues/704#issuecomment-874842479 from linked issue. It mentions that serialization is currently being used with continuations.

EduFrazao:
Hi Folks. What is the goal/benefit of remove serialization support?
My application makes heavy use of server-side scripting, and they are hold into the application-server sessions.
We are evaluating support the distrubuition of http-sessions, wich will rely on serialize everything. This change will cause impact those scenario rigth?

p-bakker:
Converted to draft, as has conflicts and open discussions

PR #728, opened 2020-06-25
regular expressions are not functions in the context of string replace

### LINKED ISSUE #726: An issue about String.prototype.replace
##### Version

rhino-1.7.12

##### Test case

```js
var NISLFuzzingFunc=function (){
    var a = 'test'.replace('e', /^/); 
    print(a); 
    var b = 'test'.replace('e', String(/^/))
    print(b);
};
NISLFuzzingFunc();
```

##### Execution steps

##### Output

```
tst
t/^/st
```

##### Expected behavior

```
t/^/st
t/^/st
```

##### Description

The ES standard stipulates that when the `String.prototype.replace (searchValue, replaceValue)` method is executed, if the second parameter`replaceValue` is not a function, it should be converted into a string first. rhino seems to be missing this step.

Contributor: [@Wen Yi](https://github.com/YiWen-y)

rbri:
fixes #726


gbrail:
Thanks!

Issue #352, opened 2017-10-26
Make Rhino work with Java 9


gbrail:
It just doesn't. Java 9 is supposed to be backward compatible, so we have to figure this out.

gbrail:
Upgrading Gradle helps a lot.

There are a number of new test failures around String.prototype.trim, and a few date-parsing test failures. The rest passes.

gbrail:
Sounds like the main test failures are coming from the fact that Java 9 uses a different set of Unicode encodings by default than previous Java releases. These might not be actual test failures but a different interpretation of Unicode from what was done before versus what's done in Java 9.

http://openjdk.java.net/jeps/252

p-bakker:
@gbrail Think this one can be closed? Builds on Java versions >8 & < 16 seem to work just fine no?

gbrail:
Yep -- works and we're adding CI.

PR #729, opened 2020-06-26
Enable java 11 build support

syjer:
Hi, this PR add java 11 in the travis test matrix.

There are 3 fixes:
 - removal of a character in `string.trim.doctest` which is no more considered as a whitespace. See https://github.com/mozilla/rhino/commit/7ec73c035668398aa02839eaa0717b917adb0f36 . To be noted, the test with the problematic character fail on chrome and FF, so it was likely obsolete/no more correct.
 - use of explicit format for DateFormat in `NativeDate`. During the java8 -> java9 transition, the pattern defined by using `FormatStyle.LONG` has changed. See https://github.com/mozilla/rhino/commit/64790dde7908e84a8b87328d99d87de076598b33 and https://stackoverflow.com/q/53317365
 - fix on the javadoc to make the linter happy

Everything build successfully: https://travis-ci.org/github/syjer/rhino/builds/702403837 :)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ language: java
 dist: xenial
 jdk: 
   - openjdk8
+  - openjdk11
 script:
   - travis_wait 30 ./gradlew check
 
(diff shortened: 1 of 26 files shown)


gbrail:
Thanks for tracking this down!

syjer:
you're welcome. I guess https://github.com/mozilla/rhino/issues/352 can be considered closed? :)

PR #731, opened 2020-06-27
Add String Support

### LINKED ISSUE #730: Add String.fromCodePoint
Add String.fromCodePoint()

Docs:
 - [String.fromCodePoint() - MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/fromCodePoint)
 - [ECMAScript (ECMA-262)
The definition of 'String.fromCodePoint' in that specification.](https://tc39.es/ecma262/#sec-string.fromcodepoint)

ian4hu:
fixed #730

### DIFF
--- a/src/org/mozilla/javascript/NativeString.java
+++ b/src/org/mozilla/javascript/NativeString.java
@@ -236,23 +236,25 @@ final class NativeString extends IdScriptableObject
                 }
 
                 case ConstructorId_fromCodePoint: {
-                    int N = args.length;
-                    if (N < 1)
+                    int n = args.length;
+                    if (n < 1) {
                         return "";
-                    int[] codePoints = new int[N];
-                    StringBuilder sb = new StringBuilder(N);
-                    for (int i = 0; i != N; i++) {
+                    }
+                    int[] codePoints = new int[n];
+                    StringBuilder sb = new StringBuilder(n);
+                    for (int i = 0; i != n; i++) {
                         codePoints[i] = ScriptRuntime.toInt32(args[i]);
                     }
-                    return new String(codePoints, 0, N);
+                    return new String(codePoints, 0, n);
                 }
 
                 case ConstructorId_fromCharCode: {
-                    int N = args.length;
-                    if (N < 1)
+                    int n = args.length;
+                    if (n < 1) {
                         return "";
-                    char[] chars = new char[N];
-                    for (int i = 0; i != N; ++i) {
+                    }
+                    char[] chars = new char[n];
+                    for (int i = 0; i != n; ++i) {
                         chars[i] = ScriptRuntime.toUint16(args[i]);
                     }
                     return new String(chars);
--- a/src/org/mozilla/javascript/NativeString.java
+++ b/src/org/mozilla/javascript/NativeString.java
@@ -251,11 +251,11 @@ final class NativeString extends IdScriptableObject
                     int N = args.length;
                     if (N < 1)
                         return "";
-                    StringBuilder sb = new StringBuilder(N);
+                    char[] chars = new char[N];
                     for (int i = 0; i != N; ++i) {
-                        sb.append(ScriptRuntime.toUint16(args[i]));
+                        chars[i] = ScriptRuntime.toUint16(args[i]);
                     }
-                    return sb.toString();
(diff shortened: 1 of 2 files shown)


src/org/mozilla/javascript/NativeString.java
```diff
@@ -251,11 +251,11 @@ public Object execIdCall(IdFunctionObject f, Context cx, Scriptable scope,
                     int N = args.length;
                     if (N < 1)
                         return "";
-                    StringBuilder sb = new StringBuilder(N);
+                    char[] chars = new char[N];
```
ian4hu, line 254: Since we know how many chars, we just new a string on it

src/org/mozilla/javascript/NativeString.java
```diff
@@ -231,6 +233,18 @@ public Object execIdCall(IdFunctionObject f, Context cx, Scriptable scope,
                     continue again;
                 }
 
+                case ConstructorId_fromCodePoint: {
+                    int N = args.length;
```
rbri, line 239: please let us follow the java code style and start variable names with lowercase char

src/org/mozilla/javascript/NativeString.java
```diff
@@ -231,6 +233,18 @@ public Object execIdCall(IdFunctionObject f, Context cx, Scriptable scope,
...
+                    if (N < 1)
```
rbri, line 240: Please add the {}

testsrc/jstests/es6/string.js
```diff
@@ -22,4 +22,8 @@ load("testsrc/assert.js");
   assertEquals(2, "abcdEfg".search("cd", "not used"));
 })();
 
+(function TestFromCodePoint() {
+    assertEquals('☃★♲你', String.fromCodePoint(9731, 9733, 9842, 0x2F804))
```
rbri, line 26: maybe we can make this more robust by using the hex value here - otherwise we have to fight with the js file encoding sooner or later

src/org/mozilla/javascript/NativeString.java
ian4hu: Oh, I just keep the style as exists code, is there a schema for formatter?

src/org/mozilla/javascript/NativeString.java
ian4hu: done

src/org/mozilla/javascript/NativeString.java
```diff
@@ -231,15 +235,29 @@ public Object execIdCall(IdFunctionObject f, Context cx, Scriptable scope,
                     continue again;
                 }
 
+                case ConstructorId_fromCodePoint: {
+                    int n = args.length;
+                    if (n < 1) {
+                        return "";
+                    }
+                    int[] codePoints = new int[n];
+                    StringBuilder sb = new StringBuilder(n);
```
gbrail, line 244: This StringBuilder is not used. You can remove it!

ian4hu:
> Thanks! This is looking good but there's a little more to do.
> 
> Can you please make sure that you run:
> 
> git submodule init
> git submodule update
> 
> so that we sync the test262 tests, and then run
> 
> ./gradlew check
> 
> to run all the tests and checks.
> 
> SpotBugs figured out about the dead local store in NativeString.java.
> 
> Also, at the end of the tests, the test framework is telling us about a bunch of test262 tests that this change fixes -- congratulations!
> 
> So, can you please look at those warnings and edit "test262.properties" to un-disable those tests. Let us know if that doesn't make sense and someone else can look at it too.

done, all tests are passed.

gbrail:
Thanks!

ian4hu:
> Thanks!

I just learned about the test262 and is trying to enable relative tests case and making it green. I will make a pr later.

rbri:
> I just learned about the test262 and is trying to enable relative tests case and making it green. I will make a pr later.

There is no real magic - usually you only have to delete the lines for the now passing tests from the file.
image

As you can see there are still two failing test for fromCodePoint

    fromCodePoint/argument-is-not-integer.js
    fromCodePoint/number-is-out-of-range.js

Will be great if you can make this two cases also working.

PR #735, opened 2020-07-01
improved regex range handling

### LINKED ISSUE #723: RegEx difference between ECMAScript and other RegEx engines.
According to some online RegEx parsers, the expression `[_-\\s]+` is invalid in most languages, but valid in ECMAScript.  Most browsers will handle the expression fine, but the Rhino implementation complains that an invalid character range is being specified.  Many other RegEx parsers also complain.  But since Rhino is primarily used to execute Javascript code, then curious if this expressions should be legitimate and acceptable to the Rhino engine.

rbri:
fixes #723


gbrail:
Looks good. Thanks!

PR #736, opened 2020-07-01
fix tests in test262/built-ins/String/fromCodePoint/*

ian4hu:
Additional tests for #731

### DIFF
--- a/src/org/mozilla/javascript/NativeString.java
+++ b/src/org/mozilla/javascript/NativeString.java
@@ -242,7 +242,13 @@ final class NativeString extends IdScriptableObject
                     }
                     int[] codePoints = new int[n];
                     for (int i = 0; i != n; i++) {
-                        codePoints[i] = ScriptRuntime.toInt32(args[i]);
+                        Object arg = args[i];
+                        int codePoint = ScriptRuntime.toInt32(arg);
+                        double num = ScriptRuntime.toNumber(arg);
+                        if (!ScriptRuntime.eqNumber(num, codePoint) || !Character.isValidCodePoint(codePoint)) {
+                            throw rangeError("Invalid code point " + ScriptRuntime.toString(arg));
+                        }
+                        codePoints[i] = codePoint;
                     }
                     return new String(codePoints, 0, n);
                 }
--- a/testsrc/test262.properties
+++ b/testsrc/test262.properties
@@ -847,16 +847,6 @@ built-ins/String
     ! prototype/toLocaleLowerCase/special_casing_conditional.js
     ! prototype/toLowerCase/special_casing_conditional.js
     ! prototype/Symbol.iterator/this-val-non-obj-coercible.js
-    # not implemented
-    ! fromCodePoint/argument-is-not-integer.js
-    ! fromCodePoint/argument-not-coercible.js
-    ! fromCodePoint/arguments-is-empty.js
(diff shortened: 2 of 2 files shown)


testsrc/test262.properties
```diff
@@ -848,15 +848,15 @@ built-ins/String
...
-    ! fromCodePoint/argument-is-not-integer.js
-    ! fromCodePoint/argument-not-coercible.js
-    ! fromCodePoint/arguments-is-empty.js
-    ! fromCodePoint/fromCodePoint.js
-    ! fromCodePoint/length.js
-    ! fromCodePoint/name.js
-    ! fromCodePoint/number-is-out-of-range.js
-    ! fromCodePoint/return-string-value.js
-    ! fromCodePoint/to-number-conversions.js
+    # fromCodePoint/argument-is-not-integer.js
```
rbri, line 851: please remove the lines complety

testsrc/test262.properties
```diff
@@ -848,15 +848,15 @@ built-ins/String
     ! prototype/toLowerCase/special_casing_conditional.js
     ! prototype/Symbol.iterator/this-val-non-obj-coercible.js
     # not implemented
```
rbri, line 850: you can remove this line also

src/org/mozilla/javascript/NativeString.java
```diff
@@ -242,7 +242,13 @@ public Object execIdCall(IdFunctionObject f, Context cx, Scriptable scope,
                     }
                     int[] codePoints = new int[n];
                     for (int i = 0; i != n; i++) {
-                        codePoints[i] = ScriptRuntime.toInt32(args[i]);
+                        Object arg = args[i];
+                        int codePoint = ScriptRuntime.toInt32(arg);
+                        double num = ScriptRuntime.toNumber(arg);
+                        if (!ScriptRuntime.eqNumber(num, codePoint) || !Character.isValidCodePoint(codePoint)) {
+                            throw rangeError("invalid code point " + ScriptRuntime.toString(arg));
```
rbri, line 249: I think the error messages start usually with an uppercase letter

gbrail:
Thanks! Can you address those two small suggestions above please?

I spent a bunch of time refreshing my memory on how Infinity and NaN behave in Java but it looks like the test cases have it covered.

Also (not for this PR) if someone wants something to do, we could probably use a "ScriptRuntime.toLegalInt32" method that converts an Object into an int32 but throws the appropriate exception if the integer is not valid. It would probably do to make this code slightly more efficient and I bet we could clean up 10 other places as well.

src/org/mozilla/javascript/NativeString.java
ian4hu: u r right.

ian4hu:
> Thanks! Can you address those two small suggestions above please?
> 
> I spent a bunch of time refreshing my memory on how Infinity and NaN behave in Java but it looks like the test cases have it covered.
> 
> Also (not for this PR) if someone wants something to do, we could probably use a "ScriptRuntime.toLegalInt32" method that converts an Object into an int32 but throws the appropriate exception if the integer is not valid. It would probably do to make this code slightly more efficient and I bet we could clean up 10 other places as well.

You are not the only one who spent time digging into the `ScriptRuntime.toXXX`

gbrail:
Thanks for contributing this!

PR #740, opened 2020-07-04
reduce autoboxing

rbri:
Have removed mostly all auto boxing
* the have the boxing explicit
* remember the developers that boxing takes place and hopefully reduce boxing to the minimum

Have reduced the boxing at some places.
Also made a lot more functions static (only private ones to have no interface changes).
Hopefully this helps a bit to make this more faster and robust.

### DIFF
--- a/src/org/mozilla/classfile/SuperBlock.java
+++ b/src/org/mozilla/classfile/SuperBlock.java
@@ -110,7 +110,7 @@ final class SuperBlock {
    * constraint that is imposed on merging operand stacks (the local variable
    * array is always the same size).
    */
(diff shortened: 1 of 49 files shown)


gbrail:
I think it makes sense to have a style that discourages autoboxing since it's hard to keep track of when it's happening in an efficient way and when it isn't. Let me look more carefully and merge it later in the week since it's a lot of changes and I want to merge carefully. There are some optional SpotBugs checks from Facebook that we can enable that might be able to help detect problems automatically but when I tried them recently they were mostly things we don't want to check for. Perhaps we can cherry pick a few.

PR #741, opened 2020-07-04
Debugger fixes for FlatLaf:

DevCharly:
This PR fixes issues reported here https://github.com/JFormDesigner/FlatLaf/issues/120 in debugger Swing GUI when using [FlatLaf](https://github.com/JFormDesigner/FlatLaf) L&F (or other 3rd party L&Fs):
- make renderer tree row height same as table row height
- increase monospaced font size in script source and evaluation view if L&F uses larger font
- remove renderer tree border if L&F sets one (built in L&F do not)

This screenshot shows the problems of the debugger using FlatLaf running in Java 8 on a screen that is scales 150%:

grafik

With this PR applied:

grafik

(the gray background on "Boolean" will be fixed in FlatLaf)

Test program:

~~~java
import org.mozilla.javascript.Context;
import org.mozilla.javascript.ContextFactory;
import org.mozilla.javascript.ImporterTopLevel;
import org.mozilla.javascript.tools.debugger.Main;

public class RhinoDebuggerTest {
    public static void main(String[] args) throws Exception {
        // init FlatLaf
        com.formdev.flatlaf.FlatLightLaf.install();

        ContextFactory factory = new ContextFactory();
        Main dbg = Main.mainEmbedded("A Sample Test");
        dbg.attachTo(factory);

        Context context = factory.enterContext();
        ImporterTopLevel scope = new ImporterTopLevel(context);
        dbg.setBreakOnEnter(true);
        dbg.setVisible(true);
        dbg.setExitAction(() -> {
            // close the debug window
            dbg.dispose();
        });

        context.evaluateString(scope,
                "importPackage(javax.swing);"
                        + "\n// JOptionPane.showMessageDialog(null, 'Hello');\n",
                "A Sample Test", 1, null);
    }
}
~~~

Also tested with Windows L&F and Metal L&F without issues.

### DIFF
--- a/toolsrc/org/mozilla/javascript/tools/debugger/SwingGui.java
+++ b/toolsrc/org/mozilla/javascript/tools/debugger/SwingGui.java
@@ -1119,7 +1119,7 @@ class EvalTextArea
         doc.addDocumentListener(this);
         addKeyListener(this);
         setLineWrap(true);
-        setFont(new Font("Monospaced", 0, 12));
+        setFont(new Font("Monospaced", 0, Math.max(12, UIManager.getFont("Label.font").getSize())));
         append("% ");
         outputMark = doc.getLength();
     }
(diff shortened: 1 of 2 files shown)


gbrail:
Makes sense to me. Thanks!

I do wish we had a way to automate debugger testing but that sounds like an awful lot of work!

PR #742, opened 2020-07-04
fix wrong start object for getter in Object.assign

### LINKED ISSUE #711: Object.assign doen't update property in time
If the property definition is changed during the assign progress, rhino doesn't reflect the change in time. The code below(LanguageVersion=Context.VERSION_ES6) ends in different results with rhino and browser(Firefox):
```javascript
var oa=Object.defineProperty({},"a", {
                        enumerable: true,
                        get: function () {
                            Object.defineProperty(this, "b", {
                                value: 3,
                                enumerable: false
                            });
                        }
                    });

var ob=Object.assign(oa, {b: 2});
Object.assign({b: 1},ob);
print('ob.b='+ob.b+', enum='+Object.getOwnPropertyDescriptor(ob,"b").enumerable);
```
Rhino: ob.b=2, enum=true
Firefox: ob.b=3, enum=false

The problem is at the second call of the Object.assign. ob.a has a getter function which re-defines property b. During the value assignment, the getter is called. In rhino, the the new definition of b doesn't take effective so that ob.b still holds the old definition.

rbri:
fixes #711

### DIFF
--- a/src/org/mozilla/javascript/NativeObject.java
+++ b/src/org/mozilla/javascript/NativeObject.java
@@ -544,29 +544,29 @@ public class NativeObject extends IdScriptableObject implements Map
             if (args.length < 1) {
               throw ScriptRuntime.typeError1("msg.incompat.call", "assign");
             }
-            Scriptable t = ScriptRuntime.toObject(cx, thisObj, args[0]);
+            Scriptable targetObj = ScriptRuntime.toObject(cx, thisObj, args[0]);
             for (int i = 1; i < args.length; i++) {
               if ((args[i] == null) || Undefined.instance.equals(args[i])) {
                 continue;
               }
-              Scriptable s = ScriptRuntime.toObject(cx, thisObj, args[i]);
-              Object[] ids = s.getIds();
+              Scriptable sourceObj = ScriptRuntime.toObject(cx, thisObj, args[i]);
+              Object[] ids = sourceObj.getIds();
               for (Object key : ids) {
                 if (key instanceof String) {
-                  Object val = s.get((String) key, t);
+                  Object val = sourceObj.get((String) key, sourceObj);
                   if ((val != Scriptable.NOT_FOUND) && (val != Undefined.instance)) {
-                    t.put((String) key, t, val);
+                    targetObj.put((String) key, targetObj, val);
                   }
                 } else if (key instanceof Number) {
                   int ii = ScriptRuntime.toInt32(key);
-                  Object val = s.get(ii, t);
+                  Object val = sourceObj.get(ii, sourceObj);
                   if ((val != Scriptable.NOT_FOUND) && (val != Undefined.instance)) {
-                    t.put(ii, t, val);
+                    targetObj.put(ii, targetObj, val);
                   }
                 }
               }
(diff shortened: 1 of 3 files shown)


src/org/mozilla/javascript/NativeObject.java
```diff
@@ -544,29 +544,29 @@ public Object execIdCall(IdFunctionObject f, Context cx, Scriptable scope,
             if (args.length < 1) {
               throw ScriptRuntime.typeError1("msg.incompat.call", "assign");
             }
-            Scriptable t = ScriptRuntime.toObject(cx, thisObj, args[0]);
+            Scriptable targetObj = ScriptRuntime.toObject(cx, thisObj, args[0]);
             for (int i = 1; i < args.length; i++) {
               if ((args[i] == null) || Undefined.instance.equals(args[i])) {
```
gbrail, line 549: In other parts of the code I think that we started moving to using "Undefined.isUndefined()" for this test to ensure that the check works in all situations.

src/org/mozilla/javascript/NativeObject.java
```diff
@@ -544,29 +544,29 @@ public Object execIdCall(IdFunctionObject f, Context cx, Scriptable scope,
...
               }
-              Scriptable s = ScriptRuntime.toObject(cx, thisObj, args[i]);
-              Object[] ids = s.getIds();
+              Scriptable sourceObj = ScriptRuntime.toObject(cx, thisObj, args[i]);
+              Object[] ids = sourceObj.getIds();
               for (Object key : ids) {
                 if (key instanceof String) {
-                  Object val = s.get((String) key, t);
+                  Object val = sourceObj.get((String) key, sourceObj);
                   if ((val != Scriptable.NOT_FOUND) && (val != Undefined.instance)) {
```
gbrail, line 557: Same here: Undefined.isUndefined()?

src/org/mozilla/javascript/NativeObject.java
```diff
@@ -544,29 +544,29 @@ public Object execIdCall(IdFunctionObject f, Context cx, Scriptable scope,
...
-                    t.put((String) key, t, val);
+                    targetObj.put((String) key, targetObj, val);
                   }
                 } else if (key instanceof Number) {
                   int ii = ScriptRuntime.toInt32(key);
-                  Object val = s.get(ii, t);
+                  Object val = sourceObj.get(ii, sourceObj);
                   if ((val != Scriptable.NOT_FOUND) && (val != Undefined.instance)) {
```
gbrail, line 563: Undefined.isUndefined()?

rbri:
> Thanks for fixing this -- I probably broke it in the first place!

Yes you did - i was super happy to see you name in the history for this instead of my one :-)

> Since we're opening up this function, would you mind making the handling of Undefined consistent? Let me know what you think. Thanks!

Good point will have a look.

rbri:
@gbrail - any chance to merge this?
I like to do a new HtmlUnit release this weekend and it is easier if the is merged.

gbrail:
Looks good -- thanks!

rbri:
Thanks

PR #743, opened 2020-07-05
do not inherit strict mode when parsing a function body

rbri:


gbrail:
Looks good -- thanks!

PR #747, opened 2020-07-20
fix String.indexOf and String.includes when searching for an empty st…

### LINKED ISSUE #746: A possible issue of String.prototype.indexOf
##### Version

rhino-1.7.12

##### Test case

```js
var NISLFuzzingFunc = function () {
    var a = 'test'.indexOf('', 5);
    print(a);
};
NISLFuzzingFunc();
```

##### Execution steps

##### Output

```
-1
```

##### Expected behavior

```
4
```

##### Description

 The `indexOf()` method returns the index within the calling string object of the first occurrence of the specified value, starting the search at `fromIndex`. Returns `-1` if the value is not found. When the first parameter of `String.prototype.indexOf` is an empty string and the second parameter is greater than the length of the string, other engines return the length of this string (like chakraCore, v8 and spiderMonkey), but rhino returns `-1`. I suspect this is a bug of rhino.

Contributor: [@Wen Yi](https://github.com/YiWen-y)

rbri:
…ring

fixes #746

### DIFF
--- a/src/org/mozilla/javascript/NativeString.java
+++ b/src/org/mozilla/javascript/NativeString.java
@@ -689,7 +689,7 @@ final class NativeString extends IdScriptableObject
         }
 
         if (methodId != Id_startsWith && methodId != Id_endsWith
-        		&& position > target.length()) {
+                && position > target.length()) {
             return -1;
         }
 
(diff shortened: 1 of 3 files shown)


gbrail:
Thanks! I'm going to squash the three commits since they're mainly formatting.

PR #751, opened 2020-07-21
Object.setPrototypeOf() arg[0] has to be coercible

### LINKED ISSUE #750: An issue about the Object.setPrototypeOf()
**Revision**

rhino-1.7.12

**Test case**

```javascript
var NISLFuzzingFunc = function(e,t){
    return Object.setPrototypeOf(e,t);
};
var NISLParameter0 = null;
var NISLParameter1 = new Object();
var NISLCallingResult = NISLFuzzingFunc(NISLParameter0, NISLParameter1);
print(NISLCallingResult);
```

**Execution steps**

```javascript
java -jar rhino-1.7.12.jar -debug -version 200 testcase.js
```

**Output**

```javascript
null
```

**Expected behavior**

```javascript
Throw a TypeError exception.
```

**Description**

According to ES standard, it should throw a TypeError exception when the first parameter is null.But in the above test case, Rhino returns null.

rbri:
fixes #750

### DIFF
--- a/src/org/mozilla/javascript/NativeArray.java
+++ b/src/org/mozilla/javascript/NativeArray.java
@@ -1910,7 +1910,9 @@ public class NativeArray extends IdScriptableObject implements List
         // as a result we have to handle ConstructorId_xxx calls (negative id)
         // the same way and always us the abs value of the id for method selection
         int id = Math.abs(idFunctionObject.methodId());
-        if (Id_find == id || Id_findIndex == id) o = requireObjectCoercible(cx, o, idFunctionObject);
+        if (Id_find == id || Id_findIndex == id) {
+            requireObjectCoercible(cx, o, idFunctionObject);
+        }
 
         long length = getLengthProperty(cx, o, id == Id_map);
         Object callbackArg = args.length > 0 ? args[0] : Undefined.instance;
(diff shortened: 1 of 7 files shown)


gbrail:
When I build this on top of the latest master, I get a compilation error in NativeString:

/Users/gregbrail/src/rhino/src/org/mozilla/javascript/NativeString.java:547: error: incompatible types: Object cannot be converted to Scriptable
                  return new NativeStringIterator(scope, requireObjectCoercible(cx, thisObj, f));

Can you look at that? Thanks!

rbri:
looks like this line

> return new NativeStringIterator(scope, requireObjectCoercible(cx, thisObj, f));

is like this in my code base

return new NativeStringIterator(scope, thisObj);

Maybe you have a local change on top?

rbri:
Looks like i was able to finish your work here ;-)

rbri:
During my local repo cleanup if found a branch for that introduces requireObjectCoercible in the line where you got the compile error.

    return new NativeStringIterator(scope, requireObjectCoercible(cx, thisObj, f));

Maybe this is a second case where your merge of my pr was not perfect :-)
But the commit from the morning should fix both.

gbrail:
OK, works now, thanks.

PR #752, opened 2020-07-21
minimal RegEx[Symbol.search]() and RegEx[Symbol.match]() support

rbri:


gbrail:
Thanks. Are we still missing a few test cases? I'm OK with adding this functionality in its current form, but do you think you'd have time to make the support more complete?

rbri:
At the moment i like to have another look at this - but i fear going forward with this implementation requires some refactoring.

rbri:
Have different plans for the next days - please merge as starting point. Will come back to this later.

PR #754, opened 2020-07-30
switch value and done

rbri:
I know this is a bit strange but the order of this two properties is reversed in all current browsers. Will be great if you merge this.
Please....


gbrail:
Sure -- it's perfectly safe to do that.

rbri:
Thanks

PR #755, opened 2020-08-06
make some method protected to support rhino-external implementations

rbri:
Hi Greg,
for HtmlUnit i have to implement my own Iterator subclass.
https://github.com/HtmlUnit/htmlunit/blob/master/src/main/java/com/gargoylesoftware/htmlunit/javascript/host/URLSearchParams.java#L60
This is about the minimal way to support this.

Another option might be to generalize the NativeCollectionIterator in a way to work with Map.Entry instead of Hashtable.Entry. But there is a problem because the NativeCollectionIterator calls clear on the entries. Maybe you have an idea about a more general solution....

### DIFF
--- a/src/org/mozilla/javascript/ES6Iterator.java
+++ b/src/org/mozilla/javascript/ES6Iterator.java
@@ -10,7 +10,7 @@ public abstract class ES6Iterator extends IdScriptableObject {
 
     private static final long serialVersionUID = 2438373029140003950L;
 
-    static void init(ScriptableObject scope, boolean sealed, IdScriptableObject prototype, String tag) {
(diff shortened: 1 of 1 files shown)


gbrail:
Well, ES6Iterator is supposed to be subclassed, so I don't think that making some methods protected is a problem. It might make it more difficult to maintain backward compatibility going forward but I'm not sure that doing something more clever with Map.Entry or something would be better.

So if you agree that this is the best choice then I'm happy to merge this.

rbri:
Agree - please merge....

rbri:
Thanks

PR #761, opened 2020-08-21
Add feature flag for changes to Function.__proto__

gbrail:
This change broke some legacy code previously. Add a version guard
to the new behavior so that it only works the "new way" for version
"ES6" and up.

Also, change a test that was expecting the new behavior for
version "1_8".

@rbri this adds a version guard to a change you made a while back. I noticed that you explicitly set version 1_8 in the test for this change -- this PR changes it to only work the new way for version ES6. Does that work for your other use cases?


gbrail:
@rbri here's a better patch that includes more comprehensive tests for both the new (spec-compliant) and old behavior. Can you take a look?

rbri:
Great

PR #769, opened 2020-09-20
[TypedArray].prototype.subarray() should not throw an exception

### LINKED ISSUE #768: An issue about the %TypedArray%.prototype.subarray().
**Revision**

rhino-1.7.12

**Test case**

```javascript
    var a = new Uint32Array(1)
    return a.subarray();
}
var NISLFuzzingResult = NISLFuzzingFunc();
print(NISLFuzzingResult);
```

**Execution steps**

```javascript
```

**Output**

```javascript
Error: invalid arguments.
```

**Expected behavior**

```javascript
Output 0.
```

**Description**

According to ES standard, the parameters of %TypedArray%.prototype.subarray() is optional,but rhino reported an error when I didn't pass parameters to this function.Therefore,I think there may be a problem here.

Contributor : @Haobin-Lee

rbri:
[TypedArray].prototype.subarray() should not throw an exception if called without parameters (ES6)

fixes #768


gbrail:
LGTM. Thanks!

PR #772, opened 2020-09-23
make sure the this object is coercible for String.prototype.big and similar calls

### LINKED ISSUE #765: An issue about String.prototype.big
According to ES6.0, when using method [[call]] to modify 'this' of `String.prototype.big`, update 'this' into 'null' or 'undefined' will raise a TypeError. And many other methods like `String.prototype.small`, `String.prototype.fontsize`, which are defined in Annex B have the same situations. This may be an issue of Rhino.

#### version

```
1.7.12
```

#### command

```
```

#### testcase

```javascript
var v0 = function(){
    return String.prototype.big.call(null);
};
var result = v0();
print(result);
```

#### output

```
<big>null</big>
```

#### expected output

```
TypeError: big: 'this' is null or undefined
```

contributor:[@YuanWangC](https://github.com/YuanWangC)

rbri:
add missing attribute escaping for all the methods

fixes #765

### DIFF
--- a/src/org/mozilla/javascript/NativeString.java
+++ b/src/org/mozilla/javascript/NativeString.java
@@ -383,43 +383,43 @@ final class NativeString extends IdScriptableObject
                 }
 
                 case Id_bold:
-                    return tagify(thisObj, "b", null, null);
+                    return tagify(cx, thisObj, f, "b", null, null);
 
                 case Id_italics:
-                    return tagify(thisObj, "i", null, null);
+                    return tagify(cx, thisObj, f, "i", null, null);
 
                 case Id_fixed:
-                    return tagify(thisObj, "tt", null, null);
+                    return tagify(cx, thisObj, f, "tt", null, null);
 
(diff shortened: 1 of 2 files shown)


rbri:
@gbrail please this also - or do you have any concerns about backward compatibility?

gbrail:
I am always worried about backward compatibility ;-) However this particular use is obscure enough that I think it's fine for us to just update and start matching the spec. (I didn't know about these special prototype functions either!)

rbri:
> I am always worried about backward compatibility

I'm always worried about your worries about backward compatibility :-)

> However this particular use is obscure enough that I think it's fine for us to just update and start matching the spec.

Yes strange stuff from the good old days.

> (I didn't know about these special prototype functions either!)

Same here - but i fear they are still in use somewhere (e.g. nuclear power plants)

rbri:
Oh BTW: - Thanks for merging

PR #785, opened 2020-10-19
Add comments when transforming tree

brecht-yperman-tb:


gbrail:
Sorry I didn't look at this before. It looks very straightforward and seems to address a regression so let me check it out. Thanks!

PR #787, opened 2020-11-08
Fix encoding to UTF-8

anatawa12:
assert.js was saved in ISO-8859-1 so I converted the file to UTF-8


gbrail:
Simple enough. If you find other more significant UTF-8 problems in the codebase (this one is just in a test) then please add PRs!

PR #792, opened 2020-11-30
Fix the variable which has been defined null can not found, in script engine.

monrachen:
Changed the if condition from null-checking to Map::containsKey.

### DIFF
--- a/src/org/mozilla/javascript/engine/BindingsObject.java
+++ b/src/org/mozilla/javascript/engine/BindingsObject.java
@@ -31,11 +31,10 @@ public class BindingsObject
 
   @Override
   public Object get(String name, Scriptable start) {
-    Object ret = bindings.get(name);
-    if (ret == null) {
+    if (!bindings.containsKey(name)) {
       return Scriptable.NOT_FOUND;
     }
-    return Context.jsToJava(ret, Object.class);
+    return Context.jsToJava(bindings.get(name), Object.class);
   }
 
   @Override
--- a/testsrc/org/mozilla/javascript/tests/scriptengine/ScriptEngineTest.java
+++ b/testsrc/org/mozilla/javascript/tests/scriptengine/ScriptEngineTest.java
@@ -273,4 +273,13 @@ public class ScriptEngineTest {
     Object result = engine.eval("file.getAbsolutePath();");
     assertEquals(absVal, result);
   }
+  
+  @Test
+  public void testJsObjectDefinedNull() throws ScriptException {
+      String script = "" +
+          "var obj = null;" +
+          "if (obj === null) { print(\"null\"); }" +
+          "else { print(\"notnull\"); }";
+      engine.eval(script);
+  }
 }
--- a/testsrc/org/mozilla/javascript/tests/scriptengine/ScriptEngineTest.java
+++ b/testsrc/org/mozilla/javascript/tests/scriptengine/ScriptEngineTest.java
@@ -1,6 +1,7 @@
 package org.mozilla.javascript.tests.scriptengine;
 
 import java.io.File;
+import java.io.FileNotFoundException;
 import java.io.FileReader;
 import java.io.IOException;
 import java.io.StringReader;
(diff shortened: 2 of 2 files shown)


rbri (commented):
Can you please add a test case for this.

monrachen:
> 
> 
> Can you please add a test case for this.

monrachen:
> 
> 
> Can you please add a test case for this.

[Test.txt](https://github.com/mozilla/rhino/files/5635932/Test.txt)
You can eval the js code in this file. It's the correct js code, but throw exception.

gbrail:
Thanks -- I understand what you're trying to fix and it's small, but in order to merge this we need a test for it in the test suite that fails before the fix and works with the fix in. I'd suggest looking at the tests in:

https://github.com/mozilla/rhino/tree/master/testsrc/org/mozilla/javascript/tests/scriptengine

and either adding a new test or modifying an existing test. If you don't write a test case, then someone else has to do it, and that may take some time.

monrachen:
> 
> 
> Thanks -- I understand what you're trying to fix and it's small, but in order to merge this we need a test for it in the test suite that fails before the fix and works with the fix in. I'd suggest looking at the tests in:
> 
> https://github.com/mozilla/rhino/tree/master/testsrc/org/mozilla/javascript/tests/scriptengine
> 
> and either adding a new test or modifying an existing test. If you don't write a test case, then someone else has to do it, and that may take some time.

Oh, sorry, I am the new. Should I cancel this pull request, create a new commit contain this fix and test case, then redo the pull request?

gbrail:
Oh, no need to get rid of the pull request. Please either write a test case using one of the existing ones as a model, or modify an existing test case, and add whatever you changed as a new commit to this PR. Thanks!

gbrail:
Ok -- I think this is an important fix and it's important that we get it right. If you look at an existing test case like this one:

https://github.com/mozilla/rhino/blob/8c6107b2c93d004fb77b0e3946085fe69060f18e/testsrc/org/mozilla/javascript/tests/scriptengine/ScriptEngineTest.java#L83

you'll see that it exercises the API, and then uses assertions to verify that it gets back what it expects. The test script that you added just runs some JavaScript code and doesn't actually use the Bindings class and doesn't assert anything.

I think what you're trying to fix can be tested with a test script like this:

  @Test
  public void testNulHandling() throws IOException, ScriptException {
    engine.put("shouldBeNull", null);
    engine.eval(new FileReader("testsrc/assert.js"));
    engine.eval("assertEquals(shouldBeNull, null);");
}

I would expect that this test or a similar one would fail without your fix, and then when you put the fix back in, it'll work.

monrachen:
> 
> 
> Ok -- I think this is an important fix and it's important that we get it right. If you look at an existing test case like this one:
> 
> https://github.com/mozilla/rhino/blob/8c6107b2c93d004fb77b0e3946085fe69060f18e/testsrc/org/mozilla/javascript/tests/scriptengine/ScriptEngineTest.java#L83
> 
> you'll see that it exercises the API, and then uses assertions to verify that it gets back what it expects. The test script that you added just runs some JavaScript code and doesn't actually use the Bindings class and doesn't assert anything.
> 
> I think what you're trying to fix can be tested with a test script like this:
> 
> @test
> public void testNulHandling() throws IOException, ScriptException {
> engine.put("shouldBeNull", null);
> engine.eval(new FileReader("testsrc/assert.js"));
> engine.eval("assertEquals(shouldBeNull, null);");
> }
> 
> I would expect that this test or a similar one would fail without your fix, and then when you put the fix back in, it'll work.

Thanks for your guide, I had corrected it accroding by your tips, but the difference is I define the variable in js script.

gbrail (changes requested):
Thanks for working on this. I think what's confusing me is that your test is just setting a variable in JavaScript -- that doesn't touch the Bindings object, which is the thing that you fixed. In order to verify that this fix actually fixes the bug, then you need to put a "null" value into the Bindings object using the Java API, don't you?

For example, if you take your fix out, then does the test case still pass? I think that it will.

gbrail:
I still think that this is an important (and small) fix and that we should merge it. Is there anyone out there who would like to volunteer to create a good test case for it?

gbrail:
I cherry-picked the commit in here that fixes ScriptEngine itself with a new test that I wrote and created a new PR: https://github.com/mozilla/rhino/pull/855. I'm going to close this one.

In other words, I'm going to merge the original fix to this problem into master but not using this particular PR. Thanks for fixing the original bug!

PR #793, opened 2020-12-04
Math.max/min - each element coerced

### LINKED ISSUE #781: An issue about the Math.max().
**Revision**

rhino-1.7.13

**Test case**

```javascript
    var a = {
        valueOf: function(){
            print(1);
        }
    };
    var b = Math.max(NaN,a);
};
NISLFuzzingFunc();
```

**Execution steps**

```javascript
java -jar rhino/rhino-1.7.13.jar -debug -version 200 testcase.js
```

**Output**

```javascript
No output.
```

**Expected behavior**

```javascript
Output 1.
```

**Description**

When I passed NaN and an object with the "valueOf" attribute value as a callable function to the first and second parameters of Math.max, chakra did not execute this function. According to the ES standard, the ToNumber operation is performed on each parameter of Math.max, and the "valueOf" attribute value function of the second parameter will be executed.

Contributor : @Haobin-Lee

morfav:
I wasn't sure about tests, I did submit a PR to test262 however not sure if that will ever get accepted.

This is my first contribution, please let me know if there is anything I should improve.

Thanks!
Vlad

resolves #781

### DIFF
--- a/src/org/mozilla/javascript/NativeMath.java
+++ b/src/org/mozilla/javascript/NativeMath.java
@@ -270,10 +270,6 @@ final class NativeMath extends IdScriptableObject
                     ? Double.NEGATIVE_INFINITY : Double.POSITIVE_INFINITY;
                 for (int i = 0; i != args.length; ++i) {
                     double d = ScriptRuntime.toNumber(args[i]);
-                    if (Double.isNaN(d)) {
-                        x = d; // NaN
-                        break;
-                    }
                     if (methodId == Id_max) {
                         // if (x < d) x = d; does not work due to -0.0 >= +0.0
                         x = Math.max(x, d);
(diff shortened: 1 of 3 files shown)


gbrail:
Thanks -- I appreciate that you looked at test262 as well.

We do need a test for this in the test suite somewhere. The first choice would be to enable an existing test in the test262 test suite by uncommenting some tests in test262.properties. The second would be to add a test in our own test suite. If you follow the pattern in https://github.com/mozilla/rhino/tree/master/testsrc/org/mozilla/javascript/tests then that's not too much work.

If this isn't already covered in test262, then can you show us where in the spec it says that Rhino is doing this incorrectly, or to different behavior in the "more popular" script engines?

Thanks -- it's a lot of work but we're trying to be consistent with all this since a lot of people depend on Rhino. Thanks for contributing!

gbrail:
Sorry -- I forgot that the attached issue clearly explains the problem and associated spec, so don't worry about "why should we fix this." But please do add a test case based on the example in that issue.

morfav:
Is this what you had in mind? Is @LanguageVersion(Context.VERSION_DEFAULT) appropriate?

gbrail:
Yes -- that looks like what I was thinking of. I'll look at it and merge it soon.

gbrail:
This looks good. Thanks!

PR #794, opened 2020-12-06
Add option to run tests once only

morfav:
I was looking for a way to speed up the tests. Running the 2 test suites at just 1 optimization level doesn't speed them up that much (from ~12.5 minutes to ~9.5 minutes), but it's still a speed up, so please let me know if you think this is a useful change.

Would it be worthwhile to pull in a recent version of the test262 project?

Also, I see 1 of the tests always failing on my machine, is anyone familiar with what's going on?
java.lang.AssertionError: In "testsrc/tests/ecma/String/15.5.4.12-3.js":
 FAILED! var s = new String( String.fromCharCode(4304) ); s.toUpperCase().charCodeAt(0) = 7312 expected: 4304
 FAILED! var s = new String( String.fromCharCode(4305) ); s.toUpperCase().charCodeAt(0) = 7313 expected: 4305
etc...

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,6 +59,14 @@ git submodule update
 ```
 Build and run all the tests.
 ```
+./gradlew test -Dquick
+```
+The tests in Test262Suite and MozillaSuiteTest are run 3 times by default (at differing optimization levels).
+The `quick` parameter can be used to only run them once.
+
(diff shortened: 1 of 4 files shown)


gbrail:
This is a nice change and speeds up the tests quite a bit when needed. Thanks!

PR #795, opened 2020-12-06
Fix JSON.stringify on a RegExp Object

### LINKED ISSUE #762: An issue about the JSON.stringify().
**Revision**

rhino-1.7.12

**Test case**

```javascript
	print(JSON.stringify(new RegExp()));
}
NISLFuzzingFunc();
```

**Execution steps**

```javascript
```

**Output**

```javascript
undefined
```

**Expected behavior**

```javascript
{}
```

**Description**

According to ES standard, it should be rendered as an opening left brace followed by zero or more properties, separated with commas, closed with a right brace when the type of parameter is object，but rhino returned undefined when the parameter was a regular expression object.

morfav:
Calling JSON.stringify() on a RegExp object should return an object, not undefined

resolves #762

### DIFF
--- a/src/org/mozilla/javascript/NativeJSON.java
+++ b/src/org/mozilla/javascript/NativeJSON.java
@@ -310,23 +310,15 @@ public final class NativeJSON extends IdScriptableObject
             }
             return "null";
         }
-
-        if (value instanceof NativeArray) {
-            return ja((NativeArray) value, state);
-        }
-
-        if (isObject(value)) {
+        
+        if (value instanceof Scriptable && !(value instanceof Callable)) {
+            if (value instanceof NativeArray) {
+                return ja((NativeArray) value, state);
+            }
             return jo((Scriptable) value, state);
         }
 
         return Undefined.instance;
-    }
-
-    private static boolean isObject(Object value) {
-        if (value instanceof NativeRegExp) {
-            return true;
-        }
-        return value instanceof Scriptable && !(value instanceof Callable);
     }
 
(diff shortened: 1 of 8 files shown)


gbrail:
This works and it definitely fixes the issue, but I wonder if there isn't something more fundamentally wrong about the way NativeRegExp is implemented that requires us to do this -- after all, it sounds like for this purpose a RegExp is just another kind of an object with properties. The ECMA script spec for stringify, for example, doesn't say anything like "if it's a RegExp, do this special thing". https://www.ecma-international.org/ecma-262/11.0/index.html#sec-json.stringify

Can you poke around more and see if there's some more fundamental reason why RegExp objects fail the tests that JSON is using to figure out if their properties can just be stringified like all other objects? If it's too deeply buried we can still do this.

morfav:
It looks like an early implementation of RegExp was callable, this was even used in rhino unit tests, for example:

```/(a*)b1+/("baaac")[0]```

There is a historical discussion here:
https://bugzilla.mozilla.org/show_bug.cgi?id=61911

Also there is a more recent unit test specifically testing for this backward compatibility.

Please let me know what you think of this second attempt and whether you see a better way to do this - I have a feeling there must be :)

gbrail:
Thanks -- that explains a lot. I can see that RegExp objects aren't supposed to be callable any more and your fix should preserve backward compatibility. Let me try it out one last time tomorrow. Thanks!

gbrail:
Thanks!

PR #799, opened 2020-12-09
fix the (confusing) exception msg if Object.prototype.setPrototypeOf …

rbri:
…is called with less than two parameters


gbrail:
Makes sense. Thanks!

PR #800, opened 2020-12-10
Fix parameter handling for Object.assign

rbri:
Another fix for wrong '"msg.incompat.call'.

@gbrail  looks like you did this some years ago but because you have added this function without a version check i think it is ok to chance the function. From my point of view this is only available in ES6 we do not need a version check here to be backward compatible. In fact this changes only the msg text.


gbrail:
Yes, this is good and I can't imagine it would introduce any kind of compatibility problem. Thanks!

PR #802, opened 2020-12-12
make typechecking and casting a bit more generic to be able to provide better error messages

rbri:
It's really hard to find the reason for issues like https://github.com/HtmlUnit/htmlunit/issues/274 in huge libraries. Because of this i tried to make the error message a bit more explicit.

Hope you like it.

Will do a performance test for this change by applying this to the htmlunit rhino fork and check if there is some negative impact.
Will report the results here.


rbri:
Looks like this change has no negative effect regarding performance - at least not for the  HtmlUnit test suites.

gbrail:
I agree -- this makes sense and I wouldn't expect this to have a real performance impact. I also ran some benchmarks and they were inconclusive, so let's go with this.

Issue #803, opened 2020-12-15
EvaluatorException parsing with comments between try/catch statements


xabylr:
With the following piece of code I am trying to build an AST tree keeping JavaScript comments.
```java
String script = "some JavaScript code";

CompilerEnvirons env = new CompilerEnvirons();
env.setRecordingComments(true);
Parser p = new Parser(env);
AstRoot ast = p.parse(script, "test", 1);
```

When parsing the following JavaScript codes, the parse method throws different exceptions.

- More than one comment between try token and its block.

  ```javascript
  try //this comment is fine
  //this will produce an exception
  {} catch(e) {}
  ```
  This code throws `org.mozilla.javascript.EvaluatorException: missing { before try block`

- One or more comments between try block and catch token.
  ```javascript
  try {} 
  //this will produce an exception
  catch(e) {}
  ```
  This code throws `org.mozilla.javascript.EvaluatorException: 'try' without 'catch' or 'finally'`
  
 The same happens when using multiline comments or JSDocs. It doesn't matter the number of lines they take but the number of comment tokens.

rbri:
@gbrail can we close this?

gbrail:
Thanks for fixing!

PR #804, opened 2020-12-16
Fix comment parsing between try and catch statements. #803

xabylr:
This is a fix for #803 
More than one comment is allowed between try token and its block and also between try block and catch token.

I also added one test that pass if the exceptions are not found or the comments are not retrieved.


gbrail:
Looks good and a good test case. Thanks!

xabylr:
Thank you!

Issue #725, opened 2020-06-23
An issue about JSON.stringify


NWU-NISL:
##### Version

rhino-1.7.12

##### Test case

```js
var NISLFuzzingFunc=function(nislMutationParameter1){
    var a={"0":1, "1":2, "2":3, "name":4};
    var b=JSON.stringify(a, nislMutationParameter1); 
    print(b);
}; 
var nislMutationArgument1=["0", "1", "name", "name"]; 
NISLFuzzingFunc(nislMutationArgument1);
```

##### Execution steps

` java -jar rhino/rhino-1.7.12.jar -debug -version 200 testcase.js`

##### Output

```
 {"name":4,"name":4}
```

##### Expected behavior

```
{"0":1,"1":2,"name":4}
```

##### Description

The `JSON.stringify(value[, replacer[, space]])` method converts a JavaScript object or value to a JSON string, optionally including only the specified properties if a replacer array is specified, and the ES standard specifies that the output attributes should not contain duplicate attributes. When executing this test case, rhino has two problems: the first is that the attribute named by numeric characters cannot be output, and the second is to repeatedly output the same attribute.

contributor：  [@Wen Yi](https://github.com/YiWen-y)

gbrail:
By the way -- I appreciate all the tricky but important bugs you are finding. (I also appreciate @rbri for fixing them!)

Who or what is "NWU NISL?" Are you a professional, a student, a research project... I'm just curious what methodology you are using to find these.

YiWen-y:
We are a Chinese research group, and we use fuzz testing to find bugs of JS engines.

tom08zehn:
I found a similar problem with the replacer argument but with slightly different output.

I'm using Rhino with ES6 (language version 200) but also tested with lower version 160.

I experienced a bug in Rhino with `JSON.stringify()` function when you provide an array of strings as 2nd argument that is used as a whitelist for the property names that you want to keep **and if the property names are numbers as strings** (stringified numbers). You can see a full code example here and compare Rhino result with any other JS engine (browser):

https://js.do/code/rhino-issue-object-keys-stringnumber-keys

Summary:

```js
var obj1 = {"2":"b","3":"c","1":"a"};
JSON.stringify(obj1, ["1","2","3"]); // returns {} in Rhino but works in JS returning {"1":"a","2":"b","3":"c"}
[...]

PR #806, opened 2020-12-17
Stringify with more than 3 arguments should respect the first 3 argum…

### LINKED ISSUE #767: Issue with JSON.stringify when called with more than 3 arguments
JSON.stringify returns the string 'null' when called with more than three arguments, the same as if it had been called with zero arguments.

This is a regression introduced in f94fb0d54616e0ec22ae653d8bc07f713ea13415 in the file NativeJSON.java when the default case in the switch statement was moved from the top to the bottom.

morfav:
…ents

resolves #767

### DIFF
--- a/src/org/mozilla/javascript/NativeJSON.java
+++ b/src/org/mozilla/javascript/NativeJSON.java
@@ -89,14 +89,18 @@ public final class NativeJSON extends IdScriptableObject
             }
 
             case Id_stringify: {
-                Object value = null, replacer = null, space = null;
-                switch (args.length) {
-                    default:
-                    case 3: space = args[2];
-                    /* fall through */ case 2: replacer = args[1];
-                    /* fall through */ case 1: value = args[0];
-                    /* fall through */ case 0:
+                Object value = Undefined.instance, replacer = null, space = null;
+
+                if (args.length > 0) {
+                    value = args[0];
+                    if (args.length > 1) {
+                        replacer = args[1];
+                        if (args.length > 2) {
+                            space = args[2];
+                        }
+                    }
                 }
+
                 return stringify(cx, scope, value, replacer, space);
             }
 
--- a/src/org/mozilla/javascript/NativeJSON.java
+++ b/src/org/mozilla/javascript/NativeJSON.java
@@ -91,11 +91,11 @@ public final class NativeJSON extends IdScriptableObject
             case Id_stringify: {
                 Object value = null, replacer = null, space = null;
                 switch (args.length) {
+                    default:
                     case 3: space = args[2];
                     /* fall through */ case 2: replacer = args[1];
                     /* fall through */ case 1: value = args[0];
                     /* fall through */ case 0:
-                    /* fall through */ default:
                 }
                 return stringify(cx, scope, value, replacer, space);
             }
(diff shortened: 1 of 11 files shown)


src/org/mozilla/javascript/NativeJSON.java
```diff
@@ -91,11 +91,11 @@ public Object execIdCall(IdFunctionObject f, Context cx, Scriptable scope,
             case Id_stringify: {
                 Object value = null, replacer = null, space = null;
                 switch (args.length) {
+                    default:
```
gbrail, line 94: I'm a bit conflicted about this since it's a very clever switch statement, but it's also a bit confusing. Would a series of "if (args.length >= 1)", "if (args.length >= 2)", and such statements be easier to follow for future maintainers? I realize that I'm suggesting three comparisons versus (theoretically) one, but it does seem easier to follow. I don't have a strong opinion, though.

gbrail (commented):
While you're looking at this, do you mind looking carefully at:

https://www.ecma-international.org/ecma-262/11.0/index.html#sec-json.stringify

and seeing if there are other things that we should fix? For instance, I see in here that we initialize value to null, rather than Undefined.instance, which means that we're likely not handling undefined properly -- and we're not, since JSON.stringify(undefined) returns null and not undefined.

morfav:
I replaced the switch (which admittedly also took me some time to figure out), with some if statements that arguably look a bit like an abomination anyway :)

I did some extra investigation, things like:
JSON.stringify()
JSON.stringify({})
JSON.stringify(null)
JSON.stringify(undefined)
JSON.stringify([,1,undefined])
JSON.stringify(NaN)
JSON.stringify(Number.NaN)
JSON.stringify(Infinity)

work well, apart from the first one (which you pointed out). I added a fix for this case.

I also ran the latest test262 tests, some of the recurring themes are to do with the replacer, and have already been pointed out in [this issue](https://github.com/mozilla/rhino/issues/725)
Namely:
1. rhino doesn't remove duplicates from the replacer array, which should be relatively straightforward to fix:
For example, JSON.stringify({"key":"value"},["key", "key"]) would output "key" twice
2. from some debugging, it looks like JSON.stringify({"1":"value"}, ["1"]) treats the "1" in the object as a number, vs the "1" in the array as a string. So the output is actually empty since the indexOrHash values end up different and are not matched.
I was thinking of looking to address these two in a separate commit, if/when I can figure number 2 out.

src/org/mozilla/javascript/NativeJSON.java
tonygermano: I mentioned in the issue that I opened that before the regression bug was introduced by moving the default case to the bottom, the original switch statement was written with the default case at the top. I think it looked a lot cleaner than the cascading if statements. Perhaps it only needs a comment that the default case is for >= 3?
 ```
              switch (args.length) {
                    /* >= 3     */ default: space = args[2];
                    /* fallthru */ case 2: replacer = args[1];
                    /* fallthru */ case 1: value = args[0];
                    /* fallthru */ case 0:
                }  
```

gbrail:
Thanks for making those improvements -- I think that it makes sense now.

If you have time to investigate other issues with Stringify then that'd be very helpful of course, but I agree that if you do it should be a separate PR. Thanks!

PR #807, opened 2020-12-18
Memberbox delegate

rbri:
This is related to https://github.com/HtmlUnit/htmlunit/issues/274.

We do some magic at least with the stack trace handling - in  fact so far the delegate stuff of the member box was ignored at some places.

This pr merges some HtmlUnit changes and improves these fixes.
Additionally i have change the impl. of captureStackTrace.

Have done a test with all this changes using the HtmlUnit test suite - works fine so far and makes a lot more pojo tests passing (see https://jenkins.wetator.org/job/HtmlUnit%20-%20Headless/ and https://jenkins.wetator.org/job/HtmlUnit%20-%20Library%20Build/).

@gbrail  Not sure if it will be possible to improve this more but if i have an idea i will submit another pr. Will be great if you can merge this soon to let me go forward in HtmlUnit.

### DIFF
--- a/src/org/mozilla/javascript/MemberBox.java
+++ b/src/org/mozilla/javascript/MemberBox.java
@@ -135,11 +135,14 @@ final class MemberBox implements Serializable
     /**
      * Function returned by calls to __lookupGetter__
      */
-    Function asGetterFunction(final String name, final Scriptable scope, final Scriptable prototype) {
+    Function asGetterFunction(final String name, final Scriptable scope) {
+        // Note: scope is the scriptable this function is related to; therefore this function
+        // is constant for this member box.
+        // Because of this we can cache the function in the attribute
         if (asGetterFunction == null) {
-            asGetterFunction = new BaseFunction(scope, prototype) {
+            asGetterFunction = new BaseFunction(scope, ScriptableObject.getFunctionPrototype(scope)) {
                 @Override
-                public Object call(Context cx, Scriptable scope, Scriptable thisObj, Object[] originalArgs) {
+                public Object call(Context cx, Scriptable callScope, Scriptable thisObj, Object[] originalArgs) {
                     MemberBox nativeGetter = MemberBox.this;
                     Object getterThis;
                     Object[] args;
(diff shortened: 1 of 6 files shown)


src/org/mozilla/javascript/MemberBox.java
```diff
@@ -130,6 +132,68 @@ public String toString()
         return memberObject.toString();
     }
 
+    /**
+     * Function returned by calls to __lookupGetter__
+     */
+    Function asGetterFunction(final String name, final Scriptable scope, final Scriptable prototype) {
```
gbrail, line 138: Would you consider naming "scope" and "prototype" here something else, even "s" and "p", to make it clear that they are unused (at least I assume that they are, and if not, then I don't think that the lazy construction makes sense).

src/org/mozilla/javascript/NativeObject.java
```diff
@@ -306,8 +306,16 @@ public Object execIdCall(IdFunctionObject f, Context cx, Scriptable scope,
                       else
                           break;
                   }
-                  if (gs != null)
+                  if (gs != null) {
+                      if (gs instanceof MemberBox) {
```
gbrail, line 310: Is there a way that we can do this without the instanceof check -- with that check, now NativeObject is aware of the implementation of MemberBox.

src/org/mozilla/javascript/ScriptableObject.java
```diff
@@ -249,7 +249,7 @@ ScriptableObject getPropertyDescriptor(Context cx, Scriptable scope) {
             String fName = name == null ? "f" : name.toString();
             if (getter != null) {
                 if ( getter instanceof MemberBox ) {
-                    desc.defineProperty("get", new FunctionObject(fName, ((MemberBox)getter).member(), scope), EMPTY);
+                    desc.defineProperty("get", ((MemberBox) getter).asGetterFunction(fName, scope, ScriptableObject.getFunctionPrototype(scope)), EMPTY);
```
gbrail, line 252: Same here I wish that we didn't have to do "instanceof MemberBox" but I do realize that this code is very very old.

gbrail (commented):
Thanks! This is fine and I appreciate the attention to detail on the tests and on making this work in general. I am starting to worry about all the "instanceof" checks that we have all over the codebase that make fundamental objects like ScriptableObject and NativeObject know all sorts of special things about other objects.

I realize that this is an old problem but if you see an opportunity to improve in this PR then would you consider trying to make it better?

src/org/mozilla/javascript/NativeObject.java
rbri: I fear that was the way MemberBox was introduces (the instanceof  from your comment below was already there).
No idea how we can solve it without introducing many new interfaces and breaking the backward compatibility.

BTW: Simply count the occurrences of 'instanceof MemberBox' and compare it with 'instanceof Symbol' ;-)

src/org/mozilla/javascript/ScriptableObject.java
rbri: See above - and instanceof is really fast in java since years now

src/org/mozilla/javascript/MemberBox.java
rbri: Good point, maybe there is a better way to deal with this - will have a look (code was taken from the htmlunit fork of rhino written years ago).

Will write some more tests to make sure we have the right scope/protoype for the function object.

rbri:
Have done a bit regarding your comments - including a fix for the scope. Hope you like it; feel free to improve the code/and or the comments

gbrail:
Thanks! That makes it more clear for the future.

rbri:
Thanks for merging and as always thanks for the review and the hints.

PR #809, opened 2020-12-27
Fix bug with node position in labeled loops

### LINKED ISSUE #808: Source position incorrect within a labeled for-in loop
I'm using Rhino 1.7.13.  Input:
```js
var obj = {};
function bar(i) {}
function foo() {
  loop1:
  for (el in obj) {
    bar(el);
  }
}
```
When Rhino parses this input, it gives the absolute position (character offset) of the `bar(el)` call as 31, which is incorrect; it should be 83.  The absolute position of the `foo()` function declaration is given as 33, which is correct.  I debugged a bit, and the error is apparent when calling `getAbsolutePosition()` on the `FunctionCall` node:

https://github.com/mozilla/rhino/blob/41e57c749027ff20b3977a3b3b25e1b7ad6d45df/src/org/mozilla/javascript/ast/AstNode.java#L191-L199

The `FunctionCall` node has a `Scope` node as an ancestor, and that `Scope` node returns a negative value for `getPosition()`, messing up the calculation.  If I remove the for-in loop, the bug goes away.

If there is a known workaround for this issue, please let me know.  Thanks!

msridhar:
Fixes #808 

The best fix I could come up with was to restore the relative position of the loop earlier.  The test is based heavily on `Issue129Test`

### DIFF
--- a/src/org/mozilla/javascript/Parser.java
+++ b/src/org/mozilla/javascript/Parser.java
@@ -1597,9 +1597,9 @@ public class Parser
             try {
                 AstNode body = getNextStatementAfterInlineComments(pn);
                 pn.setLength(getNodeEnd(body) - forPos);
+                restoreRelativeLoopPosition(pn);
                 pn.setBody(body);
             } finally {
-                restoreRelativeLoopPosition(pn);
                 exitLoop();
             }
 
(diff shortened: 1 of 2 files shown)


msridhar:
Not sure why CI didn't run for my latest commit...

gbrail:
Thanks -- these changes make sense to me and when I test locally they all pass. I don't know why circleci is not running them on the last commit.

PR #810, opened 2020-12-28
Fix doc link in readme

Friss:
Previous doc link was 404ing.


Friss:
Actually looks like this page is deprecated as well. Is there a new page that should be linked to?

gbrail:
Thanks -- someday we should figure out a way to get those legacy docs onto GitHub...

PR #811, opened 2020-12-29
Refactor: Variadic getMessage

tuchida:
Before
```java
    public static String getMessage0(String messageId)
    public static String getMessage1(String messageId, Object arg1)
    public static String getMessage2(String messageId, Object arg1, Object arg2)
```
After
```java
   // use variadic of JDK1.5
    public static String getMessage(String messageId, Object... args)
```
If there was a code to pass `Object[]` to `getMessage1`, It may be incompatible.

### DIFF
--- a/src/org/mozilla/javascript/ScriptRuntime.java
+++ b/src/org/mozilla/javascript/ScriptRuntime.java
@@ -56,7 +56,7 @@ public class ScriptRuntime {
 
           @Override
           public Object call(Context cx, Scriptable scope, Scriptable thisObj, Object[] args) {
-            throw typeError0("msg.op.not.allowed");
+            throw typeErrorById("msg.op.not.allowed");
           }
           @Override
           public int getLength() {
@@ -437,7 +437,7 @@ public class ScriptRuntime {
             if (val instanceof Boolean)
                 return ((Boolean) val).booleanValue() ? 1 : +0.0;
             if (val instanceof Symbol)
-                throw typeError0("msg.not.a.number");
+                throw typeErrorById("msg.not.a.number");
             if (val instanceof Scriptable) {
                 val = ((Scriptable) val).getDefaultValue(NumberClass);
                 if ((val instanceof Scriptable) && !isSymbol(val))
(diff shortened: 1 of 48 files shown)


tuchida:
The object it want to print in the message may has better call `ScriptRuntime.toString`.
`String... args` may be better than `Object... args`.

src/org/mozilla/javascript/ScriptRuntime.java
```diff
@@ -4152,46 +4152,14 @@ public static boolean isArrayObject(Object obj)
...
-            String msg = getMessage1("msg.deprec.ctor", name);
+            String msg = getMessage("msg.deprec.ctor", name);
             if (version == Context.VERSION_DEFAULT)
                 Context.reportWarning(msg);
             else
                 throw Context.reportRuntimeError(msg);
         }
     }
 
-    public static String getMessage0(String messageId)
```
gbrail, line 4163: Unfortunately, we have no way of knowing how much third-party code built on Rhino uses this and the other "public static" methods. I'm totally sure that it's more than zero. So while I think it's great to modernize the codebase and update all the places that use these functions, I don't think that we should delete them. I would prefer if you instead marked them with a deprecated annotation.

src/org/mozilla/javascript/ScriptRuntime.java
```diff
@@ -4212,9 +4180,9 @@ public static String getMessage4(
 
     public static final MessageProvider messageProvider = new DefaultMessageProvider();
 
-    public static String getMessage(String messageId, Object[] arguments)
+    public static String getMessage(String messageId, Object... args)
```
gbrail, line 4183: Sadly here I am also sure that this will break someone, but if you instead deprecated the old one and created a new one then we could modernize the codebase while still being safe.

src/org/mozilla/javascript/ScriptRuntime.java
```diff
@@ -4291,60 +4259,40 @@ public static EcmaError typeError(String message)
         return constructError("TypeError", message);
     }
 
-    public static EcmaError typeError0(String messageId)
```
gbrail, line 4294: Same question here about the need to deprecate and delete these public methods.

gbrail (changes requested):
Thanks -- I'm all for modernizing the codebase but I am sure that if we remove all these public methods we will break various people's projects. Can you instead create new functions and deprecate the old ones? Thanks!

src/org/mozilla/javascript/ScriptRuntime.java
tuchida: Thanks. I made getMessageById and deprecated old method.

gbrail:
OK -- this makes sense now. Thanks!

Issue #783, opened 2020-10-09
An issue about Object.assign


NWU-NISL:
###### Version

rhino-1.7.13

###### Test case

```js
var o = {
    get x(){}
}; 
Object. assign(o,o);
```

###### Execution steps

    java -jar rhino-1.7.13.jar -debug -version 200 testcase.js

###### Output

```

```

###### Expected behavior

```js
TypeError: Cannot set property x of [object Object] which has only a getter
```

###### Description

When the source object and target object of `Object.assign` are both objects that only contain getter properties, other engines (such as v8, spiderMonkey, JavaScriptCore, etc.) threw a `TypeError `exception, but rhino passed it normally.

tuchida:
I think this is probably.
https://github.com/mozilla/rhino/blob/41e57c749027ff20b3977a3b3b25e1b7ad6d45df/testsrc/test262.properties#L693
https://github.com/tc39/test262/blob/041da54c02ae7d17edb8c134ab7691c4f643bafe/test/built-ins/Object/assign/target-set-not-writable.js

rbri:
Will have a look

tuchida:
@rbri I fixed it #812. Could you review if you like?

gbrail:
Thanks!

PR #812, opened 2020-12-29
ref. #783 target set not writable in Object.assign

tuchida:
ref. #783

https://www.ecma-international.org/ecma-262/11.0/index.html#sec-object.assign
> Perform ? Set(to, nextKey, propValue, true).

https://www.ecma-international.org/ecma-262/11.0/index.html#sec-set-o-p-v-throw
> Let success be ? O.[[Set]](P, V, O).
> If success is false and Throw is true, throw a TypeError exception.

https://www.ecma-international.org/ecma-262/11.0/index.html#sec-ordinary-object-internal-methods-and-internal-slots-set-p-v-receiver
https://www.ecma-international.org/ecma-262/11.0/index.html#sec-ordinaryset
https://www.ecma-international.org/ecma-262/11.0/index.html#sec-ordinarysetwithowndescriptor
> If IsDataDescriptor(ownDesc) is true, then
> If ownDesc.[[Writable]] is false, return false.

### DIFF
--- a/src/org/mozilla/javascript/NativeObject.java
+++ b/src/org/mozilla/javascript/NativeObject.java
@@ -567,6 +567,12 @@ public class NativeObject extends IdScriptableObject implements Map
               Scriptable sourceObj = ScriptRuntime.toObject(cx, thisObj, args[i]);
               Object[] ids = sourceObj.getIds();
               for (Object key : ids) {
+                if (targetObj instanceof ScriptableObject) {
+                  ScriptableObject desc = ((ScriptableObject) targetObj).getOwnPropertyDescriptor(cx, key);
+                  if (desc != null && isDataDescriptor(desc) && isFalse(desc.get("writable"))) {
+                      throw ScriptRuntime.typeErrorById("msg.change.value.with.writable.false", key);
+                  }
+                }
                 if (key instanceof String) {
                   Object val = sourceObj.get((String) key, sourceObj);
                   if ((val != Scriptable.NOT_FOUND) && !Undefined.isUndefined(val)) {
--- a/testsrc/test262.properties
+++ b/testsrc/test262.properties
@@ -690,7 +690,6 @@ built-ins/Number
     ! parseInt.js
 
 built-ins/Object
-    ! assign/target-set-not-writable.js
     ! create/15.2.3.5-4-311.js
     ! defineProperties/15.2.3.7-6-a-112.js
     ! defineProperties/15.2.3.7-6-a-113.js


tuchida:
How do you check the property descriptor?

For example:
https://github.com/mozilla/rhino/blob/41e57c749027ff20b3977a3b3b25e1b7ad6d45df/src/org/mozilla/javascript/NativeObject.java#L479-L480
Do you need to use `ScriptRuntime.toBoolean` or `ScriptableObject.getProperty`?
```js
Object.defineProperty(o, 'a', {
  configurable: 1 // truthy
});
```
```js
Object.defineProperty(o, 'a', {
  __proto__: {
    configurable: true // prototype chain
  }
});
```

tuchida:
Sorry, My misunderstanding.
https://github.com/mozilla/rhino/blob/41e57c749027ff20b3977a3b3b25e1b7ad6d45df/src/org/mozilla/javascript/ScriptableObject.java#L216-L218
I'll correct it.

src/org/mozilla/javascript/NativeObject.java
```diff
@@ -567,6 +567,13 @@ public Object execIdCall(IdFunctionObject f, Context cx, Scriptable scope,
               Scriptable sourceObj = ScriptRuntime.toObject(cx, thisObj, args[i]);
               Object[] ids = sourceObj.getIds();
               for (Object key : ids) {
+                if (targetObj instanceof ScriptableObject) {
+                  ScriptableObject desc = ((ScriptableObject) targetObj).getOwnPropertyDescriptor(cx, key);
+                  if (desc != null && isDataDescriptor(desc) &&
+                      Boolean.FALSE.equals(desc.get("writable"))) {
```
gbrail, line 573: Can you take a look at the other implementation that checks for something like this, which is in ScriptableObject.checkPropertyChange and make it as much like that as you can?
That code uses an "isFalse" method to test for false, which is likely a more consistent way to do this -- these things tend to be very picky for edge cases.

gbrail (changes requested):
Just a minor thing. Thanks for looking at this!

src/org/mozilla/javascript/NativeObject.java
tuchida: Thanks.  I changed to use isFalse. 959147a

src/org/mozilla/javascript/NativeObject.java
```diff
@@ -567,6 +567,12 @@ public Object execIdCall(IdFunctionObject f, Context cx, Scriptable scope,
               Scriptable sourceObj = ScriptRuntime.toObject(cx, thisObj, args[i]);
               Object[] ids = sourceObj.getIds();
               for (Object key : ids) {
+                if (targetObj instanceof ScriptableObject) {
+                  ScriptableObject desc = ((ScriptableObject) targetObj).getOwnPropertyDescriptor(cx, key);
+                  if (desc != null && isDataDescriptor(desc) && isFalse(desc.get("writable"))) {
+                      throw ScriptRuntime.typeError1("msg.change.value.with.writable.false", key);
```
gbrail, line 573: Thanks! This looks great except that since I just merged your other PR this is now deprecated. Can you please update it?

gbrail (changes requested):

src/org/mozilla/javascript/NativeObject.java
tuchida: I changed it to use `ScriptRuntime.typeErrorById` and squashed it.

gbrail:
Great. Thanks!

Issue #815, opened 2021-01-01
Can't convert BoundFunction to Java interface


tuchida:
ArrowFunction can convert to Java interface.
```
js> java.util.Arrays.asList(1, 2, 3).stream().filter(i => i % 2 == 0)
java.util.stream.ReferencePipeline$2@506e6d5e
```
But if bound it, can't convert.
```
js> java.util.Arrays.asList(1, 2, 3).stream().filter((i => i % 2 == 0).bind())
js: "<stdin>", line 3: Cannot convert org.mozilla.javascript.BoundFunction@60f82f98 to java.util.function.Predicate
	at <stdin>:3
```
The case of BoundFunction is missing.
https://github.com/mozilla/rhino/blob/5cf8b7f22e331d458985e56adf59b422e726e200/src/org/mozilla/javascript/NativeJavaObject.java#L697-L701

tuchida:
Probably better `value instanceof Callable`.
https://github.com/mozilla/rhino/blob/5cf8b7f22e331d458985e56adf59b422e726e200/src/org/mozilla/javascript/InterfaceAdapter.java#L39

PR #816, opened 2021-01-01
ref. #815 convert Callable to Java interface

tuchida:
ref. #815

### DIFF
--- a/src/org/mozilla/javascript/NativeJavaObject.java
+++ b/src/org/mozilla/javascript/NativeJavaObject.java
@@ -694,9 +694,9 @@ public class NativeJavaObject
                     return value;
                 reportConversionError(value, type);
             }
-            else if (type.isInterface() && (value instanceof NativeObject
-                    || value instanceof NativeFunction
-                    || value instanceof ArrowFunction)) {
+            else if (type.isInterface()
+                     && (value instanceof NativeObject
+                         || (value instanceof Callable && value instanceof ScriptableObject))) {
                 // Try to use function/object as implementation of Java interface.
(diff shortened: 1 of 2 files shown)


src/org/mozilla/javascript/NativeJavaObject.java
```diff
@@ -695,8 +695,7 @@ else if (value instanceof Wrapper) {
                 reportConversionError(value, type);
             }
             else if (type.isInterface() && (value instanceof NativeObject
-                    || value instanceof NativeFunction
-                    || value instanceof ArrowFunction)) {
+                    || value instanceof Callable)) {
```
gbrail, line 698: I'm pretty sure that this is OK as long as value is also a ScriptableObject -- see the cast below and the comment in "createInterfaceAdapter." It's possible for something to be Callable but not a ScriptableObject, and in that case this change will result in a ClassCastException.

gbrail (changes requested):

src/org/mozilla/javascript/NativeJavaObject.java
gbrail: So sorry, my comment wasn't clear -- I think it's fine to check "instanceof Callable" here and I like that it fixes a number of other problems, but we also need to check "instanceof ScriptableObject" in case people have written code that is Callable but does not derive from ScriptableObject.

tuchida: 7fdd75e
Thanks. Could you check it again?

gbrail:
Yes, that does it. Thanks!

PR #818, opened 2021-01-02
Can't convert function to interface of all default methods.

tuchida:
Cannot convert function to interface with no methods.
```java
    public interface EmptyInterface {}
    public static void receiveEmptyInterface(EmptyInterface i) {}
```
```js
// org.mozilla.javascript.EvaluatorException: Cannot convert function to interface org.mozilla.javascript.tests.InterfaceAdapterTest$EmptyInterface with no methods (testNativeFunction.js#1)
a.receiveEmptyInterface(a => a)
```
But in interface case of all default methods, can convert.
```java
    public interface AllDefaultMethodInterface {
        default void a() {}
        default void b() {}
    }
    public static void receiveAllDefaultMethodInterface(AllDefaultMethodInterface i) {}
```
```js
// no errors
a.receiveAllDefaultMethodInterface(a => a)
```
I think it should be handled as if there were no methods.

### DIFF
--- a/src/org/mozilla/javascript/InterfaceAdapter.java
+++ b/src/org/mozilla/javascript/InterfaceAdapter.java
@@ -9,6 +9,8 @@ package org.mozilla.javascript;
 import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
 
+import java.util.HashSet;
+
 /**
  * Adapter to use JS function as implementation of Java interfaces with
  * single method or multiple methods with the same signature.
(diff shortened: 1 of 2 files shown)


tuchida:
Should I change the error message? Could you let me know if you have a better message.
https://github.com/mozilla/rhino/blob/5cf8b7f22e331d458985e56adf59b422e726e200/src/org/mozilla/javascript/resources/Messages.properties#L183-L184

gbrail (commented):
Thanks for looking at this. Default methods make this complicated. 

Should we be able to convert a function into:

An interface that has only one method, but it's "default"? (No? Yes?)

An interface that has multiple "default" methods but only one non-default method (Yes?)

An interface that extends an interface that has only default methods, but which adds a single non-default method (Yes?)

I'm really not sure and I don't work with code that depends on this aspect of Rhino very often. What do you think of the use cases above? Would you consider adding additional test cases?

tuchida:
NG
```java
interface I {
}
```
OK
```java
interface I {
  void a();
}
```
NG
```java
interface I {
  void a();
  void b();
}
```
OK
```java
interface I {
  void a(String s);
  void a(int i);
}
```
OK
```java
interface I {
  void a();
  default void b();
}
```
???
```java
interface I {
  default void a();
}
```
NG?
```java
interface I {
  default void a();
  default void b();
}
```
???
```java
interface I {
  default void a(String s);
  default void a(int i);
}
```
???
```java
interface I {
  void a(String s);
  default void a(int i);
}
```

gbrail:
I think that there are two reasonable answers:

1) Ignore the "default" keyword and use the same logic that we have now. That would imply that the answers are "OK, NG, OK, and OK" for the last four items in your list.
2) Ignore all methods with the "default" keyword, which would instead imply "NG, NG, NG, OK".

I think that people might want to use default functions, so my preference would be the first.

gbrail:
More thoughts -- I bet there's a pretty reasonable case in which an interface (which may inherit from other interfaces) has multiple default methods and only one non-default method, and we want to support that because that's the way that new methods are added to an interface in a backward-compatible way. That would argue for ignoring all methods with a "default" keyword.

gbrail:
Thanks for pulling together the list. I would love to see that be a series of tests! 

Would it be possible to do the following:

If an interface has only one method AND that method is "default", then the conversion is OK.
Otherwise, ignore any methods on the interface that are marked "default" and do the check like we did before.

That means the following:

OK, because there's only one method even though it's default:
interface I {
  default void a() {}
}

Not OK, because other than in the case above we ignore all default methods and that leaves us with an empty interface: 
interface I {
  default void a() {}
  default void b() {}
}

Not OK, because other than in the case above we ignore all default methods and that leaves us with an empty interface: 
interface I {
  default void a(String s) {}
  default void a(int i) {}
}

OK, because there is only one non-default method:
interface I {
  void a(String s) {}
  default void a(int i) {}
}

tuchida:
853f94d
Fixed. Are there any more test cases?

gbrail:
This looks good to me now. Thanks!

PR #819, opened 2021-01-04
Add a sanity check for DocumentBuilder's resettability

drubanovich-soti:
Some DocumentBuilder.reset implementations do not work as expected,
and do not restore the builder to the original state (when it was
created with DcoumentBuilderFactory.newDocument). An example of
such implementation is Android - see https://issuetracker.google.com/issues/176560895

This change adds a sanity check which treats such implementations
as non-resettable.

### DIFF
--- a/testsrc/org/mozilla/javascript/xmlimpl/XmlNonResettableDocumentBuilderTest.java
+++ b/testsrc/org/mozilla/javascript/xmlimpl/XmlNonResettableDocumentBuilderTest.java
@@ -1,0 +1,43 @@
+package org.mozilla.javascript.xmlimpl;
+
+import org.junit.After;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+import org.mozilla.javascript.Context;
+import org.mozilla.javascript.ContextFactory;
+import org.mozilla.javascript.Scriptable;
+
+public class XmlNonResettableDocumentBuilderTest {
+    private static final String XML_PROPERTY = "javax.xml.parsers.DocumentBuilderFactory";
+    private final String originalDocumentBuilderFactory = System.getProperty(XML_PROPERTY);
+
+    @Before
+    public void setUp() {
+        System.setProperty(XML_PROPERTY, NonResettableDocumentBuilderFactory.class.getName());
+    }
+
+    @After
+    public void tearDown() {
+        if (originalDocumentBuilderFactory == null) {
+            System.clearProperty(XML_PROPERTY);
+        } else {
+            System.setProperty(XML_PROPERTY, originalDocumentBuilderFactory);
+        }
+    }
+
+    @Test
+    public void testNonResettableDocumentBuilder() {
+        Context cx = new ContextFactory().enterContext();
+        try {
+            Scriptable scope = cx.initStandardObjects();
+            Object result = cx.evaluateString(scope,
+                    "var employees = new XML('<employees><employee><name>John</name></employee></employees>');" +
+                    "employees.employee.name;",
+                    "source", 1, null);
+            Assert.assertEquals("John", String.valueOf(result));
+        } finally {
+            Context.exit();
+        }
+    }
+}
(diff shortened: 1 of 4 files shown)


testsrc/org/mozilla/javascript/xmlimpl/XmlNonResettableDocumentBuilderTest.java
```diff
@@ -0,0 +1,43 @@
...
+            System.setProperty(XML_PROPERTY, originalDocumentBuilderFactory);
+        }
+    }
+
+    @Test
+    public void testNonResettableDocumentBuilder() {
+        Context cx = new ContextFactory().enterContext();
+        try {
+            Scriptable scope = cx.initStandardObjects();
+            Object result = cx.evaluateString(scope,
```
gbrail, line 34: I've been reading but I don't understand how this test case tests the resetting of the document builder. Don't we need to at least run the XML processor twice so that we can observe the attempt to reset the builder and then cache the result? Or is there something about how we configure the builder in the XML processing code that will fail if it doesn't happen here? In other words, is this a test that actually failed before this change was added?

xmlimplsrc/org/mozilla/javascript/xmlimpl/XmlProcessor.java
```diff
@@ -247,6 +247,11 @@ private DocumentBuilder getDocumentBuilderFromPool()
     private void returnDocumentBuilderToPool(DocumentBuilder db) {
         try {
             db.reset();
+            // DocumentBuilders are supposed to be namespace-aware.
+            // This is a sanity check for DocumentBuilder's resettability (a known bug in Android).
+            if (!db.isNamespaceAware()) {
```
gbrail, line 252: Is there somewhere else in the XML code where we set "isNamespaceAware" so that we can be sure that this test will not be triggered every time and caching will not work any more?

gbrail (commented):
Thanks -- I mainly have questions. I agree that it's important to fix this and I'm fine with us doing something to support a bug in many Android implementations, but my questions are about trying to understand how this fix fixes the problem and how the test tests for this.

testsrc/org/mozilla/javascript/xmlimpl/XmlNonResettableDocumentBuilderTest.java
drubanovich-soti: Yes, this is a real test: if you comment out the workaround in XmlProcessor.returnDocumentBuilderToPool - the test will fail.

The thing is both XmlProcessor.getDocumentBuilderFromPool() and XmlProcessor.returnDocumentBuilderToPool() are called twice during the test:
1. When XML library is initialized, for empty XML string
2. When the actual XML string from the test is read

xmlimplsrc/org/mozilla/javascript/xmlimpl/XmlProcessor.java
drubanovich-soti: Rhino gets all DocumentBuilders using DocumentBuilderFactory.newBuilder. The DocumentBuilderFactory is obtained using DocumentBuilderFactory.newInstance(). There are two places in non-test Rhino code where this newInstance() is called, and in both places newInstance() is followed by setNamespaceAware(true) 
```
        this.dom = DocumentBuilderFactory.newInstance();
        this.dom.setNamespaceAware(true);
```
setNamespaceAware(false) is never called in Rhino code.

So at the moment, all DocumentBuilders, created by Rhino, are namespace-aware. 

Is it possible that in future Rhino will also use namespace-unaware DocumentBuilders? My test shows that such builder won't be able to parse XML properly, so we cannot have namespace-unaware builders in the cache.

I thought about FEATURE-based solutions as well for this problem, like having something like FEATURE_DISABLE_DOCUMENT_BUILDER_CACHE or FEATURE_NON_RESETTABLE_DOCUMENT_BUILDER, but I picked the one above since it's less intrusive and doesn't require client changes. If you prefer FEATURE-based solution - I can upload another PR.

testsrc/org/mozilla/javascript/xmlimpl/XmlNonResettableDocumentBuilderTest.java
gbrail: Thanks -- didn't realize that it's invoked multiple times anyway.

xmlimplsrc/org/mozilla/javascript/xmlimpl/XmlProcessor.java
gbrail: Thanks for pointing that out. I think that this change is OK, then.

PR #822, opened 2021-01-11
make function __proto__ writable

rbri:
Made this PR based on the suggestion in https://github.com/HtmlUnit/htmlunit/issues/291.

### DIFF
--- a/src/org/mozilla/javascript/SpecialRef.java
+++ b/src/org/mozilla/javascript/SpecialRef.java
@@ -87,8 +87,7 @@ class SpecialRef extends Ref
                     Scriptable search = obj;
                     do {
                         if (search == target) {
-                            throw Context.reportRuntimeErrorById(
-                                "msg.cyclic.value", name);
+                            throw Context.reportRuntimeErrorById("msg.cyclic.value", name);
                         }
                         if (type == SPECIAL_PROTO) {
                             search = search.getPrototype();
@@ -106,6 +105,9 @@ class SpecialRef extends Ref
 
                     if (cx.getLanguageVersion() >= Context.VERSION_ES6) {
                         final String typeOfValue = ScriptRuntime.typeof(value);
+                        if (value instanceof Double) {
+                            return value;
+                        }
                         if (!"object".equals(typeOfValue) && !"function".equals(typeOfValue)) {
                             return obj;
                         }
--- a/src/org/mozilla/javascript/SpecialRef.java
+++ b/src/org/mozilla/javascript/SpecialRef.java
@@ -105,11 +105,8 @@ class SpecialRef extends Ref
 
                     if (cx.getLanguageVersion() >= Context.VERSION_ES6) {
                         final String typeOfValue = ScriptRuntime.typeof(value);
-                        if (value instanceof Number) {
-                            return value;
-                        }
                         if (!"object".equals(typeOfValue) && !"function".equals(typeOfValue)) {
-                            return obj;
+                            return value;
                         }
 
                         final String typeOfTarget = ScriptRuntime.typeof(target);
@@ -147,4 +144,3 @@ class SpecialRef extends Ref
         return false;
     }
 }
-
--- a/src/org/mozilla/javascript/SpecialRef.java
+++ b/src/org/mozilla/javascript/SpecialRef.java
@@ -105,11 +105,13 @@ class SpecialRef extends Ref
                     }
 
                     if (cx.getLanguageVersion() >= Context.VERSION_ES6) {
-                        if ((value != null && !"object".equals(ScriptRuntime.typeof(value))) ||
-                            !"object".equals(ScriptRuntime.typeof(target))) {
-                            return Undefined.instance;
+                        if (value != null && !"object".equals(ScriptRuntime.typeof(value))) {
+                            return obj;
                         }
-                        target.setPrototype(obj);
+                        final String typeOfTarget = ScriptRuntime.typeof(target);
+                        if ("object".equals(typeOfTarget) || "function".equals(typeOfTarget)) {
+                            target.setPrototype(obj);
+                        }
                     } else {
                         target.setPrototype(obj);
                     }
--- a/src/org/mozilla/javascript/SpecialRef.java
+++ b/src/org/mozilla/javascript/SpecialRef.java
@@ -105,6 +105,11 @@ class SpecialRef extends Ref
 
                     if (cx.getLanguageVersion() >= Context.VERSION_ES6) {
                         final String typeOfValue = ScriptRuntime.typeof(value);
+                        if (value == null) {
+                            target.setPrototype(Undefined.SCRIPTABLE_UNDEFINED);
+                            return null;
+                        }
+
                         if (!"object".equals(typeOfValue) && !"function".equals(typeOfValue)) {
                             return value;
                         }
--- a/src/org/mozilla/javascript/SpecialRef.java
+++ b/src/org/mozilla/javascript/SpecialRef.java
@@ -105,9 +105,11 @@ class SpecialRef extends Ref
                     }
 
                     if (cx.getLanguageVersion() >= Context.VERSION_ES6) {
-                        if (value != null && !"object".equals(ScriptRuntime.typeof(value))) {
+                        final String typeOfValue = ScriptRuntime.typeof(value);
+                        if (!"object".equals(typeOfValue) && !"function".equals(typeOfValue)) {
                             return obj;
                         }
+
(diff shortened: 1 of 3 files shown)


atnak (approved):
Thanks, conditions look good.

tuchida:
https://www.ecma-international.org/ecma-262/11.0/index.html#sec-set-object.prototype.__proto__
> 2. If Type(proto) is neither Object nor Null, return undefined.
> 3. If Type(O) is not Object, return undefined.

Probably `Type(A) is Object` equals `typeof A === 'object' || typeof A === 'function'`.

```js
var a = function() {};
a.b = 1;

var c = { __proto__: a };
c.b; // SpiderMonkey returns `1`. But Rhino `undefined`
```
Do you also need `"function".equals(ScriptRuntime.typeof(value))`?
https://github.com/mozilla/rhino/blob/41df10d6562649b99479eea072a851b70096e7e7/src/org/mozilla/javascript/SpecialRef.java#L108

atnak:
> 2. If Type(proto) is neither Object nor Null, return undefined.
> 3. If Type(O) is not Object, return undefined.

That's rather Interesting.  However, that linked documentation also says:

> 6. Return undefined.

Yet [we're returning obj](https://github.com/mozilla/rhino/blob/4af1f97438eb25c381315bf90418cbe1965b8eff/src/org/mozilla/javascript/SpecialRef.java#L119) in the main path, so I think it could be a compatibility thing, or the docs are different from practical implementations.

I did `({}).__proto__ = 42` in Chrome 87 / FF 84 / IE 11 and noticed the return value appears to be whatever I put on the RHS regardless of its type or whether its actually used by `__proto__`.

> Do you also need `"function".equals(ScriptRuntime.typeof(value))`?

👍

tuchida:
> Yet we're returning obj in the main path, so I think it could be a compatibility thing, or the docs are different from practical implementations.

I think it's probably like this:
```js
var desc = Object.getOwnPropertyDescriptor(Object.prototype, '__proto__');
desc.set(1);  // undefined
```
SpiderMonkey return undefined. In rhino, the `desc` is now undefined.

atnak:
> I think it's probably like this:

Oh now I understand.  The docs you were linking to were about `Object.prototype.__proto__`, and not about `Object.__proto__`, hence the "return undefined". 👍 

I misunderstood and thought you were commenting we shouldn't change the [return value from `undefined` to `obj` here](https://github.com/mozilla/rhino/pull/822/files#diff-b6176c40c3e5c64a0ddd2647ea8e1c28bd0608e0612e05b8e951e21c5f76c576L110) but I guess that wasn't the point.

I agree with you about [this condition for `value`](https://github.com/mozilla/rhino/pull/822/files#diff-b6176c40c3e5c64a0ddd2647ea8e1c28bd0608e0612e05b8e951e21c5f76c576R108) should also include function objects.  I now understand your citation was to point out both how both `Type(proto)` and `Type(O)` talk about whether or not it's an "Object", and we included functions for `Type(O)` so naturally the definition of "Object" should be expanded to include functions in `Type(proto)` too.

src/org/mozilla/javascript/SpecialRef.java
```diff
@@ -105,11 +105,13 @@ public Object set(Context cx, Scriptable scope, Object value)
                     }
 
                     if (cx.getLanguageVersion() >= Context.VERSION_ES6) {
-                        if ((value != null && !"object".equals(ScriptRuntime.typeof(value))) ||
-                            !"object".equals(ScriptRuntime.typeof(target))) {
-                            return Undefined.instance;
+                        if (value != null && !"object".equals(ScriptRuntime.typeof(value))) {
```
atnak, line 108:
```suggestion
                        final String typeOfValue = ScriptRuntime.typeof(value);
                        if (!"object".equals(typeOfValue) && !"function".equals(typeOfValue)) {
```

rbri:
Have added a new commit that includes the change from @atnak and enhances the test case.

Hope i got you right - and if you have more or more detailed test cases please add.

src/org/mozilla/javascript/SpecialRef.java
tuchida: > 2. If Type(proto) is neither Object nor Null, return undefined.

Since it is written like this, I think it is easier to see that `value != null` is more in line with the specification.
If you want to remove this, I would like the following test case.
```js
var a = {};
a.__proto__ = null;
a.__proto__ === undefined; // true
```

tuchida:
I would like to have this test case.
```js
(({}).__proto__ = 42) === 42
```

atnak:
> ```
> var a = {};
> a.__proto__ = null;
> a.__proto__ === undefined; // true
> ```

I saw this deleted comment.  I think a better test is the following, but I agree it shouldn't be included in this PR as it's too much to chew, since Rhino appear to lack a lot of code to support when `__proto__` is set to undefined.

```js
var a = {__proto__: {foo: function () {}}}
a.__proto__ = null; // null
a.__proto__ === undefined; // true
try { a.foo() } catch (e) { e } // TypeError: Cannot find function foo in object [object Object].

File: src/org/mozilla/javascript/SpecialRef.java [hunk 2]
```diff
@@ -105,11 +104,18 @@ public Object set(Context cx, Scriptable scope, Object value)
...
                     if (cx.getLanguageVersion() >= Context.VERSION_ES6) {
-                        if ((value != null && !"object".equals(ScriptRuntime.typeof(value))) ||
-                            !"object".equals(ScriptRuntime.typeof(target))) {
-                            return Undefined.instance;
+                        final String typeOfValue = ScriptRuntime.typeof(value);
+                        if (value instanceof Number) {
+                            return value;
+                        }
+                        if (!"object".equals(typeOfValue) && !"function".equals(typeOfValue)) {
+                            return obj;
```
atnak, line 112: Probably for non-objects, this should always be `value` rather than `obj`:
```suggestion
                        if (!"object".equals(typeOfValue) && !"function".equals(typeOfValue)) {
                            return value;
```

I tested using this code:

```js
[undefined, null, true, 123, "abc", Symbol(), function () {}, {}].forEach(value => {
    var obj = {}
    var res = (obj.__proto__ = value);
    console.log(typeof value + ": returned=" + (res === value))
})
```

rbri:
Ok, i fear there is more....

atnak:
I didn't think we should attempt case as properly supporting `__proto__ = undefined` probably requires patching places in Rhino that use `__proto__` to do the prototyping stuff and that could be somewhat involved.
https://github.com/mozilla/rhino/pull/822/files#diff-b6176c40c3e5c64a0ddd2647ea8e1c28bd0608e0612e05b8e951e21c5f76c576R108-R111

However, if we're going to add this then I think we need to check `typeOfTarget` first:

```java
final String typeOfTarget = ScriptRuntime.typeof(target);
if ("object".equals(typeOfTarget) || "function".equals(typeOfTarget)) {

    if (value == null) {
        target.setPrototype(Undefined.SCRIPTABLE_UNDEFINED);
        return value;
    }

    final String typeOfValue = ScriptRuntime.typeof(value);
    if ("object".equals(typeOfValue) || "function".equals(typeOfValue)) {
        target.setPrototype(obj);
    }
}
return value;
```

Alternatively we could set it to an empty object for the `null` case.  It's not entirely correct but it probably works better with the rest of Rhino and it probably won't make much difference to real world application.

```java
if (value == null) {
    target.setPrototype(new NativeObject());
    return value;
}
```

rbri:
> However, if we're going to add this then I think we need to check typeOfTarget first:

Good point - i was in a hurry a bit :-)

> I didn't think we should attempt this case as properly supporting __proto__ = undefined probably requires patching places in Rhino that use __proto__ to do the prototyping stuff and that could be somewhat involved.

As always i like to have more test cases for this. Will apply this patch to the HtmlUnit fork of Rhino and check the effect. HtmlUnit has huge test suites (e.g. the dojo and jQuery) included. Maybe this gives us a hint about the danger of this change.

atnak (approved):

rbri:
The whole HtmlUnit test suite passes with these code changes (including the null handling).

gbrail:
I like this and I am not too worried about compatibility because it maps to our general philosophy of making ">= VERSION_ES6" code match the spec without making changes that will affect older code.

Are there no test262 tests that we can enable that will further exercise this functionality?

If not then I'm happy to merge this.

rbri:
> Are there no test262 tests that we can enable that will further exercise this functionality?

Of course if have checked this but seems like we have no tests for __proto__ in there.

Waiting for your merge :-)

PR #826, opened 2021-01-18
ADD: sealed shared scope support

rPraml:
We want to use a sealed shared scope this way as documented here: https://developer.mozilla.org/en-US/docs/Mozilla/Projects/Rhino/Scopes_and_Contexts#Sharing_Scopes
```
       +---------------------+
       | sealed shared scope | <- defines importClass/importPackage
       +---------------------+ <- stores importClass/importPackage info
           ^ Prototype ^
           |           |
+---------------+  +---------------+
| ThreadScope 1 |  | ThreadScope 2 |
+---------------+  +---------------+
```
Our problem was, that `importClass` and `importPackage` was defined in the sealed shared scope and tries also to store the `importedPackages` and importedClassles, which was not possible if the scope was sealed.

There was also a problem if thread1 imports something, it was visible for thread2.
Especially, if thread1 imports `package1.MyClass` and thread2 imports `package2.MyClass` one of them wins and the other fails with a "Property already defined" error.

This PR changes, that the info is stored in the thread scopes now:
```
       +---------------------+
       | sealed shared scope | <- defines importClass/importPackage
       +---------------------+ 
           ^ Prototype ^
           |           |
+---------------+  +---------------+
| ThreadScope 1 |  | ThreadScope 2 |
+---------------+  +---------------+
  ^- Stores importClass    ^- Stores importClass
  importPackage info          importPackage info
  for threadScope1            for threadScope2
```

Instead of holding the `importedPackages` in a dedicated variable, I changed this to an "associatedValue".
As on this value was synchronized, I synchronize now on the scope that holds the associated value.

Someone may take a look at this, if this is correct.

Cheers
Roland

### DIFF
--- a/src/org/mozilla/javascript/ImporterTopLevel.java
+++ b/src/org/mozilla/javascript/ImporterTopLevel.java
@@ -103,9 +103,13 @@ public class ImporterTopLevel extends TopLevel {
 
     private Object getPackageProperty(String name, Scriptable start) {
         Object result = NOT_FOUND;
-        Object[] elements;
-        synchronized (importedPackages) {
-            elements = importedPackages.toArray();
+        Scriptable scope = start;
+        if (topScopeFlag) {
+            scope = ScriptableObject.getTopLevelScope(scope);
+        }
+        Object[] elements = getNativeJavaPackages(scope);
+        if (elements == null) {
+            return result;
         }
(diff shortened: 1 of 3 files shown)


gbrail:
This all seems fine to me. If no one else in the community wants to experiment with it I'll approve it tomorrow.

PR #829, opened 2021-01-24
Make sure setting "functionName = null" is propagated to thisObject.

rbri:


gbrail:
Sounds like a good compatibility fix, and I doubt that it will break old stuff. 

However, could you improve the commit message so that it tells what it fixes? That'll be the basis of the release notes (or, update this PR at least). That way people will know what happens if they did indeed affect this.

rbri:
Ok, what about:

Make sure setting "functionName = null" is propagated to thisObject

rbri:
@gbrail any hope for this?

gbrail:
Yes, there is always hope. And thank you!

PR #830, opened 2021-01-25
Improve NativeJavaList to behave more like NativeArray

rPraml:
This PR extends the NativeJavaList to behave more like NativeArray.

Some ideas are extracted from https://github.com/mozilla/rhino/pull/827

@tuchida maybe you can take a look at this.


testsrc/org/mozilla/javascript/tests/NativeJavaListTest.java
```diff
@@ -88,6 +88,263 @@ public void testUpdatingJavaListStringValues() {
...
+        list.add("b");
+        list.add("c");
+        assertEquals("1", runScriptAsString("value.indexOf('b')", list));
+        assertEquals("-1", runScriptAsString("value.indexOf('d')", list));
+        
+        List<Double> listD = new ArrayList<>();
+        listD.add(1.0);
+        listD.add(2.0);
+        listD.add(3.0);
+        assertEquals("1", runScriptAsString("value.indexOf(2)", listD));
```
tuchida, line 216: Isn't this calling Java's `indexOf` method?

https://docs.oracle.com/javase/8/docs/api/java/util/List.html#indexOf-java.lang.Object-
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/indexOf

They differ in the following ways.
- JS can receive `fromIndex` as the second argument.
```js
list.indexOf(1, 2);
// js: Can't find method java.util.ArrayList.indexOf(number,number).
```
- The equivalence check is different.
```js
list.add(NaN);
list.indexOf(NaN); // 0

[NaN].indexOf(NaN); // -1
```
https://docs.oracle.com/javase/8/docs/api/java/lang/Double.html#equals-java.lang.Object-

There may be compatibility issues.

tuchida:
In Rhino, all methods of an instance can be accessed.
```java
// java
public static List getList() {
    return new ArrayList();
}
```
```js
// js
var list = Foo.getList();
// ArrayList method, not a List method.
list.forEach;
// function forEach() {/*
// void forEach(java.util.function.Consumer)
// */}
```
When calling Java's methods against the return value of `getList`, it is safe to use only the `List`'s methods. However, Java methods have priority, so if you use JS's methods, you need to avoid `ArrayList`'s methods. If `getList` is changed to return a `List` implementation that is not an `ArrayList`, then you may not be able to call the JS's methods.

tuchida:
I think it would be better to implement `Symbol.iterator` like `NodeList` and convert it with `Array.from` to avoid problems.
https://developer.mozilla.org/en-US/docs/Web/API/NodeList

rPraml:
> When calling Java's methods against the return value of getList, it is safe to use only the List's methods.
> However, Java methods  have priority, so if you use JS's methods, you need to avoid ArrayList's methods.
> If getList is changed to return a List implementation that is not an ArrayList, then you may not be able to call the JS's methods.

On the one hand I think, that JS methods should have always priority, maybe we should not inherit from NativeJavaObject at all??
On the other hand, implementing a method in a Java class (e.g. join()) is like overloading in JS.

> I think it would be better to implement `Symbol.iterator` like `NodeList` and convert it with `Array.from` to avoid problems.
> https://developer.mozilla.org/en-US/docs/Web/API/NodeList

Hmm, you mean:
```javascript
var javaList = new java.util.ArrayList()
var jsList = Array.from(javaList)
```

This means, that changes on `jsList` are not reflected to `javaList

tuchida:
If it don't prioritize Java methods, it may run into problems with code written in the past.
However, whichever it prioritize, it will behave strangely in corner cases as long as the names conflict.

rPraml:
> If it don't prioritize Java methods, it may run into problems with code written in the past.

My bad. Changes in this way should not be done.

> However, whichever it prioritize, it will behave strangely in corner cases as long as the names conflict.

In most cases, there should be no conflicts, as in most cases, ArrayLists are used. And if there are conflicts, it "works as designed", respectively, this should be specified as "design"

BTW: Sry. I assume most of my "review" comments were not yet visible for you, because they were "pending"

gbrail:
I'm sorry -- I've lost track of what we collectively want to do with this collection of this PR. @rPraml I am hoping that we can still resolve this but I'm not sure where you left it on the review comments. Do we still want to pursue this?

rPraml:
@gbrail I updated this PR (I had not much time in the past)
Not much is left, because many smaller changes are already implemented (Thanks to @tuchida and others)
So what's left:
- Fixed `ScriptRuntime.sameZero` und `ScriptRuntime.shallowEq` when a Wrapper is passed (in my case a wrapped integer was compared with a primitive in Array.indexOf)
- NativeJavaList does a [setPrototype(ScriptableObject.getClassPrototype(scope, "Array"))](https://github.com/mozilla/rhino/pull/830/files#diff-0816273ed7d515a6f035acc7c6b55efb9a40774f30544df6648af9d1ad753b73R21), so all JavaScript Array methods works also on a `java.util.List`.
Note: If there is a name clash between Array prototype and `java.util.List` (e.g. `Array.indexOf()` and `java.util.List.indexOf()`) the java method is called. See [here](https://github.com/mozilla/rhino/pull/830/files#diff-0816273ed7d515a6f035acc7c6b55efb9a40774f30544df6648af9d1ad753b73R21)
This is required to be backward compatible. 
- NativeJavaList has methods to set length and to delete entries (`delete javaArr[0]`)

src/org/mozilla/javascript/NativeJavaList.java
```diff
@@ -39,6 +42,13 @@ public boolean has(int index, Scriptable start) {
         }
         return super.has(index, start);
     }
+    
+    public void delete(int index) {
+        if (isWithValidIndex(index)) {
```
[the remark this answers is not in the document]
rPraml: IMHO this is a tradeOff, I would not add this complexity to NativeArray to simulate "empty slots"

src/org/mozilla/javascript/NativeJavaList.java
```diff
@@ -17,6 +18,7 @@ public NativeJavaList(Scriptable scope, Object list) {
         super(scope, list, list.getClass());
         assert list instanceof List;
         this.list = (List<Object>) list;
+        setPrototype(ScriptableObject.getClassPrototype(scope, "Array"));
```
tuchida, line 21: What happens if you are using instanceof to determine if it is an array?
```js
if (value instanceof Array) {
  // some code to operate `value`
}
```
How about `Array.isArray`?

rPraml: Hello Tuchida,  `value instanceof Array` will return `true`, while `Array.isArray(value)` will return false.

I get different results in firefox

class inheritance

```javaScript
class Foo extends Array {};
var y = new Foo();
y instanceof Array; // returns true
Array.isArray(y); // returns true
```

vs. prototype

```javascript
var Bar = function(){};
Bar.prototype = Array.prototype;
var x = new Bar();
x instanceof Array; // returns true
Array.isArray(x); // returns false
```

What do you think, is ok, when we return true for `instanceof` and false for `isArray`?

tuchida: I think `javaList instanceof Array` should be `false`.

Java's `List` and JavaScript's `Array` are different, so their detailed behaviors are different.(For example, `.indexOf`)
If the existing code is written as follows, a `List` will be passed to the code that is expecting an `Array`.
```js
if (value instanceof Array) {
  // some code to operate `value`
}
```
I think it may cause compatibility issues.

rPraml:
@tuchida I made JavaList/JavaObject available as prototypes and made some fixes in class inheritance, so all of them is true:
```javascript
var x= new java.util.ArrayList();
x instanceof JavaList; 
x instanceof JavaObject; 
x instanceof Object;
x instanceof java.util.List
```

To add the array functions, I can define them now on the `JavaList`
```javascript
Object.getOwnPropertyNames(Array.prototype).forEach(function(x) { JavaList.prototype[x] = Array.prototype[x]})
```
I think this would be the nicest solutuion, when someone wants to handle JavaList as an array.

What do you think about that change?

tuchida (commented):
To what extent do you want Java's `List` and JS's `Array` to be the same?
If we make such a change, it would be better to discuss the specification before implementing it.
For example, as p-bakker commented, I think there are issues that should be considered for the specification, such as investigating how GraalJS and Nashorn are doing.

tuchida:
I consider Java's List and JS's Array to be different things, so I think mutual conversion is sufficient.
```js
var javaList = java.util.ArrayList([1, 2, 3]);
var jsArray = Array.from(javaList);
```

testsrc/org/mozilla/javascript/tests/NativeJavaListTest.java
```diff
@@ -134,13 +482,31 @@ private String runScriptAsString(String scriptSourceText, Object value) {
...
                 .call(
                         context -> {
-                            Scriptable scope = context.initStandardObjects(global);
+                            Scriptable scope = context.newObject(global);
                             scope.put("value", scope, Context.javaToJS(value, scope));
                             return convert.apply(
-                                    context.evaluateString(scope, scriptSourceText, "", 1, null));
+                                    context.evaluateString(
+                                            scope,
+                                            "Object.getOwnPropertyNames(Array.prototype).forEach(function(x) { JavaList.prototype[x] = Array.prototype[x]});"
```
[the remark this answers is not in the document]
tuchida: Extending the prototype of a built-in object is an anti-pattern.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Inheritance_and_the_prototype_chain#1_new_initialization

src/org/mozilla/javascript/NativeJavaList.java
```diff
@@ -9,9 +9,17 @@
 import java.util.List;
 
 public class NativeJavaList extends NativeJavaObject {
+    private static final long serialVersionUID = 1L;
 
     private List<Object> list;
 
+    static void init(ScriptableObject scope, boolean sealed) {
+        NativeJavaList obj = new NativeJavaList();
+        obj.exportAsJSClass(scope, "JavaObject", sealed);
```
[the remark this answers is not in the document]
tuchida: Why these three? There are many interfaces and classes in Java. For example, if you want to extend a `Set` or `Stream` with a prototype, do you have to add more JavaXXX each time? Why `List` instead of `Collection`?

If you have a problem you want to solve, it is better to fully discuss what specifications you want to use before implementing them.

rPraml (commented):
> Haven't followed along here, but you might want to look at how GraalJS handles Java List interoperability, if only to see how they solved things

I made some tests with GraalVM - I think this is the way we should also go in 
```javascript
> var ArrayList = Java.type("java.util.ArrayList");
> var List = Java.type("java.util.List")
> var Map = Java.type("java.util.Map")
> typeof ArrayList
function
> var list = new ArrayList();
> typeof list
object
> var arr = []
> typeof arr
object
> arr instanceof Object
true
> list instanceof Object
false
> list instanceof List
true
> list instanceof Map
false
> arr instanceof List
false

> Object.getPrototypeOf(list)
null

> Object.getOwnPropertyNames(arr)
["length"]
> Object.getOwnPropertyNames(list)
(31)["add", "remove", "get", "clone", "indexOf", "clear", "isEmpty", ...

> list[0] = 'a'
> list[1] = 'b'
> list[4] = 'c' // can only add element to the end
> list
(2)["a", "b"]
Array.prototype.includes.call(list, 'b')
true
Array.prototype.includes.call(list, 'c')
false
```

I will update the PR and remove unneccesary things

src/org/mozilla/javascript/NativeJavaMap.java
```diff
@@ -33,7 +37,7 @@ public String getClassName() {
     @Override
     public boolean has(String name, Scriptable start) {
         Context cx = Context.getCurrentContext();
-        if (cx != null && cx.hasFeature(Context.FEATURE_ENABLE_JAVA_MAP_ACCESS)) {
+        if (map != null && cx != null && cx.hasFeature(Context.FEATURE_ENABLE_JAVA_MAP_ACCESS)) {
```
tonygermano, line 40: I don't think we should be introducing the possibility of map being null.

rPraml:
I would close this PR and create a new one

rPraml:
closing this PR - all changes are in other PRs

Issue #832, opened 2021-01-31
Addition with string literal does not conform to ECMAScript spec


pastelmind:
* Rhino version: 1.7.13

If an object has both a `toString()` and a `valueOf()` method, which return different values, then adding (`+`) with a string literal causes an unexpected behavior:

```js
// Not standards-compliant:
({ toString: () => "foo", valueOf: () => 123 }) + "bar";
// Rhino:                   "foobar"
// SpiderMonkey (Firefox):  "123bar"
// V8 (Node.js, Chrome):    "123bar"
```

Using anything other than a string literal appears to trigger the "correct" behavior. For example, the following works:

```js
// Okay:
({ toString: () => "foo", valueOf: () => 123 }) + 1000;
// Rhino:                   1123
// SpiderMonkey (Firefox):  1123
// V8 (Node.js, Chrome):    1123

// Okay:
let stuff = "bar";
({ toString: () => "foo", valueOf: () => 123 }) + stuff;
// Rhino:                   "123bar"
// SpiderMonkey (Firefox):  "123bar"
// V8 (Node.js, Chrome):    "123bar"

// Okay:
({ toString: () => "foo", valueOf: () => 123 }) + String("bar");
// Rhino:                   "123bar"
// SpiderMonkey (Firefox):  "123bar"
// V8 (Node.js, Chrome):    "123bar"
```

## What spec is this violating?
According to the ECMA spec (I checked [ES5.1](https://262.ecma-international.org/5.1/#sec-11.6.1) and [ES6](https://262.ecma-international.org/6.0/#sec-addition-operator-plus)), the behavior in this case can be roughly summarized to:

1. Call `ToPrimitive(left)` with no hint &rarr; `lprim`
2. Call `ToPrimitive(right)` with no hint &rarr; `rprim`
3. If either `lprim` or `rprim` is string, use string concatenation
4. Otherwise, use numeric addition

The `ToPrimitive()` operation tries both `valueOf()` and `toString()` to determine the primitive value of an object:

* If the type hint is String:
   1. If `toString()` returns a primitive value, use it
   2. If `valueOf()` returns a primitive value, use it
* If the type hint is Number (with no hint, this is the default):
   1. If `valueOf()` returns a primitive value, use it
   2. If `toString()` returns a primitive value, use it

## Why is this happening?
I don't know.

Hypothesis: If either operand is a string literal...

* Rhino supplies a "String" hint to the other object, or...
[...]

PR #833, opened 2021-02-01
ref. #832 Call toPrimitive when adding a string

tuchida:
ref. #832
- Do you need to check the `LanguageVersion >= Context.VERSION_ES6`?
- This test case did not exist in test262. But I added tests for addition to test262.properties.

### DIFF
--- a/src/org/mozilla/javascript/ScriptRuntime.java
+++ b/src/org/mozilla/javascript/ScriptRuntime.java
@@ -2964,10 +2964,32 @@ public class ScriptRuntime {
         return new ConsString(toCharSequence(val1), toCharSequence(val2));
     }
 
+    /**
+     * https://262.ecma-international.org/11.0/#sec-addition-operator-plus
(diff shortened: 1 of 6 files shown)


pastelmind:
The [ECMA 5.1 spec](https://262.ecma-international.org/5.1/#sec-11.6.1) also indicates that `ToPrimitive()` must be applied to both operands of the Addition Operator. I think it shouldn't be an ES6+ thing.

tuchida:
ref. #814
Regardless of the ECMAScript specification, there is a policy of maintaining the old behavior for compatibility.

gbrail:
I was going to suggest that this be changed so only take effect with language version >= ES6. But I can see that you only had to change the BodyCodegen, but not the interpreter. Interestingly, the problem only happens with the classfile compiler and not in interpreted mode.

IMO that just makes this a Rhino bug that y'all found and fixed, so unless some weird thing comes up in my tests I'm going to merge it. Thanks!

PR #836, opened 2021-02-07
Disable Java Map accessing by properties

tuchida:
ref. #820
This will have an impact on these Pull Requests. #824, #827
___
#713 had compatibility issues because entry access by properties took precedence over Java methods. If Java methods are prioritized, compatibility issues will probably disappear.
But here's what happens.
```java
// Java
public Map getMap() {
    return new TreeMap();
}
```
```js
// Rhino
var m = Foo.getMap();
m.firstKey();  // TreeMap method
```
`getMap` returns `java.util.Map`, but an instance of it is `java.util.TreeMap`, so you can use the `java.util.TreeMap` method. If Java methods are prioritized, in order to access it by property, you need to avoid the `TreeMap` method instead of `Map`. You always need to know what the instance is. Also, if `getMap` returns a different Map implementation, your existing code may not work.
___
Nashorn has probably extended the use of indirect-eval-call to distinguish between property access and method invocation. I'm also checking for behavior such as the distinction between `m[0]` and `m['0']`, and that Java methods are not JavaScript Functions.

### DIFF
--- a/src/org/mozilla/javascript/Context.java
+++ b/src/org/mozilla/javascript/Context.java
@@ -349,6 +349,26 @@ public class Context
      */
     public static final int FEATURE_ENABLE_XML_SECURE_PARSING = 20;
 
+    /**
+     * Configure whether the entries in a Java Map can be accessed by properties.
+     *
+     * WARNING: This feature is similar to the one in Nashorn, but incomplete.
+     *
+     * 1. A entry has priority over method.
+     *
+     *   map.put("put", "abc");
+     *   map.put;  // abc
+     *   map.put("put", "efg"); // ERROR
+     *
+     * 2. The distinction between numeric keys and string keys is ambiguous.
+     *
+     *   map.put('1', 123);
+     *   map['1']; // Not found. This means `map[1]`.
+     *
+     * @since 1.7 Release 14
+     */
+    public static final int FEATURE_ENABLE_JAVA_MAP_ACCESS = 21;
+
     public static final String languageVersionProperty = "language version";
     public static final String errorReporterProperty   = "error reporter";
 
--- a/src/org/mozilla/javascript/Context.java
+++ b/src/org/mozilla/javascript/Context.java
@@ -352,6 +352,18 @@ public class Context
     /**
      * Configure whether the entries in a Java Map can be accessed by properties.
      *
+     * Not enabled:
+     *
+     *   var map = new java.util.HashMap();
+     *   map.put('foo', 1);
+     *   map.foo; // undefined
+     *
+     * Enabled:
+     *
+     *   var map = new java.util.HashMap();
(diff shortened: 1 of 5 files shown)


src/org/mozilla/javascript/Context.java
```diff
@@ -349,6 +349,26 @@
...
+     *   map.put;  // abc
+     *   map.put("put", "efg"); // ERROR
+     *
+     * 2. The distinction between numeric keys and string keys is ambiguous.
+     *
+     *   map.put('1', 123);
+     *   map['1']; // Not found. This means `map[1]`.
+     *
+     * @since 1.7 Release 14
+     */
```
gbrail, line 369: Can you add to this description a bit so that we can show what the flag does with a small example? Or we can write it up in the release notes or steal from the previous feature. I'd imagine we'd show an example like:

Before:
  var m = new java.util.HashMap();
  m.put('foo', 1);
  print(m.foo)
    undefined

After:
  var m = new java.util.HashMap();
  m.put('foo', 1);
  print(m.foo)
    1

src/org/mozilla/javascript/NativeJavaMap.java
```diff
@@ -25,63 +25,86 @@ public String getClassName() {
         return "JavaMap";
     }
 
-
     @Override
     public boolean has(String name, Scriptable start) {
-        if (map.containsKey(name)) {
-            return true;
+        Context cx = Context.getContext();
```
gbrail, line 30: "getContext" will throw a RuntimeError if there is no Context on the stack. That's usually incorrect but I know that there's code out there that might do it. Could you please replace these tests with:

Context cx = Context.getCurrentContext();
if (cx != null && cx.hasFeature(Context.FEATURE_ENABLE_JAVA_MAP_ACCESS))

(or make a helper function in Context or ScriptRuntime since we do that ll the time!)

gbrail (changes requested):
I think that this is the next PR that we should merge as we make List and Map work better. Can you please take a look at a few improvements? Thanks!

src/org/mozilla/javascript/Context.java
tuchida: 4322742 How about this?

src/org/mozilla/javascript/NativeJavaMap.java
tuchida: dc43827
I fixed `NativeJavaMap` and `NativeJavaList` since I based it on #713.

src/org/mozilla/javascript/Context.java
gbrail: Great. Thanks!

gbrail (commented):
Yes, those changes help. Thanks!

PR #837, opened 2021-02-13
ES2020 BigInt

tuchida:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt

# Compatibility
Raw `java.math.BigInteger` is now handled as JavaScript `BigInt`.
This leads to the following problems.
```java
final String script = "d + bi";
Utils.runWithAllOptimizationLevels(cx -> {

        Scriptable scope = cx.initStandardObjects();
        scope.put("d", scope, Double.valueOf(123));
        scope.put("bi", scope, new BigInteger("234"));

        Object result =  cx.evaluateString(scope, script, "", 1, null);
        Assert.assertEquals(Double.valueOf(357), result);
        return null;
});
```
It used to be interpreted as `123 + 234`, but now it is interpreted as `123 + 234n`, which results in an error. This will happen regardless of the `LanguageVersion`. It may be able to work around this by version checking if always pass the `Context` to `ScriptRuntime`.

# Benchmark
## master
```
SunSpiderBenchmark.accessBinaryTrees     avgt    5    7339.020 ±  163.986  us/op
SunSpiderBenchmark.accessFannkuch        avgt    5   24465.875 ±  486.237  us/op
SunSpiderBenchmark.accessNBody           avgt    5   18369.340 ±  834.781  us/op
SunSpiderBenchmark.accessNsieve          avgt    5    3967.743 ±   83.396  us/op
SunSpiderBenchmark.bitops3BitBitsInByte  avgt    5    2434.747 ±   69.952  us/op
SunSpiderBenchmark.bitopsBitsInByte      avgt    5   11837.858 ± 1274.065  us/op
SunSpiderBenchmark.bitopsBitwiseAnd      avgt    5   24415.034 ±  450.579  us/op
SunSpiderBenchmark.bitopsNsieveBits      avgt    5   19733.613 ±  825.524  us/op
SunSpiderBenchmark.controlflowRecursive  avgt    5    2924.783 ±   62.720  us/op
SunSpiderBenchmark.cryptoAes             avgt    5   10444.501 ±  305.440  us/op
SunSpiderBenchmark.cryptoMd5             avgt    5    7597.363 ±  538.611  us/op
SunSpiderBenchmark.cryptoSha1            avgt    5    5274.564 ±  555.156  us/op
SunSpiderBenchmark.dateFormatToFte       avgt    5   17574.064 ±  786.713  us/op
SunSpiderBenchmark.dateFormatXparb       avgt    5   24559.370 ± 2241.143  us/op
SunSpiderBenchmark.mathCordic            avgt    5    9568.241 ±  756.833  us/op
SunSpiderBenchmark.mathPartialSums       avgt    5   21454.715 ± 2229.847  us/op
SunSpiderBenchmark.mathSpectralNorm      avgt    5   10544.219 ±   77.957  us/op
SunSpiderBenchmark.regexpDna             avgt    5  110944.817 ± 1671.568  us/op
SunSpiderBenchmark.stringBase64          avgt    5   24295.712 ± 1267.732  us/op
SunSpiderBenchmark.stringFasta           avgt    5   29006.286 ±  533.307  us/op
SunSpiderBenchmark.stringTagcloud        avgt    5   37410.147 ±  825.179  us/op
SunSpiderBenchmark.stringUnpackCode      avgt    5   27798.423 ± 2576.532  us/op
SunSpiderBenchmark.stringValidateInput   avgt    5   16280.353 ± 1294.283  us/op
SunSpiderBenchmark.threeDCube            avgt    5   21931.051 ±  290.129  us/op
SunSpiderBenchmark.threeDMorph           avgt    5   14004.732 ±  513.980  us/op
SunSpiderBenchmark.threeDRayTrace        avgt    5   15150.070 ±  209.269  us/op

V8Benchmark.boyer          avgt    5  70481.542 ±  262.282  us/op
V8Benchmark.cryptoDecrypt  avgt    5  87530.566 ± 1282.020  us/op
V8Benchmark.cryptoEncrpyt  avgt    5   4824.302 ±   99.392  us/op
V8Benchmark.deltaBlue      avgt    5   6164.140 ±   54.848  us/op
V8Benchmark.earley         avgt    5  20443.091 ±  462.214  us/op
V8Benchmark.rayTrace       avgt    5  40234.364 ± 2703.675  us/op
V8Benchmark.richards       avgt    5   2912.571 ±   10.596  us/op
V8Benchmark.splay          avgt    5   2007.140 ±  383.734  us/op
```
## bigint
```
SunSpiderBenchmark.accessBinaryTrees     avgt    5    6857.694 ±  115.087  us/op
SunSpiderBenchmark.accessFannkuch        avgt    5   25297.146 ±  358.333  us/op
SunSpiderBenchmark.accessNBody           avgt    5   17830.198 ± 1607.994  us/op
SunSpiderBenchmark.accessNsieve          avgt    5    3661.630 ±  410.265  us/op
SunSpiderBenchmark.bitops3BitBitsInByte  avgt    5    1970.694 ±   57.911  us/op
SunSpiderBenchmark.bitopsBitsInByte      avgt    5   12308.511 ±  132.202  us/op
SunSpiderBenchmark.bitopsBitwiseAnd      avgt    5   22380.243 ±  531.675  us/op
SunSpiderBenchmark.bitopsNsieveBits      avgt    5   15186.223 ±  685.243  us/op
SunSpiderBenchmark.controlflowRecursive  avgt    5    2983.580 ±   30.732  us/op
SunSpiderBenchmark.cryptoAes             avgt    5   10921.327 ±  123.993  us/op
SunSpiderBenchmark.cryptoMd5             avgt    5    8026.452 ±  691.531  us/op
SunSpiderBenchmark.cryptoSha1            avgt    5    5661.606 ±  427.463  us/op
SunSpiderBenchmark.dateFormatToFte       avgt    5   17969.765 ±  314.480  us/op
SunSpiderBenchmark.dateFormatXparb       avgt    5   25856.467 ± 2411.417  us/op
SunSpiderBenchmark.mathCordic            avgt    5   10083.659 ±  980.310  us/op
SunSpiderBenchmark.mathPartialSums       avgt    5   22052.303 ±  724.467  us/op
SunSpiderBenchmark.mathSpectralNorm      avgt    5   12104.923 ±  434.196  us/op
SunSpiderBenchmark.regexpDna             avgt    5  107707.159 ± 7605.636  us/op
SunSpiderBenchmark.stringBase64          avgt    5   23835.379 ±  535.228  us/op
SunSpiderBenchmark.stringFasta           avgt    5   29075.257 ±  546.614  us/op
SunSpiderBenchmark.stringTagcloud        avgt    5   37760.985 ±  789.787  us/op
SunSpiderBenchmark.stringUnpackCode      avgt    5   29698.743 ±  717.608  us/op
SunSpiderBenchmark.stringValidateInput   avgt    5   16533.185 ±  278.765  us/op
SunSpiderBenchmark.threeDCube            avgt    5   22863.344 ±  459.779  us/op
SunSpiderBenchmark.threeDMorph           avgt    5   13546.212 ±  167.729  us/op
SunSpiderBenchmark.threeDRayTrace        avgt    5   15353.174 ±  224.194  us/op

V8Benchmark.boyer          avgt    5   68474.731 ± 3457.683  us/op
V8Benchmark.cryptoDecrypt  avgt    5  117865.924 ± 2687.926  us/op
V8Benchmark.cryptoEncrpyt  avgt    5    4895.171 ±  371.395  us/op
V8Benchmark.deltaBlue      avgt    5    6215.524 ±  255.234  us/op
V8Benchmark.earley         avgt    5   21386.777 ±  584.409  us/op
V8Benchmark.rayTrace       avgt    5   43347.104 ±  777.577  us/op
V8Benchmark.richards       avgt    5    2849.553 ±  141.519  us/op
V8Benchmark.splay          avgt    5    2041.655 ±  248.059  us/op
```
These two things seem to have changed a lot.
```
SunSpiderBenchmark.bitopsNsieveBits: 19733.613 => 15186.223
V8Benchmark.cryptoDecrypt: 87530.566 => 117865.924
```

### DIFF
--- a/src/org/mozilla/javascript/Icode.java
+++ b/src/org/mozilla/javascript/Icode.java
@@ -132,8 +132,17 @@ abstract class Icode {
             Icode_GENERATOR_RETURN = -65,
             Icode_YIELD_STAR = -66,
 
+            // Load BigInt register to prepare for the following BigInt operation
+            Icode_REG_BIGINT_C0 = -67,
+            Icode_REG_BIGINT_C1 = -68,
+            Icode_REG_BIGINT_C2 = -69,
+            Icode_REG_BIGINT_C3 = -70,
+            Icode_REG_BIGINT1 = -71,
+            Icode_REG_BIGINT2 = -72,
+            Icode_REG_BIGINT4 = -73,
+
             // Last icode
-            MIN_ICODE = -66;
+            MIN_ICODE = -73;
 
     static String bytecodeName(int bytecode) {
         if (!validBytecode(bytecode)) {
@@ -281,6 +290,20 @@ abstract class Icode {
                 return "GENERATOR_RETURN";
             case Icode_YIELD_STAR:
                 return "YIELD_STAR";
+            case Icode_REG_BIGINT_C0:
+                return "REG_BIGINT_C0";
+            case Icode_REG_BIGINT_C1:
+                return "REG_BIGINT_C1";
+            case Icode_REG_BIGINT_C2:
+                return "REG_BIGINT_C2";
+            case Icode_REG_BIGINT_C3:
+                return "REG_BIGINT_C3";
+            case Icode_REG_BIGINT1:
+                return "LOAD_BIGINT1";
+            case Icode_REG_BIGINT2:
+                return "LOAD_BIGINT2";
+            case Icode_REG_BIGINT4:
+                return "LOAD_BIGINT4";
         }
 
         // icode without name
--- a/src/org/mozilla/javascript/Icode.java
+++ b/src/org/mozilla/javascript/Icode.java
@@ -6,143 +6,136 @@
 
 package org.mozilla.javascript;
 
-/**
- * Additional interpreter-specific codes
- */
+/** Additional interpreter-specific codes */
 abstract class Icode {
 
     static final int
 
-    // delete operator used on a name
-        Icode_DELNAME                    = 0,
+            // delete operator used on a name
+            Icode_DELNAME = 0,
 
-    // Stack: ... value1 -> ... value1 value1
-        Icode_DUP                       = -1,
+            // Stack: ... value1 -> ... value1 value1
+            Icode_DUP = -1,
 
-    // Stack: ... value2 value1 -> ... value2 value1 value2 value1
-        Icode_DUP2                      = -2,
+            // Stack: ... value2 value1 -> ... value2 value1 value2 value1
+            Icode_DUP2 = -2,
 
-    // Stack: ... value2 value1 -> ... value1 value2
-        Icode_SWAP                      = -3,
+            // Stack: ... value2 value1 -> ... value1 value2
+            Icode_SWAP = -3,
 
-    // Stack: ... value1 -> ...
-        Icode_POP                       = -4,
+            // Stack: ... value1 -> ...
+            Icode_POP = -4,
 
-    // Store stack top into return register and then pop it
-        Icode_POP_RESULT                = -5,
+            // Store stack top into return register and then pop it
+            Icode_POP_RESULT = -5,
 
-    // To jump conditionally and pop additional stack value
-        Icode_IFEQ_POP                  = -6,
+            // To jump conditionally and pop additional stack value
+            Icode_IFEQ_POP = -6,
 
-    // various types of ++/--
-        Icode_VAR_INC_DEC               = -7,
-        Icode_NAME_INC_DEC              = -8,
-        Icode_PROP_INC_DEC              = -9,
-        Icode_ELEM_INC_DEC              = -10,
-        Icode_REF_INC_DEC               = -11,
+            // various types of ++/--
+            Icode_VAR_INC_DEC = -7,
+            Icode_NAME_INC_DEC = -8,
+            Icode_PROP_INC_DEC = -9,
+            Icode_ELEM_INC_DEC = -10,
+            Icode_REF_INC_DEC = -11,
 
-    // load/save scope from/to local
-        Icode_SCOPE_LOAD                = -12,
-        Icode_SCOPE_SAVE                = -13,
+            // load/save scope from/to local
+            Icode_SCOPE_LOAD = -12,
+            Icode_SCOPE_SAVE = -13,
+            Icode_TYPEOFNAME = -14,
 
-        Icode_TYPEOFNAME                = -14,
+            // helper for function calls
+            Icode_NAME_AND_THIS = -15,
+            Icode_PROP_AND_THIS = -16,
+            Icode_ELEM_AND_THIS = -17,
+            Icode_VALUE_AND_THIS = -18,
 
-    // helper for function calls
-        Icode_NAME_AND_THIS             = -15,
-        Icode_PROP_AND_THIS             = -16,
-        Icode_ELEM_AND_THIS             = -17,
-        Icode_VALUE_AND_THIS            = -18,
+            // Create closure object for nested functions
+            Icode_CLOSURE_EXPR = -19,
+            Icode_CLOSURE_STMT = -20,
 
-    // Create closure object for nested functions
-        Icode_CLOSURE_EXPR              = -19,
-        Icode_CLOSURE_STMT              = -20,
+            // Special calls
+            Icode_CALLSPECIAL = -21,
 
-    // Special calls
-        Icode_CALLSPECIAL               = -21,
+            // To return undefined value
+            Icode_RETUNDEF = -22,
 
-    // To return undefined value
-        Icode_RETUNDEF                  = -22,
+            // Exception handling implementation
+            Icode_GOSUB = -23,
+            Icode_STARTSUB = -24,
+            Icode_RETSUB = -25,
 
-    // Exception handling implementation
-        Icode_GOSUB                     = -23,
-        Icode_STARTSUB                  = -24,
-        Icode_RETSUB                    = -25,
+            // To indicating a line number change in icodes.
+            Icode_LINE = -26,
 
-    // To indicating a line number change in icodes.
-        Icode_LINE                      = -26,
+            // To store shorts and ints inline
+            Icode_SHORTNUMBER = -27,
+            Icode_INTNUMBER = -28,
 
-    // To store shorts and ints inline
-        Icode_SHORTNUMBER               = -27,
-        Icode_INTNUMBER                 = -28,
+            // To create and populate array to hold values for [] and {} literals
+            Icode_LITERAL_NEW = -29,
+            Icode_LITERAL_SET = -30,
 
-    // To create and populate array to hold values for [] and {} literals
-        Icode_LITERAL_NEW               = -29,
-        Icode_LITERAL_SET               = -30,
+            // Array literal with skipped index like [1,,2]
+            Icode_SPARE_ARRAYLIT = -31,
 
-    // Array literal with skipped index like [1,,2]
-        Icode_SPARE_ARRAYLIT            = -31,
+            // Load index register to prepare for the following index operation
+            Icode_REG_IND_C0 = -32,
+            Icode_REG_IND_C1 = -33,
+            Icode_REG_IND_C2 = -34,
+            Icode_REG_IND_C3 = -35,
+            Icode_REG_IND_C4 = -36,
+            Icode_REG_IND_C5 = -37,
+            Icode_REG_IND1 = -38,
+            Icode_REG_IND2 = -39,
+            Icode_REG_IND4 = -40,
 
-    // Load index register to prepare for the following index operation
-        Icode_REG_IND_C0                = -32,
-        Icode_REG_IND_C1                = -33,
-        Icode_REG_IND_C2                = -34,
-        Icode_REG_IND_C3                = -35,
-        Icode_REG_IND_C4                = -36,
-        Icode_REG_IND_C5                = -37,
-        Icode_REG_IND1                  = -38,
-        Icode_REG_IND2                  = -39,
-        Icode_REG_IND4                  = -40,
+            // Load string register to prepare for the following string operation
+            Icode_REG_STR_C0 = -41,
+            Icode_REG_STR_C1 = -42,
+            Icode_REG_STR_C2 = -43,
+            Icode_REG_STR_C3 = -44,
+            Icode_REG_STR1 = -45,
+            Icode_REG_STR2 = -46,
+            Icode_REG_STR4 = -47,
 
-    // Load string register to prepare for the following string operation
-        Icode_REG_STR_C0                = -41,
-        Icode_REG_STR_C1                = -42,
-        Icode_REG_STR_C2                = -43,
-        Icode_REG_STR_C3                = -44,
-        Icode_REG_STR1                  = -45,
-        Icode_REG_STR2                  = -46,
-        Icode_REG_STR4                  = -47,
+            // Version of getvar/setvar that read var index directly from bytecode
+            Icode_GETVAR1 = -48,
+            Icode_SETVAR1 = -49,
 
-    // Version of getvar/setvar that read var index directly from bytecode
-        Icode_GETVAR1                   = -48,
-        Icode_SETVAR1                   = -49,
+            // Load undefined
+            Icode_UNDEF = -50,
+            Icode_ZERO = -51,
+            Icode_ONE = -52,
 
-    // Load undefined
-        Icode_UNDEF                     = -50,
-        Icode_ZERO                      = -51,
-        Icode_ONE                       = -52,
+            // entrance and exit from .()
+            Icode_ENTERDQ = -53,
+            Icode_LEAVEDQ = -54,
+            Icode_TAIL_CALL = -55,
 
-    // entrance and exit from .()
-       Icode_ENTERDQ                    = -53,
-       Icode_LEAVEDQ                    = -54,
+            // Clear local to allow GC its context
+            Icode_LOCAL_CLEAR = -56,
 
-       Icode_TAIL_CALL                  = -55,
+            // Literal get/set
+            Icode_LITERAL_GETTER = -57,
+            Icode_LITERAL_SETTER = -58,
 
-    // Clear local to allow GC its context
-       Icode_LOCAL_CLEAR                = -56,
+            // const
+            Icode_SETCONST = -59,
+            Icode_SETCONSTVAR = -60,
+            Icode_SETCONSTVAR1 = -61,
 
-    // Literal get/set
-       Icode_LITERAL_GETTER             = -57,
-       Icode_LITERAL_SETTER             = -58,
+            // Generator opcodes (along with Token.YIELD)
+            Icode_GENERATOR = -62,
+            Icode_GENERATOR_END = -63,
+            Icode_DEBUGGER = -64,
+            Icode_GENERATOR_RETURN = -65,
+            Icode_YIELD_STAR = -66,
 
-    // const
-       Icode_SETCONST                   = -59,
-       Icode_SETCONSTVAR                = -60,
-       Icode_SETCONSTVAR1               = -61,
+            // Last icode
+            MIN_ICODE = -66;
 
-    // Generator opcodes (along with Token.YIELD)
-       Icode_GENERATOR                  = -62,
-       Icode_GENERATOR_END              = -63,
-
-       Icode_DEBUGGER                   = -64,
-
-       Icode_GENERATOR_RETURN           = -65,
-       Icode_YIELD_STAR                 = -66,
-
-       // Last icode
-        MIN_ICODE                       = -66;
-
-    static String bytecodeName(int bytecode)
-    {
+    static String bytecodeName(int bytecode) {
         if (!validBytecode(bytecode)) {
             throw new IllegalArgumentException(String.valueOf(bytecode));
         }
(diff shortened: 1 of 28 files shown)


src/org/mozilla/javascript/optimizer/Block.java
```diff
@@ -529,6 +529,7 @@ private static int findExpressionType(OptFunctionNode fn, Node n,
             case Token.ARRAYCOMP:
             case Token.ARRAYLIT:
             case Token.OBJECTLIT:
+            case Token.BIGINT:
```
tuchida, line 532: I wasn't sure if `BigInt` should be `AnyType` or not.
Should these also be `AnyType` as the result can be `BigInt`?
https://github.com/mozilla/rhino/blob/21198a516ee818c56dbd689198fdfcb4ec3dc79e/src/org/mozilla/javascript/optimizer/Block.java#L482-L497

testsrc/org/mozilla/javascript/tests/WrapFactoryTest.java
```diff
@@ -48,12 +48,12 @@ public void primitiveWrapFalse() {
     test(false, new Integer(1), "number", "number", "number");
     test(false, new Long(2L), "number", "number", "number");
 
-    // I want to treat BigInteger / BigDecimal as BigInteger / BigDecimal. But fails.
-    test(false, new BigInteger("30"), "number", "object", "object");
+    test(false, new BigInteger("30"), "bigint", "bigint", "bigint");
```
tuchida, line 51: Compatibility issue.

src/org/mozilla/javascript/optimizer/BodyCodegen.java
```diff
@@ -3421,26 +3431,55 @@ private static boolean isArithmeticNode(Node node)
...
             if (!isArithmeticNode(child.getNext()))
-                addObjectToDouble();
-            cfw.add(opCode);
-            if (!childOfArithmetic) {
-                addDoubleWrap();
+                addObjectToNumeric();
+
+            switch (type) {
+            case Token.SUB:
+                addScriptRuntimeInvoke("subtract", "(Ljava/lang/Number;Ljava/lang/Number;)Ljava/lang/Number;");
```
gbrail, line 3470: I bet that this is where some of the performance regressions might be coming in -- before, in most cases (because I have rarely seen the "numeric optimization" above really work -- we convert both operands to a double and then do the operation in one bytecode. This replaces all of those with a call to a function that does a bunch of instanceof checks before invoking the instruction. 

I wonder if this would benchmark better if we first called a method that checked the operand types for big-integerness in the most efficient way we can, then either used the old code path or the new one. Eventually invokedynamic could be used to make the same thing happen.

gbrail (commented):
Thanks! This is a very thorough implementation and it seems to do all the things. I have some smaller comments, in this review. If it didn't affect performance it'd be almost ready to go.

Other than that, the performance drop-off worries me. It might be because this PR replaces some arithmetic operations that were previously a single bytecode with a ScriptRuntime invocation that first does a few instanceof checks. Those are hot code paths so it's hard to optimize, but at the same time JavaScript doesn't give us much choice but to assume that any number might be a BigInt.

src/org/mozilla/javascript/optimizer/BodyCodegen.java
```diff
@@ -3421,26 +3431,55 @@ private static boolean isArithmeticNode(Node node)
...
+            default:
+                throw Kit.codeBug(Token.typeToName(type));
+            }
         }
         else {
             boolean childOfArithmetic = isArithmeticNode(parent);
             generateExpression(child, node);
             if (!isArithmeticNode(child))
-                addObjectToDouble();
+                addObjectToNumeric();
```
tuchida, line 3463:
```js
(print(1), { valueOf(){ print(2)}}) - (print(3), { valueOf(){ print(4)}})
// actual
// 1
// 2
// 3
// 4

// actual
// 1
// 3
// 2
// 4
```
This is correct.
https://github.com/mozilla/rhino/blob/d02cff98b66bbc9347874720a6d1dbca8195ab8f/src/org/mozilla/javascript/optimizer/BodyCodegen.java#L3460-L3467

src/org/mozilla/javascript/Token.java
```diff
@@ -124,116 +124,116 @@
         REF_MEMBER     = 78, // Reference for x.@y, x..y etc.
         REF_NS_MEMBER  = 79, // Reference for x.ns::y, x..ns::y etc.
         REF_NAME       = 80, // Reference for @y, @[y] etc.
-        REF_NS_NAME    = 81; // Reference for ns::y, @ns::y@[y] etc.
+        REF_NS_NAME    = 81, // Reference for ns::y, @ns::y@[y] etc.
+        BIGINT         = 82; // ES2020 BigInt
```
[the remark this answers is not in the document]
tuchida: Is it possible to change this to be defined enum or like itoa in go?
https://golang.org/ref/spec#Iota

src/org/mozilla/javascript/optimizer/BodyCodegen.java
tuchida: b3170b6

tuchida: Thanks. Sorry for the late reply.

I too have concerns about performance. However, I couldn't figure out what was slowed down by this change.

The bigint branch and master were each measured five times by `./gradlew testBenchmark`.
https://docs.google.com/spreadsheets/d/1sYM7EhGONY5-oQLFjuHwJou4DZEwJ7o9KCGB3-jKisc/edit?usp=sharing

There doesn't seem to be much difference except for `SunSpiderBenchmark.bitopsNsieveBits`. (`SunSpiderBenchmark.bitopsNsieveBits` is it getting faster?)

tuchida: Do you have any data or samples that show that using `invokedynamic` is faster?
How does `invokedynamic` work on Android?

gbrail:
Sorry I haven't responded -- I spent a bunch of time trying various things
to improve performance in general for arithmetic operations but all I
learned is that our benchmark suites have so much noise in them that it's
very hard to see small improvements. (The other thing I've learned,
however, is that it's easy to make Rhino slower and a lot harder to make it
faster.)

I do, however, have a handle on a PR that will use invokedynamic so that we
can decide in runtime whether to use a fairly optimized code path for
arithmetic that assumes certain things, or a less optimized one that
assumes nothing.

With that said, on Android the only choice for Rhino is interpreted mode,
so I suspect that none of these changes are going to make a big difference.

I'd like to spend more time on the other ideas I have before we merge this,
however -- I think that by then we'll have a way to make things fast
enough. However, I can't get to that for a week or two. I think it's
important that we figure out a good way to include this feature, but I'm
not ready for it yet. Thanks!

src/org/mozilla/javascript/Icode.java
```diff
@@ -138,8 +138,17 @@
        Icode_GENERATOR_RETURN           = -65,
        Icode_YIELD_STAR                 = -66,
 
+    // BigInt
```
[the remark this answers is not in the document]
tuchida: Is that okay?
https://github.com/mozilla/rhino/blob/3e5eea461b7500f3c25793553f76c6a831b7f372/src/org/mozilla/javascript/Icode.java#L141

src/org/mozilla/javascript/Interpreter.java
```diff
@@ -655,6 +657,29 @@ static void dumpICode(InterpreterData idata)
                 out.println(tname+" "+indexReg);
                 ++pc;
                 break;
+              // TODO: Icode_REG_STR_C0-3 is not dump. I made this the same it.
```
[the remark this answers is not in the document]
tuchida: Is that okay?
https://github.com/mozilla/rhino/blob/3e5eea461b7500f3c25793553f76c6a831b7f372/src/org/mozilla/javascript/Interpreter.java#L660-L666

src/org/mozilla/javascript/NativeBigInt.java
```diff
@@ -0,0 +1,216 @@
...
+
+package org.mozilla.javascript;
+
+import java.math.BigInteger;
+import java.util.Arrays;
+
+/**
+ * This class implements the BigInt native object.
+ */
+final class NativeBigInt extends IdScriptableObject
```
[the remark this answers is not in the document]
tuchida: I made in the same way as `NativeNumber`.

src/org/mozilla/javascript/NativeBigInt.java
```diff
@@ -0,0 +1,216 @@
...
+{
+    private static final long serialVersionUID = 1L;
```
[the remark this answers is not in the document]
tuchida:
```sh
$ serialver -classpath buildGradle/libs/rhino-1.7.14-SNAPSHOT.jar org.mozilla.javascript.NativeBigInt
```

https://github.com/mozilla/rhino/blob/3e5eea461b7500f3c25793553f76c6a831b7f372/src/org/mozilla/javascript/NativeBigInt.java#L17

src/org/mozilla/javascript/NativeBigInt.java
```diff
@@ -0,0 +1,216 @@
...
+        return 0;
+    }
+
+// #string_id_map#
+
+    @Override
+    protected int findPrototypeId(String s)
+    {
+        int id;
+// #generated# Last update: 2007-05-09 08:15:50 EDT
```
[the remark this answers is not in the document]
tuchida: https://github.com/mozilla/rhino/blob/3e5eea461b7500f3c25793553f76c6a831b7f372/src/org/mozilla/javascript/NativeBigInt.java#L179

Is this really fast? I feel it would be easier for the JVM to optimize using Strings in switch in Java7.
https://docs.oracle.com/javase/8/docs/technotes/guides/language/strings-switch.html

src/org/mozilla/javascript/Token.java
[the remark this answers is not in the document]
tuchida: I fixed.
https://github.com/mozilla/rhino/blob/3e5eea461b7500f3c25793553f76c6a831b7f372/src/org/mozilla/javascript/Token.java#L129

src/org/mozilla/javascript/optimizer/BodyCodegen.java
tuchida: We need to figure out what to do with `isArithmeticNode`, so we'll leave that for next time.

tuchida:
Here's what I'm doing now.
```java
if (val1 instanceof BigInteger && val2 instanceof BigInteger) {
  // calc
} else if (val1 instanceof Number && val2 instanceof Number) {
  // calc
}
```
It may be faster this way.
```java
if (val1 instanceof Double && val2 instanceof Double) {
  // calc
} else if (val1 instanceof BigInteger && val2 instanceof BigInteger) {
  // calc
} else if (val1 instanceof Number && val2 instanceof Number) {
  // calc
}
```

tuchida:
> SunSpiderBenchmark.bitopsNsieveBits

```
Benchmark                            Mode  Cnt      Score     Error  Units
SunSpiderBenchmark.bitopsNsieveBits  avgt    5  18908.779 ± 148.335  us/op

commit 5181accd8a5eeb8304c5efda7840798ac4b16f35 (HEAD)
    bit operations for BigInt
```
```
Benchmark                            Mode  Cnt      Score     Error  Units
SunSpiderBenchmark.bitopsNsieveBits  avgt    5  14542.023 ± 286.907  us/op

commit c8656ddb467a48d3a666ada3c2a9d0e9c63f97bc (HEAD)
    bitwise NOT for BigInt
```
c8656ddb467a48d3a666ada3c2a9d0e9c63f97bc seems to have made it faster.
https://github.com/mozilla/rhino/blob/c8656ddb467a48d3a666ada3c2a9d0e9c63f97bc/testsrc/benchmarks/sunspider-0.9.1/bitops-nsieve-bits.js

tuchida:
b525566 This is faster.
```
Benchmark                                Mode  Cnt       Score      Error  Units
SunSpiderBenchmark.accessBinaryTrees     avgt    5    6896.580 ±  349.558  us/op
SunSpiderBenchmark.accessFannkuch        avgt    5   25249.452 ± 1179.327  us/op
SunSpiderBenchmark.accessNBody           avgt    5   18699.092 ±  727.417  us/op
SunSpiderBenchmark.accessNsieve          avgt    5    3902.934 ±  165.557  us/op
SunSpiderBenchmark.bitops3BitBitsInByte  avgt    5    2085.102 ±  200.521  us/op
SunSpiderBenchmark.bitopsBitsInByte      avgt    5   12264.252 ±  459.047  us/op
SunSpiderBenchmark.bitopsBitwiseAnd      avgt    5   22322.207 ±  842.092  us/op
SunSpiderBenchmark.bitopsNsieveBits      avgt    5   10660.768 ±  426.337  us/op
SunSpiderBenchmark.controlflowRecursive  avgt    5    2946.532 ±  204.685  us/op
SunSpiderBenchmark.cryptoAes             avgt    5   10951.822 ±  315.705  us/op
SunSpiderBenchmark.cryptoMd5             avgt    5    8082.551 ±  399.776  us/op
SunSpiderBenchmark.cryptoSha1            avgt    5    5523.482 ±   77.351  us/op
SunSpiderBenchmark.dateFormatToFte       avgt    5   18337.982 ± 2044.831  us/op
SunSpiderBenchmark.dateFormatXparb       avgt    5   25343.045 ± 3758.116  us/op
SunSpiderBenchmark.mathCordic            avgt    5   10282.664 ±  515.868  us/op
SunSpiderBenchmark.mathPartialSums       avgt    5   22594.311 ± 1290.731  us/op
SunSpiderBenchmark.mathSpectralNorm      avgt    5   11819.411 ±  874.842  us/op
SunSpiderBenchmark.regexpDna             avgt    5  111098.051 ± 7397.927  us/op
SunSpiderBenchmark.stringBase64          avgt    5   23172.763 ±  856.285  us/op
SunSpiderBenchmark.stringFasta           avgt    5   28796.824 ±  339.958  us/op
SunSpiderBenchmark.stringTagcloud        avgt    5   37317.970 ± 1359.425  us/op
SunSpiderBenchmark.stringUnpackCode      avgt    5   27426.340 ±  596.161  us/op
SunSpiderBenchmark.stringValidateInput   avgt    5   17481.931 ± 1348.579  us/op
SunSpiderBenchmark.threeDCube            avgt    5   22605.476 ± 3175.912  us/op
SunSpiderBenchmark.threeDMorph           avgt    5   11958.454 ± 1150.110  us/op
SunSpiderBenchmark.threeDRayTrace        avgt    5   16087.591 ± 1250.476  us/op

Benchmark                  Mode  Cnt      Score      Error  Units
V8Benchmark.boyer          avgt   10  70875.019 ± 1976.893  us/op
V8Benchmark.cryptoDecrypt  avgt   10  89292.307 ± 2279.035  us/op
V8Benchmark.cryptoEncrpyt  avgt   10   4724.938 ±  106.322  us/op
V8Benchmark.deltaBlue      avgt   10   6032.947 ±   78.235  us/op
V8Benchmark.earley         avgt   10  19685.939 ±  714.160  us/op
V8Benchmark.rayTrace       avgt   10  42013.522 ± 1908.857  us/op
V8Benchmark.richards       avgt   10   2985.615 ±   73.166  us/op
V8Benchmark.splay          avgt   10   2294.107 ±  194.044  us/op
```
It seems that `bitopsNsieveBits` is now about twice as fast as master.
`Token.POS`/`Token.NEG` also looks like it could be optimized.
Shall I create another Pull Request?

src/org/mozilla/javascript/NativeBigInt.java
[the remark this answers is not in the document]
tuchida: ref. #847

rbri:
> Token.POS/Token.NEG also looks like it could be optimized.

for me any optimization is welcome.

@gbrail - is there something that blocks merging this? While working on the 626 update i saw a lot of bigint stuff that might work if this is done

tonygermano:
@tuchida In PR #860 I made a more generalized message for when JSON.stringify is unable to serialize an object and throws a TypeError. I based it on the message that Chrome returns when trying to stringify a BigInt. Not sure if you would want to make that small adjustment to avoid having two separate messages for the same purpose (assuming my PR eventually gets merged.)

https://github.com/mozilla/rhino/blob/663dbe2ad5106105c5c914db413e84b73b3261cb/src/org/mozilla/javascript/resources/Messages.properties#L254-L262

tuchida:
Thanks! If your PR is merged, I will fix the message.

src/org/mozilla/javascript/Interpreter.java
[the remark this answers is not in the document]
gbrail: This is great. Thanks.

src/org/mozilla/javascript/NativeBigInt.java
```diff
@@ -0,0 +1,212 @@
...
+        if (SymbolKey.TO_STRING_TAG.equals(k)) {
+            return SymbolId_toStringTag;
+        }
+        return 0;
+    }
+
+// #string_id_map#
+
+    @Override
+    protected int findPrototypeId(String s)
```
gbrail, line 176: Since you already worked to replace these bits with string switches, can you update this method as well?

src/org/mozilla/javascript/ScriptRuntime.java
```diff
@@ -733,6 +743,148 @@ public static double toNumber(String s) {
         }
     }
 
+    /**
+     * Convert the value to a BigInt.
+     */
+    public static BigInteger toBigInt(Object val)
+    {
+        for (;;) {
+            if (val instanceof BigInteger)
```
gbrail, line 752: I'd like to get to a point where we start to apply a formatting standard to the code. I think that one thing that we'll want to enforce is using brackets with conditional statements, even one-line ones. So this is a nit but please consider it.

src/org/mozilla/javascript/ScriptRuntime.java
```diff
@@ -2997,6 +3180,132 @@ public static CharSequence add(Object val1, CharSequence val2) {
...
+            throw ScriptRuntime.typeErrorById("msg.cant.convert.to.number", "BigInt");
+        } else {
+            int result = toInt32(val1.doubleValue()) ^ toInt32(val2.doubleValue());
+            return Double.valueOf(result);
+        }
+    }
+
+    public static Number leftShift(Number val1, Number val2) {
+        if (val1 instanceof BigInteger && val2 instanceof BigInteger) {
+            // TODO: val2 is supported only in the range of 32-bit integer.
```
gbrail, line 3278: Do we need to enter a bug or something to fix this TODO? What will happen now if val2 is out of range?

src/org/mozilla/javascript/optimizer/BodyCodegen.java
gbrail: If there are performance regressions here, then I think this is where they'd come, because there is at least a theoretical optimization here that we're partially undoing. If you have any ideas how to keep the old code path in some more cases then I'd be more comfortable. For instance, if "childOfArithmetic" is true at all, can we assume that there is no BigInteger object involved? but I might be OK just doing it this way and then trying to restore the old optimizations later.

src/org/mozilla/javascript/optimizer/BodyCodegen.java
```diff
@@ -3408,26 +3421,54 @@ private static boolean isArithmeticNode(Node node)
...
-            if (!childOfArithmetic) {
-                addDoubleWrap();
+                addObjectToNumeric();
+
+            switch (type) {
+            case Token.SUB:
+                addScriptRuntimeInvoke("subtract", "(Ljava/lang/Number;Ljava/lang/Number;)Ljava/lang/Number;");
+                break;
+            case Token.MUL:
+                addScriptRuntimeInvoke("multiply", "(Ljava/lang/Number;Ljava/lang/Number;)Ljava/lang/Number;");
```
gbrail, line 3462: Can we assume that the parameter types are Number objects for these methods? I'm not sure what we do elsewhere but it might be important that we assume that they are Objects instead.

gbrail (changes requested):
This is very complete and very close -- thanks for doing all the work.

Can you address a few of my small comments here, merge master again, and let me look at it one more time.

If I continue to be worried about bytecode performance (which will not affect Android BTW) then I might plan to make a few small changes on top of it myself, but we'll see then.

Thanks!

src/org/mozilla/javascript/ScriptRuntime.java
tonygermano: https://docs.oracle.com/javase/8/docs/api/java/math/BigInteger.html#intValue--

>if this BigInteger is too big to fit in an int, only the low-order 32 bits are returned

In Chrome when I try `1n << 2147483647n` (java.lang.Integer.MAX_VALUE) I get a RangeError: Maximum BigInt size exceeded.

It looks like java also throws an error with numbers that large:
```
js> var b = new java.math.BigInteger(1)
js> b.shiftLeft(java.lang.Integer.MAX_VALUE)
org.mozilla.javascript.WrappedException: Wrapped java.lang.ArithmeticException: BigInteger would overflow supported range
```
If the BigInteger is larger than Ingeter.MAX_VALUE you can get some pretty wacky incorrect results:

```
js> new java.math.BigInteger(java.lang.Integer.MAX_VALUE).shiftLeft(1).intValue()
-2
```

It looks like you can use BigInteger.intValueExact() instead, which will throw an ArithmeticException if the number is too large to fit in an int.

gbrail: Can you change it to throw RangeError if the size is out of range? That sounds quite reasonable, and then we're done. I think that's a lot more preferable to letting Java exceptions leak out.

tonygermano: I looked at it again, and have another suggestion.

I think you will want to check if `val2.signum() == -1`, return  `signedRightShift(val1, val2.negate())`. Else if either `val1.shiftLeft()` or `val2.intValueExact()` throw an ArithmeticException, rethrow as a RangeError.

Likewise, in signedRightShift, if `val2.signum() == -1`, return `leftShift(val1, val2.negate())`. Else if `val2.intValueExact()` throws an ArithmeticException, return BigInteger.ZERO.

tuchida:
Thank you for your review. I will fix it in the near future.

By the way, are you sure about the compatibility? `java.math.BigInteger` inherits from `java.lang.Number`. Until now, `java.math.BigInteger` has been treated as a Double. It's a buggy code with a potential error, but this time, `TypeError` will be raised.

src/org/mozilla/javascript/ScriptRuntime.java
tuchida: I too like the writing style you suggested.
I want to automatically check what kind of writing style I should use. ref. #863

gbrail:
Thanks for the question -- no, I'm not sure about compatibility. I'd be more comfortable if the new capability only worked if the language version was >= ES6. However, I don't think it's worth putting that check into every arithmetic operation! That would be quite bad for performance.

For pure JavaScript code, there's no problem -- it's only a problem if someone is doing embedded code and creating BigInteger objects instead of Double or Integer objects, and then expecting to be able to mix them freely with other numbers. As you said that's actually incorrect code and to me it sounds like it'd also be very rare.

So unless you can think of a clever way to restrict this whole thing by language level I think that the risk of making this change is low.

src/org/mozilla/javascript/optimizer/BodyCodegen.java
tuchida: The parameter passed here is a Number or BigInt by `addObjectToNumeric`. With Object, it is hard to know explicitly where to do ToNumeric. If you use ToNumeric in `ScriptRuntime.multiply` method, the evaluation order will change, which violates the specification. ref. https://github.com/mozilla/rhino/pull/837#discussion_r579555719

src/org/mozilla/javascript/NativeBigInt.java
tuchida: The idswitch tool does not support the new coding style, so the code could not be generated correctly.
Do you mind if I correct it by hand?

src/org/mozilla/javascript/optimizer/Block.java
```diff
@@ -530,6 +530,7 @@ private static int findExpressionType(OptFunctionNode fn, Node n,
...
```
tuchida, line 533: It seems that there is still a problem with optimization.
I will fix it and redo the measurement.
```js
// $ java -jar buildGradle/libs/rhino-1.7.14-SNAPSHOT.jar -version 200 -opt 9

const one = 1n;
(() => 1n - one)(); // js: uncaught JavaScript runtime exception: TypeError: Cannot convert BigInt to a number
```

gbrail:
Thanks for continuing to iterate on this. Unfortunately it needs another merge and conflict resolution...

I'll hold off on doing anything for this until you tell me it's ready but I think it's very close now.

src/org/mozilla/javascript/ScriptRuntime.java
tuchida: https://github.com/mozilla/rhino/pull/837/commits/13dcf61bd7eddd57d1d858b42cc7d5b370e42ba3
I did this because not only `intValueExact`, but also `shiftLeft`, etc. sometimes threw `ArithmeticException`.
 I'm checking for error patterns in `testsrc/jstests/es2020/bigint.js`, please let me know if I'm missing anything.

src/org/mozilla/javascript/NativeBigInt.java
tuchida: https://github.com/mozilla/rhino/pull/837/commits/49009648ddbc0b3bff8636eaf01b3919785286b9
I did not use idswitch.

src/org/mozilla/javascript/ScriptRuntime.java
tuchida: Apparently, the current coding style allows one-line. I have checked and corrected what I have added.
https://github.com/mozilla/rhino/blob/a25c5176d37fad4da77e97a2e89953258801293e/src/org/mozilla/javascript/ScriptRuntime.java#L413-L415

tuchida:
Here's what the benchmark looks like.
https://docs.google.com/spreadsheets/d/17H7OstklzKuS7qBApHtpmD2IlQRUfEvhNUJqmxGIFYs/edit?usp=sharing

V8Benchmark.cryptoDecrypt is slower because it is faster when optimized by #851, which may have opened up the difference.
SunSpiderBenchmark.mathSpectralNorm seems to be caused by each operation being a little slower than the other, but I haven't looked into it too much.
Other than that, we are still investigating.

### Speedup suggestions.

First of all, this did not work. ref. https://github.com/mozilla/rhino/pull/837#issuecomment-792261727
I haven't tried any other methods, but I can think of the following
- Dynamic type inference using `invokedynamic`
- More types of static type inference
  - https://github.com/mozilla/rhino/blob/13dcf61bd7eddd57d1d858b42cc7d5b370e42ba3/src/org/mozilla/javascript/optimizer/Optimizer.java#L14-L16
    - NumericType (Number or BigInt)
    - IntType
- I've only seen the benchmark. How much faster or slower will it be in real world code?

gbrail:
Thanks -- let's keep at it. I have several experiments that have tried to use invokedynamic and none have helped yet, but with biginteger support in there it might. I'll see if I can find time next week to try them out.

rbri:
@tuchida sorry but i fear you have to adapt at least the TokenStream a bit because of the new numeric separator support - hope we have the bigint stuff finally in soon.

tonygermano:
What is the status of this? Should we put a pause on PRs that create merge conflicts until this one goes in?

gbrail:
I have been dragging my feet because of performance. But I recently
convinced myself that the performance is fine and that there are other
performance things that we can fix once it's in.

So I'm going to make the time today or tomorrow to fix the conflicts and
merge it.

src/org/mozilla/javascript/optimizer/BodyCodegen.java
```diff
@@ -3328,112 +3276,145 @@ private void visitIncDec(Node node)
...
+                    cfw.add(ByteCode.DREM);
+                    break;
+                default:
+                    throw Kit.codeBug(Token.typeToName(type));
+            }
+        } else {
             generateExpression(child, node);
-            if (!isArithmeticNode(child))
-                addObjectToDouble();
+            if (!isArithmeticNode(child)) addObjectToNumeric();
```
tuchida, line 3389: On optimization with `isArithmeticNode`:

1. This PR
2. Make the condition of `isArithmeticNode` the same as this
    https://github.com/mozilla/rhino/blob/0c05d30e134aed65b9c3536d8ac3764590840a35/src/org/mozilla/javascript/optimizer/Block.java#L482-L498
3. Delete `isArithmeticNode`

When I compared these results, `./gradlew testBenchmark` results were almost the same.
This is just a guess, but this benchmark was run with `OptimizationLevel` of 9, so it seems to be optimized by `childNumberFlag`. We may be able to find a way to speed up the process by setting the `OptimizationLevel` to 0 and verifying it.

tuchida:
I'm not a native speaker of English. I would appreciate it if you could take a look at Messages.properties to see if there are any strange sentences, even if they are trivial.

src/org/mozilla/javascript/resources/Messages.properties
```diff
@@ -519,6 +519,9 @@ msg.destruct.default.vals =\
 msg.no.old.octal.strict =\
     Old octal numbers prohibited in strict mode.
 
+msg.no.old.octal.bigint =\
+    Old octal numbers prohibited in BigInt.
+
```
tonygermano, line 524: I don't think this message is necessary. Below is the current behavior, and it is only triggered in strict mode, so I think the existing message is good enough, and you don't need a special case for bigint in the parser.

```
js> (function() {return 013})()
11
js> (function() {"use strict"; return 013})()
js: "<stdin>", line 24: Old octal numbers prohibited in strict mode.
js: (function() {"use strict"; return 013})()
js: .....................................^
js: "<stdin>", line 24: missing } after function body
js: (function() {"use strict"; return 013})()
js: ........................................^
js: "<stdin>", line 24: Compilation produced 2 syntax errors.

js> (function() {return 013n})() // bigint conversion to octal when not in strict mode
11
js> (function() {"use strict"; return 013n})()
js: "<stdin>", line 26: Old octal numbers prohibited in BigInt.
js: (function() {"use strict"; return 013n})()
js: ......................................^
js: "<stdin>", line 26: missing } after function body
js: (function() {"use strict"; return 013n})()
js: .........................................^
js: "<stdin>", line 26: Compilation produced 2 syntax errors.
```

Technically, the spec doesn't allow legacy octal notation at all, but lists is as a browser extension when not in strict mode. Chrome supports it in the same way that we do for numbers, but reports a generic Syntax Error for bigints.

Node.js reports a generic Syntax Error for both numbers and bigints, even when not in strict mode, per the spec. I think we should match this behavior for language version 200+. I would open an issue for this after merging, though. I don't think it needs to stop this PR from going in.

tonygermano: The rest of the messages look fine to me.

src/org/mozilla/javascript/NativeJSON.java
```diff
@@ -305,6 +317,9 @@ private static Object str(Object key, Scriptable holder,
         }
 
         if (value instanceof Number) {
+            if (value instanceof BigInteger) {
+                throw ScriptRuntime.typeErrorById("msg.cant.json.serialized.from.bigint");
```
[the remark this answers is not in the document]
tuchida: Thanks! This has been fixed.
https://github.com/mozilla/rhino/pull/837/commits/56d380232be4885d13e3195d68b116d0be633302#diff-ecefe1882290fb24e0b96edd98a83e5c7446d94a851ceeb829c94b94dc30bb10R334

I'm sorry that I rebased it so many times that it was hard to understand the additional corrections I made.

src/org/mozilla/javascript/resources/Messages.properties
tuchida: The latest test262 had this test case.
https://github.com/tc39/test262/blob/2ee3864136747ee69401b2d266e234cdd0a95965/test/language/literals/bigint/legacy-octal-like-invalid-00n.js

tuchida: There seems to have been a mistake when I fixed the conflicts with #814.

tonygermano: I think it's working well enough to merge now, and then fix after. Existing code won't break, because it previously would not parse at all.

gbrail: That's fine with me -- like we said the "123n" syntax didn't work at all before.

gbrail:
This is great. Thanks for all the hard work!

rbri:
Just a note about performance here. Did the merge into the HtmlUnit fork some days ago and rerun the whole HtmlUnit test suite (including various js framwork test suites. The whole test runs for > 3h and did not show any notable difference to the versions before - in terms of runtime.

Great work and thanks to @tuchida, @tonygermano and @gbrail. Nice too see all the progress here - this is really motivating.

Issue #817, opened 2021-01-01
Can we discuss how to update test262?


morfav:
Dear maintainers,

I thought it may be a good idea to update test262 to its current point.

It's taken me a lot longer (many hours) than I thought it would. The following issues cropped up:
1. test262 has introduced extra "helper" (harness) files which use syntax that rhino does not understand (3-4 such files). It's necessary to create a custom version of these that rhino would run (or use a previous version of these files if it's an existing one). Do you think it's viable that these custom versions will be maintained with potentially further divergence in future?
2. There are a lot of new, and some changed tests. Some of these pass, while others fail. It's a VERY time-consuming task to update the test262.properties file. I added all the new tests from git history to be excluded, however there were also quite a few (order of 1000) that pass, so need to be located and removed from the exclusions.
3. A couple of new tests seem to never complete. To overcome this, I added a new status of "hard" exclude, which prevents them from being run (since currently tests marked to be excluded with an exclamation mark are still run, which causes the tests to hang).

I feel like it may make this whole process easier if we change the approach to running these tests. Namely, keep a version-controlled list of tests that PASS (rather than what is currently done, which is keep a list of failing tests).
When fixing bugs, it would be possible to add to the list of passing tests.
When syncing with the test262 repo, we have a different mode which runs all the non-passing tests to see which ones pass and can be (automatically) included (we still need a hard exclude to prevent running the ones that never complete). This mode can also be triggered to search for tests that are passing following commits to rhino. One potential complication would be that currently we can include whole directories, so would need to see how this fits with the new approach.

Please let me know what you think about the above, and whether there is an easier way to maintain test262 in sync.

Thanks!

tuchida:
Could you please show me the rhino branch with the updated test262?
I'm currently implementing this in BigInt for ES2020, but I'm having trouble with a lack of test cases.
(I make it as a hobby, so don't expect it.)
https://github.com/mozilla/rhino/compare/master...tuchida:es2020-bigint

gbrail:
Sorry I didn't answer this before -- yes, I love the idea of making test262
easier to keep up to date and thanks, Vlad, for all your work on this in
the past.

Keeping a list of "passing" tests seems reasonable -- I hope that over time
we can expand this.

Do we know what we're missing in order to pass those new test files? I know
that there's some syntax that we're missing but I wonder how hard it'd be
to implement really.

Finally, are the tests that never complete possibly related to Promise
support? (that's the case with the compatibility table code). I had made
[...]

PR #838, opened 2021-02-13
Exponentiation Operator

tuchida:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Exponentiation

https://github.com/mozilla/rhino/issues/817#issuecomment-778128456
Although not directly related to this feature, the test for checking the limit of the array length became very time consuming, so it has been fixed.

### DIFF
--- a/src/org/mozilla/javascript/NativeArray.java
+++ b/src/org/mozilla/javascript/NativeArray.java
@@ -1930,8 +1930,7 @@ public class NativeArray extends IdScriptableObject implements List
         }
 
         long length = getLengthProperty(cx, o);
-        // TODO: In the case of Id_filter, CreateDataPropertyOrThrow throws a TypeError instead of ArraySpeciesCreate.
-        if ((id == Id_map || id == Id_filter) && length > Integer.MAX_VALUE) {
+        if (id == Id_map && length > Integer.MAX_VALUE) {
             String msg = ScriptRuntime.getMessageById("msg.arraylength.bad");
             throw ScriptRuntime.rangeError(msg);
         }
--- a/src/org/mozilla/javascript/NativeArray.java
+++ b/src/org/mozilla/javascript/NativeArray.java
@@ -723,7 +723,7 @@ public class NativeArray extends IdScriptableObject implements List
           }
         }
 
-        final long length = getLengthProperty(cx, items, false);
+        final long length = getLengthProperty(cx, items);
         final Scriptable result = callConstructorOrCreateArray(cx, scope, thisObj, length, true);
         for (long k = 0; k < length; k++) {
             Object temp = getRawElem(items, k);
@@ -846,7 +846,7 @@ public class NativeArray extends IdScriptableObject implements List
      * getLengthProperty returns 0 if obj does not have the length property
      * or its value is not convertible to a number.
      */
-    static long getLengthProperty(Context cx, Scriptable obj, boolean throwIfTooLarge) {
+    static long getLengthProperty(Context cx, Scriptable obj) {
         // These will both give numeric lengths within Uint32 range.
         if (obj instanceof NativeString) {
             return ((NativeString)obj).getLength();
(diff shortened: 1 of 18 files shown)


src/org/mozilla/javascript/NativeArray.java
```diff
@@ -862,17 +862,15 @@ static long getLengthProperty(Context cx, Scriptable obj, boolean throwIfTooLarg
         }
 
         double doubleLen = ScriptRuntime.toNumber(len);
+
+        // ToLength
         if (doubleLen > NativeNumber.MAX_SAFE_INTEGER) {
-            if (throwIfTooLarge) {
```
tuchida, line 866: It seems that this is probably the policy.
- get length
  - Up to 53 ** 2 - 1
  - https://262.ecma-international.org/11.0/#sec-tolength
- set length
  - Up to 31 ** 2 - 1
  - https://262.ecma-international.org/11.0/#sec-arraycreate

src/org/mozilla/javascript/Parser.java
```diff
@@ -2579,15 +2580,36 @@ private AstNode addExpr()
...
+    private AstNode expExpr()
+        throws IOException
+    {
+        AstNode pn = unaryExpr();
+        for (;;) {
+            int tt = peekToken(), opPos = ts.tokenBeg;
+            switch (tt) {
+            case Token.EXP:
+                if (pn instanceof UnaryExpression) {
+                    reportError("msg.no.unary.expr.on.left.exp", AstNode.operatorToString(pn.getType()));
```
tuchida, line 2608:
```
js> -1 ** 2
js: "<stdin>", line 2: "-" is not allowed on the left-hand side of "**".
js: -1 ** 2
js: .....^
```
The error appears to be in the wrong position.
I tried `void reportError(String messageId, int position, int length)`, but it didn't change.

gbrail:
This change looks OK to me. I think that you've handled the various limits on the lengths properly according to the spec that you linked. I see that you have a "TODO" about an error format -- do you want to fix that first?

tuchida:
Do you mean this?
https://github.com/mozilla/rhino/blob/c61469d7a17fbaf96c29d77172b4c3552f4bc1ad/src/org/mozilla/javascript/NativeArray.java#L1933-L1934
I think it's probably rare to have a problem, so I'd like to have a next time.

https://262.ecma-international.org/11.0/#sec-array.prototype.filter
In the specification, an error occurs here.
> 1. Perform ? CreateDataPropertyOrThrow(A, ! ToString(to), kValue).

Not here
> 4. Let A be ? ArraySpeciesCreate(O, 0).

src/org/mozilla/javascript/Parser.java
gbrail: I think that the shell does this for everything -- so fixing this shouldn't be in scope for this PR assuming that errors loaded from a regular file show up on the right line.

src/org/mozilla/javascript/NativeArray.java
```diff
@@ -1914,7 +1929,13 @@ private static Object iterativeMethod(Context cx, IdFunctionObject idFunctionObj
             requireObjectCoercible(cx, o, idFunctionObject);
         }
 
-        long length = getLengthProperty(cx, o, id == Id_map);
+        long length = getLengthProperty(cx, o);
+        // TODO: In the case of Id_filter, CreateDataPropertyOrThrow throws a TypeError instead of ArraySpeciesCreate.
```
gbrail, line 1933: This is what I'm asking about -- previously, we only threw an error in the case of Id_map. Is there a reason to also throw it for Id_filter too? If this is a change then you know I'm going to ask if we should check the language version too since if we didn't previously throw in this situation.

gbrail:
Sorry I wasn't clear -- the second TODO was about exception handling, so please check out my comments.

src/org/mozilla/javascript/NativeArray.java
tuchida: 25f317e
Sorry. There should have been a slow test, but I checked and it didn't slow down without throws an error.

gbrail:
Thanks for pulling this together!

Issue #842, opened 2021-02-24
Object.assign.apply doesn't execute correctly.


QuXing9:
## Description

As the following test case showed, when the value of the first argument to the function Object.assign.apply is `null`, rhino throws a java exception while all other engines such as V8, SpiderMonkey, and JavascriptCore execute successfully. According to the section [9.2.1.2](https://262.ecma-international.org/#sec-ordinarycallbindthis) of ES2020, when `thisArgument` is undefined or `null`, the JS engine should execute without an error.  Is this a normal behavior of Rhino?

## Version

[Rhino1.7.13](https://github.com/mozilla/rhino/releases/download/Rhino1_7_13_Release/rhino-1.7.13.jar)

## Run Command

~~~
java -jar rhino-1.7.13.jar testcase.js
~~~

## Testcase

~~~javascript
var NISLFuzzingFunc = function (nislMutationParameter0) {
    return Object.assign.apply(null,nislMutationParameter0);
};
var nislMutationArgument0 = [1,2,3,4];
var NISLCallingResult = NISLFuzzingFunc(nislMutationArgument0);
print(NISLCallingResult);
~~~

## Actual Output

~~~
at org.mozilla.javascript.ScriptableObject.getTopLevelScope(ScriptableObject.java:2245)
at org.mozilla.javascript.ScriptRuntime.setBuiltinProtoAndParent(ScriptRuntime.java:3994)
at org.mozilla.javascript.ScriptRuntime.toObject(ScriptRuntime.java:1125)
at org.mozilla.javascript.NativeObject.execIdCall(NativeObject.java:551)
at org.mozilla.javascript.IdFunctionObject.call(IdFunctionObject.java:100)
at org.mozilla.javascript.ScriptRuntime.applyOrCall(ScriptRuntime.java:2774)
at org.mozilla.javascript.BaseFunction.execIdCall(BaseFunction.java:308)
at org.mozilla.javascript.IdFunctionObject.call(IdFunctionObject.java:100)
at org.mozilla.javascript.optimizer.OptRuntime.call2(OptRuntime.java:55)
at org.mozilla.javascript.gen.___demo_js_1._c_anonymous_1(Unknown Source)
at org.mozilla.javascript.gen.___demo_js_1.call(Unknown Source)
at org.mozilla.javascript.optimizer.OptRuntime.callName(OptRuntime.java:76)
at org.mozilla.javascript.gen.___demo_js_1._c_script_0(Unknown Source)
[...]

PR #845, opened 2021-03-18
fix Object.assign ToObject calls to use correct scope

tonygermano:
fixes https://github.com/mozilla/rhino/issues/842

### DIFF
--- a/src/org/mozilla/javascript/NativeObject.java
+++ b/src/org/mozilla/javascript/NativeObject.java
@@ -555,16 +555,16 @@ public class NativeObject extends IdScriptableObject implements Map
           {
             Scriptable targetObj;
             if (args.length > 0) {
-              targetObj = ScriptRuntime.toObject(cx, thisObj, args[0]);
+              targetObj = ScriptRuntime.toObject(cx, scope, args[0]);
             }
             else {
-              targetObj = ScriptRuntime.toObject(cx, thisObj, Undefined.instance);
+              targetObj = ScriptRuntime.toObject(cx, scope, Undefined.instance);
             }
             for (int i = 1; i < args.length; i++) {
               if ((args[i] == null) || Undefined.isUndefined(args[i])) {
                 continue;
               }
-              Scriptable sourceObj = ScriptRuntime.toObject(cx, thisObj, args[i]);
+              Scriptable sourceObj = ScriptRuntime.toObject(cx, scope, args[i]);
               Object[] ids = sourceObj.getIds();
               for (Object key : ids) {
                 if (targetObj instanceof ScriptableObject) {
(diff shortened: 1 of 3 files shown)


tuchida:
I could not find any test cases in test262 that validate this issue.

tonygermano:
I couldn't either, but I did run all of the Object.assign tests from test262, and they passed after making my change. I think this is a rather odd way to call Object.assign in the first place, but the method signature of ScriptRuntime.toObject that is being called is `toObject(Context cx, Scriptable scope, Object val)`, so I think it was a mistake to be passing `thisObj` instead of `scope` as the scope to begin with.

tuchida:
> but I did run all of the Object.assign tests from test262, and they passed after making my change.

:+1: 

In the other `Object` methods, it seems to call `getCompatibleObject` instead of `ScriptRuntime.toObject` directly.
https://github.com/mozilla/rhino/blob/61493d5aca90f537058ac873f0fd9987d494d7e4/src/org/mozilla/javascript/NativeObject.java#L606-L613

However, `Object.assign` doesn't exist in ES5, so I think it doesn't matter either way.
https://262.ecma-international.org/5.1/#sec-15.2

gbrail:
This makes sense but I want to make sure that we have test coverage. Is there a test case that failed before you fixed this that we can add somewhere in the test suite?

tonygermano:
There is a test in referenced issue https://github.com/mozilla/rhino/issues/842

```javascript
var NISLFuzzingFunc = function (nislMutationParameter0) {
    return Object.assign.apply(null,nislMutationParameter0);
};
var nislMutationArgument0 = [1,2,3,4];
var NISLCallingResult = NISLFuzzingFunc(nislMutationArgument0);
print(NISLCallingResult);
```
I'm not sure of the appropriate place to add it.

tonygermano:
So I wrote a .jstest for this, and it actually passed. I'm not sure what language level those tests run at, but it looks like the test case in the issue fails on version 180 or 200, and passes on <180.

I tracked it down to here, where older versions apparently don't let you use `call` or `apply` with a null `this`.
 https://github.com/mozilla/rhino/blob/399b9f05a8effd48f105afc25e7f4a5dd8ffebf7/src/org/mozilla/javascript/ScriptRuntime.java#L2755-L2757

I'm not sure what to think about it passing at lower language versions, since Object.assign is technically an ES6 feature, anyway.

I just pushed a new commit with a test using `ScriptTestsBase` and `RhinoTest`, which does work. Is this the preferred way to write a test? I probably found 5 or 6 different ways that tests are bring run. Is there any guidance on writing tests? Which testing method to use, where to put them, naming conventions, etc...

gbrail:
Those tests look fine to me and setting them up the way that you did to make it work with ES6 only is fine. Thanks!

PR #846, opened 2021-03-20
fix #839 Iterable Java Object

### LINKED ISSUE #839: Iterable Java Object
I implemented `Symbol.iterator` of `NativeJavaMap` and `NativeJavaList`.
https://github.com/mozilla/rhino/compare/master...tuchida:iterable-javamap-and-javalist

This allows us to write the following.
```js
list.add(123);
for (var e of list) {
  print(e); // 123
}

var map = new java.util.HashMap();
map.put('abc', 123);
for (var [k, v] of map) {
  print([k, v]); // abc,123
}
```
However, this does not make `java.util.Set`, etc. iterable.

So I decided that `Symbol.iterator` exists in all `java.lang.Iterable`.
https://github.com/mozilla/rhino/compare/master...tuchida:iterable-javaobject
`java.util.Set`, etc., will also be iterable, but the writing style is different between JavaScript `Map`.
```js
for (var {key, value} of javaMap.entrySet()) ;
for (var [key, value] of jsMap) ;
```

Which do you think is better?

tuchida:
ref. #839
Implemented `Symbol.iterator` in `java.lang.Iterable` and `java.util.Map`.
`java.util.Iterator` is equivalent to the Iterator object in ECMAScript because it has state, so we did not implement `Symbol.iterator`.
`java.util.Map` can iterate over keys and values just like ECMAScript's `Map`. Nashorn works the same way.

### DIFF
--- a/src/org/mozilla/javascript/NativeJavaMap.java
+++ b/src/org/mozilla/javascript/NativeJavaMap.java
@@ -6,12 +6,17 @@
 package org.mozilla.javascript;
 
 import java.util.ArrayList;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 
 public class NativeJavaMap extends NativeJavaObject {
 
     private Map<Object, Object> map;
+
+    static void init(ScriptableObject scope, boolean sealed) {
+        NativeJavaMapIterator.init(scope, sealed);
+    }
 
     @SuppressWarnings("unchecked")
     public NativeJavaMap(Scriptable scope, Object map) {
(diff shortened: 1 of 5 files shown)


gbrail:
It seems like this is useful. Is it possible to update the commit message or this PR with a short sample of what is possible now that wasn't possible before so that we can put something in the release notes?

Also, how do we ensure that some existing code won't break now that iteration is easier on many types of Java objects?

tuchida:
> It seems like this is useful. Is it possible to update the commit message or this PR with a short sample of what is possible now that wasn't possible before so that we can put something in the release notes?

Can I write the release notes here?
https://github.com/mozilla/rhino/blob/master/RELEASE-NOTES.md

> Also, how do we ensure that some existing code won't break now that iteration is easier on many types of Java objects?

Since `Symbol.iterator` is a `Symbol`, it does not override Java methods.
Up until now, it was not iterable, which caused a `TypeError`.
```js
var l = new java.util.ArrayList([1, 2, 3]);
for (var v of l) print(v);
// js: uncaught JavaScript runtime exception: TypeError: [1.0, 2, 3] is not iterable
```
I don't think there are any side effects from this becoming iterable.

gbrail:
Thanks -- you have given me enough to put in the release notes when it's time. I might ask for help with the PR at that time if it's not obvious. It's cool that we're starting to use lambdas to cut down on code clutter too!

PR #852, opened 2021-03-27
fix #850 XMLList assignment by index of another node from the same XMLList

### LINKED ISSUE #850: Copying nodes in an XML tree removes the source node
The bug originally reported by Alberto Saez Torres at https://bugzilla.mozilla.org/show_bug.cgi?id=567484 is still present in 1.7.13.

According to the [e4x spec](https://www.ecma-international.org/wp-content/uploads/ECMA-357_2nd_edition_december_2005.pdf) in part 11.6.2, (emphasis mine)
> If the AssignmentExpression evaluates to an XML value, the assignment operator replaces the value of the property
identified by property name with a **_deep copy_** of the given XML object

Original text:

> When copying nodes in a node tree, removes the original Name.
> 
> For example, for the next code:
> 
> ```jsx
> var tmp =<xml>
>    <group>Group One</group>
>    <group>Group Two</group>
>    <group>Group Three</group>
> </xml>;
> 
> tmp.group[0]=tmp.group[2];
> ```
> The result XML of "tmp" is:
> ```
> <xml>
>    <group>Group Three</group>
>    <group>Group Two</group>
> </xml>
> ```
> The 3º "group" element has been removed after.
> 
> 
> 
> Reproducible: Always
> 
> Actual Results:  
> ```
> <xml>
>    <group>Group Three</group>
>    <group>Group Two</group>
> </xml>
> ```
> Expected Results:  
> tmp result as seeing in spiderMonkey
> ```
> <xml>
>    <group>Group Three</group>
>    <group>Group Two</group>
>    <group>Group Three</group>
> </xml>
> ```
> The problem is related to the function of XmlNode 
> ```java
> void replaceWith(XmlNode other) {
>   Node replacement = other.dom;
>   if (replacement.getOwnerDocument() != this.dom.getOwnerDocument()) {
>      replacement = this.dom.getOwnerDocument().importNode(replacement, true);
>   }
>   this.dom.getParentNode().replaceChild(replacement, this.dom);
> }
> ```
> When the function is aplied to a node of the same xml tree, the function importNode has the following behaviour:
> ```java
> Node replaceChild(Node newChild, Node oldChild)  throws DOMException
> ```
> Replaces the child node oldChild with newChild in the list of children, and returns the oldChild node. 
> If newChild is a DocumentFragment object, oldChild is replaced by all of the DocumentFragment children, which are inserted in the same order. If the newChild is already in the tree, it is first removed.
> 
> So, my proposal is to change the method to the next:
> ```java
> void replaceWith(XmlNode other) {
>    Node replacement = this.dom.getOwnerDocument().importNode(other.dom, true);
>    this.dom.getParentNode().replaceChild(replacement, this.dom);
> }
> ```

tonygermano:
fixes #850

Patch originally proposed by Alberto Saez Torres. This fix has been in production for ~11 years in the Nextgen (formerly mirth) Connect integration engine. User written code relies heavily on e4x.

https://github.com/nextgenhealthcare/connect/blob/a8e3768f63660c7a3e4301236e43dadb635ab006/server/src/org/mozilla/javascript/xmlimpl/XmlNode.java#L541..L557

### DIFF
--- a/testsrc/jstests/gh-850-XMLList-assign-from-same-list.jstest
+++ b/testsrc/jstests/gh-850-XMLList-assign-from-same-list.jstest
@@ -1,0 +1,22 @@
+function assertEquals(expected, actual) {
+   if (expected != actual) {
+     throw "Expected '" + expected + "' but was '" + actual + "'";
+   }
+ }
+ var xml =<xml>
+   <group>Group One</group>
+   <group>Group Two</group>
+   <group>Group Three</group>
+</xml>;
+
+var expected = <xml>
+   <group>Group Three</group>
+   <group>Group Two</group>
+   <group>Group Three</group>
+</xml>
+
+xml.group[0]=xml.group[2];
+
+assertEquals(xml, expected)
+
+"success";
--- a/xmlimplsrc/org/mozilla/javascript/xmlimpl/XmlNode.java
+++ b/xmlimplsrc/org/mozilla/javascript/xmlimpl/XmlNode.java
@@ -539,10 +539,7 @@ class XmlNode implements Serializable {
     }
 
     void replaceWith(XmlNode other) {
-        Node replacement = other.dom;
-        if (replacement.getOwnerDocument() != this.dom.getOwnerDocument()) {
-            replacement = this.dom.getOwnerDocument().importNode(replacement, true);
-        }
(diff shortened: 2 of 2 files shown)


gbrail:
Based on your experience and testing this seems like a fine thing to me. Thanks!

PR #856, opened 2021-04-04
more language/expressions test added

rbri:


rbri:
@gbrail  - knock knock :-)

gbrail:
Sorry -- I was just testing it locally when you pinged -- really!

gbrail:
For future Rhino users -- people with Rhino already checked out will need to "git submodule update" before the tests will run after this change goes in.

PR #857, opened 2021-04-06
Update JavaAdapter.java

mulingLHY:
java interfaces may extend other interfaces.Therefore, you cannot obtain the methods of upper-layer interfaces through getDeclaredmethods.

### DIFF
--- a/src/org/mozilla/javascript/JavaAdapter.java
+++ b/src/org/mozilla/javascript/JavaAdapter.java
@@ -501,6 +501,7 @@ public final class JavaAdapter implements IdFunctionCall
             ArrayList<Method> list, HashSet<String> skip)
     {
         Method[] methods = c.getDeclaredMethods();
+        if(c.isInterface())methods=c.getMethods();
         for (int i = 0; i < methods.length; i++) {
             String methodKey = methods[i].getName() +
                 getMethodSignature(methods[i],


tuchida:
Could you write a unit test that shows what you are able to do?

This seems to be the test for `JavaAdapter`.
https://github.com/mozilla/rhino/blob/db5ce1665d06a3afb9cac9d556811fb6fc1fdc05/testsrc/doctests/javaadapter.doctest
Or you may want to add the file here.
https://github.com/mozilla/rhino/tree/db5ce1665d06a3afb9cac9d556811fb6fc1fdc05/testsrc/org/mozilla/javascript/tests

src/org/mozilla/javascript/JavaAdapter.java
```diff
@@ -501,6 +501,7 @@ private static void appendOverridableMethods(Class<?> c,
             ArrayList<Method> list, HashSet<String> skip)
     {
         Method[] methods = c.getDeclaredMethods();
+        if(c.isInterface())methods=c.getMethods();
```
tuchida, line 504: In this case, `c.getDeclaredMethods()` is unnecessary.
This would be better.
https://github.com/mozilla/rhino/blob/8442718d94434e4e1c054b6228d0d4b1603b2878/src/org/mozilla/javascript/JavaAdapter.java#L491
```java
appendOverridableMethods(c.getDeclaredMethods(), list, skip); 
```
https://github.com/mozilla/rhino/blob/8442718d94434e4e1c054b6228d0d4b1603b2878/src/org/mozilla/javascript/JavaAdapter.java#L495
```java
appendOverridableMethods(c.getMethods(), list, skip); 
```

mulingLHY:
> Could you write a unit test that shows what you are able to do?
> 
> This seems to be the test for `JavaAdapter`.
> https://github.com/mozilla/rhino/blob/db5ce1665d06a3afb9cac9d556811fb6fc1fdc05/testsrc/doctests/javaadapter.doctest
> Or you may want to add the file here.
> https://github.com/mozilla/rhino/tree/db5ce1665d06a3afb9cac9d556811fb6fc1fdc05/testsrc/org/mozilla/javascript/tests

Supposing that  the program has a java abstract class as follows and I want to extend class A and override the  method methodInInterfaceC in interface C, the current method `getOverridableMethods` cannot obtain such method. Considering that the methods in the interface are all public, I It can be solved by changing `getDeclaredMethods` to `getMethods` for interface.
```
interface C{
  public void methodInterfaceC()
//...
}

interface B extends C{
//...
}

Class A implements B{
//...
}
```
I found this probem when I was about to extend `android.widget.BaseAdapter` and override methods `getCount,getView...` which is declared in interfce `android .widget.Adapter` 
I am an amateur from China and my English is not very good. If the expression is inaccurate or inappropriate, please forgive me.

tuchida:
Thanks.
Perhaps you may want to create `JavaAdapterTest.java`. This will be helpful for you.
https://github.com/mozilla/rhino/blob/8442718d94434e4e1c054b6228d0d4b1603b2878/testsrc/org/mozilla/javascript/tests/InterfaceAdapterTest.java

Can you write test?

> I am an amateur from China and my English is not very good. If the expression is inaccurate or inappropriate, please forgive me.

I don't understand English either, so I use a translation tool.

mulingLHY:
> Can you write test?

I still can’t use github proficiently, so I don’t know whether you mean this.
https://github.com/mulingLHY/rhino/blob/master/testsrc/org/mozilla/javascript/tests/TestJavaAdapter.java

tuchida:
It looks good.
I think you commit to the wrong branch. Try to commit to `patch-1` instead of `master`.

tuchida:
Good! Please wait a moment for the owner to confirm.

src/org/mozilla/javascript/JavaAdapter.java
```diff
@@ -500,7 +500,10 @@ private static ObjToIntMap getObjectFunctionNames(Scriptable obj)
     private static void appendOverridableMethods(Class<?> c,
             ArrayList<Method> list, HashSet<String> skip)
     {
-        Method[] methods = c.getDeclaredMethods();
+        Method[] methods = null;
+        if(c.isInterface())methods=c.getMethods();
+        else methods=c.getDeclaredMethods();
```
gbrail, line 505: Sadly we are not at the point in this 25-year-old project to run an automatic formatting tool on new PRs, but...

Most of the other Rhino code puts spaces around equals and brackets around all if..then..else clauses. It looks like other code in this file does the same, so could you please update to either:

    if (c.isInterface()) {
      methods = c.getMethods();
    } else {
      methods = c.getDeclaredMethods();
    }

or...

`Method[] methods = c.isInterface() ? c.getMethods() : c.getDeclaredMethods;`

testsrc/org/mozilla/javascript/tests/TestJavaAdapter.java
```diff
@@ -0,0 +1,53 @@
...
+		String testCode =
+	            "JavaAdapter(Packages."+A.class.getName()+",{methodInC:function(){},methodInB:function(){}},null)";
+		
+		NativeJavaObject aJavaObject=(NativeJavaObject)cx.evaluateString(topScope, testCode,"", 1, null);
+		
+		
+		Method overrideMethod=null;
+		try {
+			overrideMethod=aJavaObject.unwrap().getClass().getDeclaredMethod("methodInC", String.class);
+		}catch(NoSuchMethodException e) {
```
gbrail, line 47: I'm not sure that I understand this test. Are you expecting a NoSuchMerthodException? If so then I think that you should assert inside the try...catch if the throw didn't happen. Or better yet, use JUnit's "assertThrows".

gbrail (changes requested):
Thanks for doing this! I have a few specific comments.

Also, could you update the PR description with a bit longer explanation of what changed, combined with a small example like the one in your test? That will help a lot in writing the release notes.

testsrc/org/mozilla/javascript/tests/PullRequest857Test.java
```diff
@@ -0,0 +1,53 @@
...
+	
+	interface C {
+		int methodInC(String str);
+	}
+	interface B extends C {
+	}
+	public abstract class A implements B{
+	}
+	@Test
+	public void testOverrideMethodInMultiLayerInterface() {
```
tuchida, line 34: If I do this, won't it compile without the catch?
```java
	public void testOverrideMethodInMultiLayerInterface() throws NoSuchMethodException {
```

gbrail: If you do that then it'll compile without the catch and the test will fail if it actually throws the exception. The way it is now, the test will still pass even if it throws the exception.

Unless the test is explicitly checking to see if the exception is thrown, then adding the exception to the throws clause is fine.

testsrc/org/mozilla/javascript/tests/PullRequest857Test.java
```diff
@@ -0,0 +1,44 @@
...
+
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+import org.mozilla.javascript.Context;
+import org.mozilla.javascript.NativeJavaObject;
+import org.mozilla.javascript.Scriptable;
+
+
+public class PullRequest857Test{
```
tuchida, line 11: I prefer the file name `JavaAdapterTest.java`. When we change the `JavaAdapter` in the future, it is easier to modify it if we know where the tests are already written.

gbrail:
OK, you have answered lots of our questions. Thanks!

PR #859, opened 2021-04-10
Update assert.js to handle e4x objects

tonygermano:
`assert.js` should allow comparisons of e4x objects for writing e4x related tests. The e4x standard uses the `==` operator for comparison of XML, XMLList, QName and Namespace objects.

`XML` and `XMLList` objects both return `xml` when used with the typeof operator.

`QName` and `Namespace` objects both return `object` when used with the typeof operator.

Notably, two `Namespace` objects with the same `uri` values, but different `prefix` values are still considered equal, so a deep object compare produces the wrong result.

### DIFF
--- a/testsrc/assert.js
+++ b/testsrc/assert.js
@@ -132,6 +132,7 @@ var assertUnoptimized;
       case "undefined":
       case "function":
         return String(value);
+      case "xml":
       case "object":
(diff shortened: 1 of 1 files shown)


gbrail:
Thanks!

PR #860, opened 2021-04-11
Allow JSON.stringify to operate on Java objects

### LINKED ISSUE #578: Stackoverflow with NativeJSON.stringify after update from 1.7.7 to 1.7.11
NativeJSON.stringify used to throw an exception when it experienced a problem. This way it was possible to identify the problem in the json string, usually an object that was a java string. Now (since 1.7.8) it goes into a Stackoverflow. This should be changed imho. To see how i struggeld to find out what was going on i left my analysis below.

We upgraded from 1.7.7 to version 1.7.11. After that update we now get a Stackoverflow for the following code that finally uses org.mozilla.javascript.NativeJSON.stringify() on the Java side. It seems that this occurs if you add a Java object to the json you want to stringify. The following example uses trim to generate a java string, you could however also use another object like Joda DateTime etc to provoke the stackoverflow.

Here's the client code
```
var Util =
{
    getModel : function(x)
    {
        var model = {
            x : x,
        }
        return model;
    },
}

//example for a java object in the json:
var x = new Packages.java.lang.String("Hello World").trim(); //-> Stackoverfow 
//var x = "Hello World"; -> works fine
var model = Util.getModel(x);

println(new Packages.org.acme.commons.lang.RhinoUtil.jsonStringify(model));
```

And here's the code for the RhinoUtil.jsonStringify()
```
    public static final String jsonStringify(NativeObject nativeObj)
    {
        ContextFactory cf = null;
        try
        {
            cf = new ContextFactory();
        }
        catch (NoClassDefFoundError nce)
        {
            throw nce;
        }

        Context cx = cf.enterContext();
        try 
        {   
            ScriptableObject sealedSharedScope = cx.initStandardObjects(null, true); //sealed object
            String loadMe = "RegExp; getClass; java; Packages; JavaAdapter;";
            cx.evaluateString(sealedSharedScope , loadMe, "lazyLoad", 0, null);
            sealedSharedScope.sealObject();

            return NativeJSON.stringify(cx, sealedSharedScope, nativeObj, null, null) + "";
        }
        finally
        {
            Context.exit();
        }
    }
```

If we analyse the stackoverflow exception we get the following stacktrace
```
java.lang.StackOverflowError
at org.mozilla.javascript.JavaMembers.getIds(JavaMembers.java:177)
at org.mozilla.javascript.NativeJavaObject.getIds(NativeJavaObject.java:188)
at org.mozilla.javascript.NativeJSON.jo(NativeJSON.java:353)
at org.mozilla.javascript.NativeJSON.str(NativeJSON.java:322)
at org.mozilla.javascript.NativeJSON.jo(NativeJSON.java:359)
at org.mozilla.javascript.NativeJSON.str(NativeJSON.java:322)
at org.mozilla.javascript.NativeJSON.jo(NativeJSON.java:359)
at org.mozilla.javascript.NativeJSON.str(NativeJSON.java:322)
...
```

Please note that we need a way to stringify on the java side, therefore js JSON.stringify() is not an option. Please help.

Tested same code with 1.7.7 and it seems that it never worked, however the 1.7.7 error message was more clear 
```
Java class "[Ljava.lang.reflect.Constructor;" has no public instance field or method named "toJSON"
```

Note that i also changed my jdk from oracle do adopt open jdk 1.8. Will research further...

Ok, found out that our internal util already checks for this and has a fallback
```
NativeObject nativeObj = (NativeObject) object;
try
{
	//this will fail if the json contains a Java String, therefore we will use a fallback
	return RhinoUtil.jsonStringify(nativeObj);
}
catch (Exception e)
{
        // fallback to old style iteration with nativeObject.getIds()
	return JSONStringifier.jsonStringify(nativeObj);
}
```

I now changed the catch from Exception to Throwable so that it also gets the Stackoverflow which is Throwable, not Exception. Now everything works as before. So actually for me the problme is solved and we can continue to use 1.7.11. Sorry for that elaborate story. But it shows that it would be useful to avoid the Stackoverflow and change it to the previous exception that was easier to understand.

tonygermano:
- Boxed Primitives and Strings are unwrapped and converted to
  javascript primitives.
- Collections and Arrays will be converted to javascript arrays.
- Maps will be converted to javascript objects, but any non-string-
  like keys will be ignored.
- Other java classes will throw a TypeError.

Conversions occur after the replacer function is called if one is
provided, so custom handling is still possible.

Currently a StackOverflowError occurs when JSON.stringify encounters
a java object without a toJSON method.

fixes #578

### DIFF
--- a/src/org/mozilla/javascript/NativeJSON.java
+++ b/src/org/mozilla/javascript/NativeJSON.java
@@ -11,6 +11,7 @@ import java.util.Collection;
 import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
+import java.util.Map;
 import java.util.Stack;
 
 import org.mozilla.javascript.json.JsonParser;
@@ -287,13 +288,33 @@ public final class NativeJSON extends IdScriptableObject
                                         new Object[] { key, value });
         }
 
-
         if (value instanceof NativeNumber) {
             value = Double.valueOf(ScriptRuntime.toNumber(value));
         } else if (value instanceof NativeString) {
             value = ScriptRuntime.toString(value);
         } else if (value instanceof NativeBoolean) {
             value = ((NativeBoolean) value).getDefaultValue(ScriptRuntime.BooleanClass);
+        } else if (value instanceof NativeJavaObject) {
+            value = ((NativeJavaObject) value).unwrap();
+            if (value instanceof Map) {
+                Map<?,?> map = (Map<?,?>) value;
+                NativeObject nObj = new NativeObject();
+                map.forEach((k, v) -> {
+                    if (k instanceof CharSequence) {
+                        nObj.put(((CharSequence) k).toString(), nObj, v);
+                    }
+                });
+                value = nObj;
+            }
+            else {
+                if (value instanceof Collection<?>) {
+                    Collection<?> col = (Collection<?>) value;
+                    value = col.toArray(new Object[col.size()]);
+                }
+                if (value instanceof Object[]) {
+                    value = new NativeArray((Object[]) value);
+                }
+            }
         }
 
         if (value == null) return "null";
@@ -314,11 +335,15 @@ public final class NativeJSON extends IdScriptableObject
             return "null";
         }
 
-        if (value instanceof Scriptable && !(value instanceof Callable)) {
-            if (value instanceof NativeArray) {
-                return ja((NativeArray) value, state);
+        if (value instanceof Scriptable) {
+            if (!(value instanceof Callable)) {
+                if (value instanceof NativeArray) {
+                    return ja((NativeArray) value, state);
+                }
+                return jo((Scriptable) value, state);
             }
-            return jo((Scriptable) value, state);
+        } else if (!Undefined.isUndefined(value)) {
+            throw ScriptRuntime.typeErrorById("msg.json.cant.serialize", value.getClass().getName());
         }
 
         return Undefined.instance;
--- a/src/org/mozilla/javascript/resources/Messages.properties
+++ b/src/org/mozilla/javascript/resources/Messages.properties
@@ -254,6 +254,10 @@ msg.invalid.date =\
 msg.toisostring.must.return.primitive =\
     toISOString must return a primitive value, but instead returned "{0}"
 
+# NativeJSON
+msg.json.cant.serialize =\
+    Do not know how to serialize a {0}
+
 # Parser
 msg.got.syntax.errors = \
     Compilation produced {0} syntax errors.
--- a/testsrc/jstests/stringify-java-objects.js
+++ b/testsrc/jstests/stringify-java-objects.js
@@ -1,0 +1,147 @@
+load('testsrc/assert.js');
+
+function replacer(key, value) {
+    var javatype = value instanceof java.lang.Object ? value.getClass().name : null;
+    return javatype ? 'replaced: ' + javatype : value;
+}
+
+// java.lang.String
+var javaString = new java.lang.String('test');
+
+var obj = {test: javaString};
+var expected = JSON.stringify({test: 'replaced: java.lang.String'});
+var actual = JSON.stringify(obj, replacer);
+assertEquals(expected, actual);
+
+var obj = {test: javaString};
+var expected = JSON.stringify({test: 'test'});
+var actual = JSON.stringify(obj);
+assertEquals(expected, actual);
+
+// java.lang.Double
+var javaDouble = java.lang.Double.valueOf(12.34);
+
+var obj = {test: javaDouble};
+var expected = JSON.stringify({test: 'replaced: java.lang.Double'});
+var actual = JSON.stringify(obj, replacer);
+assertEquals(expected, actual);
+
+var obj = {test: javaDouble};
+var expected = JSON.stringify({test: 12.34});
+var actual = JSON.stringify(obj);
+assertEquals(expected, actual);
+
+// java.lang.Boolean
+var javaBoolean = java.lang.Boolean.valueOf(false);
+
+var obj = {test: javaBoolean};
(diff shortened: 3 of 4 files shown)


src/org/mozilla/javascript/NativeJSON.java
```diff
@@ -314,11 +335,15 @@ private static Object str(Object key, Scriptable holder,
...
-                return ja((NativeArray) value, state);
+        if (value instanceof Scriptable) {
+            if (!(value instanceof Callable)) {
+                if (value instanceof NativeArray) {
+                    return ja((NativeArray) value, state);
+                }
+                return jo((Scriptable) value, state);
             }
-            return jo((Scriptable) value, state);
+        } else if (!Undefined.isUndefined(value)) {
```
gbrail, line 345: I'm worried that throwing this exception will break backward compatibility. Previously, someone with a complex object who is trying to stringify it might end up with a weird result, and now they will get a less weird result. But if there's a chance that people who were previously getting weird results will now get an exception, then we could easily break a lot of code. Unless the ECMA spec says that we should throw an exception in this case, then do we really need it?

gbrail (commented):
I'm worried that the new exception will break a lot of existing code. Can we do this without throwing an exception in a new place?

src/org/mozilla/javascript/NativeJSON.java
tonygermano: The current code sees it as a NativeJavaObject, which is Scriptable, and so it sends it down the path to try to stringify it as a javascript object, because it is not Callable. This is throwing the StackOverflowError mentioned in the linked issue. I believe the ECMA spec says that all Objects that are not Functions should throw a TypeError if they are cyclic or return a JSON string representing an object.

Since I am unwrapping the NativeJavaObject, it is no longer Scriptable. I think at this point it either has to be Undefined or some other Java Object that can't be serialized, so the options are to either throw a more descriptive exception than it had been throwing, return Undefined, or return "`{}`".

I believe prior to https://github.com/mozilla/rhino/commit/4c76ef2b816350a1bc9a553fa5860669e0abfb04 it never made it this far and threw the other exception described in the linked issue:
```
Java class "[Ljava.lang.reflect.Constructor;" has no public instance field or method named "toJSON"
```
That commit let the object make it through to the replacer, but created a worse exception if a replacer did not exist to do something with the object.

I think my TypeError is in line with the historical behavior before the StackOverflowError, but I am open to suggestions. This also mimics the behavior of BigInt, which throws a TypeError when you try to stringify it without a replacer or adding a custom BigInt.prototype.toJSON method.

gbrail:
OK -- the fact that it used to fail in worse ways convinces me. This looks good. Thanks!

tuchida:
This is a minor issue, though.

`java.lang.Long` and `java.math.BigDecimal` inherit from `java.lang.Number`, so they seem to output `doubleValue` values.
```
js> java.lang.Long.MAX_VALUE
9223372036854776000
js> java.lang.Long.valueOf(java.lang.Long.MAX_VALUE)
9223372036854775807

js> JSON.stringify(java.lang.Long.MAX_VALUE)
9223372036854776000
js> JSON.stringify(java.lang.Long.valueOf(java.lang.Long.MAX_VALUE))
9223372036854776000
```
The JSON specification seems to recommend that you do not exceed the range of double.
https://tools.ietf.org/html/rfc8259#section-6

tonygermano:
I'm not sure if it's an issue. Anywhere else those types are used as numbers they are converted to their double value as well, e.g.,
```js
js> java.lang.Long.MAX_VALUE === java.lang.Long.MAX_VALUE + 1
true
```

I believe the recommendation in the spec is to not (as a JSON user) attempt to stringify numbers representing integer values greater than Number.MAX_SAFE_INTEGER for maximum interoperability. A java.lang.Long doesn't exceed the range of Double, but it does exceed the integer range of Double in which there is no loss of precision.

As a JSON user, if you expect to have numbers that are integers which exceed Number.MAX_SAFE_INTEGER and they must be parsed back to their exact value, they should be represented in JSON as strings instead.

This was in the Chrome console. JSON itself can definitely handle integers that large, but javascript can't.
```js
> JSON.parse("9223372036854775807")
9223372036854776000
```
In the case of storing a java.lang.Long without any loss of precision, I think that would mean doing something like:
```js
js> l1 = java.lang.Long.valueOf(java.lang.Long.MAX_VALUE)
9223372036854775807
js> json = JSON.stringify(l1.toString())
"9223372036854775807"
js> l2 = java.lang.Long.valueOf(JSON.parse(json))
9223372036854775807
js> l1.equals(l2)
true
```
I would put the responsibility on the person calling JSON.stringify to make sure that any numbers are in a reasonable range.

tonygermano:
For what it's worth, my most common use case is expecting this to work, and it previously did not:
```js
js> var hm = new java.util.HashMap()
js> hm.put('a', 1)
null
js> hm.put('b', 'two')
null
js> hm.put('c', false)
null
js> JSON.stringify({a: hm.get('a'), b: hm.get('b'), c: hm.get('c')})
{"a":1,"b":"two","c":false}
```

tuchida:
Thank you. I think it's a good idea too.
As you are probably seeing, if #837 is merged, `java.math.BigInteger` will error out. ref. https://github.com/mozilla/rhino/pull/837/commits/a44b33bf8be8718ca033a3dcb3cc6a6e7078aa8f

tonygermano:
@tuchida Yes, that is a good point. I think that will be consistent with other BigInt related changes and java.math.BigInteger.

p-bakker:
@tonygermano any chance you could determine if this PR makes #12 obsolete?

p-bakker:
@tonygermano any chance you could see if PR #12 is now obsolete due to merging this PR and thus could be closed?

PR #864, opened 2021-04-13
Make Context Closeable

gbrail:
This makes it possible to write:

try (Context cx = Context.enter()) { ... }

instead of using try...finally.

### DIFF
--- a/src/org/mozilla/javascript/Context.java
+++ b/src/org/mozilla/javascript/Context.java
@@ -10,6 +10,7 @@ package org.mozilla.javascript;
 
 import java.beans.PropertyChangeEvent;
 import java.beans.PropertyChangeListener;
+import java.io.Closeable;
 import java.io.IOException;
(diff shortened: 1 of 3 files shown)


rbri (commented):
Should we make exit deprecated?
Maybe we can make close() the default implementation and exit() the forwarder method.

gbrail:
Yes -- we could encourage people to call "close" instead of exit if they aren't using a try...catch (for instance, in lots of tests we create the context in a @Before method and exit in @After. But my gut feeling is that making it deprecated will just mean that we get deprecation warnings for the next 20 years of Rhino development and it might not be worth the effort unless we want to actually go in and remove the code some day in the future.

rbri:
not sure what we all do in 20 years but i can live without the deprecated warning (hopefully for the next 20 years)

gbrail:
Thanks for taking a look at this!

PR #866, opened 2021-04-15
Introduce automated code formatting

gbrail:
This PR uses google-java-format, via the "spotless" plugin, to format
the Java code.

It uses the "ratchet" mode of Spotless. That means that all Java
code committed AFTER this commit will be required to be formatted
using google-java-format. This will let us reformat the codebase
gradually rather than in one huge chunk.

We are keeping "checkstyle" as it checks for a few other things like
"star imports" that we'd still like to avoid.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -7,6 +7,7 @@ plugins {
     id 'jacoco'
     id 'distribution'
     id 'checkstyle'
+    id 'com.diffplug.spotless' version "5.12.1"
     id 'com.github.spotbugs' version "4.6.2"
 }
 
@@ -379,6 +380,13 @@ spotbugs {
     effort = "less"
     reportLevel = "medium"
     excludeFilter = file("./spotbugs-exclude.xml")
+}
+
+spotless {
+    ratchetFrom 'code-formatting-required'
+    java {
+        googleJavaFormat().aosp()
+    }
 }
 
 jacocoTestReport.dependsOn test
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -381,6 +382,13 @@ spotbugs {
     excludeFilter = file("./spotbugs-exclude.xml")
 }
 
+spotless {
+    ratchetFrom 'code-formatting-required'
+    java {
+        googleJavaFormat().aosp()
```
tuchida, line 388: Is the coding style ok with Android Open Source Project (AOSP) style?
https://source.android.com/setup/contribute/code-style

You can also change to the Google Java Style Guide.
https://google.github.io/styleguide/javaguide.html

I chose AOSP because the style over here had 4 spaces indentations.
This way, you can also change the indentation in the Google Java Style Guide. However, it failed to change some of the files.
```groovy
spotless {
    ratchetFrom 'code-formatting-required'
    java {
        googleJavaFormat()
        indentWithTabs(2)
        indentWithSpaces(4)
    }
}
```

gbrail:
I don't think that those other options to spotless are actually passed to google-java-format -- I think that they do something else. 

I picked AOSP because it looks like most of the old Rhino code uses four spaces for indents instead of two. (I would prefer two if this code was new.) 

I search quickly through the source for google-java-format and the only difference I could see for AOSP was the way that it sorts import statements -- I think that everything else is the same.

tuchida:
OK. Thank you.

PR #868, opened 2021-04-16
fix nested java objects in JSON.stringify

tonygermano:
I realized that my previous PR #860 did not allow for nested Maps and Lists. Can someone please review to see that I am now creating objects and arrays the correct way? I have a good deal of experience working with java objects in javascript, but very little in the other direction. The new tests are passing.

### DIFF
--- a/src/org/mozilla/javascript/NativeJSON.java
+++ b/src/org/mozilla/javascript/NativeJSON.java
@@ -7,7 +7,9 @@
 package org.mozilla.javascript;
 
 import java.util.Arrays;
+import java.util.Calendar;
 import java.util.Collection;
+import java.util.Date;
 import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
@@ -201,7 +203,7 @@ public final class NativeJSON extends IdScriptableObject
             this.propertyList = propertyList;
         }
 
-        Stack<Scriptable> stack = new Stack<Scriptable>();
+        Stack<Object> stack = new Stack<Object>();
         String indent;
         String gap;
         Callable replacer;
@@ -269,6 +271,8 @@ public final class NativeJSON extends IdScriptableObject
                               StringifyState state)
     {
         Object value = null;
+        Object wrappedJavaValue = null;
+        
         if (key instanceof String) {
             value = getProperty(holder, (String) key);
         } else {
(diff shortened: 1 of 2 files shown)


src/org/mozilla/javascript/NativeJSON.java
```diff
@@ -298,10 +298,10 @@ private static Object str(Object key, Scriptable holder,
             value = ((NativeJavaObject) value).unwrap();
             if (value instanceof Map) {
                 Map<?,?> map = (Map<?,?>) value;
-                NativeObject nObj = new NativeObject();
+                Scriptable nObj = state.cx.newObject(state.scope);
                 map.forEach((k, v) -> {
                     if (k instanceof CharSequence) {
-                        nObj.put(((CharSequence) k).toString(), nObj, v);
+                        nObj.put(((CharSequence) k).toString(), nObj, state.cx.getWrapFactory().wrap(state.cx, state.scope, v, v.getClass()));
```
tonygermano, line 304: Should I be using `ScriptRuntime.toObject` instead of calling wrap directly? Or something else?

tonygermano:
I had overlooked until now PR https://github.com/mozilla/rhino/pull/824, which looks very similar in function to #860 , and also appears to cover the shortcoming I mentioned above. I will take a closer look at it, and see how the two approaches might be incorporated, and possibly roll back some of my changes.

I think for the time being this PR is still good as it corrects an issue in the current master branch.

src/org/mozilla/javascript/NativeJSON.java
tonygermano: It appears that Context.javaToJS does the right thing more than any of the alternatives I found. I didn't originally choose it because it's only available as a static method, and I already had a Context object, so I didn't think it needed to go through the process of finding the current Context.

gbrail:
This looks fine. Since @rPraml (who I can't seem to tag here) wrote the original PR I wonder if he can verify whether it fixes those issues?

Otherwise, since I merged your other PR for this file, can you please resolve the conflicts and merge that one?

tonygermano:
Putting this one into draft for now. I'm still considering some other changes after the discussion on his PR.

tonygermano:
Closing this PR in favor of #875

PR #869, opened 2021-04-20
Allow JSON.stringify to operate on e4x XML values.

tonygermano:
Will simply return the toString() value of the XML object. User still has the ability to handle the object in the replacer function prior to converting to a string.

This has historically thrown various TypeErrors.


tonygermano:
@gbrail How can I see what failed in the checks? When I ran `.gradlew test` locally on the branch everything passed.

gbrail:
Looks like you ran "./gradlew check" and that helped you fix the issues!

Not sure what to suggest about CircleCI -- it's what the Mozilla people said to use, and I find it pretty trouble-free when I log in using my GitHub account.

This change makes sense to me and I'll merge it shortly.

PR #853, opened 2021-04-03
Shorthand property names

tuchida:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer#new_notations_in_ecmascript_2015

We can now do this.
```js
var a = 123;
({a});
var fn = ({ a }) => print(a);
```

This has always been possible.
```js
var { a } = { a: 123 };
a; // 123

function fn({ a }) { print(a); }
fn({ a: 123 }); // 123
```

PR #872, opened 2021-04-24
Number separator support

rbri:
spend some time to bring the numeric separator support to rhino (https://github.com/tc39/proposal-numeric-separator)

* 600 more tests in the Test262 test suite (no notable runtime degradation at my local machine)
* i have not added an es6 switch because i think this will not break any old code
* the code is a bit strange and chatty but the idea was to do as less checks as possible

feedback welcome

### DIFF
--- a/src/org/mozilla/javascript/TokenStream.java
+++ b/src/org/mozilla/javascript/TokenStream.java
@@ -1211,10 +1211,7 @@ class TokenStream
      * and ignore the number separator if there is one.
      */
     private int readDigits(int base, int c) throws IOException {
-        if ((base == 10 && isDigit(c))
-                || (base == 16 && isHexDigit(c))
-                || (base == 8 && isOctalDigit(c))
-                || (base == 2 && isDualDigit(c))) {
+        if (isValidDigit(base, c)) {
             addToString(c);
 
             c = getChar();
@@ -1236,19 +1233,13 @@ class TokenStream
                         return REPORT_NUMBER_FORMAT_ERROR;
                     }
 
-                    if (!((base == 10 && isDigit(c))
-                            || (base == 16 && isHexDigit(c))
-                            || (base == 8 && isOctalDigit(c))
-                            || (base == 2 && isDualDigit(c)))) {
+                    if (!isValidDigit(base, c)) {
                         // bad luck we have to roll back
                         ungetChar(c);
                         return NUMERIC_SEPARATOR;
                     }
                     addToString(NUMERIC_SEPARATOR);
-                } else if ((base == 10 && isDigit(c))
-                                || (base == 16 && isHexDigit(c))
-                                || (base == 8 && isOctalDigit(c))
-                                || (base == 2 && isDualDigit(c))) {
+                } else if (isValidDigit(base, c)) {
                     addToString(c);
                     c = getChar();
                     if (c == EOF_CHAR) {
@@ -1260,6 +1251,13 @@ class TokenStream
             }
         }
         return c;
+    }
+
+    private boolean isValidDigit(int base, int c) throws IOException {
+        return (base == 10 && isDigit(c))
+                || (base == 16 && isHexDigit(c))
+                || (base == 8 && isOctalDigit(c))
+                || (base == 2 && isDualDigit(c));
     }
 
     private static boolean isAlpha(int c)
--- a/src/org/mozilla/javascript/TokenStream.java
+++ b/src/org/mozilla/javascript/TokenStream.java
@@ -28,8 +28,13 @@ class TokenStream
      * to check.  (And checking EOF by exception is annoying.)
      * Note distinction from EOF token type!
      */
-    private final static int
-        EOF_CHAR = -1;
+    private final static int EOF_CHAR = -1;
+
+    /*
+     * Return value for readDigits() to signal the caller has
+     * to return an number format problem.
+     */
+    private final static int REPORT_NUMBER_FORMAT_ERROR = -2;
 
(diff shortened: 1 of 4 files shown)


tuchida:
(This is difficult to deal with right now), the following is a violation of the specification.
```js
// Rhino
(() => 1_2_3).toString() // () => 123
```
```js
// SpiderMonkey
(() => 1_2_3).toString() // () => 1_2_3
```
Now the AST is reverted to the source code by the `Decompiler`, but I think it would be better to return the source code directly as `Function#toString`.

rbri:
> ```js
> (() => 1_2_3).toString() // () => 1_2_3
> ```

I think this is a separate issue.

tuchida:
Sorry. It's about this.
https://github.com/mozilla/rhino/blob/15c4d52019bab829f92de517cdbf45eae6291fc8/src/org/mozilla/javascript/Parser.java#L3827-L3835

Here, parsed values are reverting, but halfway.
```js
(() => 0xffff).toString() // () => 65535
```

rbri:
ok will add a commit that preserves the parsed string 

Am 25. April 2021 13:13:19 MESZ schrieb tuchida ***@***.***>:
>Sorry. It's about this.
>https://github.com/mozilla/rhino/blob/15c4d52019bab829f92de517cdbf45eae6291fc8/src/org/mozilla/javascript/Parser.java#L3827-L3835
>
>Here, parsed values are reverting, but halfway.
>```js
>(() => 0xffff).toString() // () => 65535
>```
>
>-- 
>You are receiving this because you authored the thread.
>https://github.com/mozilla/rhino/pull/872#issuecomment-826306442

-- 
Diese Nachricht wurde von meinem Android-Gerät mit K-9 Mail gesendet.

rbri:
@tuchida do you have an idea about testing the NumberLiteral string part?

tuchida:
I've added tests here before.
https://github.com/tuchida/rhino/commit/a243df983b6231ab563016cc119fa42a3e988fc7

Sorry after the correction, but I don't know what value AST is expecting. Also, the value of `Function#toString` does not change after modifying here.

rbri:
I think we have too many open PR's at the moment...

Suggestion:
* let us open a new issue for the toString() stuff
* i have no time this week to work any further on this but maybe next week
* hopefully @gbrail will merge this - i see no test that brakes because of this

tuchida:
> let us open a new issue for the toString() stuff

I agree.
There are probably some pretty big problems with `IRFactory` and `Decompiler`, and if we are going to be ECMA262 compliant in the future, we will need to make some big fixes. ref. https://github.com/mozilla/rhino/pull/853#issuecomment-816597763

src/org/mozilla/javascript/TokenStream.java
```diff
@@ -1185,6 +1206,62 @@ final int getToken() throws IOException
...
 
+    /*
+     * Helper to read the next digits according to the base
+     * and ignore the number separator if there is one.
+     */
+    private int readDigits(int base, int c) throws IOException {
+        if ((base == 10 && isDigit(c))
+                || (base == 16 && isHexDigit(c))
+                || (base == 8 && isOctalDigit(c))
+                || (base == 2 && isDualDigit(c))) {
```
gbrail, line 1217: Minor, but I see that we have repeated this four-line conditional three times in this code. Would you consider DRYing this and moving it into a separate private function?

src/org/mozilla/javascript/TokenStream.java
```diff
@@ -1185,6 +1206,62 @@ final int getToken() throws IOException
...
+            addToString(c);
+
+            c = getChar();
+            if (c == EOF_CHAR) {
+                return EOF_CHAR;
```
gbrail, line 1222: FWIW according to "./gradlew jacocoTestReport" this is never hit in this function -- everything else is.

gbrail (commented):
Thanks! I attached one minor note if you wouldn't mind taking a look.

Also, with the latest master this will require a "./gradlew spotlessApply" to reformat it. However, I don't mind doing that since I like to separate out reformatting commits from more substantial ones.

src/org/mozilla/javascript/TokenStream.java
rbri: I think this is because there is no test for this case - last char of a script has to be a digit. I dont like to add test to the 262 suite - should we write a separate test for this?

src/org/mozilla/javascript/TokenStream.java
rbri: done

src/org/mozilla/javascript/TokenStream.java
rbri: Ok, just writing a test, give me 10 min.

rbri: done

gbrail:
That is great -- I'm going to pull this into a set of commits in a minute, so hold on. Thanks!

gbrail:
I merged this manually. Thanks for doing it!

src/org/mozilla/javascript/TokenStream.java
```diff
@@ -1194,11 +1259,33 @@ private static boolean isAlpha(int c)
         return 'a' <= c && c <= 'z';
     }
 
-    static boolean isDigit(int c)
+    private boolean isDigit(int base, int c) throws IOException {
```
tuchida, line 1262: (It is already merged, but) I don't think `throws IOException` is necessary here.
Also, I think static method is better, just like `isDigit`.

Issue #870, opened 2021-04-20
JPMS modularization


carlosame:
When this project is used in JPMS modular builds, it causes either warnings or errors because it is not declaring a module name. For example, a Gradle build of a project that depends on rhino is producing the following:
```
> Could not resolve all files for configuration ':project:testRuntimeClasspath'.
   > Failed to transform rhino-1.7.13.jar (org.mozilla:rhino:1.7.13) to match attributes {artifactType=jar, javaModule=true, org.gradle.category=library, org.gradle.libraryelements=jar, org.gradle.status=release, org.gradle.usage=java-runtime}.
      > Execution failed for ExtraModuleInfoTransform: C:\***\.m2\repository\org\mozilla\rhino\1.7.13\rhino-1.7.13.jar.
         > Not a module and no mapping defined: rhino-1.7.13.jar
```
I'm working around this with the `extra-java-module-info` Gradle plugin, but it would be handier if this project provided a module name.

One possibility would be to declare an automatic module name for the main `rhino` jar, see the following blog post for context:

http://branchandbound.net/blog/java/2017/12/automatic-module-name/

The idea would be to add the following to `build.gradle`, in the section for the jar manifest attributes:
```gradle
"Automatic-Module-Name": "org.mozilla.rhino",
```
it could go below the `Bundle-*` field declarations (_i.e._ near the line 192 at this moment).

However this project is also providing other jar files, so instead of the automatic name a full [JPMS modularization](https://www.oracle.com/corporate/features/understanding-java-9-modules.html) could be performed. That would imply to split the code in a way that two or three modules (with their own `module-info` files) can be produced, _e.g._ `org.mozilla.rhino.runtime`, `org.mozilla.rhino.engine` and `org.mozilla.rhino` (although one of the first two could be skipped, leaving only two modules).

If you are curious about what the other projects are doing on JPMS modularization, you may want to check the following blog post:

https://blog.frankel.ch/update-state-java-modularization/

Any thoughts?

gbrail:
It seems like the simplest thing and one that would solve most problems would be to put the "automatic module name" into build.gradle for the main jar, and again for the "engine JAR."

The runtime jar is a funny case because I'd love to split the code here, but I also know that changing rhino.jar to not include everything will break many many things, so it might make sense to leave the module name out of that (or also make it the same as the main JAR).

What do you think, and what would you think of submitting a PR for this?

carlosame:
[...]

PR #873, opened 2021-04-24
Set automatic module names to rhino.jar and rhino-engine.jar

carlosame:
The name assigned to `rhino.jar` is the same as in the OSGi metadata (`org.mozilla.rhino`).

Closes issue #870.


gbrail:
LGTM. Thanks for doing this!

PR #876, opened 2021-04-30
fix for stringify array replacer mozilla/rhino#725

### LINKED ISSUE #725: An issue about JSON.stringify

tonygermano:
- Per spec, Numbers in an array replacer should be converted to strings
  prior to adding to the property list
- Duplicates should be removed during construction of the property list
- Changed some variables declared as List  to Collection instead because
  that is sufficient for how they are being used
- Moved LinkedList to array conversion from happening in every iteration of
  the jo method to happening once in the stringify method
- Convert strings items representing integers to their Integer value for
  correct property lookup

fixes #725

### DIFF
--- a/src/org/mozilla/javascript/NativeJSON.java
+++ b/src/org/mozilla/javascript/NativeJSON.java
@@ -9,8 +9,8 @@ package org.mozilla.javascript;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Iterator;
+import java.util.LinkedHashSet;
 import java.util.LinkedList;
-import java.util.List;
 import java.util.Map;
 import java.util.Stack;
 import org.mozilla.javascript.json.JsonParser;
@@ -198,7 +198,7 @@ public final class NativeJSON extends IdScriptableObject {
                 String indent,
                 String gap,
                 Callable replacer,
-                List<Object> propertyList) {
+                Object[] propertyList) {
             this.cx = cx;
             this.scope = scope;
 
@@ -212,7 +212,7 @@ public final class NativeJSON extends IdScriptableObject {
         String indent;
         String gap;
         Callable replacer;
-        List<Object> propertyList;
+        Object[] propertyList;
 
         Context cx;
         Scriptable scope;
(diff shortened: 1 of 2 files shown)


tonygermano:
Relevant tests which both pass are:
- https://github.com/tc39/test262/blob/70bc32edab22b44db9d671ce505db8842ae200b6/test/built-ins/JSON/stringify/replacer-array-duplicates.js
- https://github.com/tc39/test262/blob/70bc32edab22b44db9d671ce505db8842ae200b6/test/built-ins/JSON/stringify/replacer-array-number.js

Tests are not yet run automatically, but are part of the effort in #817

src/org/mozilla/javascript/NativeJSON.java
```diff
@@ -374,12 +388,12 @@ private static String jo(Scriptable value, StringifyState state) {
         state.indent = state.indent + state.gap;
         Object[] k = null;
         if (state.propertyList != null) {
-            k = state.propertyList.toArray();
```
tonygermano, line 377: `propertyList.toArray()` is called for every object and child object being stringified where propertyList is a LinkedList.

gbrail (commented):
This seems fine to me, but I'd like to see that some tests are included. Does it make sense to wait and merge this after we get more tests from test262 working, or are there some other tests that you can add?

tonygermano:
I can add a test using the examples from the linked issue. I didn't know if I needed to since it's going to be covered by test262, hopefully in the near-ish future.

rbri:
I plan to go forward with test262 updates in small steps on a weekly basis

tonygermano:
@gbrail I added some tests. The first time CircleCI tried to build, it failed with:
```
java.lang.AssertionError: In "testsrc/tests/js1_5/Regress/regress-211590.js":
 FAILED! Math.random should be random : Expected value 'between 48% and 52%', Actual value ' is 52.060' 
 ```
 
I reran the CircleCI build without changing any code, and it passed the second time. I'm guessing it was just a fluke with the rng.

gbrail:
Works for me. Thanks!

PR #877, opened 2021-05-01
Fixed evaluation order of greater than

tuchida:
Rhino was converting greater-than to less-than-or-equal by swap, so the evaluation order was wrong.
https://github.com/tc39/test262/blob/845407e0b6a9d7c97fef632a66f2f90fce7ae1f4/test/language/expressions/greater-than/11.8.2-1.js


rbri:
Great :-)

gbrail (approved):
I have to admit you had me really scared, and that switch statement in Interpreter.java really scared me even more as it takes a very careful look to see that it's not a 22-year-old cut-and-paste bug.

This looks like a fine change -- thanks for fixing it!

PR #879, opened 2021-05-02
Implement String.raw

tonygermano:
Fixed merge conflict and bugs from 3rd commit of #81

Bug fixes were likely due to changes in spec since the original PR.

I separated this out since it isn't actually dependent on template literals.

All test262 tests are now passing except for the four which depend on template literals themselves.

### DIFF
--- a/src/org/mozilla/javascript/NativeString.java
+++ b/src/org/mozilla/javascript/NativeString.java
@@ -81,6 +81,7 @@ final class NativeString extends IdScriptableObject {
     protected void fillConstructorProperties(IdFunctionObject ctor) {
         addIdFunctionProperty(ctor, STRING_TAG, ConstructorId_fromCharCode, "fromCharCode", 1);
         addIdFunctionProperty(ctor, STRING_TAG, ConstructorId_fromCodePoint, "fromCodePoint", 1);
+        addIdFunctionProperty(ctor, STRING_TAG, ConstructorId_raw, "raw", 1);
         addIdFunctionProperty(ctor, STRING_TAG, ConstructorId_charAt, "charAt", 2);
         addIdFunctionProperty(ctor, STRING_TAG, ConstructorId_charCodeAt, "charCodeAt", 2);
         addIdFunctionProperty(ctor, STRING_TAG, ConstructorId_indexOf, "indexOf", 2);
@@ -392,6 +393,9 @@ final class NativeString extends IdScriptableObject {
                         }
                         return new String(chars);
                     }
+
+                case ConstructorId_raw:
+                    return js_raw(cx, scope, args);
 
                 case Id_constructor:
                     {
@@ -1106,6 +1110,47 @@ final class NativeString extends IdScriptableObject {
         return 0;
     }
 
+    /**
+     *
+     *
+     * <h1>String.raw (template, ...substitutions)</h1>
+     *
+     * <p>22.1.2.4 String.raw [Draft ECMA-262 / April 28, 2021]
+     */
+    private static CharSequence js_raw(Context cx, Scriptable scope, Object[] args) {
+        /* step 1-2 */
+        Object arg0 = args.length > 0 ? args[0] : Undefined.instance;
+        Scriptable cooked = ScriptRuntime.toObject(cx, scope, arg0);
+        /* step 3 */
+        Object rawValue = ScriptRuntime.getObjectProp(cooked, "raw", cx);
+        Scriptable raw = ScriptRuntime.toObject(cx, scope, rawValue);
+        /* step 4-5 */
+        long rawLength = NativeArray.getLengthProperty(cx, raw);
+        if (rawLength > Integer.MAX_VALUE) {
+            throw ScriptRuntime.rangeError("raw.length > " + Integer.toString(Integer.MAX_VALUE));
+        }
+        int literalSegments = (int) rawLength;
+        if (literalSegments <= 0) return "";
+        /* step 6-7 */
+        StringBuilder elements = new StringBuilder();
+        int nextIndex = 0;
+        for (; ; ) {
+            /* step 8 a-i */
+            Object next;
+            next = ScriptRuntime.getObjectIndex(raw, nextIndex, cx);
+            String nextSeg = ScriptRuntime.toString(next);
+            elements.append(nextSeg);
+            nextIndex += 1;
+            if (nextIndex == literalSegments) {
+                break;
+            }
+            next = args.length > nextIndex ? args[nextIndex] : "";
+            String nextSub = ScriptRuntime.toString(next);
+            elements.append(nextSub);
+        }
+        return elements;
+    }
+
     // #string_id_map#
 
     @Override
@@ -1270,6 +1315,7 @@ final class NativeString extends IdScriptableObject {
 
     private static final int ConstructorId_fromCharCode = -1,
             ConstructorId_fromCodePoint = -2,
+            ConstructorId_raw = -3,
             Id_constructor = 1,
             Id_toString = 2,
             Id_toSource = 3,
--- a/src/org/mozilla/javascript/NativeString.java
+++ b/src/org/mozilla/javascript/NativeString.java
@@ -12,31 +12,27 @@ import static org.mozilla.javascript.ScriptRuntimeES6.requireObjectCoercible;
 import java.text.Collator;
 import java.text.Normalizer;
 import java.util.Locale;
-
 import org.mozilla.javascript.ScriptRuntime.StringIdOrIndex;
 import org.mozilla.javascript.regexp.NativeRegExp;
 
 /**
  * This class implements the String native object.
  *
- * See ECMA 15.5.
+ * <p>See ECMA 15.5.
  *
- * String methods for dealing with regular expressions are
- * ported directly from C. Latest port is from version 1.40.12.19
- * in the JSFUN13_BRANCH.
+ * <p>String methods for dealing with regular expressions are ported directly from C. Latest port is
+ * from version 1.40.12.19 in the JSFUN13_BRANCH.
  *
  * @author Mike McCabe
  * @author Norris Boyd
  * @author Ronald Brill
  */
-final class NativeString extends IdScriptableObject
-{
+final class NativeString extends IdScriptableObject {
     private static final long serialVersionUID = 920268368584188687L;
 
     private static final Object STRING_TAG = "String";
 
-    static void init(Scriptable scope, boolean sealed)
-    {
+    static void init(Scriptable scope, boolean sealed) {
         NativeString obj = new NativeString("");
         obj.exportAsJSClass(MAX_PROTOTYPE_ID, scope, sealed);
     }
@@ -50,19 +46,15 @@ final class NativeString extends IdScriptableObject
         return "String";
     }
 
-    private static final int
-        Id_length                    =  1,
-        MAX_INSTANCE_ID              =  1;
+    private static final int Id_length = 1, MAX_INSTANCE_ID = 1;
 
     @Override
-    protected int getMaxInstanceId()
-    {
+    protected int getMaxInstanceId() {
         return MAX_INSTANCE_ID;
     }
 
     @Override
-    protected int findInstanceIdInfo(String s)
-    {
+    protected int findInstanceIdInfo(String s) {
         if (s.equals("length")) {
             return instanceIdInfo(DONTENUM | READONLY | PERMANENT, Id_length);
         }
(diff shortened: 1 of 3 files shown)


rbri:
Great, another step forward

src/org/mozilla/javascript/NativeString.java
```diff
@@ -941,6 +946,47 @@ protected int findPrototypeId(Symbol k)
...
+        long literalSegments = NativeArray.getLengthProperty(cx, raw);
+        if (literalSegments == 0) return "";
+        /* step 6-7 */
+        StringBuilder elements = new StringBuilder();
+        long nextIndex = 0;
+        for (;;) {
+            /* step 8 a-i */
+            Object next;
+            if (nextIndex > Integer.MAX_VALUE) {
+                next = raw.get(Long.toString(nextIndex), raw);
```
gbrail, line 1139: Would you mind looking to see if there are any other tests, perhaps in the v8 project, that we could use to get better coverage? For example, I see that this case (of the index being greater than a 32-bit int) isn't being tested.

src/org/mozilla/javascript/NativeString.java
```diff
@@ -941,6 +946,47 @@ protected int findPrototypeId(Symbol k)
...
+     * <p>22.1.2.4 String.raw [Draft ECMA-262 / April 28, 2021]</p>
+     */
+    private static CharSequence js_raw(Context cx, Scriptable scope, Object[] args) {
+        final Object undefined = Undefined.instance;
+        /* step 1-2 */
+        Object arg0 = args.length > 0 ? args[0] : undefined;
+        Scriptable cooked = ScriptRuntime.toObject(cx, scope, arg0);
+        /* step 3 */
+        Object rawValue = cooked.get("raw", cooked);
+        if (rawValue == NOT_FOUND) rawValue = undefined;
```
gbrail, line 1127: FWIW if I am reading "./gradlew jacocoTestReport" correctly, the "NOT_FOUND" case here isn't actually being tested.

src/org/mozilla/javascript/NativeString.java
```diff
@@ -941,6 +946,47 @@ protected int findPrototypeId(Symbol k)
...
     }
 
+    /**
+     * <h1>String.raw (template, ...substitutions)</h1>
+     * <p>22.1.2.4 String.raw [Draft ECMA-262 / April 28, 2021]</p>
+     */
+    private static CharSequence js_raw(Context cx, Scriptable scope, Object[] args) {
+        final Object undefined = Undefined.instance;
+        /* step 1-2 */
+        Object arg0 = args.length > 0 ? args[0] : undefined;
```
gbrail, line 1123: I'm not sure if the "undefined" case is being tested here.

gbrail (commented):
Would it be possible to cast around for a test suite that hits more code coverage? For instance:

https://github.com/v8/v8/blob/master/test/mjsunit/es6/string-raw.js

We've used a bunch of these in the "testsrc/jstests" directory. They don't always pass due to other things that we're missing, and when that happens I put a "TODO Rhino" comment around them, but that might help, since I see a bunch of the corner cases here not being covered. (And JavaScript, it turns out, is about 90% corner cases!)

Sorry, I am getting gradually more picky since I see so much good new stuff coming ;-)

src/org/mozilla/javascript/NativeString.java
tonygermano: Can I actually just change nextIndex from a long to an int, and get rid of that step all together? It will be impossible to hit that number, because `args` would have to be an `Object[]` with `Integer.MAX_VALUE+1` elements.

tonygermano: Hmm, I guess raw can have a length up to `2^53-1`, and if that happens, it will stop doing substitutions from args and use empty strings for the remainder of the substitutions. Do we really want a test that has to iterate that many times? It will probably cause an out of memory error.

tonygermano: I think it would get an out of memory error because either raw would have to have a massive amount of properties, or it is sparse, and the output will repeat "undefined" for each index that doesn't have an element.

gbrail:
Thanks, this looks good -- I agree that we don't want to have a unit test that has more than 2^32 elements anyway.

src/org/mozilla/javascript/NativeString.java
tonygermano: just to confirm, I ran the following in a shell
```js
Rhino 1.7.14-SNAPSHOT 2021 05 05
js> String.raw({raw:{length:5}})
undefinedundefinedundefinedundefinedundefined
js> String.raw({raw:{length:Math.pow(2,32)+1}})
java.lang.OutOfMemoryError: Java heap space
        at java.base/java.util.Arrays.copyOf(Arrays.java:3745)
        at java.base/java.lang.AbstractStringBuilder.ensureCapacityInternal(AbstractStringBuilder.java:172)
        at java.base/java.lang.AbstractStringBuilder.append(AbstractStringBuilder.java:538)
        at java.base/java.lang.StringBuilder.append(StringBuilder.java:174)
        at org.mozilla.javascript.NativeString.js_raw(NativeString.java:1145)
        at org.mozilla.javascript.NativeString.execIdCall(NativeString.java:398)
        at org.mozilla.javascript.IdFunctionObject.call(IdFunctionObject.java:100)
        at org.mozilla.javascript.optimizer.OptRuntime.call1(OptRuntime.java:45)
        at org.mozilla.javascript.gen._stdin__2._c_script_0(Unknown Source)
        at org.mozilla.javascript.gen._stdin__2.call(Unknown Source)
        at org.mozilla.javascript.ContextFactory.doTopCall(ContextFactory.java:415)
        at org.mozilla.javascript.ScriptRuntime.doTopCall(ScriptRuntime.java:3603)
        at org.mozilla.javascript.gen._stdin__2.call(Unknown Source)
        at org.mozilla.javascript.gen._stdin__2.exec(Unknown Source)
        at org.mozilla.javascript.tools.shell.Main.processSource(Main.java:497)
        at org.mozilla.javascript.tools.shell.Main.processFiles(Main.java:181)
        at org.mozilla.javascript.tools.shell.Main$IProxy.run(Main.java:101)
        at org.mozilla.javascript.Context.call(Context.java:534)
        at org.mozilla.javascript.ContextFactory.call(ContextFactory.java:525)
        at org.mozilla.javascript.tools.shell.Main.exec(Main.java:163)
        at org.mozilla.javascript.tools.shell.Main.main(Main.java:138)
js: exception from uncaught JavaScript throw: java.lang.OutOfMemoryError: Java heap space
js> 
```

gbrail: I'm find with leaving this untested, but it raises a bigger problem that we have in a lot of places -- JavaScript strings are supposed to have a maximum length of 2^53 but Java strings have a maximum length of 2^31. I don't think that we can change that all over Rhino. But if someone DOES try to create a raw string that's longer than that, then we'll likely leak a runtime exception from somewhere in Java.

I think we're better off here throwing an exception if the length is larger than the maximum integer size rather than let some other kind of exception bubble up.

tonygermano: So, if literalSegments > Interger.MAX_VALUE throw a RangeError? And then I could make nextIndex an int and remove this check?

gbrail: Yes! I think that would prevent worse problems later.

tonygermano: One other question that I wasn't able to find an answer to by searching...
Is there a negative performance impact for checking `if (a == b)` many times when one variable is an int and another is a long? My understanding is that the int is implicitly converted to a long. Would it be better to do this to avoid the conversions later, or does it not really matter?
```java
long rawLength = NativeArray.getLengthProperty(cx, raw);
if (rawLength > Integer.MAX_VALUE) {/*throw RangeError*/}
int literalSegments = (int) rawLength;
```

src/org/mozilla/javascript/NativeString.java
tonygermano: Updated this to use `ScriptRuntime.getObjectProp` instead, which does the correct ecma `[[Get]]` operation. It now searches up the prototype chain and returns undefined if not found in a single call. The test that covers raw being undefined is `template-raw-not-object-throws.js`

src/org/mozilla/javascript/NativeString.java
tonygermano: `template-not-object-throws.js` tests `String.raw(undefined)`. Is that good enough or do you want a `String.raw()`, too?

src/org/mozilla/javascript/NativeString.java
```diff
@@ -1106,6 +1110,47 @@ protected int findPrototypeId(Symbol k) {
...
+        }
+        int literalSegments = (int) rawLength;
+        if (literalSegments <= 0) return "";
+        /* step 6-7 */
+        StringBuilder elements = new StringBuilder();
+        int nextIndex = 0;
+        for (; ; ) {
+            /* step 8 a-i */
+            Object next;
+            next = ScriptRuntime.getObjectIndex(raw, nextIndex, cx);
```
tonygermano, line 1140: This also changed from the original PR to use `ScriptRuntime.getObjectIndex` instead of `Scriptable.get(int Scriptable)`, which searches up the prototype chain and returns undefined when not found, rather than only searching the current object, and returning Scriptable.NOT_FOUND.

src/org/mozilla/javascript/NativeString.java
tonygermano: I went ahead and implemented it the way I asked about in my last question. In addition to removing the implicit conversion in the comparison, it also removed the need to explicitly cast long to int in two places further down in the code.

gbrail:
The code looks great and thanks for the attention to detail. It looks like this fixes a bunch of test262 tests -- I can see the output here:

Could you enable those String.raw tests in test262 in this PR, or if you're really tired of iterating, in another PR? Thanks!

org.mozilla.javascript.tests.Test262SuiteTest STANDARD_OUT
    Test is marked as failing but it does not: test262/test/built-ins/DataView/prototype/getBigInt64/return-abrupt-from-tonumber-byteoffset-symbol.js
    Test is marked as failing but it does not: test262/test/built-ins/RegExp/prototype/Symbol.match/builtin-failure-g-set-lastindex-err.js
    Test is marked as failing but it does not: test262/test/built-ins/RegExp/prototype/Symbol.match/g-init-lastindex-err.js
    Test is marked as failing but it does not: test262/test/built-ins/String/raw/length.js
    Test is marked as failing but it does not: test262/test/built-ins/String/raw/name.js
    Test is marked as failing but it does not: test262/test/built-ins/String/raw/raw.js
    Test is marked as failing but it does not: test262/test/built-ins/String/raw/return-empty-string-from-empty-array-length.js
    Test is marked as failing but it does not: test262/test/built-ins/String/raw/return-empty-string-if-length-is-negative-infinity.js
    Test is marked as failing but it does not: test262/test/built-ins/String/raw/return-empty-string-if-length-is-not-defined.js
    Test is marked as failing but it does not: test262/test/built-ins/String/raw/return-empty-string-if-length-is-undefined.js
    Test is marked as failing but it does not: test262/test/built-ins/String/raw/return-empty-string-if-length-is-zero-NaN.js
    Test is marked as failing but it does not: test262/test/built-ins/String/raw/return-empty-string-if-length-is-zero-boolean.js
    Test is marked as failing but it does not: test262/test/built-ins/String/raw/return-empty-string-if-length-is-zero-null.js
    Test is marked as failing but it does not: test262/test/built-ins/String/raw/return-empty-string-if-length-is-zero-or-less-number.js
    Test is marked as failing but it does not: test262/test/built-ins/String/raw/return-empty-string-if-length-is-zero-or-less-string.js
    Test is marked as failing but it does not: test262/test/built-ins/String/raw/return-the-string-value.js
    Test is marked as failing but it does not: test262/test/built-ins/String/raw/returns-abrupt-from-next-key-toString.js
    Test is marked as failing but it does not: test262/test/built-ins/String/raw/returns-abrupt-from-next-key.js
    Test is marked as failing but it does not: test262/test/built-ins/String/raw/returns-abrupt-from-substitution.js
    Test is marked as failing but it does not: test262/test/built-ins/String/raw/substitutions-are-appended-on-same-index.js
    Test is marked as failing but it does not: test262/test/built-ins/String/raw/substitutions-are-limited-to-template-raw-length.js
    Test is marked as failing but it does not: test262/test/built-ins/String/raw/template-length-throws.js
    Test is marked as failing but it does not: test262/test/built-ins/String/raw/template-raw-throws.js
    Test is marked as failing but it does not: test262/test/language/expressions/greater-than-or-equal/bigint-and-string.js
    Test is marked as failing but it does not: test262/test/language/expressions/less-than/bigint-and-string.js

tonygermano:
Oops. I had enabled all of the String.raw ones in the original PR, but I must have killed it in a rebase. I'll try to get that back in in a bit. I'll throw those few extra ones that aren't related to this PR in a separate commit so we don't need to make a whole new PR for them.

tonygermano:
test262 updates are now back in this PR. I think it's ready.

PR #880, opened 2021-05-02
Update test262

rbri:
another small move forward of the Test262 suite incl. updated unsupported features and a minor date fix

### DIFF
--- a/src/org/mozilla/javascript/NativeDate.java
+++ b/src/org/mozilla/javascript/NativeDate.java
@@ -12,185 +12,312 @@ import java.util.Date;
 import java.util.TimeZone;
 
 /**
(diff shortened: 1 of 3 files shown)


rbri:
btw have not done the reformatting of NativeDate to make an integration please decide if we should do this for this minor change.

gbrail (changes requested):
Thanks for sticking with this. Of course we introduced new conflicts while you were doing this ;-)

Yes, can you please reformat NativeDate too? It does seem like we are going to have less reformatting pain soon.

(Here's a longer-term idea -- should we try to separate test262.properties into multiple files so it's not always being merged?)

gbrail:
This is super helpful. Thanks!

PR #881, opened 2021-05-04
Allow abstract LengthOfArrayLike to operate on XMLObject

tonygermano:
`XMLObject` is indexed like an array, but it has a `length()` method instead of a `length` property. This allows `XML` and `XMLList` objects to be used as arrays in several places, including many `Array.prototype` methods (see tests.)

This abstract method is implemented as `static long NativeArray.getLengthProperty(Context, Scriptable)`.

### DIFF
--- a/src/org/mozilla/javascript/NativeArray.java
+++ b/src/org/mozilla/javascript/NativeArray.java
@@ -19,6 +19,7 @@ import java.util.ListIterator;
 import java.util.NoSuchElementException;
 
 import org.mozilla.javascript.regexp.NativeRegExp;
+import org.mozilla.javascript.xml.XMLObject;
 
 /**
  * This class implements the Array native object.
@@ -847,12 +848,16 @@ public class NativeArray extends IdScriptableObject implements List
      * or its value is not convertible to a number.
      */
     static long getLengthProperty(Context cx, Scriptable obj) {
-        // These will both give numeric lengths within Uint32 range.
+        // These will give numeric lengths within Uint32 range.
         if (obj instanceof NativeString) {
             return ((NativeString)obj).getLength();
         }
         if (obj instanceof NativeArray) {
             return ((NativeArray)obj).getLength();
+        }
+        if (obj instanceof XMLObject) {
+            Callable lengthFunc = (Callable) ((XMLObject) obj).get("length", obj);
+            return ((Number) lengthFunc.call(cx, obj, obj, ScriptRuntime.emptyArgs)).longValue();
         }
 
         Object len = ScriptableObject.getProperty(obj, "length");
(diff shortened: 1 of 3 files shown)


src/org/mozilla/javascript/NativeArray.java
```diff
@@ -847,13 +848,17 @@ private void setLength(Object val) {
...
-        // These will both give numeric lengths within Uint32 range.
+        // These will give numeric lengths within Uint32 range.
         if (obj instanceof NativeString) {
             return ((NativeString)obj).getLength();
         }
         if (obj instanceof NativeArray) {
             return ((NativeArray)obj).getLength();
         }
+        if (obj instanceof XMLObject) {
+            Callable lengthFunc = (Callable) ((XMLObject) obj).get("length", obj);
```
gbrail, line 917: This seems to assume that "length" is present on all XMLObject instances and that it's a function. Is there a possibility that it is really something else, or not present at all?

Also, no big deal but the additional cast to "XMLObject" isn't doing anything here since you have to cast it to Callable anyway.

src/org/mozilla/javascript/NativeArray.java
```diff
@@ -847,13 +848,17 @@ private void setLength(Object val) {
...
+            return ((Number) lengthFunc.call(cx, obj, obj, ScriptRuntime.emptyArgs)).longValue();
```
gbrail, line 918: Is it possible that the result will be not a number, or out of range? There's code before to validate "doubleLen". Does it make sense to set "doubleLen" and use the rest of the logic in this function instead?

gbrail (commented):
Just a few questions since for native JavaScript type, the question "what is the length" turns out to be very complicated!

src/org/mozilla/javascript/NativeArray.java
tonygermano: The e4x implementation is kind of odd, and I don't really understand why it's done the way that it is. XMLObject is a public class, but does not really represent what an e4x object is. The real e4x api is in XMLObjectImpl, but that class is package-private (and in a different package,) so I can't call the length method on it directly. Instead I need to go the round-about way to get the method as a property from the object. The only two implementations of XMLObjectImpl, XML and XMLList, will both always have a length method, per the e4x spec.

I'm casting `obj` to an XMLObject, and the return value to a Callable.

src/org/mozilla/javascript/NativeArray.java
tonygermano: XMLObjectImpl.length() returns an int.

src/org/mozilla/javascript/NativeArray.java
gbrail: Agreed -- the XML implementation is weird but probably not worth making un-weird at this point. I think this is fine to go for now.

p-bakker: Just for future reference: I think the separation of XMLObject and XMLObjectImpl stems from the time where the actual implementation of E4X relied on 3rd party libraries for all the XML handling and when not enabling E4X in Rhino the 3rd party xml libs shouldn't be a hard dependancy

PR #884, opened 2021-05-05
Optimize addition in generated bytecode

gbrail:
Use invokedynamic via a new "FallbackCallSite" class that can try various
    methods in order and adjust at runtime to fall back to less-specific
    methods. This makes it possible to try a set of efficient methods that
    assume what their operand types and often arrive at the most efficient
    after a few iterations.

    Use this method to optimize the "add" operation for numbers and strings.

    Also optimize the "OptRuntime.add" operation for better performance.

    Taken together, in microbenchmarks, integer and float addition is about
    twice as fast and string concatenation in a loop is about four times faster.

### DIFF
--- a/src/org/mozilla/javascript/optimizer/DynamicRuntime.java
+++ b/src/org/mozilla/javascript/optimizer/DynamicRuntime.java
@@ -1,0 +1,99 @@
+package org.mozilla.javascript.optimizer;
+
+import java.lang.invoke.CallSite;
+import java.lang.invoke.MethodHandle;
+import java.lang.invoke.MethodHandles;
+import java.lang.invoke.MethodType;
+import org.mozilla.classfile.ByteCode;
+import org.mozilla.classfile.ClassFileWriter.MHandle;
+import org.mozilla.javascript.ConsString;
+import org.mozilla.javascript.Context;
+import org.mozilla.javascript.Kit;
+import org.mozilla.javascript.ScriptRuntime;
+
+public class DynamicRuntime {
+    public static final String ADD_OP = "add";
+
+    public static final String BOOTSTRAP_SIGNATURE =
+            "(Ljava/lang/invoke/MethodHandles$Lookup;"
+                    + "Ljava/lang/String;"
+                    + "Ljava/lang/invoke/MethodType;"
+                    + ")Ljava/lang/invoke/CallSite;";
+    public static final String METHOD_SIGNATURE_OOC =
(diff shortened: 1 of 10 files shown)


src/org/mozilla/javascript/optimizer/DynamicRuntime.java
```diff
@@ -0,0 +1,99 @@
...
+    public static Object addIntegers(Object o1, Object o2, Context cx) {
+        try {
+            return (Integer) o1 + (Integer) o2;
+        } catch (ClassCastException | NullPointerException e) {
+            throw new FallbackException();
+        }
+    }
+
+    /** Perform an add assuming both arguments are Double objects and fail otherwise. */
+    public static Object addFloats(Object o1, Object o2, Context cx) {
```
tuchida, line 42: Why is it `float`? In Java, `float` and `double` are different things.

src/org/mozilla/javascript/optimizer/DynamicRuntime.java
```diff
@@ -0,0 +1,99 @@
...
+
+    public static final MHandle BOOTSTRAP_HANDLE =
+            new MHandle(
+                    ByteCode.MH_INVOKESTATIC,
+                    "org.mozilla.javascript.optimizer.DynamicRuntime",
+                    "bootstrap",
+                    DynamicRuntime.BOOTSTRAP_SIGNATURE);
+
+    /** Perform an add assuming both arguments are Integer objects and fail otherwise. */
+    public static Object addIntegers(Object o1, Object o2, Context cx) {
```
tuchida, line 33: How often do you use `java.lang.Integer` in Rhino?
As far as I know, the result of numeric operations is almost always a `double` or `java.lang.Double`.
https://github.com/mozilla/rhino/blob/3a29b988e7e918ac6478d72c1e6e14af9e25e547/src/org/mozilla/javascript/optimizer/BodyCodegen.java#L3365
Do you mean you want to use `java.lang.Integer` as much as possible in the future?

tonygermano: I'm also curious what the motivation was for a special case for Integer addition and for removing ScriptRuntime.wrapNumber from the result of adding two Numbers

gbrail:
I'd like to more extensively explore using invokedynamic later when I have time to properly try Java 9's "dynalink" module. Until then I'm going to close this in favor of a simpler PR.

PR #851, opened 2021-03-26
Improve speed of bitwiseNOT.

tuchida:
https://github.com/mozilla/rhino/pull/837#issuecomment-798295199
When it don't need to convert to Object, it don't.

before
```
# JMH version: 1.27
# VM version: JDK 1.8.0_282, OpenJDK 64-Bit Server VM, 25.282-b08
# VM invoker: /usr/lib/jvm/java-8-openjdk-amd64/jre/bin/java
# VM options: -Dfile.encoding=UTF-8 -Duser.country=JP -Duser.language=ja -Duser.variant
# JMH blackhole mode: full blackhole + dont-inline hint
# Warmup: 5 iterations, 10 s each
# Measurement: 5 iterations, 10 s each
# Timeout: 10 min per iteration
# Threads: 1 thread, will synchronize iterations
# Benchmark mode: Average time, time/op
# Benchmark: org.mozilla.javascript.benchmarks.SunSpiderBenchmark.bitopsNsieveBits

# Run progress: 26.92% complete, ETA 00:31:47
# Fork: 1 of 1
# Warmup Iteration   1: 19254.470 us/op
# Warmup Iteration   2: 18990.886 us/op
# Warmup Iteration   3: 19167.778 us/op
# Warmup Iteration   4: 19168.355 us/op
# Warmup Iteration   5: 19200.321 us/op
Iteration   1: 19247.105 us/op
Iteration   2: 19062.072 us/op
Iteration   3: 19205.020 us/op
Iteration   4: 19181.690 us/op
Iteration   5: 19168.668 us/op


Result "org.mozilla.javascript.benchmarks.SunSpiderBenchmark.bitopsNsieveBits":
  19172.911 ±(99.9%) 264.802 us/op [Average]
  (min, avg, max) = (19062.072, 19172.911, 19247.105), stdev = 68.768
  CI (99.9%): [18908.110, 19437.713] (assumes normal distribution)
```

after
```
# JMH version: 1.27
# VM version: JDK 1.8.0_282, OpenJDK 64-Bit Server VM, 25.282-b08
# VM invoker: /usr/lib/jvm/java-8-openjdk-amd64/jre/bin/java
# VM options: -Dfile.encoding=UTF-8 -Duser.country=JP -Duser.language=ja -Duser.variant
# JMH blackhole mode: full blackhole + dont-inline hint
# Warmup: 5 iterations, 10 s each
# Measurement: 5 iterations, 10 s each
# Timeout: 10 min per iteration
[...]

PR #886, opened 2021-05-06
Optimize Token.POS and Token.NEG

tuchida:
Optimized the unary operators plus and minus, just like #851.

SunSpiderBenchmark.accessFannkuch is now slower, but V8Benchmark.boyer is now faster.
https://docs.google.com/spreadsheets/d/1cjeRNOJWjxEroZmRmuP-42FGL9--fLQLyILLxPfAKAM/edit?usp=sharing

In particular, this kind of code is now faster.
```js
function f() {
  var a = 0;
  for (var i = 0; i < 10000000; i++) {
    a = +a;
    a = -a;
    a = +a;
    a = -a;
    a = +a;
    a = -a;
    a = +a;
    a = -a;
    a = +a;
    a = -a;
  }
}
f();
```
```sh
# master
$ time java -jar buildGradle/libs/rhino-1.7.14-SNAPSHOT.jar -opt 9 -version 200 test.js

real	0m0.634s
user	0m0.855s
sys	0m0.227s

# opt-posneg
$ time java -jar buildGradle/libs/rhino-1.7.14-SNAPSHOT.jar -opt 9 -version 200 test.js

real	0m0.213s
user	0m0.359s
sys	0m0.053s

```

### DIFF
--- a/src/org/mozilla/javascript/optimizer/Optimizer.java
+++ b/src/org/mozilla/javascript/optimizer/Optimizer.java
@@ -303,6 +303,8 @@ class Optimizer {
                 }
 
             case Token.BITNOT:
+            case Token.POS:
+            case Token.NEG:
                 {
                     Node child = n.getFirstChild();
                     int type = rewriteForNumberVariables(child, NumberType);
--- a/src/org/mozilla/javascript/optimizer/BodyCodegen.java
+++ b/src/org/mozilla/javascript/optimizer/BodyCodegen.java
@@ -1233,16 +1233,29 @@ class BodyCodegen {
                 break;
 
             case Token.POS:
-                generateExpression(child, node);
-                addObjectToDouble();
-                addDoubleWrap();
-                break;
+                {
+                    int childNumberFlag = node.getIntProp(Node.ISNUMBER_PROP, -1);
+                    generateExpression(child, node);
+                    if (childNumberFlag == -1) {
+                        addObjectToDouble();
+                        addDoubleWrap();
+                    }
+                    break;
+                }
 
             case Token.NEG:
-                generateExpression(child, node);
-                addObjectToNumeric();
-                addScriptRuntimeInvoke("negate", "(Ljava/lang/Number;" + ")Ljava/lang/Number;");
-                break;
+                {
+                    int childNumberFlag = node.getIntProp(Node.ISNUMBER_PROP, -1);
+                    generateExpression(child, node);
+                    if (childNumberFlag == -1) {
+                        addObjectToNumeric();
+                        addScriptRuntimeInvoke(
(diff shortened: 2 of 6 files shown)


src/org/mozilla/javascript/optimizer/Optimizer.java
```diff
@@ -303,6 +303,8 @@ private int rewriteForNumberVariables(Node n, int desired) {
                 }
 
             case Token.BITNOT:
+            case Token.POS:
+            case Token.NEG:
                 {
```
tonygermano, line 308: I've never looked at this area of the code before, but should BITNOT, POS, and NEG be moved up to the same section as INC and DEC, since they are all unary operators?

I could be wrong, because I've been looking at this for a while and still can't figure out how it works, but intuitively, it seems like these should all be the same.

tonygermano: I realized the difference may be in that INC and DEC mutate the variable, but I still don't understand how this works :)

tuchida: I used binary operators such as `Token.SUB` and unary operators such as `Token.INC` as a reference here.

I'm not sure if `NoType` or `AnyType` is the right choice here.
https://github.com/mozilla/rhino/blob/ae3563a99b88585d8d4ee72ca03466a33a58dd79/src/org/mozilla/javascript/optimizer/Optimizer.java#L316

tonygermano: From what I can tell, in the case where `(child.getType() == Token.GETELEM || child.getType() == Token.GETPROP)` then type will always be NoType, so that extra check in INC and DEC seems unnecessary. Though, I wonder if it _should_ be possible for them to return something else.

For example in the INC case, I would think if there ever were a case where the child.getType() is GETELEM and type is NumberType, it would be for something like:
```js
(function() {
  var a = [1,2];
  a[0]++;
})()
```

I do feel like I'm in a bit over my head here, though, since I don't understand exactly how everything is supposed to work. I don't suppose anyone knows of any design docs or references that explain the optimizer?

gbrail (commented):
I get that this works in the tests and speeds things up but I have to think some more about whether it's safe.

Previously, the code generated for Token.POS always left a Double on the stack, and now it might leave a Double or a double I think. Same goes for Token.NEG. It may well be the case that the way this is used in the actual codegen it doesn't matter, since if you look at the actual code there's all sorts of weird stuff in it, but I'm not sure how to verify that.

tuchida:
ref. 9d8c669, 4d94617
I added tests.

It certainly speeds up benchmarking, but it is questionable how much faster it will be in real-world code.

gbrail:
You've tested this carefully -- let's go with it. Thanks!

PR #888, opened 2021-05-09
add new StackStyle.MOZILLA_LF

rbri:
HtmlUnit is mainly a testing framework and we are using the mozilla StackStyle. To have stable tests the stack trace should not be system dependent. And the modern browsers are also using \n on all platforms.

To be backward compatible i have introduced a new StackStyle.

See https://github.com/HtmlUnit/htmlunit/issues/284.


gbrail:
This one makes sense to me and it's pretty low risk. Thanks!

PR #889, opened 2021-05-10
Add AutoGrow for NativeJavaList / fix NPE in NativeJavaList/Map

rPraml:
This PR contains 2 code changes which should be unproblematic

1. When a List/Map contains null values there was an NPE on `cx.getWrapFactory().wrap(cx, this, obj, obj.getClass());` See: https://github.com/mozilla/rhino/commit/084908502006e3039c867dcf950c7c9548e8c34e
2. Support for auto grow, so that you can set values in a java.util.List. See https://github.com/mozilla/rhino/commit/cf84dbf60e417378eb171e20769ee92407cac808

The rest are spotless fixes

### DIFF
--- a/src/org/mozilla/javascript/NativeJavaList.java
+++ b/src/org/mozilla/javascript/NativeJavaList.java
@@ -24,7 +24,6 @@ public class NativeJavaList extends NativeJavaObject {
         return "JavaList";
     }
 
-
     @Override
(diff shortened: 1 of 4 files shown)


gbrail:
Those changes look very good and coverage is fine for the new stuff. Thanks!

PR #892, opened 2021-05-13
Optimize two specific addition cases

gbrail:
This adds a microbenchmark for math operations and two specific optimizations to the "add" optimization.

The first fixes some cases in which optimized math opeations would take a slower path before a faster one (and also would not fall back properly to BigInteger operations).

The second fixes string concatenation cases, especially when strings are concatenated in a loop.

### DIFF
--- a/src/org/mozilla/javascript/optimizer/OptRuntime.java
+++ b/src/org/mozilla/javascript/optimizer/OptRuntime.java
@@ -2,12 +2,10 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-
 package org.mozilla.javascript.optimizer;
 
 import org.mozilla.javascript.ArrowFunction;
 import org.mozilla.javascript.Callable;
-import org.mozilla.javascript.ConsString;
 import org.mozilla.javascript.Context;
 import org.mozilla.javascript.ContextFactory;
 import org.mozilla.javascript.ES6Generator;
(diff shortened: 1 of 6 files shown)


src/org/mozilla/javascript/optimizer/OptRuntime.java
```diff
@@ -22,116 +20,88 @@
...
 
-    public static Object add(double val1, Object val2)
-    {
-        if (val2 instanceof Scriptable)
-            val2 = ((Scriptable) val2).getDefaultValue(null);
-        if (!(val2 instanceof CharSequence))
-            return wrapDouble(toNumber(val2) + val1);
-        return new ConsString(toString(val1), (CharSequence)val2);
+    public static Object add(double val1, Object val2, Context cx) {
+        if (val2 instanceof Integer) {
```
tuchida, line 87: I think `java.lang.Double` is used more than `java.lang.Integer` inside Rhino.
I grep with `wrapInt`, I found 14 matches, but with `wrapNumber`, I found 90 matches.

In the future, do you want to use integers like v8 to make it faster?
https://medium.com/fhinkel/v8-internals-how-small-is-a-small-integer-e0badc18b6da

gbrail: Yes, in general I'd like to try and start steering Rhino into using Integers more when we can. (For example, the standard constant for "zero" is a double, which means that we use doubles in every "for (var i = 0; i < foo; i++)" loop!

I agree that doubles are more common and I'll try switching them. 

Over time I want to go back to using invokedynamic in the bytecode for all the super-complex operations in ScriptRuntime like "add," "compare," "eq," etc. That will allow the JVM at runtime to re-order which operations are most likely to happen at a particular call site. However I still haven't come up with a single place where this makes a big difference yet, as opposed to simpler optimizations like this one.

PR #895, opened 2021-05-15
update to version 1056d8fd (2019-09-18)

rbri:
can't move any further because in the next step the test harness assert.js switches to string templates


gbrail:
Works great. Sounds like a few people might be interested in figuring out string templates. Thanks!

PR #896, opened 2021-05-17
Refactor SlotMap and Slot classes

gbrail:
Ensure that Slot and all related classes are separate classes now and not just nested inside ScriptableObject.

Also, break the two Slot implementations (regular and "GetterSlot") into four so we can use inheritance rather than complex "instanceof" logic to implement the different types of slots.

This simplifies the structure and logic and will make it easier to extend stuff later.

### DIFF
--- a/src/org/mozilla/javascript/MemberBoxSlot.java
+++ b/src/org/mozilla/javascript/MemberBoxSlot.java
@@ -28,22 +28,8 @@ public class MemberBoxSlot extends Slot {
         // but the spec is super pedantic about things like the order of properties here,
         // so we need special support here.
         ScriptableObject desc = (ScriptableObject) cx.newObject(scope);
+        desc.setCommonDescriptorProperties(getAttributes(), getter == null && setter == null);
         String fName = name == null ? "f" : name.toString();
-        desc.defineProperty(
-                "enumerable",
-                (getAttributes() & ScriptableObject.DONTENUM) == 0,
-                ScriptableObject.EMPTY);
-        desc.defineProperty(
-                "configurable",
-                (getAttributes() & ScriptableObject.PERMANENT) == 0,
-                ScriptableObject.EMPTY);
-        if (getter == null && setter == null) {
-            desc.defineProperty(
-                    "writable",
-                    (getAttributes() & ScriptableObject.READONLY) == 0,
-                    ScriptableObject.EMPTY);
-        }
-
         if (getter != null) {
             desc.defineProperty(
                     "get", getter.asGetterFunction(fName, scope), ScriptableObject.EMPTY);
--- a/src/org/mozilla/javascript/MemberBoxSlot.java
+++ b/src/org/mozilla/javascript/MemberBoxSlot.java
@@ -56,16 +56,12 @@ public class MemberBoxSlot extends Slot {
             Class<?> valueType = pTypes[pTypes.length - 1];
             int tag = FunctionObject.getTypeTag(valueType);
             Object actualArg = FunctionObject.convertArg(cx, start, value, tag);
-            Object setterThis;
-            Object[] args;
+
             if (setter.delegateTo == null) {
-                setterThis = start;
-                args = new Object[] {actualArg};
+                setter.invoke(start, new Object[] {actualArg});
             } else {
-                setterThis = setter.delegateTo;
-                args = new Object[] {start, actualArg};
+                setter.invoke(setter.delegateTo, new Object[] {start, actualArg});
             }
-            setter.invoke(setterThis, args);
             return true;
         }
         return super.setValue(value, owner, start);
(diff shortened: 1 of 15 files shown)


src/org/mozilla/javascript/ThreadSafeSlotMapContainer.java
```diff
@@ -62,13 +62,21 @@ public boolean isEmpty() {
     }
 
     @Override
-    public Slot get(Object key, int index, SlotAccess accessType) {
+    public Slot modify(Object key, int index, int attributes) {
+        final long stamp = lock.writeLock();
+        try {
+            checkMapSize();
```
tonygermano, line 68: It's probably fine to call checkMapSize here, but should the implementation of checkMapSize in this class actually do anything? Calling super doesn't seem right because at best it won't do anything (current behavior,) and at worst it could swap the map for a non-threadsafe version.

rbri (approved):
looks like a real improvement, did not found the time to dive into the details - but one suggestion:
What about placing all the different slot implementations in an separate sub package?

src/org/mozilla/javascript/SlotMap.java
```diff
@@ -25,21 +25,21 @@
 
     /**
      * Return the Slot that matches EITHER "key" or "index". (It will use "key" if it is not null,
-     * and otherwise "index". "accessType" is one of the constants defined in ScriptableObject.
+     * and otherwise "index".) If no slot exists, then create a default slot class.
```
rbri, line 28: If we touching the javadoc here what about using @retruns and @param.

src/org/mozilla/javascript/Slot.java
```diff
@@ -0,0 +1,117 @@
+package org.mozilla.javascript;
+
+import java.io.IOException;
+import java.io.ObjectInputStream;
+import java.io.Serializable;
+
+/**
+ * A Slot is the base class for all properties stored in the ScriptableObject class. There are a
+ * number of different types of slots. This base class represents an "ordinary" property such a s
```
rbri, line 9: such a s???

src/org/mozilla/javascript/MemberBoxSlot.java
```diff
@@ -0,0 +1,120 @@
...
+            Object getterThis;
+            Object[] args;
+            if (getter.delegateTo == null) {
+                getterThis = start;
+                args = ScriptRuntime.emptyArgs;
+            } else {
+                getterThis = getter.delegateTo;
+                args = new Object[] {start};
+            }
+            return getter.invoke(getterThis, args);
```
rbri, line 86: Maybe we can let this look more simple

    if (getter.delegateTo == null) {
        return getter.invoke(start, ScriptRuntime.emptyArgs);
    }
    return getter.invoke(getter.delegateTo, new Object[] {start});

src/org/mozilla/javascript/MemberBoxSlot.java
```diff
@@ -0,0 +1,120 @@
...
+            Object[] args;
+            if (setter.delegateTo == null) {
+                setterThis = start;
+                args = new Object[] {actualArg};
+            } else {
+                setterThis = setter.delegateTo;
+                args = new Object[] {start, actualArg};
+            }
+            setter.invoke(setterThis, args);
+            return true;
```
rbri, line 69: see my comment below - seems to be the same pattern here

src/org/mozilla/javascript/ThreadSafeSlotMapContainer.java
gbrail: SlotMapContainer and ThreadSafeSlotMapContainer are just wrappers for either an EmbeddedSlotMap or a HashSlotMap. Neither of those last two types are thread safe -- we're relying on the thread-safe wrapper to do the locking. So there isn't any chance of the map on a particular object being replaced with a non-thread safe version.

src/org/mozilla/javascript/SlotMap.java
gbrail: Thanks -- I did a little bit of cleanup.

src/org/mozilla/javascript/Slot.java
gbrail: Fixed weird comment.

src/org/mozilla/javascript/MemberBoxSlot.java
gbrail: Yeah, I cut and pasted and I don't know why that code was so weird.

src/org/mozilla/javascript/MemberBoxSlot.java
gbrail: Ack.

gbrail:
I'm going to experiment with a different package -- but it's going to make this an even bigger PR.

gbrail:
Moving the package is going to result in a ton of changes because of use of package-protected fields and methods. I'm going to leave these classes in the main package and we might consider a change at a future time.

PR #897, opened 2021-05-18
Improve test run capabilities

p-bakker:
Added ability to specify a custom (test262).properties file, so you can create a custom one containing only the tests (for the feature you're trying to implement).

Also extended the docs/readme with regards to running tests

### DIFF
--- a/testsrc/org/mozilla/javascript/tests/Test262SuiteTest.java
+++ b/testsrc/org/mozilla/javascript/tests/Test262SuiteTest.java
@@ -103,7 +103,8 @@ public class Test262SuiteTest {
         }
 
         String propFile = System.getProperty("test262properties");
-        testProperties = propFile != null && !propFile.equals("") ? propFile : "testsrc/test262.properties";
+        testProperties = 
+                propFile != null && !propFile.equals("") ? propFile : "testsrc/test262.properties";
     }
 
     private static String getOverriddenLevel() {
--- a/testsrc/org/mozilla/javascript/tests/Test262SuiteTest.java
+++ b/testsrc/org/mozilla/javascript/tests/Test262SuiteTest.java
@@ -103,7 +103,7 @@ public class Test262SuiteTest {
         }
 
         String propFile = System.getProperty("test262properties");
-        testProperties = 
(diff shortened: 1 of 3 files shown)


testsrc/org/mozilla/javascript/tests/Test262SuiteTest.java
```diff
@@ -51,7 +51,7 @@
 
     private static final File testDir = new File("test262/test");
     private static final String testHarnessDir = "test262/harness/";
-    private static final String testProperties = "testsrc/test262.properties";
+    private static String testProperties = "testsrc/test262.properties";
```
tuchida, line 54: Is this to remain `static`? I'm not familiar with the behavior of JUnit, but if the variable changes with each execution, I'm worried that `static` will be a problem.

tuchida: Sorry. It needs to be `static`.
https://github.com/mozilla/rhino/blob/a84308f16abd10276a42ce1b7b3211cc0c674a48/testsrc/org/mozilla/javascript/tests/Test262SuiteTest.java#L96

gbrail: You can make this work if you make testProperties final again, but don't assign a value in the initializer -- instead add an "else" down in your static block to assign the default value if the system property isn't there. This is exactly the same thing that we do in that same static block for OPT_LEVELS.

This minor but we're trying to be more careful and consistent with the little things across the codebase. Thanks!

gbrail:
I'd also like to see the use of "static final" here being consistent with the rest of the things in this test class, but otherwise this is helpful and I especially appreciate the improved documentation.

testsrc/org/mozilla/javascript/tests/Test262SuiteTest.java
p-bakker: Changed

p-bakker:
@gbrail both your reply and comment are about one and the same thing, no?

If so, I think its fixed now

gbrail:
Thanks!

PR #899, opened 2021-05-19
Fix for pr 897

p-bakker:
1. -Dtest262properties option wasn't working
2. added -DoptLevel commandline option (and made it take prio over the TEST_262_OPTLEVEL env. var)


gbrail:
Sounds helpful -- thanks!

PR #901, opened 2021-05-20
ADD: NativeArray.subList()

rPraml:
This adds the missing NativeArray.subList() functionality

### DIFF
--- a/src/org/mozilla/javascript/NativeArray.java
+++ b/src/org/mozilla/javascript/NativeArray.java
@@ -9,6 +9,7 @@ package org.mozilla.javascript;
 import static org.mozilla.javascript.ScriptRuntimeES6.requireObjectCoercible;
 
 import java.io.Serializable;
+import java.util.AbstractList;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
@@ -2214,11 +2215,7 @@ public class NativeArray extends IdScriptableObject implements List {
 
     @Override
     public ListIterator listIterator(final int start) {
-        long longLen = length;
-        if (longLen > Integer.MAX_VALUE) {
-            throw new IllegalStateException();
-        }
-        final int len = (int) longLen;
+        final int len = size();
 
         if (start < 0 || start > len) {
             throw new IndexOutOfBoundsException("Index: " + start);
@@ -2333,7 +2330,23 @@ public class NativeArray extends IdScriptableObject implements List {
 
     @Override
     public List subList(int fromIndex, int toIndex) {
-        throw new UnsupportedOperationException();
+        if (fromIndex < 0) throw new IndexOutOfBoundsException("fromIndex = " + fromIndex);
+        if (toIndex > size()) throw new IndexOutOfBoundsException("toIndex = " + toIndex);
+        if (fromIndex > toIndex)
+            throw new IllegalArgumentException(
+                    "fromIndex(" + fromIndex + ") > toIndex(" + toIndex + ")");
+        return new AbstractList() {
+
+            @Override
+            public Object get(int index) {
+                return NativeArray.this.get(index + fromIndex);
+            }
+
+            @Override
+            public int size() {
+                return toIndex - fromIndex;
+            }
+        };
     }
 
     @Override
--- a/src/org/mozilla/javascript/NativeArray.java
+++ b/src/org/mozilla/javascript/NativeArray.java
@@ -2101,11 +2101,7 @@ public class NativeArray extends IdScriptableObject implements List {
 
     @Override
     public Object[] toArray(Object[] a) {
-        long longLen = length;
-        if (longLen > Integer.MAX_VALUE) {
-            throw new IllegalStateException();
-        }
-        int len = (int) longLen;
+        int len = size();
         Object[] array =
                 a.length >= len
                         ? a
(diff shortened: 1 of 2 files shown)


src/org/mozilla/javascript/NativeArray.java
```diff
@@ -2214,11 +2215,7 @@ public ListIterator listIterator() {
 
     @Override
     public ListIterator listIterator(final int start) {
-        long longLen = length;
-        if (longLen > Integer.MAX_VALUE) {
-            throw new IllegalStateException();
-        }
-        final int len = (int) longLen;
+        final int len = size();
```
rbri, line 2218: Looks like we have this pattern at more places, maybe we can remove the duplicated code at all places

toArray()
indexOf()
lastIndexOf()

And maybe we can add some text to the illegal state exception thrown in the size() method.

gbrail:
Can you address the comment from @rbri ? Other than that this seems fine, but I don't know enough about the guts of NativeArray to understand what this enables.

src/org/mozilla/javascript/NativeArray.java
```diff
@@ -2333,7 +2330,23 @@ public Object remove(int index) {
...
     @Override
     public List subList(int fromIndex, int toIndex) {
-        throw new UnsupportedOperationException();
+        if (fromIndex < 0) throw new IndexOutOfBoundsException("fromIndex = " + fromIndex);
+        if (toIndex > size()) throw new IndexOutOfBoundsException("toIndex = " + toIndex);
+        if (fromIndex > toIndex)
+            throw new IllegalArgumentException(
+                    "fromIndex(" + fromIndex + ") > toIndex(" + toIndex + ")");
+        return new AbstractList() {
+
```
tonygermano, line 2339: This is probably better than the UnsupportedOperationException we had before, but I don't think the List returned by subList is supposed to be immutable.

If we look at the ArrayList implementation, you can modify the original list through the subList, and if you make structural changes to the original list, subList operations fail-fast.
```js
js> al = new java.util.ArrayList([1,2,3,4,5])
[1.0, 2, 3, 4, 5]
js> sl = al.subList(0,5)
[1.0, 2, 3, 4, 5]
js> sl.remove(3)
4
js> al
[1.0, 2, 3, 5]
js> al.remove(3)
5
js> sl.get(0)
org.mozilla.javascript.WrappedException: Wrapped java.util.ConcurrentModificationException
  .
  .
  .
Caused by: java.util.ConcurrentModificationException
        at java.base/java.util.ArrayList$SubList.checkForComodification(ArrayList.java:1445)
        at java.base/java.util.ArrayList$SubList.get(ArrayList.java:1180)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
        at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.base/java.lang.reflect.Method.invoke(Method.java:566)
        at org.mozilla.javascript.MemberBox.invoke(MemberBox.java:208)
```

tonygermano: From the javadoc for [java.util.List.subList](https://docs.oracle.com/javase/8/docs/api/java/util/List.html#subList-int-int-):

> This method eliminates the need for explicit range operations (of the sort that commonly exist for arrays). Any operation that expects a list can be used as a range operation by passing a subList view instead of a whole list. For example, the following idiom removes a range of elements from a list:
> 
>       list.subList(from, to).clear();

That example would obviously not work with this implementation, because the `clear` method is not implemented.

tonygermano: You may want to refer to the implementation of ArrayList.SubList

https://github.com/openjdk/jdk/blob/jdk8-b120/jdk/src/share/classes/java/util/ArrayList.java#L995

src/org/mozilla/javascript/NativeArray.java
rPraml: Done: https://github.com/mozilla/rhino/pull/901/commits/e6ba6409ce0399c3ae27682fa4f4d365d975cb2f

src/org/mozilla/javascript/NativeArray.java
rPraml: I thought NativeArray is similar to Collections.unmodifiableList (at least when access through List interface).
So it is unmodifiable from the java perspective and
`nativeArray.sublist(x,y).clear()` will fail with an UOE (the same as `unmodifiableList .sublist(x,y).clear()`)

Nevertheless, you can get a sublist (or an iterator) in java and do a modification e.g. with `list.pop()` in javascript.

So I added a modCount to each place, where the length property is modified and throw a CME when the underlying list of a sublist/iterator is modified:

https://github.com/mozilla/rhino/pull/901/commits/895aa34fe82631e5765cef8c136b9dee081276ad

tonygermano: You are correct. I did not realize that NativeArray's List implementation did not implement any of the operations to modify the list. What I now realize is that the javadoc for this class is basically non-existent and doesn't explain any of the List behavior (an issue for another PR.)

I think your changes to throw CME are sufficient for subList and an improvement to the existing Iterator.

rPraml:
@gbrail can you take a look at this PR again? In my humble oppinion, all mentioned points should be solved now.

gbrail:
Yes, this looks good. Thanks for working on it!

PR #902, opened 2021-05-20
ADD: ES6 Object.values / Object.entries / Object.fromEntries

rPraml:
Adds the missing Object.values / Object.entries / Object.fromEntries methods.


tuchida:
After deleting these lines, try running `./gradlew test --tests Test262SuiteTest`.
https://github.com/mozilla/rhino/blob/b0fb5f8f775127cabbd8168bfe99d6ec3c2f309f/testsrc/test262.properties#L732-L771
https://github.com/mozilla/rhino/blob/b0fb5f8f775127cabbd8168bfe99d6ec3c2f309f/testsrc/test262.properties#L826-L841

rPraml:
I could enable some of the tests. I also applied spotless in an extra commit. (is it correct, that spotless has to be applied on each modified file?)
How should I proceed? I can try to see, if I can get more tests working next week.
Can this PR get merged and can/should I fix the tests in an extra commit?

tuchida:
The JavaScript specification is defined in ECMA262.
https://262.ecma-international.org/11.0/#sec-object.values

It is difficult to read ECMA262 right off the bat, so I recommend reading MDN first.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/values

rbri:
Regarding spotless - yes every modified file needs spotless applied. Greg merges as soon as he finds the time for reviewing. It will be great if you like to work further on this making more test passing.

gbrail:
Thanks for doing this!

Now that we're getting better at doing this stuff, I'd prefer if we could first figure out if we can implement a feature with as many of the test262 tests passing as possible before merging it.

Looking at the spec, Object.entries and Object.values are pretty straightforward and the implementation looks like it's probably pretty close. 

On the other hand, Object.fromEntries is one of those JS functions with a bunch of edge cases and things that have to be generalized. You might want to take a look at the implementation of Array.from, for example.

One way I've found to figure out what I'm missing is to start with the V8 "mjsunit" test suite, and comment out things that don't work because other parts of the spec aren't implemented yet. For example:

https://github.com/v8/v8/blob/dc712da548c7fb433caed56af9a021d964952728/test/mjsunit/harmony/object-fromentries.js

It's often more productive to get things working by running those tests over and over before trying to wade through test262.

What would you think of just trying to finish "entries" and "values" to pass as many tests as we can and then tackling "fromEntries" separately?

src/org/mozilla/javascript/NativeObject.java
```diff
@@ -368,6 +375,61 @@ public Object execIdCall(IdFunctionObject f, Context cx, Scriptable scope,
...
+
+          case ConstructorId_entries:
+              {
+                  Object arg = args.length < 1 ? Undefined.instance : args[0];
+                  Scriptable obj = getCompatibleObject(cx, scope, arg);
+                  Object[] ids = obj.getIds();
+                  for (int i = 0; i < ids.length; i++) {
+                      if (ids[i] instanceof Integer) {
+                          Integer key = (Integer) ids[i]; 
+                          ids[i] = cx.newArray(scope, new Object[]{ ids[i], obj.get(key, scope) });
```
tonygermano: I think the first element of the new Array should be a String, even when you need the integer index to do the lookup. Have you thought about implementing the [EnumerableOwnPropertyNames](https://tc39.es/ecma262/#sec-enumerableownpropertynames) abstract method and using it for keys, entries, and values?

rbri:
Looks like someone is waiting for this :-)

https://stackoverflow.com/questions/67627106/typeerror-cannot-find-function-entries-in-object-function-object

rPraml:
So, I got all of the "entries" and "values" test working.

Most of the "fromEntries" tests are using computed-properties like in this example:
```javascript
var iterable = {
  [Symbol.iterator]: function() {
  ...
  }
}
```
Unfortunately, they are currently not yet parseable. I gave it a try in this commit https://github.com/mozilla/rhino/commit/3756fa17012bfb46c9343d1f3181eec416bd3123 and extended the parser and codegenerators to handle computed properties.

Somehow it seems to work, but I need help here because I'm not sure why ;)
Maybe someone with more experience in parser/bytecode generation can grab this commit and continue my work.

gbrail:
The fact that we don't have support for new syntax is a fine reason to disable some test262 tests.

But if you have entries() and values() working I'll reiterate my suggestion that you put those in their own PR so that we can all look at it and possibly merge it.

rbri:
Why not also merge the fromEntries(). Most of the tests are passing - we have the same situation for other methods/tests.

rbri:
Btw: have done a test of this whole pr using it in HtmlUnit and it works great

rPraml (commented):
@gbrail @rbri 
I see, the "fromEntries" has a lot of edge cases and discovered other bugs or unimplemented features in rhino.
So I am also torn between leaving the code "as it ist" or remove "fromEntries" again.

Some of the test262 test cases I got working:

They require some special code, e.g. to check `Symbol.toPrimitive` or if "key" is a function, that it is invoked.
I'm not sure, if this is a speciality of "fromEntries" or should be handled at some other place.
 
The other test cases often use "computed properties", so either the tests have to be rewritten or computed property support has to be added. I've done this in a very experimental commit
https://github.com/FOCONIS/rhino/commit/98ef8154bc056589daecbcc98c47d0b45f2b1f4a

Are there plans, that computed-properties will be added? Maybe someone with more experience can take that commit as basis, as it will fix a lot of other testcases, too (See [test262.properties](https://github.com/FOCONIS/rhino/commit/98ef8154bc056589daecbcc98c47d0b45f2b1f4a#diff-fb67fedcbc562c1e228576ac029f72ebd7a87b15a767a9fcb2c67aed42b06ad6)

gbrail:
I'm sorry that I didn't realize this sooner, but according to the spec, the basic algorithm for pulling items from the iterator and loading the object is the same basic algorithm (in the spec) as the constructors for Map and WeakMap objects.

So, you can look at NativeMap's constructor, specifically the "loadFromIterable" function. You could possibly move that function somewhere else, like ScriptRuntime, and just use it -- it may do a lot of what you already need including closing the iterator.

src/org/mozilla/javascript/NativeMap.java
```diff
@@ -191,33 +187,20 @@ static void loadFromIterable(Context cx, Scriptable scope, ScriptableObject map,
         // Find the "add" function of our own prototype, since it might have
         // been replaced. Since we're not fully constructed yet, create a dummy instance
         // so that we can get our own prototype.
-        ScriptableObject dummy = ensureScriptableObject(cx.newObject(scope, map.getClassName()));
-        final Callable set =
-                ScriptRuntime.getPropFunctionAndThis(dummy.getPrototype(), "set", cx, scope);
+        Scriptable proto = ScriptableObject.getClassPrototype(scope, map.getClassName());
```
rPraml, line 190: These are the only code change in this class.
I have changed this to `getClassPrototype` instead of creating a dummy object and I reuse the code in 
`ScriptRuntime.loadFromIterable`

src/org/mozilla/javascript/NativeObject.java
```diff
@@ -224,11 +229,7 @@ public Object execIdCall(
 
                     if (arg instanceof Symbol) {
                         result = ((SymbolScriptable) thisObj).has((Symbol) arg, thisObj);
-                        if (result && thisObj instanceof ScriptableObject) {
-                            ScriptableObject so = (ScriptableObject) thisObj;
-                            int attrs = so.getAttributes((Symbol) arg);
-                            result = ((attrs & ScriptableObject.DONTENUM) == 0);
-                        }
+                        result = result && isEnumerable((Symbol) arg, thisObj);
```
rPraml, line 232: moved code to "isEnumerable" method

src/org/mozilla/javascript/ScriptRuntime.java
```diff
@@ -4293,22 +4338,30 @@ public static Scriptable newObjectLiteral(
...
-                } else {
-                    int index = ((Integer) id).intValue();
-                    object.put(index, object, value);
                 }
             } else {
                 ScriptableObject so = (ScriptableObject) object;
                 Callable getterOrSetter = (Callable) value;
                 boolean isSetter = getterSetter == 1;
+                // XXX: Do we have to handle Symbol here.
+                // This will be required, when conputedprops are supported.
```
rPraml, line 4364: Yes, we need to check this. See https://github.com/FOCONIS/rhino/commit/6d2a459e0d44f8efc59a0b1498da56beec1d2e9c
But I think this should be done in a separate PR

rPraml:
@gbrail Thanks for the hint. Maybe you can also take a look at this PR again.
It will fix some issues, when Symbol is used as key. But there are still some left, which I would do in a separate PR.

gbrail:
Thanks for cleaning this up. This looks good for now...

However, we had a whole bunch of new formatting changes and this requires another merge. Can you please try to merge with the latest master and resolve conflicts? Thanks!

rPraml:
@gbrail Conflicts are solved and build is passing again

rbri:
Thanks a lot for all the work on this and all the others

gbrail:
Thanks for working all that out!

PR #905, opened 2021-05-21
Nativeerror cleanup

rbri:
only a minor cleanup


gbrail:
Now that I made other changes in NativeError, do you still want this change? In general I have avoided using constants to stand in for property names (like "message" and "name") that are well-defined and part of the spec, but there may be other reasons why this is important.

rbri:
This is now obsolete with the reworking of NativError done by @gbrail .

Issue #284, opened 2016-09-07
new Error() has no .stack


ntrrgc:
This is a unit test from Rhino (https://github.com/mozilla/rhino/blob/master/testsrc/org/mozilla/javascript/tests/ErrorPropertiesTest.java#L66-L75):

```
@Test
public void mozillaStack() {
    RhinoException.useMozillaStackStyle(true);
    testScriptStackTrace("null.method()", "@myScript.js:1" + LS);
    final String script = "function f() \n{\n  null.method();\n}\nf();\n";
    testScriptStackTrace(script, "f()@myScript.js:3" + LS + "@myScript.js:5" + LS);
    testIt("try { null.method() } catch (e) { e.stack }", "@myScript.js:1" + LS);
    final String expectedStack = "f()@myScript.js:2" + LS + "@myScript.js:4" + LS;
    testIt("function f() {\n null.method(); \n}\n try { f() } catch (e) { e.stack }", expectedStack);
}
```

Currently it passes, but if you threw a constructed error like this, it would not:

```
testIt("try { throw new Error('app exception') } catch (e) { e.stack }", "@myScript.js:1" + LS);
```

The error output is this:

```
expected:<@myScript.js:1
> but was:<org.mozilla.javascript.Undefined@6ac6998d>
Expected :@myScript.js:1

Actual   :org.mozilla.javascript.Undefined@6ac6998d
```

I can reproduce the issue easily with minimal Java code, but it works fine in the CLI. Is this intended behavior? Why?

```
package com.example.test;

import org.mozilla.javascript.Context;
import org.mozilla.javascript.RhinoException;
import org.mozilla.javascript.Scriptable;

public class Main {
    public static void main(String[] args) {
        Context ctx = Context.enter();
        ctx.setGeneratingDebug(true);
        ctx.setOptimizationLevel(-1);
        Scriptable global = ctx.initStandardObjects();

        String stack = (String)ctx.evaluateString(global, "try{throw new Error('')} catch(e){ e.stack }", "code.js", 1, null);
        System.out.println(stack);
    }
}
```

shturec:
I'm not a contributor and cannot answer you question but in case you need a workaround here is what I found out. 

[...]

Issue #362, opened 2017-11-14
Make errors have stack traces by default


gbrail:
It confuses new users of Rhino that errors don't have stack traces enabled by default. This is controlled by the "FEATURE_LOCATION_INFORMATION_IN_ERROR" context "feature." Unless there is some major reason why this has a performance impact then we should change it.

PR #907, opened 2021-05-21
Rewrite NativeError to use lambda functions

gbrail:
This is a rewrite of the NativeError class to use the LambdaConstructor and LambdaFunction objects instead of IdScriptableObject. The resulting code is smaller, as you'll see, and I think easier to follow, but might not be that way for everyone.

The work is not done, so consider this a draft to use as an example if anyone else wants to try using this.

### DIFF
--- a/benchmarks/org/mozilla/javascript/benchmarks/BuiltinBenchmark.java
+++ b/benchmarks/org/mozilla/javascript/benchmarks/BuiltinBenchmark.java
@@ -359,17 +359,17 @@ public class BuiltinBenchmark {
                             "DumbLambdaClass",
                             0,
                             (Context cx, Scriptable s, Object[] args) -> new DumbLambdaClass());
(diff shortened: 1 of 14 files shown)


gbrail:
This will fix https://github.com/mozilla/rhino/issues/284 and https://github.com/mozilla/rhino/issues/362 because it always puts a stack on the error.

It also ensures that all errors have stack traces, even if "debug mode" is not on -- in that case, errors will not have file names, but they will still have stack traces that contain line numbers and function names for some stack frames.

My only concern with this change is that code that creates its own subclasses of NativeError in Java by extending the NativeError class may have to change -- I'd appreciate a look at this PR from anyone who might have that issue.

gbrail:
I am worried about breaking anyone who has subclassed NativeError with this change, so I'm shortly going to replace it with a smaller PR.

Issue #906, opened 2021-05-21
get doesnt work for indexed properties


p-bakker:
works:
```
var a = {get name(){return 'me'}}
a.name // yields me
```

fails:
```
var a = {get 0(){return 'me'}}
a[0] // yields the function
```

This bug hangs the Test262 suite when running against the master of the Test262 suite repo

tuchida:
ref. #775
Could this have anything to do with it?

p-bakker:
Indeed seems similar, albeit me getting back the getter function

p-bakker:
Bug seems located here: https://github.com/mozilla/rhino/blob/master/src/org/mozilla/javascript/ScriptRuntime.java#L4308

If the id is a String, there's logic for getter/setters, but if the if is an integer, the value is just added as a regular property.

Issue #775, opened 2020-10-02
Calling to getter/setter method incorrectly


NWU-NISL:
#### Description

As the following test case shown, when I get the attribute `2` of the object `v0` in the tenth line, Rhino returns the setter method instead of calling the getter method. According to ECMA262, getter method should be called here, so `v0[2]` should return `1` instead of the setter method.  

#### Version

```
1.7.13
```

#### Command

```
java -jar  rhino/rhino-1.7.13.jar -debug -version 200 testcase.js
```

#### testcase

~~~javascript
function f() {
    var v0 = {
        get 2(){
            return 1;
        },
        set 2(value){
            throw new Error(2);
        }
    }
    var v1 = v0[2];
    print(v1);
}
f()
~~~

#### Output

~~~javascript
function (value) {
    throw new Error(2);
}

~~~

#### Expected output

~~~javascript
1
~~~

contributor:@QuXing9

PR #908, opened 2021-05-23
fix support indexed accessor properties in Object Literals

p-bakker:
Adds support for indexed accessor properties.

Without this, running the the current master of Test262 hangs on https://github.com/tc39/test262/blob/main/test/built-ins/Array/prototype/concat/arg-length-near-integer-limit.js, as the getter for index 0 is never called

Closes https://github.com/mozilla/rhino/issues/906
Closes https://github.com/mozilla/rhino/issues/775


gbrail:
Thanks for the timely fix!

This method has a ton of undocumented prerequisites that predate your fix -- it assumes that the index is either a String or an Integer,  for example. I think that's fine since it's tightly embedded with the interpreter and code generation.

PR #912, opened 2021-05-25
Fix bug with calling function on custom global object from function defined within a function within a module

youngj:
Rhino makes it possible to execute a script with a custom object as the root scope, which makes it easy to add custom functions at the top level. Inside a CommonJS module, these custom functions are also available at the root scope. However, when inside a CommonJS module, if the custom global function is called from a function defined within another function, Rhino unexpectedly throws the error:

```
org.mozilla.javascript.EcmaError: TypeError: Method "test" called on incompatible object.
```

In the test code below, the function `test(x)` is added at the root scope by defining the method `CustomGlobal.jsFunction_test` and setting a CustomGlobal instance as the root scope. In the current version of Rhino, it works everywhere, except when called within a module from a function defined within another function, like this:

```js
var testWrap2 = function() { return function(arg) { return test(arg); }; }
testWrap2()(6);
```

Removing the condition `if (thisObj == scope)` from FunctionObject.call fixes the bug.

When calling `testWrap2()(6);` within a module, the variables in FunctionObject.call have the following types:

* thisObj: org.mozilla.javascript.commonjs.module.ModuleScope
* scope: org.mozilla.javascript.NativeCall
* parentScope: CustomGlobal

Because `thisObj != scope` within a function defined inside another function, the previous code in FunctionObject.call did not try to invoke "test" on the CustomGlobal object returned by `getParentScope()`.

I'm not sure if there is a reason for the `thisObj != scope` condition, but removing the check for `thisObj != scope` doesn't cause any tests to fail.

Spotless wants to make a lot of changes to both of these Java files, but I didn't include these changes in this PR to improve readability.

### DIFF
--- a/testsrc/org/mozilla/javascript/tests/commonjs/module/testCustomGlobal.js
+++ b/testsrc/org/mozilla/javascript/tests/commonjs/module/testCustomGlobal.js
@@ -1,0 +1,13 @@
+var assert = require("assert");
+
+assert.strictEqual(test(1), 2);
+
+var testWrap = function(arg) { return test(arg); };
+
+assert.strictEqual(testWrap(2), 3);
+
+var testWrap2 = function() { return function(arg) { return test(arg); }; };
+
+assert.strictEqual(testWrap2()(3), 4);
+
+require('testCustomGlobal2');
--- a/testsrc/org/mozilla/javascript/tests/commonjs/module/testCustomGlobal.js
+++ b/testsrc/org/mozilla/javascript/tests/commonjs/module/testCustomGlobal.js
@@ -11,3 +11,14 @@ var testWrap2 = function() { return function(arg) { return test(arg); }; };
 assert.strictEqual(testWrap2()(3), 4);
 
 require('testCustomGlobal2');
+
+var throws = false;
+try {
+    test.call({}, "oops");
+} catch (e) {
+    throws = true;
+    assert.strictEqual("Method \"test\" called on incompatible object.", e.message);
+}
+if (!throws) {
+    assert.fail("test() did not throw error when called on invalid object");
+}
--- a/src/org/mozilla/javascript/FunctionObject.java
+++ b/src/org/mozilla/javascript/FunctionObject.java
@@ -14,6 +14,7 @@ import java.lang.reflect.Constructor;
 import java.lang.reflect.Member;
 import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
+import org.mozilla.javascript.commonjs.module.ModuleScope;
 
 public class FunctionObject extends BaseFunction {
     private static final long serialVersionUID = -5332312783643935019L;
@@ -353,13 +354,15 @@ public class FunctionObject extends BaseFunction {
                 Class<?> clazz = member.getDeclaringClass();
                 if (!clazz.isInstance(thisObj)) {
                     boolean compatible = false;
-                    Scriptable parentScope = getParentScope();
-                    if (scope != parentScope) {
-                        // Call with dynamic scope for standalone function,
-                        // use parentScope as thisObj
-                        compatible = clazz.isInstance(parentScope);
-                        if (compatible) {
-                            thisObj = parentScope;
+                    if (thisObj == scope || thisObj instanceof ModuleScope) {
+                        Scriptable parentScope = getParentScope();
+                        if (scope != parentScope) {
+                            // Call with dynamic scope for standalone function,
+                            // use parentScope as thisObj
+                            compatible = clazz.isInstance(parentScope);
+                            if (compatible) {
+                                thisObj = parentScope;
+                            }
                         }
                     }
                     if (!compatible) {
--- a/src/org/mozilla/javascript/FunctionObject.java
+++ b/src/org/mozilla/javascript/FunctionObject.java
@@ -353,15 +353,13 @@ public class FunctionObject extends BaseFunction {
                 Class<?> clazz = member.getDeclaringClass();
                 if (!clazz.isInstance(thisObj)) {
                     boolean compatible = false;
-                    if (thisObj == scope) {
-                        Scriptable parentScope = getParentScope();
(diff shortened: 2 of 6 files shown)


gbrail:
I just looked at the code coverage for FunctionObject before your change, and that whole section that you changed isn't covered by any of the Rhino tests. (I didn't look with your change yet, but I can imagine that you might want to run "./gradlew jacocoTestReport" yourself and look.

So that means that either no one ever tested it and it never worked until you figured it out (the most likely scenario), or there is some weird bit of code out there that depends on Rhino and the old broke behavior.

So I'd say that this is a low-risk change and it looks good to me, but let's leave this PR open for a day or two in case there are people out there who are more familiar with this use case who might have an opinion.

youngj:
Sounds good to me. Here is the jacocoTestReport output: 

image

Should I submit a separate PR with the result of running spotlessApply on FunctionObject.java and RequireTest.java?

rbri:
Will do a quick test with the HtmlUnit tests tomorrow and inform about the results.

Am 26. Mai 2021 18:34:16 MESZ schrieb Greg Brail ***@***.***>:
>I just looked at the code coverage for FunctionObject before your
>change, and that whole section that you changed isn't covered by any of
>the Rhino tests. (I didn't look with your change yet, but I can imagine
>that you might want to run "./gradlew jacocoTestReport" yourself and
>look.
>
>So that means that either no one ever tested it and it never worked
>until you figured it out (the most likely scenario), or there is some
>weird bit of code out there that depends on Rhino and the old broke
>behavior.
>
>So I'd say that this is a low-risk change and it looks good to me, but
>let's leave this PR open for a day or two in case there are people out
>there who are more familiar with this use case who might have an
>opinion.
>
>-- 
>You are receiving this because you are subscribed to this thread.
>https://github.com/mozilla/rhino/pull/912#issuecomment-848924103

-- 
Diese Nachricht wurde von meinem Android-Gerät mit K-9 Mail gesendet.

gbrail:
Thanks!

Can you add a test case that will result in "msg.incompat.call" being thrown?

youngj:
msg.incompat.call would normally be triggered if the function is called with an incompatible `this`, e.g:

```
test.call({}, "oops");
test.call(null, "oops");
```

However, the change to remove `if (thisObj == scope)` also causes the above calls to succeed instead of fail.

If the above calls should still fail, perhaps then the test should be `if (thisObj == scope || thisObj instanceof ModuleScope)` ?

gbrail:
It sounds like there are indeed cases in which a TypeError should be thrown, such as the ones that you mentioned like an empty object, null, or undefined. I'm just saying that we should define those places and have a test case for them so that we have code coverage in that case. For instance, perhaps somewhere in that set of code we need to call something like ScriptRuntime.toObject, which does some other conversions.

youngj:
Added test cases to test "msg.incompat.call" being thrown.

When testing, I encountered an issue where the test results displayed incorrect error messages when assertion errors were thrown. 

When an assertion error was triggered, the actual exception thrown was `org.mozilla.javascript.EcmaError: ReferenceError: "__filename" is not defined.` from assert.js . This occurred because typeof __filename is `"undefined"` (a string) rather than `undefined` (the undefined value). It appeared that this code was not actually necessary, so I removed it from assert.js.

However, after fixing the __filename issue, when an assertion error was thrown and written to buildGradle/reports/tests/test/index.html , it displayed "org.mozilla.javascript.JavaScriptException: [object Object]" instead of the value returned by assert.AssertionError.prototype.toString .

Although JavaScriptException.details tries to call toString on the AssertionError object, ScriptRuntime.toString was throwing an exception because the Gradle exception messages are displayed in a different thread than the thread associated with the JavaScript Context. To fix this issue, I updated the JavaScriptException class to store the exception message when the exception is created, so that the exception can be printed correctly even if it is later printed from a different thread.

Even after fixing that issue, the assert.AssertionError.prototype.toString method defined in assert.js was still not being called because assert.js set assert.AssertionError.prototype after setting assert.AssertionError.prototype.toString . Changing the order to set assert.AssertionError.prototype  before setting assert.AssertionError.prototype.toString fixed that issue, so if the tests throw an AssertionError it will be displayed correctly (e.g. calling assert.strictEqual(1,2) will show "org.mozilla.javascript.JavaScriptException: AssertionError: 2 === 1" in the test output).

testsrc/org/mozilla/javascript/tests/commonjs/module/testCustomGlobal.js
```diff
@@ -0,0 +1,24 @@
...
+
+assert.strictEqual(testWrap(2), 3);
+
+var testWrap2 = function() { return function(arg) { return test(arg); }; };
+
+assert.strictEqual(testWrap2()(3), 4);
+
+require('testCustomGlobal2');
+
+var throws = false;
```
gbrail, line 15: I'm not sure if this helps, but did you try using "assertThrows" here? It ensures that a function throws an exception and it lets you check the assertion's type and message.

gbrail (commented):
I'm not sure about your other issue yet, but I wonder if "assertThrows" would make it easier to write those test cases.

testsrc/org/mozilla/javascript/tests/commonjs/module/testCustomGlobal.js
youngj: I tried using `assert["throws"]`, and it works, but it only checks the assertion's type. The message parameter is used as the error message if the error is not thrown. I implemented it this way so that I could check the assertion's message.

gbrail: You're right -- assertThrows has a third argument that's purportedly trying to test the message, but it seems to predate all modern JavaScript and doesn't actually do that. I think that we should fix it to check the "message" property of the error object.

But I don't think it's worth holding up this PR for that reason, and otherwise this looks good. Thanks!

Issue #878, opened 2021-05-02
Fix or delete idswitch


tuchida:
ref. #863
Use `./gradlew spotlessApply`.

before:
```java
    private static final int
            Id_constructor = 1,
```
after:
```java
    private static final int Id_constructor = 1,
```
In the case of after:, `idswitch` does not seem to find the `Id_constructor`.
https://github.com/mozilla/rhino/tree/0c05d30e134aed65b9c3536d8ac3764590840a35/toolsrc/org/mozilla/javascript/tools/idswitch

ref. #847
Since `findPrototypeId` is now a string switch, I don't see much advantage in using `idswitch` in the future, should I modify it?

rbri:
For me my last hack was only to generate the new code. I think we can drop the idswitch tool and the generator hints from the code.

rbri:
"Deleted code is debugged code." – Jeff Sickel

PR #914, opened 2021-05-29
Delete idswitch

tuchida:
ref.  #878
Removed idswitch because it does not support spotless, and also because it is now using string switch from #847.


rbri:
If my memories are correct we can use in most of the cases a pattern like

    case "constructor":
        return Id_constructor;

instead of assigning a var and returning later.
There are only two or there classes where this simplification does not work.

gbrail:
How do others watching this PR feel about all this reformatting? This PR basically reformats every file that had an idswitch tag in it.

I'm actually OK with it but I wanted to make sure that it won't impact anyone else who is on the verge of having to merge something big.

rbri:
Ok for me, at the moment i have a hard time with all the merges with my rhino fork. Will be great if the reformat ends soon - so this brings me closer to the light at the end of the tunnel.

gbrail:
Just looking at the queue of incoming PRs, there's going to be a lot of merging going on if we import such a large set of formatting changes at once. I'd rather just get rid of the idswitch tool now, and then remove the annotations more gradually as we change other things.

gbrail:
Ok, I'm sold on ripping off the Band-Aid sooner in this case. Thanks for all the work to pull this together!

Issue #918, opened 2021-06-01
Spotless on Windows formats with the wrong line endings 


p-bakker:
.,/gradlew :spottlessApply looks at .gitattributes for determining the proper line endings, which is configured with `* text=auto`

In Windows this results in Spotless changing all line endings to CRLF, instead of leaving them on LF.

Would like to change the setup  in .gitattributes to `* text=auto eol=lf`. With that setting, Spotless leaves the LF line endings in place

Agree?

PR #919, opened 2021-06-01
force LF as newLine character

p-bakker:
spotlessApply would otherwise convert all newLine characters to CRLF on
windows

Closes: #918

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,5 +1,5 @@
 # Set default behavior, in case users don't have core.autocrlf set.
-* text=auto
+* text=auto eol=lf
 
 .classpath text
 .project text


gbrail:
This is of course a fine idea but I am worried since we have files in the test directories (and some test262 files) with CRLF line endings. What happens on Windows when one checks those out with this flag set? And do all the tests run on Windows?

I don't have Windows set up to test this, unless it's using Ubuntu on Windows which I assume still uses LF.

p-bakker:
@gbrail  I'll get back to you on that one

tonygermano:
Would it make sense to only set `eol=lf` for `*.java` files?

gbrail:
Definitely -- that's lower risk I think.

p-bakker:
Surpassed by #1584

PR #920, opened 2021-06-01
Remove test that checks whether toUpperCase returns the same character

p-bakker:
as lowerCase character if Unicode defines that there's no uppercase
variant

The tests break on Windows, because windows uses the latest Unicode
standard, while Unix doesn't and in Unicode 11 some Georgian characters
that were previously only available in lowercase have now received an
uppercase variant

So, this test would either fail on Windows or if fixed (because
EcmaScript states to use the latest UniCode standard) would break on
Linux.

That said, the test basically tests whether the OS on which Rhino runs
is providing the proper Unicode data to Java and whether Java handles
this correctly. so it's not really testing anythign specific in the
Rhino implementation


gbrail:
I think it's fine to delete this test, since as you point out it just tests that Java works.

Issue #243, opened 2015-09-19
Implement template strings


gbrail:
The Node.js 4.0 developers seem to love these.

tuchida:
ref. https://github.com/mozilla/rhino/pull/81
What would you do to @anba's PR?

gbrail:
Yeah, I'm trying to merge that and it does not look terribly difficult. Of course, it's not a simple merge either given that it goes in the same few files where everything else has changed. But let me see if I can make progress.

p-bakker:
@gbrail I've merged the Patch 2 & 3 part of #81 onto the latest master in my fork: https://github.com/p-bakker/rhino/tree/template-strings

I ran the latest tests from String.raw from test262 main's branch against it and the ones that fail seem to be related to tagged templates, see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#Tagged_template_literals

Am trying to grasp the changes @anba made, but so far I think that Tagged Template support just wasn't covered. Might not have been part of the draft spec at the time he implemented these changes

gausie:
@p-bakker I know its been a while; do you have any intention on resuming your work on this? If not would you be alright with someone picking up from where you left off?

p-bakker:
Hi @gausie 

It's indeed been a while, priorities shifted, will be a few months before I'd be able to get back to it.

I managed to get it to pass all tests from the Test262 testsuite, not sure if I committed all the changes I made to my fork yet. Also don't recall from the top of my head the exact reason why I haven't put it up as a PR yet. 

Would need to spent some time on it to see where things stand

gausie:
It looks so close to being done and a project I'm involved with would make massive use of the feature, so I'm looking forward to seeing a PR when you're ready! Thanks for your work so far.

PR #922, opened 2021-06-03
Extract the Object.seal/freeze/isSealed/isFrozen logic, fixing a few bugs in the process

p-bakker:
For https://github.com/mozilla/rhino/issues/243 the logic to freeze an object needs to be accessible from Java code

This PR extracts the internal logic of Object.seal/freeze/isSealed/isFrozen to the abstract object operations testIntegrityLevel & setIntegrityLevel as defined by EcmaScript, as on that level the logic needs to be reused in Tagged Template Literals

In the process, the following issues were encountered and fixed:
- Three of the aforementioned 4 methods of Object were not taking Symbols into account
- getOwnPropertyDescriptor for length property of array didn't properly represent its actual state (due to using a cached attributes value that wasn't being kept up to date)
- proper handling of attempt to update length property on frozen array: again due to stale cached attributes value
- IdScriptableObject overwrote `.defineOwnProperty(Context cx, Object key, ScriptableObject desc)` instead of `.defineOwnProperty(Context cx, Object key, ScriptableObject desc, boolean checkValid)`, causing it's logic to never be triggers when Rhino code calls the later overload directly

### DIFF
--- a/src/org/mozilla/javascript/AbstractEcmaObjectOperations.java
+++ b/src/org/mozilla/javascript/AbstractEcmaObjectOperations.java
@@ -1,0 +1,114 @@
+package org.mozilla.javascript;
+
+/**
+ * Abstract Object Operations as defined by EcmaScript
+ * 
+ * @see <a href="https://262.ecma-international.org/11.0/#sec-operations-on-objects">Abstract Operations - Operations on Objects</a>
+ */
+class AbstractEcmaObjectOperations {
+    static enum INTEGRITY_LEVEL {
+        FROZEN,
+        SEALED
+    }
+
+    /**
+     * Implementation of Abstract Object operation testIntegrityLevel as defined by EcmaScript
+     * 
+     * @param o
+     * @param level
+     * 
+     * @return boolean
+     * 
+     * @see <a href="https://262.ecma-international.org/11.0/#sec-testintegritylevel">TestIntegrityLevel</a>
+     */
+    static boolean testIntegrityLevel (Object o, INTEGRITY_LEVEL level ) {
+        ScriptableObject obj = ScriptableObject.ensureScriptableObject(o);
+        Context cx = Context.getCurrentContext();
+
+        if (obj.isExtensible()) return Boolean.FALSE;
+
+        for (Object name : obj.getIds(true, true)) {
+            ScriptableObject desc = obj.getOwnPropertyDescriptor(cx, name);
+            if (Boolean.TRUE.equals(desc.get("configurable"))) return Boolean.FALSE;
+            
+            if (level == INTEGRITY_LEVEL.FROZEN && desc.isDataDescriptor(desc) && Boolean.TRUE.equals(desc.get("writable")))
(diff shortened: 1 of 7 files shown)


src/org/mozilla/javascript/AbstractEcmaObjectOperations.java
```diff
@@ -0,0 +1,111 @@
...
+     *
+     * @param o
+     * @param level
+     * @return boolean
+     * @see <a
+     *     href="https://262.ecma-international.org/11.0/#sec-testintegritylevel">TestIntegrityLevel</a>
+     */
+    static boolean testIntegrityLevel(Object o, INTEGRITY_LEVEL level) {
+        ScriptableObject obj = ScriptableObject.ensureScriptableObject(o);
+        Context cx = Context.getCurrentContext();
```
gbrail, line 26: It looks like this function is called from various "execIdCall" methods, which means that the Context should already be in the parameter list. I'd prefer if this method took Context as the first parameter too. That saves us from figuring out what to do in the unlikely event that the Context is not properly on the thread (which is incorrect but could still happen) and it also saves us from having to look up the context from the thread slot where it's stored, which is I assume a more expensive operation than having it passed as a parameter.

src/org/mozilla/javascript/AbstractEcmaObjectOperations.java
```diff
@@ -0,0 +1,111 @@
...
+
+        if (obj.isExtensible()) return Boolean.FALSE;
```
gbrail, line 28: Is there a reason not to just return "true" and "false" from this function, since it returns a boolean anyway?

src/org/mozilla/javascript/AbstractEcmaObjectOperations.java
```diff
@@ -0,0 +1,111 @@
...
+                obj.defineOwnProperty(cx, key, desc, false);
+            }
+        }
+
+        obj.preventExtensions();
+
+        return true;
+    }
+
+    static boolean definePropertyOrThrow(Object o, Object p, Scriptable desc) {
```
gbrail, line 108: Is this just a stub for later? Should we leave it out then?

gbrail (commented):
I have a few minor-ish things I'd love to have you look at, but otherwise thanks for moving this forward.

src/org/mozilla/javascript/AbstractEcmaObjectOperations.java
p-bakker: That was there already in the old code, but didn't look like it's needed. Will fix

src/org/mozilla/javascript/AbstractEcmaObjectOperations.java
p-bakker: That's just me being sloppy, will remove

p-bakker:
From my POV this branch is now ready to merge

gbrail:
LGTM for me too. Thanks!

Issue #924, opened 2021-06-05
Slot map refactoring introduces a problem with updates of getters/setters


rbri:
While working on merging the last refactoring of the slot map i found an issue.

Object.defineProperty - allows to only redefine the getter OR the setter. In the case of existing properties this really only updates one. Currently when updating the getter only the setter is removed afterwards.

Have written some test passing with the code before the change. Because i'm a bit in a hurry i have no time to write more detailed tests for now - but it looks to me for the browser engines the exitsting getters/setters are still intact - means the function printout still hast the get/set in the name. eg.

     function set onreadystatechange() { [native code] }

@gbrail will be great if you can have a deeper look. I think the problem is because we now have MemberBoxSlot and Function slot - both having getter and setter.

Will be not in front of my pc for the whole next week - sorry.

gbrail:
Thanks for catching this!

With regular Java function getters and setters, it seems to work like it did before -- redefining only the getter or setter does not replace the last one. I'm not sure that we had that right for Java functions as getters and setters but that's easy to fix.

What you test exercises is the ability to have the getter set to a Java function and the setter set to a JavaScript function, or vice versa. With the refactoring that isn't possible. It's fixable, but assuming that your test case worked before, we'll have to work on that.

rbri:
Yes that was the idea of the test case - you can run it with the code before the refactoring and it will pass. This case is not that uncommon because HtmlUnit defines a lot of this java function based getter/setters.
-- 
Diese Nachricht wurde von meinem Android-Gerät mit K-9 Mail gesendet.

rbri:
In between spend some time today to test this with HtmlUnit - works like a charm.
Thanks

gbrail:
Fixed in that above PR.

PR #925, opened 2021-06-05
Tests for redefning only getter or setter

rbri:
Test for #924


gbrail:
I included this commit in https://github.com/mozilla/rhino/pull/928, along with a fix. I'm going to close this PR since I would like to merge that one instead.

rbri:
Thanks - as always.

PR #926, opened 2021-06-09
run spotlessApply on some files

youngj:
I'd like to submit other PRs that modify these files, so I'm submitting a PR that only includes reformatting changes to these files.


gbrail:
Sure -- thanks!

PR #927, opened 2021-06-10
Fix NPE when regexp search has no result

gausie:
Not sure if I need to do anything tests-wise for this (and indeed, how to do that) but I ran into this NPE when adding rhino's compatibility data to `core-js` (https://github.com/zloirock/core-js/pull/942)

### DIFF
--- a/testsrc/org/mozilla/javascript/tests/es6/NativeRegExpTest.java
+++ b/testsrc/org/mozilla/javascript/tests/es6/NativeRegExpTest.java
@@ -15,9 +15,7 @@ import org.mozilla.javascript.EcmaError;
 import org.mozilla.javascript.ScriptableObject;
 import org.mozilla.javascript.tests.Utils;
 
-/**
- * @author Ronald Brill
- */
+/** @author Ronald Brill */
 public class NativeRegExpTest {
 
     @Test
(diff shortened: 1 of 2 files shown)


rbri:
It will be great if you can add a test method that shows the problem to rhino/testsrc/org/mozilla/javascript/tests/es6/NativeRegExpTest.java

gausie:
Where do I add that test? I'm struggling to navigate the testing setup

rbri:
I guess you can use the last method in the class I mentioned in my previous comment as a template

gausie:
@rbri ah, I misread your first message and thought that you were linking me to the source file not the test! Thank you and done - the attached test fails without my change,

rbri:
Many thanks

testsrc/org/mozilla/javascript/tests/es6/NativeRegExpTest.java
```diff
@@ -148,16 +131,41 @@ public void regExIsNotCallable() {
...
+                    assertEquals(
+                            "Cannot modify readonly property: lastIndex.",
+                            Context.toString(result));
+                    return null;
+                });
+    }
+
+    @Test
+    public void search() {
+        Context cx = Context.enter();
```
gbrail, line 153: Can you please turn this into a try..catch with resources, like this:

try (Context cx = Context.enter()) { // Do the tests }

Otherwise if your tests fail or throw an exception then the context leaks and that can create problems for other tests -- we already have to run all our tests in separate JVMs because some tests leak contexts and it'd be really nice to be able to stop doing that.

gbrail (changes requested):
Thanks -- looks mostly great, but I'd like you to look at the handling of the Context object in the text -- see the comment below.

testsrc/org/mozilla/javascript/tests/es6/NativeRegExpTest.java
rbri: Will be great if you can do this improvement for all tests in this file.

gausie: Do I still need to do `Context.exit()` when using try-with-resources here?

gausie: Looks like I do not, great!

gbrail:
Thanks for doing this work. I ended up merging and squashing the other version of this PR because both are basically the same. However I just realized that your name doesn't show up in the commit history directly (although it's in a comment). I'll check more carefully next time -- I want to make sure that we have a diversity of contributors and that people can use work on this project on their "GitHub resume" so let us know if you have other changes to make!

PR #928, opened 2021-06-14
Fix Slot class refactoring

gbrail:
Fix the slot maps so that a property can have a Java getter and a JavaScript setter, or vice versa.

Also, introduce a "Lambda property," which is a property implemented using a pair of lambda functions.


gbrail:
This fixes https://github.com/mozilla/rhino/issues/924.

rbri (approved):
LoL did the same today but your are faster :-)

src/org/mozilla/javascript/GetterSlot.java
```diff
@@ -0,0 +1,194 @@
+package org.mozilla.javascript;
+
+/**
+ * This is a specialization of Slot to store various types of values that are retrieved dynamically
+ * using Java and JavaScript functions. Unlike LambdaSlot, the fact that these values are accessed
+ * and mutated by functions is visible via the slot's property descriptor.
+ */
+public class GetterSlot extends Slot {
```
rbri, line 8: Maybe AccessorSlot?

src/org/mozilla/javascript/GetterSlot.java
```diff
@@ -0,0 +1,194 @@
...
+
+    @Override
+    Function getGetterFunction(String name, Scriptable scope) {
+        if (getter == null) {
+            return null;
+        }
+        return getter.asGetterFunction(name, scope);
+    }
+
+    abstract static class Getter {
```
rbri, line 86: Why not an interface?

src/org/mozilla/javascript/GetterSlot.java
```diff
@@ -0,0 +1,194 @@
...
+        final Object target;
+
+        FunctionGetter(Object target) {
+            this.target = target;
+        }
+
+        Object getValue(Scriptable start) {
+            if (target instanceof Function) {
+                Function t = (Function) target;
+                Context cx = Context.getContext();
```
rbri, line 124: Any chance to pass the context from the call tree?

src/org/mozilla/javascript/ScriptableObject.java
```diff
@@ -541,19 +543,19 @@ public void setGetterOrSetter(
         if (name != null && index != 0) throw new IllegalArgumentException(name);
         checkNotSealed(name, index);
 
-        FunctionSlot fslot;
+        GetterSlot fslot;
```
rbri, line 546: Please use a 'matching' var name

src/org/mozilla/javascript/ScriptableObject.java
```diff
@@ -1561,22 +1577,22 @@ protected void defineOwnProperty(
         }
 
         if (isAccessor) {
-            FunctionSlot fslot;
+            GetterSlot fslot;
```
rbri, line 1580: var name again

src/org/mozilla/javascript/GetterSlot.java
tonygermano, line 8: Would AccessorSlot be a more appropriate name for this class?

rbri (approved):
Thanks for fixing that.

src/org/mozilla/javascript/ScriptableObject.java
```diff
@@ -1597,6 +1613,37 @@ protected void defineOwnProperty(
...
+     *
+     * @param name the name of the function
+     * @param getter a function that returns the value of the property. If null, then the previous
+     *     version of the value will be returned.
+     * @param setter a function that sets the value of the property. If null, then the value will be
+     *     set directly and may not be retrieved by the getter.
+     * @param attributes the attributes to set on the property
+     */
+    public void defineOwnProperty(
+            String name, Supplier<Object> getter, Consumer<Object> setter, int attributes) {
```
tonygermano, line 1631: Do you need to have a `checkValid` flag like the `defineOwnProperty(Context cx, Object id, ScriptableObject desc, boolean checkValid)` method or otherwise determine if you are allowed to replace a property that already exists?

src/org/mozilla/javascript/GetterSlot.java
gbrail: Yeah, we had that name before but AccessorSlot is better.

src/org/mozilla/javascript/GetterSlot.java
gbrail: Sure!

src/org/mozilla/javascript/GetterSlot.java
gbrail: Sadly this gets called from Scriptable.get, which has no Context object, and it's been this way for a very long time. I think that changing this now would be pretty huge.

src/org/mozilla/javascript/ScriptableObject.java
gbrail: Ok.

src/org/mozilla/javascript/ScriptableObject.java
gbrail: Ok.

src/org/mozilla/javascript/GetterSlot.java
gbrail: Yes, see above.

src/org/mozilla/javascript/ScriptableObject.java
gbrail: I actually renamed this to "defineProperty" to better match the existing "defineProperty" mechanism, which is for Java reflection. That function doesn't do any checking either. I think that since both of these functions are really intended for use in Java code to implement things natively in Java, it's OK to forgo the spec-compliant checking.

Also, the only way we have to check now requires us to create a property descriptor and do a lot of complex comparisions -- a more efficient way to do that would be cool.

Issue #153, opened 2015-02-26
Generate stack trace on the Error constructor


gbrail:
Today in Rhino, the "stack" property of an Error is generated when the Error is thrown, and derived from the Java stack trace.

In V8, however, the "stack" property is populated when the Error is constructed.

In other words, the code below:

var e = new Error();
throw e;

In V8, the line number on the stack trace will be the line where "new Error" was called. In Rhino, it is where "throw" was called.

Similarly, after the first line above, the "stack" property in V8 will show the stack trace, whereas in Rhino it will be undefined.

However, making this change will potentially slow down exception-handling code because the stack trace will be gathered on every error, and then Java will gather the stack trace again when it is thrown.

So, I propose adding a new feature flag to Context that is off by default. When the feature is enabled, Rhino will capture the stack trace when the Error object is first constructed.

gbrail:
Thinking of making this change for V8 compatibility -- any comments would be appreciated.

oberhamsi:
This would be quite a change. RingoJs currently relies on the old behaviour, e.g. our unit testing would break https://github.com/ringo/ringojs/blob/master/modules/test.js#L455

It seems most browsers agree with v8 and add "stack" when constructing the Error http://wiki.ecmascript.org/doku.php?id=strawman:error_stack

Please at least hide it behind a flag. Do you need it for tool support?

hrj:
The stack property is not defined even after throwing the error.

```
$ java -jar js.jar
Rhino 1.7.7 2015 06 17
js> try {throw new Error() } catch (e) {print(e.stack)}
undefined
```

----
Apart from that, it seems like IE11 also behaves similar to your description. I found this code in web-platform tests:
```js
        var stack = new Error().stack;
        // IE11 does not initialize 'Error.stack' until the object is thrown.
        if (!stack) {
            try {
                throw new Error();
            } catch (e) {
                stack = e.stack;
            }
        }
```

gbrail:
Ah -- Rhino does not generate stack traces at all unless "debug" is
enabled. This is historical -- I have no idea why it is not the default but
perhaps 10 years ago it was a big performance drain (seriously)...

$ java -jar buildGradle/libs/rhino-1.7.8-SNAPSHOT.jar  -debug
Rhino 1.7.8-SNAPSHOT 2015 07 02
js> try {throw new Error() } catch (e) {print(e.stack)}
at <stdin>:2



[...]

PR #929, opened 2021-06-15
Improve stack handling for errors

gbrail:
Make a number of improvements to the Error class that have been requested for a long time:

1) All "new Error" instances have a "stack" even if they are not thrown
2) All errors have a stack trace, even when running in compiled mode (opt level >= 0) without "debug" on. When "debug" is not enabled, then file names will not appear and will be replaced with "(unknown)". However, line numbers and function names will still be present in the stack trace.
3) Fix a number of small spec compliance issues with all the Error classes
4) As a prereq for the above, make sure that the "name" and "length" properties of BaseFunction can have their attributes be modified and be deleted if so configured
5) Replace some ugly reflection-based code in NativeError with the new Lambda-based property capability


gbrail:
This fixes https://github.com/mozilla/rhino/issues/362 and https://github.com/mozilla/rhino/issues/284 and https://github.com/mozilla/rhino/issues/153

src/org/mozilla/javascript/BaseFunction.java
```diff
@@ -105,16 +105,20 @@ protected int getMaxInstanceId() {
...
     @Override
     protected int findInstanceIdInfo(String s) {
-        int id;
+        int id = 0;
         switch (s) {
             case "length":
-                id = Id_length;
+                if (lengthPropertyAttributes >= 0) {
+                    id = Id_length;
+                }
```
rbri, line 113: Because we have dropped the generator, we can think about replacing the whole ID assignment with direct return statements.

src/org/mozilla/javascript/BaseFunction.java
```diff
@@ -206,8 +215,15 @@ protected void setInstanceIdValue(int id, Object value) {
                 }
                 return;
             case Id_name:
+                if (value == NOT_FOUND) {
+                    namePropertyAttributes = -1;
+                }
+                return;
             case Id_arity:
             case Id_length:
+                if (value == NOT_FOUND) {
```
rbri, line 224: For amity also?

src/org/mozilla/javascript/NativeError.java
```diff
@@ -234,13 +223,13 @@ private Object callPrepareStack(Function prepare, ScriptStackElement[] stack) {
 
     private static Object js_toString(Scriptable thisObj) {
         Object name = ScriptableObject.getProperty(thisObj, "name");
-        if (name == NOT_FOUND || name == Undefined.instance) {
+        if (name == NOT_FOUND || Undefined.isUndefined(name)) {
```
rbri, line 226: Is name == null possible?

testsrc/test262.properties
```diff
@@ -761,12 +729,6 @@ built-ins/Object
     ! getOwnPropertyDescriptors/symbols-included.js
     ! getOwnPropertyDescriptors/tamper-with-global-object.js
     ! getOwnPropertyDescriptors/tamper-with-object-keys.js
-    ! getPrototypeOf/15.2.3.2-2-12.js
-    ! getPrototypeOf/15.2.3.2-2-13.js
-    ! getPrototypeOf/15.2.3.2-2-14.js
-    ! getPrototypeOf/15.2.3.2-2-15.js
-    ! getPrototypeOf/15.2.3.2-2-16.js
-    ! getPrototypeOf/15.2.3.2-2-17.js
```
rbri, line 769: Great progress

rbri (commented):
Can't wait for this...

rbri:
Have done the usual check with this change and the HtmlUnit test suite - works without any problems.

src/org/mozilla/javascript/BaseFunction.java
gbrail: Yes, we can delete a bunch of code now by doing that.

src/org/mozilla/javascript/BaseFunction.java
gbrail: That's an obsolete property but in the interest of completeness I put that in too.

src/org/mozilla/javascript/NativeError.java
gbrail: null is converted to the string "null" by the ScriptRuntime.toString() call, and that's what the tests expect.

rbri:
Thanks

PR #931, opened 2021-06-20
Fix npe on no match regex search

rbri:
Followup for #927


rbri:
@gbrail  now you have two options for this - its fine for my to use @gausie version :-D

PR #942, opened 2021-06-22
Implement promises

### LINKED ISSUE #160: Support ES2015 Promise
V8 has support for the ES6 "Promise" object.

gbrail:
This is a working implementation of promises, built using the new LambdaConstructor stuff. It supports:

Promise.all, resolve, reject, and race
Promise.prototype.catch, finally, and then

Promise.allSettled isn't implemented yet -- it seems to be a bit newer and some of this code is a year old or more. That can come soon.

This is a little bit of a naive implementation in that it follows the spec closely. There might be some optimization possible because we are currently not taking advantage of the fact that most of the time, we are calling various methods of NativePromise, but because this is JavaScript we have to look up methods like "then" every time in case they change.

I think that given the amount of Java interoperability people do with Rhino, we also need some way to implement a Promise in Java code, or at least call the various methods to resolve the promise easily from Java. That could be a future PR. I'm hoping that some people who do the deep embedding with Java some more will have ideas of a good way to do it.

Closes #160

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -4,5 +4,4 @@ version=1.7.14-SNAPSHOT
 buildDir=buildGradle
 mavenSnapshotRepo=https://oss.sonatype.org/content/repositories/snapshots
 mavenReleaseRepo=https://oss.sonatype.org/service/local/staging/deploy/maven2/
-org.gradle.caching=true
 org.gradle.parallel=true
--- a/gradle.properties
+++ b/gradle.properties
@@ -4,4 +4,5 @@ version=1.7.14-SNAPSHOT
 buildDir=buildGradle
 mavenSnapshotRepo=https://oss.sonatype.org/content/repositories/snapshots
 mavenReleaseRepo=https://oss.sonatype.org/service/local/staging/deploy/maven2/
+org.gradle.caching=true
 org.gradle.parallel=true
--- a/src/org/mozilla/javascript/ScriptRuntime.java
+++ b/src/org/mozilla/javascript/ScriptRuntime.java
@@ -4863,29 +4863,6 @@ public class ScriptRuntime {
         return new JavaScriptException(error, filename, linep[0]);
     }
 
-    /**
-     * Implement the ECMAScript abstract operation "SpeciesConstructor" defined in section 7.2.33 of
-     * ECMA262.
-     */
-    public static Constructable getSpeciesConstructor(
-            Scriptable s, Constructable defaultConstructor) {
-        Object constructor = ScriptableObject.getProperty(s, "constructor");
-        if (constructor == Scriptable.NOT_FOUND || Undefined.isUndefined(constructor)) {
-            return defaultConstructor;
-        }
-        if (!isObject(constructor)) {
-            throw typeErrorById("msg.arg.not.object", typeof(constructor));
-        }
-        Object species = ScriptableObject.getProperty((Scriptable) constructor, SymbolKey.SPECIES);
-        if (species == Scriptable.NOT_FOUND || species == null || Undefined.isUndefined(species)) {
-            return defaultConstructor;
-        }
-        if (!(species instanceof Constructable)) {
-            throw typeErrorById("msg.not.ctor", typeof(species));
-        }
-        return (Constructable) species;
-    }
-
     public static final Object[] emptyArgs = new Object[0];
     public static final String[] emptyStrings = new String[0];
 }
--- a/src/org/mozilla/javascript/ScriptRuntime.java
+++ b/src/org/mozilla/javascript/ScriptRuntime.java
@@ -278,6 +278,7 @@ public class ScriptRuntime {
             NativeCollectionIterator.init(scope, NativeSet.ITERATOR_TAG, sealed);
             NativeCollectionIterator.init(scope, NativeMap.ITERATOR_TAG, sealed);
             NativeMap.init(cx, scope, sealed);
+            NativePromise.init(cx, scope, sealed);
             NativeSet.init(cx, scope, sealed);
             NativeWeakMap.init(scope, sealed);
             NativeWeakSet.init(scope, sealed);
--- a/src/org/mozilla/javascript/ScriptRuntime.java
+++ b/src/org/mozilla/javascript/ScriptRuntime.java
@@ -4863,6 +4863,29 @@ public class ScriptRuntime {
         return new JavaScriptException(error, filename, linep[0]);
     }
 
+    /**
+     * Implement the ECMAScript abstract operation "SpeciesConstructor" defined in section 7.2.33 of
+     * ECMA262.
+     */
+    public static Constructable getSpeciesConstructor(
+            Scriptable s, Constructable defaultConstructor) {
+        Object constructor = ScriptableObject.getProperty(s, "constructor");
+        if (constructor == Scriptable.NOT_FOUND || Undefined.isUndefined(constructor)) {
+            return defaultConstructor;
+        }
+        if (!isObject(constructor)) {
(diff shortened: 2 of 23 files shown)


gbrail:
This will fix https://github.com/mozilla/rhino/issues/160 from back when I was a young boy.

tonygermano:
Make sure you actually link this to #160 so that it closes it :smiley: 

I haven't looked closely yet, but does this include a lot of the same changes regarding the new LambdaConstructor stuff that was in your NativeError PR that was closed?

p-bakker:
Nice!

Just as FYI: 
- Promise.allSettled was added in ES2020:  Promise.allSettled, a new Promise combinator that does not short-circuit
- ES2021 also adds Promise.any: a Promise combinator that short-circuits when an input value is fulfilled (guess this one hasn't turned up yet, as we're not yet on the latest commit on test262's main branch)

src/org/mozilla/javascript/ScriptRuntime.java
```diff
@@ -4863,6 +4863,29 @@ public static JavaScriptException throwCustomError(
         return new JavaScriptException(error, filename, linep[0]);
     }
 
+    /**
+     * Implement the ECMAScript abstract operation "SpeciesConstructor" defined in section 7.2.33 of
```
p-bakker, line 4867: Consider moving this to o.m.j.AbstractEcmaObjectOperations, as it is one of the Abstract Object Operations defined by EcmaScript. When you do, please add a first param of type Context (as per the instructions outlines at the top of o.m.j.AbstractEcmaObjectOperations): while the function might not need Context atm, it might in the future

Consider adding a link to the specific section in the specific version of the EcmaScript spec implemented: the codebase is full of textual references to a section in the spec, but as the spec evolves, new chapters get added and things get renamed, so textual reference without context become meaningfull. A link to [ES2021 Species Constructor](https://262.ecma-international.org/12.0/#sec-speciesconstructor) however will (hopefully) remain valid and allows someone looking at this code x years from now to figure out the delta between the specific version implemented and the latest spec at that point in time

gradle.properties
```diff
@@ -4,5 +4,4 @@ version=1.7.14-SNAPSHOT
 buildDir=buildGradle
 mavenSnapshotRepo=https://oss.sonatype.org/content/repositories/snapshots
 mavenReleaseRepo=https://oss.sonatype.org/service/local/staging/deploy/maven2/
-org.gradle.caching=true
```
p-bakker, line 7: Is this an intentional change? Because it means for example that when running `./gradlew jacocoReport`, it wont do that on the cached results of the last testrun, but instead needs to run all the tests again (I think).

If this is intentional and will stay this way, the docs for updating the test262.properties file need a slight adjustment, because it now states you need to run with the `--rerun-tasks` param, because the `-DupdateTest262properties` option doesn't work based on cached results. So, if we're not doing caching anymore, the `--rerun-tasks` param is no longer needed

gbrail: I'll take this out of that particular PR. But I have been finding that the SpotBugs plugins are failing randomly on my machine and it seemed to be caching related -- I'd otherwise like to leave it on.

testsrc/test262.properties
```diff
@@ -1144,7 +1585,6 @@ built-ins/RegExp 926/1464 (63.25%)
     prototype/Symbol.replace/y-init-lastindex.js
     prototype/Symbol.replace/y-set-lastindex.js
     prototype/Symbol.search/cstm-exec-return-index.js
-    prototype/Symbol.search/failure-return-val.js
```
gbrail, line 1147: I'm assuming that this is fixed because I didn't check in the file after merging the last PR that fixes a bug in Regexp.

p-bakker: I noticed that one too today...

Yeah, if https://github.com/mozilla/rhino/pull/931 fixed something due to which this test now started to pass and that was not accounted for in that PR, if you now rebased your Promises PR on the latest master and regenerated the .properties file, it makes sense that this delta now pops up

gbrail: Just in general (and we can open a new issue for this) -- if someone writes code that breaks test262 tests, I know that the tests will fail the first time, but will the test suite now "fix" test262.properties to indicate that there's a new set of failing tests?

If so, then we will have to monitor every incoming PR to see if it inadvertently disabled stuff in test262.properties.

p-bakker: If someone writes code that breaks tests and then they actively regenerate test262.properties (which is a manual thing they have to do, by running the test with the extra `-DupdateTest262properties` argument) and then ignore the delta in test262.properties and put it up as a PR, yeah, then, if if the reviewer also doesn't pay attention, new broken tests might sneak in.

But that is not something new imho: before someone could do something like that manually, by just adding the broken tests to be excluded through test262.properties and put that up for PR.

p-bakker: We could do something to prevent stuff like the `prototype/Symbol.search/failure-return-val.js` issue above: we could fail the build if tests excluded in test262.properties aren't failing

gbrail: I most worried now about people introducing new exclusions because they broke things, but the test262 stuff adapter and just added a new exclusion.

What if the default behavior was to remove exclusions from test262.properties automatically, but not add new ones unless you run it with a special system property? That way it's obvious when working that you broke something and it's going to stay broken unless you explicitly did something to add a new exclusions.

p-bakker: I guess we could do that. Would we document that 'special' system property or keep that as a sort of undocumented feature then?

gbrail: I think we should document them -- for instance, when adding promises I had to edit the file directly, and that would have had to trigger a "full regen" of the properties file, versus just removing exclusions.

gbrail: To be clear, when we updated the test262 stuff, I somehow thought that the properties file re-gen was going to be optional, but it seems to happen every time I do a "gradle check" which is why I wanted to have some more control.

p-bakker: It is optional... Or at least it should be

The regenerate should only happen when the `-DupdateTest262properties` argument is added to `./gradlew tests` and not under any other circumstances

p-bakker: Double-checked: the code to update the test262.properties file is only triggered when the `updateTest262properties` System property is set, so not sure how it would get triggered when you run `gradle check`...

p-bakker: Indeed, when running the tests through gradle, the regen of test262.properties was always triggered. Added a fix in #951

rbri:
Will check by replacing HtmlUnits Promise impl with this one - but this requires some time. Hope i have it done in two weeks from now.

gbrail:
Status update -- I'll be taking a break for the next two weeks but when I get back in mid-July I'll merge this unless anything else has come up.

p-bakker:
Enjoy your break! Just wondering: could JavaScript Promises maybe be a standard replacement for Continuations as Rhino supports them?

Feel they are somewhat similar, but not sure how similar and if promises could actually replace continuations...

gausie:
@gbrail does this still look good to merge to you?

gbrail:
Sorry -- I've been back but working on other things. I'm currently a bit
slow because I'd like the "compat-table" tests to work with this feature,
and for that I need a working "setTimeout," which I did but which doesn't
fix the problem yet. A few days more I hope!

rbri:
@gbrail found some time to try your impl with HtmlUnit - looks promising :-)

But some of my tests are failing - looks like problems with the scope/thisObj. Will make some pr's.

rbri:
just for the records - i finally found some time to do the switch from HtmlUnits own promise implementation.
Seems to work fine.

PR #951, opened 2021-06-24
Bring back output on passing tests while excluded via test262.properties

p-bakker:
Realized when I started using the new test2626 mechanism that the output of tests actually passing during a testrun while excluded in test262.properties as they are expected to fail had gone missing

While fixing that I noticed that all files from subfolders of a rolled-up folder were assumed to fail by the test runner, instead of just its own files, so fixed that as well

And did an optimize to report 'non-interpreted' instead of {(non-)strict: [0, 9]} if test is flagged with noStrict or onlyStrict

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -87,7 +87,9 @@ test {
         systemProperty 'TEST_OPTLEVEL', System.getProperty('optLevel')
     }
     systemProperty 'test262properties', System.getProperty('test262properties')
-    systemProperty 'updateTest262properties', System.getProperty('updateTest262properties')
+    if (System.getProperty('updateTest262properties') != null) {
+        systemProperty 'updateTest262properties', System.getProperty('updateTest262properties')
+    }
     maxHeapSize = "1g"
     testLogging.showStandardStreams = true
     // Many tests do not clean up contexts properly. This makes the tests much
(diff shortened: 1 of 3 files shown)


gbrail:
Not part of this review but I know what's going on: Because we have test262properties defined in build.gradle, then its default value is the empty string, not null. So as a result, "updateTest262Properties" is always being set to true even if the system property is not being set.

One way to fix it is back in the "static" block on top to not accept an empty string, but instead expect some other value. For instance, you might support "true."

Or TBH, I think that 99% of the time people either want to not update the properties, or they want "all" so you could also simplify it to do that.

p-bakker:
Think I rather do this in build.gradle:
```
   if (System.getProperty('updateTest262properties') != null) {
        systemProperty 'updateTest262properties', System.getProperty('updateTest262properties')
    }
 ```
instead of just:
```
   systemProperty 'updateTest262properties', System.getProperty('updateTest262properties')
``` 

Think that that would fix it as well, if I understand you correctly. I'd like an empty `-DupdateTest262Properties` property to trigger the default action, while still having all those other options (because those are handy to generate different flavors for analyses)

p-bakker:
Nvm, tested the above and that worked, so pushed that fix to be included in this PR

gbrail:
That makes sense to me. Thanks!

Issue #932, opened 2021-06-21
Support ES2020 `globalThis`


p-bakker:
Support the universal way to access the global this value through the `globalThis` property

PR #952, opened 2021-06-24
Implement globalThis

p-bakker:
Closes https://github.com/mozilla/rhino/issues/932


gbrail:
I imagined that would be a lot more complicated -- thanks for fixing it!

PR #953, opened 2021-06-24
Enable GitHub actions for CI

gbrail:


p-bakker:
For Java 16: https://github.com/google/google-java-format/releases/tag/v1.10.0

gbrail:
That new version of google-java-format hasn't been integrated with the Spotless Gradle plugin yet, so we'll have to stick with Java 15 for now.

tonygermano:
Is this replacing CircleCi or in addition to?

gbrail:
Let's see how it goes and whether someone from Mozilla contacts us with a
huge bill!

p-bakker:
@gbrail are you sure that the build failure for Java 16 is due to latest version of google-java-format isn't integrated yet with the Spotless Gradle plugin?

As far as I can see, we're using com.diffplug.spotless v5.12.1, which in turn uses google-java-format v1.10

I think it's 'just' a matter of adding these:
```
  --add-exports jdk.compiler/com.sun.tools.javac.api=ALL-UNNAMED
  --add-exports jdk.compiler/com.sun.tools.javac.file=ALL-UNNAMED
  --add-exports jdk.compiler/com.sun.tools.javac.parser=ALL-UNNAMED
  --add-exports jdk.compiler/com.sun.tools.javac.tree=ALL-UNNAMED
  --add-exports jdk.compiler/com.sun.tools.javac.util=ALL-UNNAMED
```

PR #964, opened 2021-06-28
Allow hiding Java exception implementation details from catch block

### LINKED ISSUE #911: Hiding Java exception implementation details from catch block
Currently, Rhino wraps native Java exceptions thrown by host functions with the WrappedException class. If a script catches a WrappedException, Rhino exposes an error object to the catch block with a message like "JavaException: java.lang.RuntimeException: exception message". For my use case, users running scripts do not know that their JavaScript code is being executed in Java, so it's confusing that some exception messages are prefixed by "JavaException: (Java class name):". 

This code is implemented in the static function `ScriptRuntime.newCatchScope` (https://github.com/mozilla/rhino/blob/master/src/org/mozilla/javascript/ScriptRuntime.java#L3962). As far as I can tell, it currently doesn't seem possible to override this behavior other than by updating every `jsFunction_` function to catch all exceptions and rethrow EvaluatorException. 

To allow some Rhino users to avoid exposing implementation details of Java exceptions, while maintaining the current behavior for users who prefer to allow scripts to see details of Java exceptions, one approach could be to add a context feature flag and update `ScriptRuntime.newCatchScope` like this:

```
            else if (t instanceof WrappedException) {
                WrappedException we = (WrappedException) t;
                re = we;
                javaException = we.getWrappedException();

                if (cx.hasFeature(Context.FEATURE_HIDE_WRAPPED_EXCEPTIONS)) {
                    type = TopLevel.NativeErrors.InternalError;
                    errorMsg = javaException.getMessage();
                } else {
                    type = TopLevel.NativeErrors.JavaException;
                    errorMsg = javaException.getClass().getName() + ": " + javaException.getMessage();
                }
            } 
```

Are there any simpler ways to do this?

I could submit a PR for this if adding a Context feature flag seems like a good approach.

youngj:
As discussed in https://github.com/mozilla/rhino/issues/911, currently, if a script catches an error that originated from a native Java Exception, Rhino exposes an error object to the catch block with a message like "JavaException: java.lang.RuntimeException: exception message". 

However, users running JavaScript code via systems using Rhino do not necessarily know that their JavaScript code is being executed with Rhino/Java, so it's confusing that some error messages are prefixed by "JavaException: (Java class name):". I'd like some way to avoid exposing these implementation details to the user, but it currently doesn't seem possible to override this behavior, which is implemented in the static function `ScriptRuntime.newCatchScope`.

This PR implements one possible way to override this behavior, by adding a feature flag to the ContextFactory. If the feature flag `Context.FEATURE_HIDE_WRAPPED_EXCEPTIONS` is set, the error message will hide the Java implementation details, and the exception message would look like "InternalError: exception message". 

Possible alternative approaches:
* change the behavior for everyone, since it doesn't appear that any tests rely on the old behavior
* create a new method in the Context class to generate the message for a Java exception, allowing a custom ContextFactory subclass to return a custom Context subclass that could override that method

Closes #911

### DIFF
--- a/src/org/mozilla/javascript/Context.java
+++ b/src/org/mozilla/javascript/Context.java
@@ -339,13 +339,6 @@ public class Context implements Closeable {
      */
     public static final int FEATURE_ENABLE_JAVA_MAP_ACCESS = 21;
 
-    /**
-     * If set, then any native exceptions thrown by Java code and caught by a JavaScript catch block
-     * will appear as 'InternalError' instead of 'JavaException', and will not show the class name
-     * of the Java exception.
-     */
-    public static final int FEATURE_HIDE_WRAPPED_EXCEPTIONS = 22;
-
     public static final String languageVersionProperty = "language version";
     public static final String errorReporterProperty = "error reporter";
 
--- a/src/org/mozilla/javascript/Context.java
+++ b/src/org/mozilla/javascript/Context.java
@@ -339,7 +339,12 @@ public class Context implements Closeable {
      */
     public static final int FEATURE_ENABLE_JAVA_MAP_ACCESS = 21;
 
-    public static final int FEATURE_HIDE_WRAPPED_EXCEPTIONS = 1001;
+    /**
+     * If set, then any native exceptions thrown by Java code and caught by a JavaScript catch block
+     * will appear as 'InternalError' instead of 'JavaException', and will not show the class name
+     * of the Java exception.
+     */
+    public static final int FEATURE_HIDE_WRAPPED_EXCEPTIONS = 22;
 
     public static final String languageVersionProperty = "language version";
     public static final String errorReporterProperty = "error reporter";
--- a/src/org/mozilla/javascript/Context.java
+++ b/src/org/mozilla/javascript/Context.java
@@ -339,6 +339,8 @@ public class Context implements Closeable {
      */
     public static final int FEATURE_ENABLE_JAVA_MAP_ACCESS = 21;
 
+    public static final int FEATURE_HIDE_WRAPPED_EXCEPTIONS = 1001;
+
     public static final String languageVersionProperty = "language version";
     public static final String errorReporterProperty = "error reporter";
 
--- a/src/org/mozilla/javascript/ContextFactory.java
+++ b/src/org/mozilla/javascript/ContextFactory.java
@@ -287,9 +287,6 @@ public class ContextFactory {
 
             case Context.FEATURE_ENABLE_JAVA_MAP_ACCESS:
                 return false;
-
-            case Context.FEATURE_HIDE_WRAPPED_EXCEPTIONS:
-                return false;
         }
         // It is a bug to call the method with unknown featureIndex
         throw new IllegalArgumentException(String.valueOf(featureIndex));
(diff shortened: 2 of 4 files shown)


tonygermano:
I think the feature flag is a good approach for this. I run Rhino in an environment where users are aware they are running on Java, and frequently utilize Java libraries in their scripts, so I would not want to lose the current behavior, as it helps with troubleshooting.

gbrail:
Yes, definitely the feature flag approach is good. There are plenty of
cases in which Java code is embedded in Rhino and executed in turn by more
Java code that needs to see the original exception.

youngj:
Sounds good to me. This PR currently names the feature flag `FEATURE_HIDE_WRAPPED_EXCEPTIONS` and it is only used when JavaScript code catches an exception thrown by Java code. Potentially there could be other parts of Rhino that expose Java/Rhino implementation details to JavaScript code as well. What would be a good name for this feature flag? Could this feature flag be used to hide other Java/Rhino implementation details?

p-bakker:
Wondering if you are also already disabling stuff like:
- Packages
- direct access to classes via org/java/com.xxxxx
- JavaImporter
- etc.

Just wondering if this calls for yet another, quite specific flag or if there's some other way we can be clever about this. For example only include the Java Exception details if the above stuff is also initiated

youngj:
Good idea. We're already using Context.setClassShutter to hide all Java classes:

```
cx.setClassShutter(new ClassShutter() {
    @Override
    public boolean visibleToScripts(String className)
    {
        return false;
    }
});
```

so we could use ScriptRuntime.isVisible without adding a new ContextFactory flag:

```
                if (!isVisible(cx, javaException)) {
                    type = TopLevel.NativeErrors.InternalError;
                    errorMsg = javaException.getMessage();
                } else {
                    type = TopLevel.NativeErrors.JavaException;
                    errorMsg =
                            javaException.getClass().getName() + ": " + javaException.getMessage();
                }
```

Thoughts?

p-bakker:
If you're not exposing any classes (visibleToScripts always returning false), how are you initializing your standard objects?

Because if you go through ScriptRuntime.initSafeStandardObjects (and never through ScriptRuntime.initStandardObjects), isn't the ability to access Java Classes disabled completely (without having to implement the classShutter)?

My idea was to only enable the Java Exception details inside JavaScript when the init was done through through ScriptRuntime.initStandardObjects, which enables Java class access, so it would make sense to then also have access to the Java Exception details

Not sure if this would work for you and what others think about this approach

gbrail:
This seems like a pretty handy feature to me and looks good. But I'll see if others have opinions on the alternative approach that's being suggested for a day or so.

youngj:
> Because if you go through ScriptRuntime.initSafeStandardObjects (and never through ScriptRuntime.initStandardObjects), isn't the ability to access Java Classes disabled completely (without having to implement the classShutter)?

We use initSafeStandardObjects, but the ClassShutter is still used in ScriptRuntime.newCatchScope to avoid making the Rhino-specific properties `javaException`, `rhinoException`, and `__exception__` available within the JavaScript catch scope.

gbrail:
This looks good to me. If no one else has any objections I'll merge it soon.

PR #904, opened 2021-05-21
Template Literal support

p-bakker:
First of all: tnx for @anba for the initial implementation in https://github.com/mozilla/rhino/pull/81

This PR rebases the work done in https://github.com/mozilla/rhino/pull/81 and updates it to be compatible with the latest 

~~This is a Draft PR, just to get it out there and get feedback. So, any feedback appreciated :-)~~

Left to be done: 
1. ~~fix 3 breaking tests from Test262 (see https://github.com/mozilla/rhino/issues/903, https://github.com/mozilla/rhino/issues/734#issuecomment-843137357 and https://groups.google.com/g/mozilla-rhino/c/StUgOIRoOFY)~~
2. ~~formatting (rather not do some of the formatting work already done in https://github.com/mozilla/rhino/pull/896)~~
3. code review

closes https://github.com/mozilla/rhino/issues/243
closes https://github.com/mozilla/rhino/issues/325
closes https://github.com/mozilla/rhino/issues/773
closes https://github.com/mozilla/rhino/issues/294

Enables: updating to a newer version of Test262 (see https://github.com/mozilla/rhino/commit/6c1582acd21a8b1ecaad6f39849a1a6e0565b942, https://github.com/mozilla/rhino/issues/817)

PR #978, opened 2021-06-30
fix #977 Support ES2019 JSON superset

### LINKED ISSUE #977: Support ES2019 JSON superset
Proposal: https://github.com/tc39/proposal-json-superset

Before running the test262 suite for testing your json superset impl., comment out/remove this line: https://github.com/mozilla/rhino/blob/master/testsrc/org/mozilla/javascript/tests/Test262SuiteTest.java#L120

tuchida:
Closes #977

Is it necessary to check `compilerEnv.getLanguageVersion()`?
These do not seem to be checked.
- https://github.com/mozilla/rhino/pull/904
- https://github.com/mozilla/rhino/pull/952

### DIFF
--- a/src/org/mozilla/javascript/TokenStream.java
+++ b/src/org/mozilla/javascript/TokenStream.java
@@ -914,11 +914,30 @@ class TokenStream {
                 quoteChar = c;
                 stringBufferTop = 0;
 
-                c = getChar(false);
+                c = getCharIgnoreLineEnd(false);
                 strLoop:
                 while (c != quoteChar) {
-                    if (c == '\n' || c == EOF_CHAR) {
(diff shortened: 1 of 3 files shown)


p-bakker:
Nice going!

IMHO it doesn't need to check the languageVersion, as this is a change that previously would be a SyntaxError, thus it cannot really break existing implementations (just like the other two you mentioned are syntax or (global) object additions that don't alter existing behavior)

gbrail:
Yes -- I agree that it's OK to add features like this because they result in code that used to not work working, as opposed to features that result in code that used to work stopping...

I'll test this myself and get back...

gbrail:
This looks good to me too. Thanks!

PR #979, opened 2021-06-30
Clean up handling of SCRIPTABLE_UNDEFINED

gbrail:
We introduced SCRIPTABLE_UNDEFINED a while back as an alternate representation
of "undefined" for passing as "this" to functions, as Rhino originally
couldn't do that.

However, the original implementation was as a Proxy object that threw
UnsupportedOperationException in some cases and which ended up causing
failures on the kangax/compat-table tests.

This makes SCRIPTABLE_UNDEFINED a real Scriptable and cleans up a few
other things as well.

### DIFF
--- a/src/org/mozilla/javascript/ScriptRuntime.java
+++ b/src/org/mozilla/javascript/ScriptRuntime.java
@@ -2430,9 +2430,7 @@ public class ScriptRuntime {
      */
     public static boolean loadFromIterable(
             Context cx, Scriptable scope, Object arg1, BiConsumer<Object, Object> setter) {
-        if ((arg1 == null) || Undefined.isUndefined(arg1)) {
-            return false;
-        }
+        if ((arg1 == null) || Undefined.isUndefined(arg1)) return false;
 
         // Call the "[Symbol.iterator]" property as a function.
         final Object ito = ScriptRuntime.callIterator(arg1, cx, scope);
--- a/src/org/mozilla/javascript/ScriptRuntime.java
+++ b/src/org/mozilla/javascript/ScriptRuntime.java
@@ -383,7 +383,7 @@ public class ScriptRuntime {
     public static boolean toBoolean(Object val) {
         for (; ; ) {
             if (val instanceof Boolean) return ((Boolean) val).booleanValue();
-            if (val == null || val == Undefined.instance) return false;
+            if (val == null || Undefined.isUndefined(val)) return false;
             if (val instanceof CharSequence) return ((CharSequence) val).length() != 0;
             if (val instanceof BigInteger) {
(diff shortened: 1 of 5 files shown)


rbri:
Looks good to me

p-bakker:
If possible if like to take a look at this tomorrow as well: running the test262 suite I noticed enon-RhinoExceptions  that I was attributing to this as well, but if my memory serves me well, it were more than the few that now started passing. So I'd like to check what's going on there

p-bakker:
Looks like the exceptions I remembered seeing indeed were for the problem fixed in this PR, but the majority of those tests are still failing, albeit not throwing an unexpected `UnsupportedOperationException` anymore

Did see some from test262 that used to throw a `UnsupportedOperationException` now throw a `nullPointerException`, for example `built-ins\Function\prototype\apply\this-not-callable.js`. Note that a similar `nullPointerException` was already being thrown on `built-ins\Function\prototype\call\S15.3.4.4_A14.js` before and still is.

If you want to take a close look at the exceptions I'm talking about, rebase on https://github.com/mozilla/rhino/pull/981 and uncomment the ex.printStackTrace() on line 560 of Test2626SuiteTest

src/org/mozilla/javascript/ScriptRuntime.java
```diff
@@ -2430,13 +2430,13 @@ private static void enumChangeObject(IdEnumeration x) {
      */
     public static boolean loadFromIterable(
             Context cx, Scriptable scope, Object arg1, BiConsumer<Object, Object> setter) {
-        if ((arg1 == null) || Undefined.instance.equals(arg1)) {
+        if ((arg1 == null) || Undefined.isUndefined(arg1)) {
```
rbri, line 2433: minor: we can remove the brackets here to unify the code

src/org/mozilla/javascript/Undefined.java
```diff
@@ -37,26 +41,106 @@ public boolean equals(Object obj) {
     @Override
     public int hashCode() {
         // All instances of Undefined are equivalent!
-        return 0;
+        return System.identityHashCode(instance);
```
rbri, line 44: because instance is final maybe we can make this a constant also

src/org/mozilla/javascript/ScriptRuntime.java
gbrail: Sure, I'm not used to doing that but I'd like it to match.

src/org/mozilla/javascript/Undefined.java
gbrail: Good idea. Thanks!

PR #981, opened 2021-07-01
Fix to exception handling of negative testcases

p-bakker:
If negative testcases were marked as failing in test262.properties, they would not get reported when they stopped failing (and also not removed from test262.properties when running with -DupdateTest262properties

Note: on line 560 of Test262SuiteTest the printing of unexpected exceptions is commented out (for now) as there are a ton of them, mostly related to the harness utilizing Es features like default param values or unicode regexes that we don't support yet.

There are however also a bunch related to bugs in the Rhino codebase, resulting in nullPointers

Note that some testcases now reported as passing, as doing so for the wrong reasons: for example `language/block-scope/syntax/redeclaration/fn-scopes-var-name-redeclaration-attempt-with-const` now passes as a SyntaxError is thrown, but that error is thrown because we don't support `class`, not necessarily because we properly wouldn't allow redeclaration. But I think this is fine

### DIFF
--- a/testsrc/org/mozilla/javascript/tests/Test262SuiteTest.java
+++ b/testsrc/org/mozilla/javascript/tests/Test262SuiteTest.java
@@ -846,13 +846,13 @@ public class Test262SuiteTest {
         private final boolean hasEarlyError;
 
         private final Set<String> flags;
-        private final Set<String> harnessFiles;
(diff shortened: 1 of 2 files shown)


gbrail:
Looks good to me. Thanks!

PR #997, opened 2021-07-11
Take the Delegator at some more places into account

rbri:
This replaces my old PR  #335 - if you like you can have a look at the conversation of this PR.

These changes are all changes is use since ages in HtmlUnit - i have done some checking and they are really needed. And btw: the Delegator support seem to be in Rhino from the beginning.

Depending on the feedback i like to merge this soon.

### DIFF
--- a/testsrc/org/mozilla/javascript/tests/DelegatorTest.java
+++ b/testsrc/org/mozilla/javascript/tests/DelegatorTest.java
@@ -1,0 +1,141 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/** */
+package org.mozilla.javascript.tests;
+
+import static org.junit.Assert.assertEquals;
+
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+import org.mozilla.javascript.Context;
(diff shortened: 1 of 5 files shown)


testsrc/org/mozilla/javascript/tests/DelegatorTest.java
```diff
@@ -0,0 +1,141 @@
...
+
+    @Test
+    public void equals() {
+        assertEquals("false", eval("TestDelegee == TestDelegator"));
+        assertEquals("false", eval("TestDelegator == TestDelegee"));
+
+        assertEquals("true", eval("TestDelegee === TestDelegee"));
+
+        assertEquals("true", eval("TestDelegee === TestDelegator"));
+        assertEquals("true", eval("TestDelegator === TestDelegee"));
```
tonygermano, line 74: I'm confused by the tests here.  If `TestDelegee === TestDelegator` is true, how could `TestDelegee == TestDelegator` be false?

rbri: Good point. Because of the discussion of the my try on this topic i have this time done the Delegator handling only at places the current HtmlUnit test suite fails. Because of this we will have many more places where the handling of Delegators is not symetric.
But for this special case it might make sense to solve this. - will have a look.

rbri: @tonygermano made a new PR - please have a look

PR #1000, opened 2021-07-23
Consstring optimization

rbri:
Two minor optimizations
* there is no need to call a method because a simple cast is sufficient here
* the second one is more complex, in the else case we already have a parameter conversion loop that also flattens the ConsStrings. Because of this we do not need to run the loop for this branch

Have done the usual test with my HtmlUnit test suite a found no notable performance change. But reducing i think we have to do this to reduce the load a bit.

### DIFF
--- a/src/org/mozilla/javascript/FunctionObject.java
+++ b/src/org/mozilla/javascript/FunctionObject.java
@@ -327,14 +327,14 @@ public class FunctionObject extends BaseFunction {
         boolean checkMethodResult = false;
         int argsLength = args.length;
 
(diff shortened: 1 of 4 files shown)


rbri:
Sorry, looks like the second part introduces a problem. have to investigate this....

rbri:
Hope this is now complete - waiting for feedback.

gbrail (approved):
Both of these optimizations look correct to me. Thanks!

rbri:
Thanks will merge this also tomorrow (if no one complains)

Issue #680, opened 2020-05-22
Locale is always French (at least on French platforms)


hervegirod:
I have a PC for which default Locale is "fr", and even if I set 
`context.setLocale(<some locale")` to any other locale, I still have messages in French. It work if I do
this:
```
Locale locale = new Locale("");
context.setLocale(locale);
```
The reason is that there is only the following resource files on Rhino 1.712:
- `org.mozilla.javascript.resources.Messages`
- `org.mozilla.javascript.resources.Messages_fr`
And the code in Rhino does:
```
ResourceBundle rb = ResourceBundle.getBundle("org.mozilla.javascript.resources.Messages", locale);
```
Which means that Rhino can only find the "fr" resource.

gbrail:
Can you suggest how we fix this?

Also, if anyone is looking to contribute to Rhino, someone generously
contributed the French message bundle many years ago and I'm sure it's up
to date. And obviously there are no other languages supported. Translating
the messages into the language you know best might help a lot of developers!

gbrail:
And sorry, I meant, I'm sure it's NOT up to date!

rbri:
I can contribute a German translation

hervegirod:
Hello, if the French localization is not up to date, I can contribute to make it up to date with the last version. Also I don't think my own specific fix is the right way to do it in the library, but I will look at the code and try to propose something to fix it.

gbrail:
I'm still not sure I understand the original issue. It sounds like the
problem is that the resource bundles aren't set correctly in the classpath,
so French messages are being displayed when they shouldn't. Is that the
case?

Otherwise, today Rhino only supports English and French messages and unless
someone contributes another translation (like German -- thanks!) then those
are the only choices we have.

hervegirod:
Yes, that's exactly that! Maybe it's only when the machine is configured as French though.

PR #1001, opened 2021-07-27
Add English messages resource

tonygermano:
Fixes https://github.com/mozilla/rhino/issues/680

Adds blank English language file so that it can be selected as a language for the Context Locale when the JVM default Locale's language is not English, but is translated.

Message keys not translated in the language specific properties file will fall through to the base (no language) file.

### DIFF
--- a/testsrc/org/mozilla/javascript/tests/MessageProviderLanguageTest.java
+++ b/testsrc/org/mozilla/javascript/tests/MessageProviderLanguageTest.java
@@ -1,0 +1,113 @@
+/* -*- Mode: java; tab-width: 8; indent-tabs-mode: nil; c-basic-offset: 4 -*-
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+package org.mozilla.javascript.tests;
+
+import static java.util.Locale.FRANCE;
+import static java.util.Locale.US;
+import static org.junit.Assert.assertEquals;
+
+import java.util.Locale;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+import org.mozilla.javascript.Context;
+import org.mozilla.javascript.ScriptRuntime;
+
+public class MessageProviderLanguageTest {
+
+    // Scottish Gaelic Language- assumption is this language does not have a translation
+    private final Locale NOT_TRANSLATED = new Locale("gd");
+
+    private final String messageId = "msg.invalid.date";
+    private final String englishMessage = "Date is invalid.";
+    private final String frenchMessage = "Date incorrecte.";
+    private final String defaultMessage = englishMessage;
+
+    private Context cx;
+
+    @Before
+    public void init() {
+        cx = Context.enter();
+    }
+
+    @After
+    public void cleanup() {
+        Context.exit();
+    }
+
+    @Test
+    public void defaultLocaleUsesDefaultLanguageAndContextLocaleIsNotSpecified() {
+        assertEquals(englishMessage, getMessageUsingDefaultLocale(US));
+    }
+
+    @Test
+    public void defaultLocaleUsesOtherTranslatedLanguageAndContextLocaleIsNotSpecified() {
+        assertEquals(frenchMessage, getMessageUsingDefaultLocale(FRANCE));
+    }
+
+    @Test
+    public void defaultLocaleUsesUntranslatedLanguageAndContextLocaleIsNotSpecified() {
+        assertEquals(defaultMessage, getMessageUsingDefaultLocale(NOT_TRANSLATED));
+    }
+
+    @Test
+    public void defaultLocaleUsesDefaultLanguageAndContextLocaleIsTheSame() {
+        assertEquals(englishMessage, getMessageUsingContextLocale(US, US));
+    }
+
+    @Test
+    public void defaultLocaleUsesDefaultLanguageAndContextLocaleUsesOtherTranslatedLanguage() {
+        assertEquals(frenchMessage, getMessageUsingContextLocale(FRANCE, US));
+    }
+
+    @Test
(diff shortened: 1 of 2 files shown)


p-bakker:
Not sure this is the correct fix: for any Locale that doesn't use French (fr) as the language, the code `ResourceBundle.getBundle("org.mozilla.javascript.resources.Messages", locale);` should already pick up the `Messages.properties`

Adding an empty `Messages_en.properties` shouldn't make any difference.

My hunch is more that the reported issue has more to do with the code at https://github.com/mozilla/rhino/blob/master/src/org/mozilla/javascript/ScriptRuntime.java#L4510: if there's no current context, it uses the default Locale, which, in the case of the reporter of #680 is French

tonygermano:
> for any Locale that doesn't use French (fr) as the language, the code ... should already pick up the `Messages.properties`

That is correct, but if more translations are added, then it would be true of any of them if they match the default Locale. Without this file, there is no way to explicitly select English. Attempting to select English when French is the default fails, because it is no different than any other language that does not have a translation. Currently English is no different than any other language that isn't translated, which the exception that the base file for untranslated messages happens to already be in English.

`defaultLocaleUsesOtherTranslatedLanguageAndContextLocaleUsesDefaultLanguage` is the only test that fails without the fix, and the only workaround is the one specified in the issue, where a local with no language is used at the Context level.

p-bakker:
I now see, odd behavior of the ResourceBundle, that seems to go back to using Java on the DeskTop.

The way to fix it though seems to be `ResourceBundle.getBundle(RESOURCE_BUNDLE_NAME, locale, 
    ResourceBundle.Control.getNoFallbackControl(
        ResourceBundle.Control.FORMAT_DEFAULT));`, which results in ignoring the default Locale of the vm

tonygermano:
I had originally implemented my own `Resource.Control` and overrode the `getFallBackLocale` method, but then I was running into issues where certain conditions would send it into an infinite loop (also having to do with the fact that it couldn't actually find an English translation.)

This solution seemed both elegant and simpler.

When a user's default Locale language is translated (only `fr` at this time) I believe it is the proper behavior that when an untranslated Locale is set on the Context, that it should fall back to the default Locale, not to English. In that case, English needs to be selectable as a language if that is what the user desires to see.

The "no language" implementation should be the sole message source only when both the default and context locales are untranslated.

The French file uses the same principle as the empty English file. For example, if you trigger one of the recently added messages, like trying to add a bigint and number together, it will display the (no language) English message, because the French file does not include an entry for that key, but the base file does. The ResourceBundle is set up as a hierarchy for that reason (you can stack these several layers deep by including country codes and modules in the file names.)

p-bakker:
> When a user's default Locale language is translated (only fr at this time) I believe it is the proper behavior that when an untranslated Locale is set on the Context, that it should fall back to the default Locale

I think that is debatable, personally I rather ignore the VM's default Locale and it seems to me the OP of #680 expects that as well

> but then I was running into issues where certain conditions would send it into an infinite loop
Does that also happen when using `ResourceBundle.Control.getNoFallbackControl( ResourceBundle.Control.FORMAT_DEFAULT)`?

tonygermano:
How I understood the OP's problem is that they were trying to select English as a language and could not, because technically we don't have an English translation. We only have a base translation that happens to be in English.

If I am a user in France seeing messages in French, and I set the context language to Spanish, when a Spanish translation does not exist, I would expect it to fall back to French, not English.

The context already returns the VM's default Locale when the Locale is not set.

https://github.com/mozilla/rhino/blob/76b37a1353b0f1b8bbe76b6e59b06878fee7f378/src/org/mozilla/javascript/Context.java#L740-L748

I guess that behavior is really what is in question, as that is what I was trying to achieve. I didn't think setting the context language to one that is not translated should be treated any differently than not setting a context language in the first place.

Another benefit of assuming this behavior is that this PR is a no-code change. It only adds a test and a blank file (ignoring comments.)

tonygermano:
In other words, this is how the code already works, most of it driven by default java behavior, and I think this is the correct way to do it.
 
- when no context language is set
  - use local language if defined (base translation is used for undefined keys)
  - else use base translation only
- when context language is set
  - use specified language if defined (base translation is used for undefined keys)
  - else use local language if defined (base translation is used for undefined keys)
  - else use base translation only

The unset context language scenario displays English messages when the local language is English because it falls back to the base translation, the same as any other untranslated language. French is selected when the local language is French.

The scenario where the context language is set lets you specify French when the local language is English, but it does not let you specify English when the local language is French, due to the missing file provided by this PR.

testsrc/org/mozilla/javascript/tests/MessageProviderLanguageTest.java
```diff
@@ -0,0 +1,113 @@
...
+        assertEquals(englishMessage, getMessageUsingContextLocale(US, NOT_TRANSLATED));
+    }
+
+    @Test
+    public void defaultLocaleIsNotTranslatedAndContextLocaleUsesOtherTranslatedLanguage() {
+        assertEquals(frenchMessage, getMessageUsingContextLocale(FRANCE, NOT_TRANSLATED));
+    }
+
+    private String getMessageUsingDefaultLocale(Locale defaultLocale) {
+        Locale.setDefault(defaultLocale);
```
rbri, line 104: I think a test should not change the system status. What about saving the current locale and restore in a finally block.

rbri:
Personally I see no reason for providing non English error messages. Maybe being English only is the simplest solution here.

testsrc/org/mozilla/javascript/tests/MessageProviderLanguageTest.java
gbrail: Yes, thanks! We already have a number of tests that don't clean up the environment properly so we have to fork them into separate JVMs, which means that the test run takes much longer.

rbri:
Any news here? Should we do a poll? ;-)

rbri:
Have merged this because
* this is an improvement for the current situation
* we can still go on with more or less languages

rbri:
and thanks for the PR

PR #1003, opened 2021-08-07
Add setTimeout and clearTimeout to Rhino shell

gbrail:
The "kangax" compat table tests need setTimeout and clearTimeout to be implemented in order for the Promise tests to run.

This requires a mini event loop, and I wasn't comfortable building that into Rhino, so these functions are only implemented in the Shell right now. That means that Rhino tests launched from the shell can use these functions, and that those scripts will not exit until all of the timer tasks are finished.

### DIFF
--- a/testsrc/org/mozilla/javascript/tests/MozillaSuiteTest.java
+++ b/testsrc/org/mozilla/javascript/tests/MozillaSuiteTest.java
@@ -17,30 +17,17 @@ import java.util.Collection;
 import java.util.HashSet;
 import java.util.List;
 
-import org.junit.Assert;
-import org.junit.Test;
-import org.junit.runner.RunWith;
-import org.junit.runners.Parameterized;
-import org.junit.runners.Parameterized.Parameters;
-import org.mozilla.javascript.drivers.JsTestsBase;
-import org.mozilla.javascript.drivers.ShellTest;
-import org.mozilla.javascript.drivers.TestUtils;
-import org.mozilla.javascript.tools.shell.ShellContextFactory;
-
 /**
- * This JUnit suite runs the Mozilla test suite (in mozilla.org CVS
- * at /mozilla/js/tests).
+ * This JUnit suite runs the Mozilla test suite (in mozilla.org CVS at /mozilla/js/tests).
  *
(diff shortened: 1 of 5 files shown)


gausie:
We should probably add this to the 1.7.14 milestone

tonygermano:
Am I correct that it won't work in the interactive shell or if the main program loops waiting for the side effects of a Timeout event to be observed?

gbrail:
I think what was stopping me from merging this was the fact that it's
really a very limited implementation that's just enough to make stuff work
in the shell for certain cases.

It doesn't help with the interactive shell at all -- the timers don't run
until after you try to exit the shell. It's really only enough to make
things work with a non-interactive program that uses these functions,
which just happens to be the case for the "kangax" compat-table.

Making this work properly in the shell would mean rewriting the shell to
have a proper event loop that can be triggered when a timer fires or when
input comes from the user. That sounds like a fun project but it's not
something that I can do soon and it's arguably not terribly important.

If everyone is OK with this very limited version of this feature then yes,
I'm happy to merge it.

I actually didn't know about that particular spec -- it's good to follow,
but I don't have time to implement it all any time soon I'm afraid.

p-bakker:
I'd say that if this helps our score on the Kangax compat table due to making (Promise) tests pass that otherwise would fail, we should merge this now as is, add some relevant comments in the Release Notes and create a followup case where we make it spec compliant and allow optionally enabling it when enbedding Rhino +  a followup case to make it work properly in the shell.

Only thing to maybe add to the PR is some notes on the limited functionality of the current iteration

tonygermano:
I'm fine with merging it to allow Promise tests to pass, but I wanted to understand the limits of what it could do before we announce that we have setTimeout support in the shell.

If you open up the console in a web browser and manually call setTimeout, it will fire while it's waiting for more input from the console when the time comes. You can test by logging to the console. The logged message will appear even if you are in the middle of typing another command.
```js
setTimeout(() => {console.log("this will fire in 5 seconds")}, 5000);
```

gbrail:
I spent more time trying to make the console work in a more interactive way but it's a pretty big project. I'm going to merge this and let the release notes describe its limited scope.

PR #1006, opened 2021-08-10
take care of Delegators when working with Sets/Maps

rbri:
Sorry for having a Delegator problem again (see https://github.com/HtmlUnit/htmlunit/issues/373 for details)

Have written some tests and a first implementation - waiting for feedback

### DIFF
--- a/testsrc/org/mozilla/javascript/tests/DelegatorTest.java
+++ b/testsrc/org/mozilla/javascript/tests/DelegatorTest.java
@@ -122,7 +122,7 @@ public class DelegatorTest {
                         + "result += map.delete(TestDelegator);\n"
                         + "result += '#';\n"
                         + "result;";
-        // assertEquals("true#test#true#", eval(script));
+        assertEquals("true#test#true#", eval(script));
 
         script =
                 "var result = '';\n"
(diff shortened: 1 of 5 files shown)


gbrail:
This looks reasonable to me. Does anyone else who works regularly with delegators have an idea?

SebastianGlass:
I like the way it is fixed, but just to understand... Why are there is many copy pasted static methods instead of an util class (for the new keys method for example)

rbri:
> 
> 
> I like the way it is fixed, but just to understand... Why are there is many copy pasted static methods instead of an util class (for the new keys method for example)

Simple answer, i was a bit in a hurry (as always) and i had some vague memories that the set/map classes had a similar problem with shared code in the past. But i was not able to find the code so i decide to start with this copy and have a look at this later to get some feedback first. (Greg an i are working (i guess) in different time zones)

testsrc/org/mozilla/javascript/tests/DelegatorTest.java
```diff
@@ -109,6 +109,134 @@ public void setPrototypeOf() {
...
+                + "result += map.has(TestDelegator);\n"
+                + "result += '#';\n"
+
+                + "result += map.get(TestDelegator);\n"
+                + "result += '#';\n"
+
+                + "result += map.delete(TestDelegator);\n"
+                + "result += '#';\n"
+                + "result;";
+        // assertEquals("true#test#true#", eval(script));
```
SebastianGlass, line 125: There is an unused assertion. Surely a mistake ;)

Issue #1004, opened 2021-08-07
unhandledrejection Promise


tuchida:
https://developer.mozilla.org/en-US/docs/Web/API/Window/unhandledrejection_event
```js
// Rhino
new Promise((resolve) => {
  print('resolve');
  resolve();
}).then(() => {
  p('');  // ReferenceError
});  // without .catch()
```
### Rhino
```
resolve
[object Promise]
```
### SpiderMonkey
```
Uncaught (in promise) ReferenceError: p is not defined
    <anonymous> debugger eval code:5
    promise callback* debugger eval code:4
```
### node
ref. https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode
```
> (node:2889648) UnhandledPromiseRejectionWarning: ReferenceError: p is not defined
    at REPL6:5:3
    at processTicksAndRejections (internal/process/task_queues.js:95:5)
```
---
I think it will be difficult to debug without displaying an error or killing the process.

gbrail:
Yes, I can see how that's supposed to result in an "uncatchable exception" actually and that Rhino should be throwing it. I can also see why existing test suites didn't catch it. Let's see what we can do.

PR #1007, opened 2021-08-10
Add support to track unhandled promise rejections

gbrail:
This does nothing by default, but adds an object on the Context class that tracks unhandled rejected promises and lets the user process them. The shell now prints out warnings about unhandled rejections as it runs and at the end of any executed script.

For example, in the interactive shell:

js> let p = new Promise((resolve, reject) => {reject('rejected')});
Unhandled rejected promise: rejected
js> let q = new Promise((resolve, reject) => {throw new Error('threw')});
Unhandled rejected promise: Error: threw
	at <stdin>:3 (anonymous)
	at <stdin>:3

  and 1 other unhandled rejected promises
js> q.catch((e)=>{print('Caught ' + e)});
Caught Error: threw
Unhandled rejected promise: rejected
[object Promise]
js> p.catch((e)=>{print('Caught ' + e)});
Caught rejected
[object Promise]

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -411,6 +411,7 @@ jacocoTestReport.dependsOn test
 jacocoTestReport {
     reports {
         csv.enabled true
+        html.enabled true
     }
 }
 
--- a/src/org/mozilla/javascript/Context.java
+++ b/src/org/mozilla/javascript/Context.java
@@ -2339,6 +2339,26 @@ public class Context implements Closeable {
         } while (head != null);
     }
 
+    /**
+     * Control whether to track unhandled promise rejections. If "track" is set to true, then the
+     * tracker returned by "getUnhandledPromiseTracker" must be periodically used to process the
+     * queue of unhandled promise rejections, or a memory leak may result.
+     *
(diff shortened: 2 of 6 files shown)


gbrail:
This addresses https://github.com/mozilla/rhino/issues/1004

@tuchida, please try it out!

tuchida:
Thanks for your quick work!
Is it okay to say that tracker is only enabled in the shell by default?

gbrail:
The tracker is only enabled in the shell by default -- but the idea is that we have an API now on the Context class that anyone embedding Rhino can use to wire rejection tracking into their own projects any way that makes sense to them.

tuchida (approved):
Thanks!
Maybe you want to enable it by default for embedding, but let's see how it goes first.

tuchida:
By the way, there is a -debug option, is it ok if tracker is enabled even when it is turned off?
https://github.com/mozilla/rhino/blob/b66e10f7e1b89726bb6463f0db8f0d464c9dba56/toolsrc/org/mozilla/javascript/tools/shell/Main.java#L373-L374

PR #1010, opened 2021-08-19
Update README.md

GrumpyCamel:
Removed rotted docs link with related text.


GrumpyCamel:
Closing to undo link removal and instead replace it with archived docs link. see #1011

PR #1021, opened 2021-09-09
FIX: Instanceof for java objects/classes

rPraml:
This fixes an error, when `instanceOf` is used with java objects.

- `java.lang.String instanceof java.lang.Class` this was wron IMHO, as java.lang.String is wrapped as NativeJavaClass. So every NativeJavaClass is instance of `Class`
- Now, the return value of a function is wrapped into a NativeJavaClass if it was instanceof class (old behaviour a function/property returning a class was wrapped into a NativeJavaObject, which looks wrong to me), so `Class.forName("java.lang.String") == java.lang.String`


rPraml:
Closing, because it is wrong

PR #1022, opened 2021-09-09
simplify handling of null values for Map

rbri:
fix: NativeMap's Map.entries() emulation can return NativeMap.NULL_VALUE instead of actual null instance
more test code

see https://github.com/HtmlUnit/htmlunit-core-js/issues/4 for a very detailed bug report

### DIFF
--- a/src/org/mozilla/javascript/Hashtable.java
+++ b/src/org/mozilla/javascript/Hashtable.java
@@ -138,13 +138,9 @@ public class Hashtable implements Serializable, Iterable<Hashtable.Entry> {
         }
     }
 
-    public Object get(Object key) {
+    public Entry getEntry(Object key) {
         final Entry e = new Entry(key, null);
-        final Entry v = map.get(e);
-        if (v == null) {
-            return null;
-        }
-        return v.value;
+        return map.get(e);
     }
 
     public boolean has(Object key) {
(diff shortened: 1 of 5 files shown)


src/org/mozilla/javascript/Hashtable.java
```diff
@@ -1,30 +1,21 @@
 package org.mozilla.javascript;
 
 import java.io.Serializable;
-/* -*- Mode: java; tab-width: 8; indent-tabs-mode: nil; c-basic-offset: 4 -*-
- *
- * This Source Code Form is subject to the terms of the Mozilla Public
- * License, v. 2.0. If a copy of the MPL was not distributed with this
- * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
-
```
tonygermano, line 9: Why did spotless remove this comment block?

rbri: that was my fault, sorry
will have a look

tonygermano:
Should there be any concern about removing or changing the signatures of public methods from Hashtable? Not sure of the likelihood of that class being accessed outside of Rhino itself, but the class itself is also declared public. It's probably ok, but I still wanted to bring that up.

rbri:
Good point I think I can make the changes backward compatible by introducing new methods. Will provide an update

tonygermano:
Maybe deprecate the methods that are no longer used internally? I think the chances are pretty slim that your current changes are going to affect anyone considering that class was only created three years ago to support Map and Set.

rbri:
Yes 5hat was my plan....

PR #1025, opened 2021-09-10
Do not init standard properties twice 

rPraml:
I found a strange behaviour while fixing `instanceof` in #830.

My problem was, while `[] instanceof Array` retured true, `[] instanceof Object` returned false (at least in the NativeJavaListTest)

After some time of debugging, I noticed, that the NativeJavaListTest does the following:

```java
Global global = new Global();
global.init(ContextFactory.getGlobal());
Scriptable scope = context.initStandardObjects(global);
context.evaluateString(scope, "[] instanceof Object",...)
```
The problem is, that `global.init(ContextFactory.getGlobal());` creates all standard objects the first time, while `context.initStandardObjects(global)` will overwrite the standard objects. (Note: initStandardObjects will return the `global` value and no new instance)
So, we have 2 "Object" prototypes:
`scope.get("Object", scope).prototypeProperty` will return an other instance than `ScriptableObject.getObjectPrototype(scope)` 
This has to do something with 
```java
        if (scope instanceof TopLevel) {
            ((TopLevel) scope).cacheBuiltins(scope, sealed);
        }
```
at the end of the `initSafeStandardObjects` method.

In the first call of "initStandardObjects",, the cache is empty, so
- Object-Prototype is created (instance 1)
- Array-prototype is created (requires Object-prototype and gets instance 1 - no cache hit)
- ...
- cacheBuiltins is called.

on the second call of "initStandardObjects", 
- Object-Prototype is created (instance 2)
- Array-prototype is created (requires Object-prototype and gets instance 1 - from cache !)
- ...
- cacheBuiltins is called. (will update cached value to instance 2, but this is too late)

What would you suggest as solution:

**Solution 1:** Not calling "initStandardObjects" twice?
I would add an assert/exception in ScriptRuntime.initSafeStandardObjects:
```java
    public static ScriptableObject initSafeStandardObjects(
            Context cx, ScriptableObject scope, boolean sealed) {
        if (scope == null) {
            scope = new NativeObject();
        } else {
            assert !scope.has("Object",scope) : "already initialized";
        }
```

**Solution 2:** Clearing the cache, when "initStandardObjects" is called a second time:
```java
    public static ScriptableObject initSafeStandardObjects(
            Context cx, ScriptableObject scope, boolean sealed) {
        if (scope == null) {
            scope = new NativeObject();
        } else if (scope instanceof TopLevel) {
            ((TopLevel)scope).clearCache(); // will set TopLevel.ctors and TopLevel.errors to null
        }
```

in my oppinion, it does not make sense, to initialize objects twice, so throwing an exception/printing a warning would make sense, but may break existing code

### DIFF
--- a/testsrc/org/mozilla/javascript/tests/JavaIterableTest.java
+++ b/testsrc/org/mozilla/javascript/tests/JavaIterableTest.java
@@ -6,6 +6,14 @@
  */
 package org.mozilla.javascript.tests;
 
+import static org.junit.Assert.*;
+
+import java.util.ArrayList;
+import java.util.LinkedHashMap;
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
 import junit.framework.TestCase;
 import org.junit.Test;
 import org.mozilla.javascript.Context;
@@ -14,15 +22,6 @@ import org.mozilla.javascript.EcmaError;
 import org.mozilla.javascript.NativeArray;
 import org.mozilla.javascript.Scriptable;
 import org.mozilla.javascript.tools.shell.Global;
-
-import java.util.ArrayList;
-import java.util.LinkedHashMap;
-import java.util.LinkedHashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
-
-import static org.junit.Assert.*;
 
 public class JavaIterableTest extends TestCase {
     protected final Global global = new Global();
@@ -35,9 +34,9 @@ public class JavaIterableTest extends TestCase {
     public void testMap() {
         Map map = new LinkedHashMap();
         String script =
-            "var a = [];\n" +
-            "for (var { key, value } of value.entrySet()) a.push(key, value);\n" +
-            "a";
+                "var a = [];\n"
+                        + "for (var { key, value } of value.entrySet()) a.push(key, value);\n"
+                        + "a";
 
         NativeArray resEmpty = (NativeArray) runScript(script, map);
         assertEquals(0, resEmpty.size());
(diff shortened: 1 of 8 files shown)


testsrc/org/mozilla/javascript/tests/JavaIterableTest.java
```diff
@@ -146,19 +139,23 @@ public void testSet() {
...
-            context.setLanguageVersion(Context.VERSION_ES6);
-            Scriptable scope = context.initStandardObjects(global);
-            scope.put("value", scope, Context.javaToJS(value, scope));
-            return context.evaluateString(scope, scriptSourceText, "", 1, null);
-        });
+        return ContextFactory.getGlobal()
+                .call(
+                        context -> {
+                            context.setLanguageVersion(Context.VERSION_ES6);
+                            Scriptable scope = context.newObject(global);
```
rPraml, line 156: was `Scriptable scope = context.initStandardObjects(global);`
this is the only code change in this class, which is wrong, as it does not create a new scope, but initializes the standardObjects a second time.
I fixed this also in other tests

tonygermano: I haven't tested it, but can that line be removed completely? I don't think it's necessary to create a new object as global already refers to an initialized scope. On the other hand, should a new global scope be initialized on each test instead of reusing the same one? I do see the problem you were trying to fix with the single global instance getting reinitialized on each test.

rPraml: > I haven't tested it, but can that line be removed completely?

for this test case, it would work. Or in other words, it was just like that:
`context.initStandardObjects` returns `global` so the code looks like this:
```java
context.initStandardObjects(global);
Scriptable scope = global;
```

> On the other hand, should a new global scope be initialized on each test instead of reusing the same one?

it depends on the test. If the test itself modifies the scope, it would be better to initialize a new one. In this test, the next line puts `value` into the scope (`scope.put("value", scope, Context.javaToJS(value, scope));`) - as this is done on each test, I would agree, that it is not neccessary in this test, but the origin problem was, that `initStandardObjects` mess up some objects if it is called multiple times

rPraml:
@tonygermano I would suggest this [fix](https://github.com/mozilla/rhino/pull/1025/files#diff-58ee1701deb704ddb7d0ff0d41aef3275244d978ad279b1e4f1b4a61ff9e563d):

I've added a simpler test, that shows the originial problem more clearly: [InitStandardObjectTest](https://github.com/mozilla/rhino/pull/1025/files#diff-ccbc702c46f886e1b48e501b8d585cb480a2cac113435d0638a9ec5799dbf317R42)

The complex test [NestedContextPrototypeTest](https://github.com/mozilla/rhino/pull/1025/files#diff-ccbc702c46f886e1b48e501b8d585cb480a2cac113435d0638a9ec5799dbf317R42) showed me, that I am using `initStandardObjects` the wrong way (I can remove that test, if you propose, or keep it for documentation)

I also modified the NativeXXXTests, which have used `initStandardObjects` the wrong way. I think the intention of the original test writer @syjer was, to create a new isolated scope on each test run. So I would suggest, to keep this change. (Of course, the test will work without this line and act on `global` scope)

gbrail:
Thanks -- it looks like you went with the second option, which is to clear the cache so that calling "initStandardObjects" twice does not result in non-functional scripts.

I think that's a good choice, given that there is lots of code out there.

Also, thanks for the comprehensive tests!

If no one else has any objections then I'm OK with merging this.

rbri:
@gbrail lets merge....

gbrail:
Agreed -- this looks good. Thanks!

PR #1031, opened 2021-09-17
ADD: delete method and set length property to NativeJavaList

rPraml:
This change is extracted from #830 and adds support for delete and setting the length
- `javaList.length = x` I think this is clear. It will grow or truncate the underlying java list to given size.
- `delete javaList[x]` which effectively set the value in the wrapped javaList to null. (It will not remove elements, so the size of the list will stay the same)

Note: on the java side, we can not distinguish between `null` and `undefined`, So increasing the size will add `null` values in java vs. empty slots in JS.
The same is also true for `delete javaList[x]`, it will be equivalent with `javaList[x] = null`

@tuchida you always have good objections:
I know the current implemetation is not exactly what the spec describes:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/delete
Especially in strict mode, `delete javaList['length']` etc. should throw `TypeError` and should return `true` if element is deleted.

What I want to achive is, that `delete` on a javaList "works". Currently it is completely ignored. So I think, this implementation is an improvement.

(Btw: The same is for NativeJavaObject. Deleting properties should throw a TypeError)

Cheers
Roland

### DIFF
--- a/src/org/mozilla/javascript/NativeJavaList.java
+++ b/src/org/mozilla/javascript/NativeJavaList.java
@@ -40,6 +40,12 @@ public class NativeJavaList extends NativeJavaObject {
         return super.has(index, start);
     }
 
+    public void delete(int index) {
+        if (isWithValidIndex(index)) {
+            list.set(index, null);
+        }
+    }
+
     @Override
     public boolean has(Symbol key, Scriptable start) {
         if (SymbolKey.IS_CONCAT_SPREADABLE.equals(key)) {
@@ -87,6 +93,15 @@ public class NativeJavaList extends NativeJavaObject {
         super.put(index, start, value);
     }
 
+    @Override
+    public void put(String name, Scriptable start, Object value) {
+        if (list != null && "length".equals(name)) {
+            setLength(value);
+            return;
+        }
+        super.put(name, start, value);
+    }
+
     private void ensureCapacity(int minCapacity) {
         if (minCapacity > list.size()) {
             if (list instanceof ArrayList) {
@@ -95,6 +110,20 @@ public class NativeJavaList extends NativeJavaObject {
             while (minCapacity > list.size()) {
                 list.add(null);
             }
+        }
+    }
+
+    private void setLength(Object val) {
+        double d = ScriptRuntime.toNumber(val);
+        long longVal = ScriptRuntime.toUint32(d);
+        if (longVal != d || longVal > Integer.MAX_VALUE) {
+            String msg = ScriptRuntime.getMessageById("msg.arraylength.bad");
+            throw ScriptRuntime.rangeError(msg);
+        }
+        if (longVal < list.size()) {
+            list.subList((int) longVal, list.size()).clear();
+        } else {
+            ensureCapacity((int) longVal);
         }
     }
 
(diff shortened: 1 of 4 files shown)


tuchida (approved):
Sounds good.

tuchida:
By the way, what do you think [Object.freeze](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/freeze) and [property descriptor](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/getOwnPropertyDescriptor) should be?
```js
var a = Object.freeze([1, 2, 3]);
a[1] = 123;
a;  // 1,2,3

var al = java.util.ArrayList([1, 2, 3]);
Object.freeze(al); // `-version 200` is required.
al[1] = 123;
al; // [1.0, 123.0, 3]

```

src/org/mozilla/javascript/NativeJavaList.java
```diff
@@ -98,6 +113,20 @@ private void ensureCapacity(int minCapacity) {
...
+        double d = ScriptRuntime.toNumber(val);
+        long longVal = ScriptRuntime.toUint32(d);
+        if (longVal != d || longVal > Integer.MAX_VALUE) {
+            String msg = ScriptRuntime.getMessageById("msg.arraylength.bad");
+            throw ScriptRuntime.rangeError(msg);
+        }
+        if (longVal < list.size()) {
+            list.subList((int) longVal, list.size()).clear();
+        } else {
+            ensureCapacity((int) longVal);
```
tonygermano, line 126: This seems dangerous since typical java Lists don't support sparse layouts like javascript arrays do. If you set length to a very large value on a javascript array it's no big deal, but do the same on an ArrayList, and you're probably going to run out of memory.

Should we at least add a javadoc description for the class that describes the additional behaviors which NativeJavaList adds to NativeJavaObject and potential pitfalls when using these features as if Lists were arrays? Is this stuff documented anywhere else other than in PRs and commit messages?

gbrail: Improving the Javadoc sounds great to me.

In general, the "embedding guide" is very very old and hasn't it been archived by Mozilla? Do we have any volunteers to resurrect it?

rPraml: How should we document this? JavaDoc, MD, wiki-page?

I would prefer a markdown documentation like this one: https://github.com/oracle/graaljs/blob/master/docs/user/JavaInteroperability.md

tonygermano: Personally, I think the javadoc should contain basic important information for using the class. Wiki/user guide type documentation is good for going into more detail if someone is willing to write that. As we know, external documentation is not always easily accessible and developers should have some idea how a class works without having to closely study the code or pull up external docs.

rPraml:
I've added some documentation and using add instead of grow/set if last element is added to list

rPraml:
@gbrail @tuchida @tonygermano 
I wanted to ask if the PR can be merged or if there are still objections?

tonygermano:
I'm ok with the added documentation.

Issue #915, opened 2021-05-29
java.lang.VerifyError in language/statements/try of test262


tuchida:
When I enable language/statements/try in test262, I get the following error.

```
org.mozilla.javascript.tests.Test262SuiteTest > test262Case[js=language/statements/try/S12.14_A7_T1.js, opt=0, strict=false] FAILED
    java.lang.VerifyError at Test262SuiteTest.java:217

...

org.mozilla.javascript.tests.Test262SuiteTest > test262Case[js=language/statements/try/S12.14_A7_T3.js, opt=0, strict=false] FAILED
    java.lang.VerifyError at Test262SuiteTest.java:217
```
With JavaScript code like this
```js
try{
  try{
    throw "ex1";
  }finally{
    try{
      throw "ex2";
    }catch(er1){
    }
  }
}catch(er2){
}
```
I got an error like the following
```
Exception in thread "main" java.lang.VerifyError: Stack map does not match the one at exception handler 91
Exception Details:
  Location:
    org/mozilla/javascript/gen/_stdin__1._c_script_0(Lorg/mozilla/javascript/gen/_stdin__1;Lorg/mozilla/javascript/Context;Lorg/mozilla/javascript/Scriptable;Lorg/mozilla/javascript/Scriptable;[Ljava/lang/Object;)Ljava/lang/Object; @91: nop
  Reason:
    Type top (current frame, locals[5]) is not assignable to 'java/lang/Object' (stack map, locals[5])
  Current Frame:
    bci: @52
    flags: { }
    locals: { 'org/mozilla/javascript/gen/_stdin__1', 'org/mozilla/javascript/Context', 'org/mozilla/javascript/Scriptable', 'org/mozilla/javascript/Scriptable', '[Ljava/lang/Object;' }
    stack: { 'org/mozilla/javascript/JavaScriptException' }
  Stackmap Frame:
    bci: @91
    flags: { }
    locals: { 'org/mozilla/javascript/gen/_stdin__1', 'org/mozilla/javascript/Context', 'org/mozilla/javascript/Scriptable', 'org/mozilla/javascript/Scriptable', '[Ljava/lang/Object;', 'java/lang/Object', top, 'org/mozilla/javascript/Scriptable' }
    stack: { 'java/lang/Throwable' }
```
https://app.circleci.com/pipelines/github/mozilla/rhino/394/workflows/06f40ab3-873d-4ffd-b26f-753722081c8a/jobs/400/tests

[...]

PR #1032, opened 2021-09-20
Optimize compilespeed

rPraml:
Hello, I've tried to improve the compile speed, especially, if many exception handlers are involved (see `CodegenTest.testManyExceptionHandlers`)

This test takes in original 33s on my machine.

- step 1: [Use binary search for getSuperBlockFromOffset](https://github.com/mozilla/rhino/commit/5ebfa5dc8456553174ccfbb935ee8ae9a61e2857)
As long as I see, the `itsSuperBlocksStarts` array contains the starts in sorted order, so I use binary search in getSuperblockFromOffset. 

This brings a performance boost down to 3.5s

- step 2: [narrowed lookup range for big exception tables](https://github.com/mozilla/rhino/commit/f0831e048bd9f23e6d51c18f10e5dec2383dc17f)
I also narrowed the lookup range for big exception tables and throw away the exception table entries, that are definitively irrelevant. See https://github.com/mozilla/rhino/commit/f0831e048bd9f23e6d51c18f10e5dec2383dc17f
Now the test runs in 1.2s

- step 3: [Optimized finding locals](https://github.com/mozilla/rhino/commit/73537d69ac15ec40cdbfc704ff8dbea9bb57e059)
A further optimization in finding the locals brings down the test case to 0.6s

- step 4: [TypeInfo.merge do simplest check first](https://github.com/mozilla/rhino/commit/1a7efb599b2dcd6bdd253087de90c9e55682f63b)
This is not really measurable, but showed up as "hot" method in the profiler, so I tried to optimize it also

The "all test time" goes down from about 11:50 to 11:20

### DIFF
--- a/src/org/mozilla/classfile/ClassFileWriter.java
+++ b/src/org/mozilla/classfile/ClassFileWriter.java
@@ -1598,27 +1598,25 @@ public class ClassFileWriter {
             // locals as the killed block's locals. Ignore uninitialized
             // handlers, because they will also be killed and removed from the
             // exception table.
+            int sbStart = sb.getStart();
             for (int i = 0; i < itsExceptionTableTop; i++) {
                 ExceptionTableEntry ete = itsExceptionTable[i];
                 int eteStart = getLabelPC(ete.itsStartLabel);
-                // we can sort out exception table entries immediately,
-                // that are before or behind current sb
-                if (sb.getEnd() <= eteStart || sb.getStart() >= getLabelPC(ete.itsEndLabel)) {
-                    continue;
-                }
-                if (sb.getStart() > eteStart) {
+                // this is "hot" code and it has been optimized so that 
+                // there are not too many function calls
+                if (sbStart > eteStart && sbStart < getLabelPC(ete.itsEndLabel)) {
                     int handlerPC = getLabelPC(ete.itsHandlerLabel);
                     SuperBlock handlerSB = getSuperBlockFromOffset(handlerPC);
                     locals = handlerSB.getLocals();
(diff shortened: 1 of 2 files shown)


gbrail:
This all looks very good to me. I was hoping it'd speed up test execution for Rhino itself, but sadly it didn't really make a difference. Do you or does anyone else have an example of some complex JavaScript that now compiles a bunch faster so that we can document that for the record?

rPraml:
@gbrail I've investigated much time in optimizing test execution (compile and run speed) and did only find a few places where optimization is done. (see other PRs) Unfortunately, most of them is not really measurable.
I've only found the `CodegenTest.testManyExceptionHandlers` test, where this change makes a huge difference:

As far as I see, the amount of exception handlers goes into compile speed with a complexity of ~ `O(n^3)` and the CodegenTest creates 1000 of them. If you double the number, the test takes ~8 times longer.

There were 3 nested for - loops, one of them could be replaced with the binary search, so that you'll get `O(n^2 log(n))`

All in all, compile speed will only increase if you have many of exception blocks (which may be an edge case)

BTW: You can increase the testexecution speed by factor ~2 if jacoco is disabled.

tuchida:
- #915

Rhino is not currently testing the try statement in test262.
I added `language/statements/try` to `test262.properties`, and found that the number of errors was the same before and after the modification.
```
language/statements/try 117/195 (60.0%)
```

gbrail:
Thank you! I merged this manually because I wanted fewer commits but I wanted to also separate the spotless commits from the real work.

PR #1033, opened 2021-09-20
Avoid unnecessary arrayCopy in Kit.readStream

rPraml:
This is a small performance optimization, when Kit.readStream is called with the correct stream length. This is always the case, when a file is read. Then fileSize is passed as buffer size.

the old code did the following:

- declare buffer with file size
- for loop step # 1  
- perform is.read(...)  (reads whole file in the buffer)
- `cursor == buffer.length` -> **grow buffer array**
- for loop step # 2 
- perform is.read(...) no more to read, break loop
- **shrink buffer array**

the new code tries to avoid the two unnecessary arrayCopies by checking if the stream is EOS on first loop.

### DIFF
--- a/src/org/mozilla/javascript/Kit.java
+++ b/src/org/mozilla/javascript/Kit.java
@@ -12,19 +12,14 @@ import java.io.InputStream;
 import java.io.Reader;
 import java.util.Map;
 
-/**
- * Collection of utilities
- */
-
-public class Kit
-{
-    public static Class<?> classOrNull(String className)
-    {
+/** Collection of utilities */
+public class Kit {
+    public static Class<?> classOrNull(String className) {
         try {
             return Class.forName(className);
-        } catch  (ClassNotFoundException ex) {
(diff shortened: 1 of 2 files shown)


rbri (approved):

rbri:
Had a second look at this, do you see any notable performance improvement using this - i  guess this is a optimization for a really special case?

rPraml:
@rbri I tried to speed up test execution and discovered this method when building the test cases in `Test262SuiteTest.test262SuiteValues()`
It calles `Test262Case fromSource` for all ~20000 test files. In this code path, capacity hint is the file size, so you will save 40000 times arrayCopy,
With this change the Test262SuiteTest.test262SuiteValues() goes down from 13s to 9s.
Unfortunately, the Test262 runs several minutes, so the overall performance gain are only a few percent.

gbrail:
Thanks -- I think it makes sense to do what we can to make the tests run faster!

PR #1034, opened 2021-09-20
Optimized XmlProcessor toXmlNewlines

rPraml:
This is a minor optimization in XmlProcessor.toXmlNewlines
(I did some performance measurements with JProfiler and this showed up as a hotspot during MozillaSuiteTest)


rbri (approved):

rbri:
Thanks

Issue #1047, opened 2021-10-07
Timer resolution issue (bugzilla 363258)


carlosame:
When I execute the test suite for the current `master` branch with Java 11 on a Windows 10 machine, I get this:

```
> Task :test

org.mozilla.javascript.tests.MozillaSuiteTest > runMozillaTest[1779, js=testsrc\tests\js1_5\extensions\regress-363258.js, opt=-1] STANDARD_OUT
    In "testsrc\tests\js1_5\extensions\regress-363258.js":
     FAILED! [reported from test()] Timer resolution : Expected value 'true', Actual value 'false'


org.mozilla.javascript.tests.MozillaSuiteTest > runMozillaTest[1779, js=testsrc\tests\js1_5\extensions\regress-363258.js, opt=-1] FAILED
    java.lang.AssertionError at MozillaSuiteTest.java:208

org.mozilla.javascript.tests.MozillaSuiteTest > runMozillaTest[2330, js=testsrc\tests\lc2\Methods\println-001.js, opt=-1] STANDARD_OUT
    output from test live/Methods/println-001.js

org.mozilla.javascript.tests.MozillaSuiteTest > runMozillaTest[4225, js=testsrc\tests\js1_5\extensions\regress-363258.js, opt=0] STANDARD_OUT
    In "testsrc\tests\js1_5\extensions\regress-363258.js":
     FAILED! [reported from test()] Timer resolution : Expected value 'true', Actual value 'false'


org.mozilla.javascript.tests.MozillaSuiteTest > runMozillaTest[4225, js=testsrc\tests\js1_5\extensions\regress-363258.js, opt=0] FAILED
    java.lang.AssertionError at MozillaSuiteTest.java:208

org.mozilla.javascript.tests.MozillaSuiteTest > runMozillaTest[4775, js=testsrc\tests\lc2\Methods\println-001.js, opt=0] STANDARD_OUT
    output from test live/Methods/println-001.js

org.mozilla.javascript.tests.MozillaSuiteTest > runMozillaTest[6670, js=testsrc\tests\js1_5\extensions\regress-363258.js, opt=9] STANDARD_OUT
    In "testsrc\tests\js1_5\extensions\regress-363258.js":
     FAILED! [reported from test()] Timer resolution : Expected value 'true', Actual value 'false'


org.mozilla.javascript.tests.MozillaSuiteTest > runMozillaTest[6670, js=testsrc\tests\js1_5\extensions\regress-363258.js, opt=9] FAILED
[...]

PR #1048, opened 2021-10-08
More robust version of regression test for bugzilla 363258 (#1047)

carlosame:
This new version should be more resilient against the effect of garbage collections, thread scheduling and perhaps some initialisations.

Let's see what Github's CI says.


gbrail:
Thanks  -- I'll try but in general I'm suspicious of all these tests. We have been gradually disabling many of these ancient "performance" tests because on a modern VM and in a cloud CI environment they don't really measure what they think they're measuring. What would you think of just disabling that test?

PR #1050, opened 2021-10-09
Upgrade deprecated Gradle configurations

carlosame:
The changes are required for the build being compatible with Gradle 8 (and avoid a warning with 7).


gbrail:
This is very helpful. Thanks!

PR #1054, opened 2021-10-14
Replace deprecated 'new Integer/Double/Long' by Integer/Double/Long.valueOf

rPraml:
The methods are deprecated with Java 9
https://docs.oracle.com/javase/9/docs/api/java/lang/Integer.html#Integer-int-
So I replaced it with the .valueOf method.

### DIFF
--- a/examples/Matrix.java
+++ b/examples/Matrix.java
@@ -6,7 +6,6 @@
 
 import java.util.ArrayList;
 import java.util.List;
-
 import org.mozilla.javascript.Context;
 import org.mozilla.javascript.Scriptable;
 import org.mozilla.javascript.ScriptableObject;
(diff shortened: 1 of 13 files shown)


carlosame:
> There are still a lot of occurrencies in https://github.com/mozilla/rhino/tree/master/testsrc/tests/src - should I fix them also

I'd leave those out for now, are these things even being compiled?

For a future PR that includes `Boolean` as a target, there are also a couple of `new Boolean()` here:
https://github.com/mozilla/rhino/blob/43793a43e7ee1eebfaa639b5f836672b4309898a/src/org/mozilla/javascript/NativeBoolean.java#L107-L108

rPraml:
> I'd leave those out for now, are these things even being compiled?

I think not, but I do not know exactly. There is a precompiled jar in the folder: https://github.com/mozilla/rhino/tree/master/testsrc/tests/src

gbrail:
Thank you! I merged this manually since I'm trying to squash long commit chains and simultaneously keep "spotless" commits separate from ones that do real work.

PR #1063, opened 2021-10-15
Account for BigInt in Hashtable (#1023)

### LINKED ISSUE #1023: Hashtable.Entry does not account for BigInt
https://github.com/mozilla/rhino/blob/7036623b8f1b4edb8450adbdd53d49dc727e403e/src/org/mozilla/javascript/Hashtable.java#L56-L63

```js
js> var bi = BigInt("100000000000000000000000000000000000000000000000000000000000000000000000000000000001")
js> var d = 1e+83
js> var m = new Map()
js> m.set(bi, 'bigint key')
[object Map]
js> m.get(d)
bigint key
js> m.set(d, 'number key')
[object Map]
js> m.get(bi)
number key
```

carlosame:
The description of issue #1023 practically says it all. The code in this PR just checks for `BigInteger` when the map entry is created.

The type comparison for `Double` uses `instanceof`, it could be marginally faster if `getClass()` was used instead (`k` is guaranteed to be non-null inside the check for `Number`, and `Double` is `final`).

I'm not sure about the name `NativeBigIntMapTest`: `BigInt` is native but what is actually being tested doesn't look very "native". The test checks that `BigInt` is always handled as its own type of object, while other numbers are handled differently and for example `9007199254740992` is considered the same as `9007199254740993` (being handled as the same `Double`).

Closes #1023


carlosame:
I'm not yet done examining the `test262` directories, but so far I have found [`implementation-contributed/v8/mjsunit/harmony/bigint/comparisons.js`](https://github.com/tc39/test262/blob/8b4382d2f2f80e3da8b87201a7444641c616f2f1/implementation-contributed/v8/mjsunit/harmony/bigint/comparisons.js) which is not under the `test` subdirectory and therefore not enabled in `test262.properties`. An excerpt from it:
```javascript
const zero = BigInt(0);
const another_zero = BigInt(0);
const one = BigInt(1);
const another_one = BigInt(1);

  assertTrue(new Map([[zero, 42]]).has(zero));
  assertTrue(new Map([[zero, 42]]).has(another_zero));

  assertFalse(new Map([[zero, 42]]).has(+0));
  assertFalse(new Map([[zero, 42]]).has(-0));

  assertFalse(new Map([[+0, 42]]).has(zero));
  assertFalse(new Map([[-0, 42]]).has(zero));

  assertTrue(new Map([[one, 42]]).has(one));
  assertTrue(new Map([[one, 42]]).has(another_one));
```
which is relevant to this fix, but the rest of the test seem to require some testing framework (and the whole directory is currently outside of the focus of the 262 Rhino tests).

tuchida:
Thank you for fixing the bug!
The fix looks fine, but I'd like to see a test of `Set` as well as `Map` just in case.

`WeakSet` and `WeakMap` don't seem to be relevant because they don't use `Hashtable` and their keys are only objects.
https://github.com/mozilla/rhino/blob/900b035145f29a466f3a9883c8ea67e871a1e159/src/org/mozilla/javascript/NativeWeakMap.java#L109-L111

src/org/mozilla/javascript/Hashtable.java
```diff
@@ -48,9 +49,16 @@
...
         Entry(Object k, Object value) {
-            if ((k instanceof Number) && (!(k instanceof Double))) {
-                // Hash comparison won't work if we don't do this
-                this.key = Double.valueOf(((Number) k).doubleValue());
+            if (k instanceof Number) {
+                if (k instanceof Double || k instanceof BigInteger) {
+                    // BigInteger needs to retain its own type, due to
+                    // "If Type(x) is different from Type(y), return false." in
+                    // ecma262/multipage/abstract-operations.html#sec-samevaluezero
+                    this.key = k;
```
rbri, line 57: Sorry but maybe one minor point. This method uses sometimes the this.key notation and sometimes only key.
I think we can cleanup the code here a bit and always use key without 'this.'

Many thanks for the fix and the test cases.

rbri (approved):

carlosame:
I've been deferring this a bit (apart from other Rhino issues that I got involved with, I manage 17 open source repositories...), but finally I opened an issue at tc39/test262 and today that test suite gained related tests in tc39/test262#3339.

PR #1068, opened 2021-10-18
Move SecurityManager relevant parts to SecurityBridge

rPraml:
To be prepared for JEP411, I've started to move all security relevant stuff to "SecurityBridge_SecurityManager".
I've also made other classes deprecated, that relies on SecurityManger (like RhinoSecurityManager) or other deprecated classes, that are marked for removal in JDK17

There are now only two usages of deprecated AccessController left:
grafik
https://github.com/mozilla/rhino/blob/master/src/org/mozilla/javascript/ContextFactory.java#L338
https://github.com/mozilla/rhino/blob/master/src/org/mozilla/javascript/LazilyLoadedCtor.java#L86

This relates to
https://github.com/mozilla/rhino/discussions/1045

### DIFF
--- a/src/org/mozilla/javascript/SecurityUtilities.java
+++ b/src/org/mozilla/javascript/SecurityUtilities.java
@@ -16,8 +16,6 @@ import java.security.ProtectionDomain;
  * or <code>org.mozilla.javascript.SecurityBridge_NoOp</code>, depending if <code>
  * java.lang.SecurityManager</code> is present or not.
  *
- * <ul>
- *
  * @author Attila Szegedi
  * @author Roland Praml, FOCONIS AG
  */


gbrail:
This looks OK to me and I think it's OK to have in 1.7.14 because it's preparing us better for the future. Do any of the other folks who are very familiar with the security manager want to try it out?

src/org/mozilla/javascript/SecurityBridge_SecurityManager.java
```diff
@@ -0,0 +1,103 @@
...
+import java.security.ProtectionDomain;
+
+/**
+ * Code moved from {@link SecurityUtilities}. This implementation makes use of {@link
+ * SecurityManager} and {@link AccessController} and so on, which is deprecated with JDK17 (see <a
+ * href='https://openjdk.java.net/jeps/411'>JEP411</a>) - so all related 'java.security' stuff
+ * should be routed over this class, so that it could be easily replaced by an other implementation
+ * like {@link SecurityBridge_NoOp}.
+ *
+ * <p>This implementation should be work up to JDK17
```
carlosame, line 23: According to JEP 411, the only API that will not work out the box in Java 18 will be `System.setSecurityManager()` (to make it work, one needs to set `java -Djava.security.manager=allow`). Rhino only calls `System.setSecurityManager()` in `SecurityControllerTest`, so this implementation should work in Java 18 as well. It is unclear when the actual APIs are going to be removed.

I'd suggest setting `java.security.manager=allow` in the tests section of `build.gradle` (or I could do that in a separate PR) to deal with `SecurityControllerTest`, and avoid mentioning which Java release will/won't work.

src/org/mozilla/javascript/SecurityUtilities.java
```diff
@@ -6,68 +6,79 @@
...
+/** @author Attila Szegedi */
+public abstract class SecurityUtilities {
+    static final SecurityBridge bridge = makeBridge();
+
+    private static SecurityBridge makeBridge() {
+
+        String[] classNames = {
+            "org.mozilla.javascript.SecurityBridge_custom",
+            // Check, if SecurityManager class exists
+            // TODO: Shoud we check JDK version here?
```
carlosame, line 32: It is not known when the actual APIs are going to be removed, so I would not check the Java version.

src/org/mozilla/javascript/SecurityBridge_SecurityManager.java
```diff
@@ -0,0 +1,103 @@
...
+    @Override
+    public String getSystemProperty(final String name) {
+        return AccessController.doPrivileged(
+                new PrivilegedAction<String>() {
+                    @Override
+                    public String run() {
+                        return System.getProperty(name);
+                    }
+                });
+    }
```
carlosame, line 47: The `getSystemProperty()` method is public and belongs to a public class in an exported package (I mean once Rhino is modularised), allowing the retrieval of system properties with the privileges of Rhino. The comment says "Use it only when the property is used from within Rhino code and is not passed out of it" but the code is not enforcing it: any code with access to the classes in the classpath/modulepath would be allowed to do that.

Probably not a big deal, but I wonder if the security bridge classes could be made package-visible.

rPraml: This could be really a problem

rPraml:
This PR is probably outdated.

Anyway, the SecurityManager stuff should be removed soon (with Rhino 2.0?)
Compiling Rhino with Java 17 gives you numerous warnings about the deprecated SecurityManager API

rPraml:
Closing this PR. There is already #1353

PR #1070, opened 2021-10-19
switch to a vmbridge for jdk_11 and above

rbri:
There is this concept of the VMBride to handle JDK specifics - let us use this for the jdk11 specific stuff to follow a unique path.

### DIFF
--- a/src/org/mozilla/javascript/VMBridge.java
+++ b/src/org/mozilla/javascript/VMBridge.java
@@ -10,23 +10,24 @@ package org.mozilla.javascript;
 
 import java.lang.reflect.AccessibleObject;
 import java.lang.reflect.Method;
+import java.util.Arrays;
+import java.util.Map;
 
-public abstract class VMBridge
-{
+public abstract class VMBridge {
 
     static final VMBridge instance = makeInstance();
 
-    private static VMBridge makeInstance()
-    {
+    private static VMBridge makeInstance() {
         String[] classNames = {
             "org.mozilla.javascript.VMBridge_custom",
+            "org.mozilla.javascript.jdk11.VMBridge_jdk11",
             "org.mozilla.javascript.jdk18.VMBridge_jdk18",
         };
         for (int i = 0; i != classNames.length; ++i) {
             String className = classNames[i];
             Class<?> cl = Kit.classOrNull(className);
             if (cl != null) {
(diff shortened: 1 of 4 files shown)


src/org/mozilla/javascript/VMBridge.java
```diff
@@ -10,23 +10,24 @@
...
             "org.mozilla.javascript.VMBridge_custom",
+            "org.mozilla.javascript.jdk11.VMBridge_jdk11",
             "org.mozilla.javascript.jdk18.VMBridge_jdk18",
         };
         for (int i = 0; i != classNames.length; ++i) {
             String className = classNames[i];
             Class<?> cl = Kit.classOrNull(className);
             if (cl != null) {
-                VMBridge bridge = (VMBridge)Kit.newInstanceOrNull(cl);
+                VMBridge bridge = (VMBridge) Kit.newInstanceOrNull(cl);
```
rPraml, line 30: I see no code that distinguish between the different JVM versions.
So also for a jdk8 the first instance that is used 
(or do we get a NoClassDefFound-error when VMBridge_jdk11 when instanttiated on jdk8)

src/org/mozilla/javascript/VMBridge.java
```diff
@@ -88,22 +85,50 @@ private static VMBridge makeInstance()
...
+    protected abstract Object newInterfaceProxy(
+            Object proxyHelper,
+            ContextFactory cf,
+            InterfaceAdapter adapter,
+            Object target,
+            Scriptable topScope);
+
+    public abstract void discoverPublicMethods(Class<?> clazz, Map<MethodSignature, Method> map);
+
+    public static final class MethodSignature {
```
carlosame, line 107: `MethodSignature` is being used only to satisfy generics in a `Map<MethodSignature, Method>` but otherwise is used just as an `Object`. Here, it becomes `public` and hence part of the public API. That's one of the reasons why I kept this stuff as a package-visible subclass to `JavaMembers` (and @gbrail mentioned something along those lines when merging).

The current `master` code does not change the public API of Rhino, and would leave it as it is for 1.7.14. Once that release is out, the `VMBridge` API could be changed if there was an obvious gain from it, like trying to address non-public reflective accesses. I have in mind something along those lines although my approach would keep `JavaMembers_jdk11` as a class.

PR #1072, opened 2021-10-21
Allow access to protected methods in modular Java

carlosame:
Currently, accessing `protected` methods from Java 16+ requires adding `add-opens` JVM arguments, as mentioned in the README. This PR aims at removing that requirement, at least when the methods are called from a declared subclass (which IMHO is the only legitimate circumstance to call a protected method, otherwise people should use `add-opens`).

A 1.7.14 Release Candidate is around the corner, so I decided to open this PR ahead of time (it is not finished, although it represents an improvement even as it is now) so the Rhino people can comment on it.

The test only passes when interpreted, not compiled, and haven't figured that out yet (just wrote the thing this afternoon). The path would be clearer for me if a library like ASM was used (as I mentioned elsewhere), but one should be able to do it using the current Rhino infrastructure. You know the compilation infrastructure far better than me and any suggestion would be appreciated.

Please ask any questions about the code, it changes the API of `VMBridge` (and even the way to instantiate it) but for good reasons.


src/org/mozilla/javascript/VMBridge.java
```diff
@@ -65,15 +80,18 @@ private static VMBridge makeInstance()
...
      * VMBridge instance should try to workaround that via, for example,
      * calling method.setAccessible(true) when it is available.
      * The implementation is responsible to catch all possible exceptions
      * like SecurityException if the workaround is not available.
      *
+     * @param accessible the constructor.
      * @return true if it was possible to make method accessible
      *         or false otherwise.
+     *
+     * @return the method, if it was possible to find one accessible or {@code null} otherwise.
```
carlosame: There is a double `@return` in the Javadoc, re-pushing.

gbrail:
Sorry, I wanted to get an RC out the door and I didn't look at this sufficiently or merge it. If you all agree that this is safe we can include it before the main release.

carlosame:
New version adds:

- A few `@SuppressWarnings("deprecation")` annotations covering the `isAccessible()` and `getProxyClass()` method calls that are present in the patched code of `JavaMembers` and `VMBridge_jdk8`. Other deprecation warnings are left.
- Improved `SecurityException` handling and reporting, not only for Java 11+ but also for 8.

p-bakker:
@carlosame I see you closed this PR. May I ask why? Aren't these changes relevant anymore?

carlosame:
> @carlosame I see you closed this PR. May I ask why? Aren't these changes relevant anymore?

I closed it because PRs aren't needed anymore.

My Rhino fork (and the relevant branches) is still available and I do not plan to remove it. In any case, anyone can prepare his "own version" of the patches, one that only differs from my commits in whitespace and little more, and commit the thing directly.

So I do not see where the problem is.

PR #1083, opened 2021-11-08
Improve NativeJavaList to behave more like NativeArray 2

rPraml:
This PR replaces https://github.com/mozilla/rhino/pull/830
(most of #830 is already meged)
This adds some fixes in sameZero and shallowEx that were discovered in testArrayIndexOf/lastIndexOf and testArrayIncludes

Also seems to make #561 obsolete, see https://github.com/mozilla/rhino/pull/561#issuecomment-964124846


rPraml:
@gbrail (or anyone else) can you re-run the CI builds. They failed, because https://twitter.com/toongeens/status/1458053129677819904
(or should I push an empty commit)

tonygermano:
I don't think this is correct. A java object should not be strictly equal to a javascript primitive.

The following behavior is observed with your changes
```js
js> new String('foo') === 'foo' // should be false
false
js> new java.lang.String('foo') === 'foo' // should be false
true
```

gbrail:
Thanks for working on this -- sounds like we need to have a bit more discussion on this.

I am going to hold off on merging this until after we have all had a chance to test 1.7.14 RC1 and I've been able to publish a non-RC version of 1.7.14. (I think that this project is too small for us to have branches for old releases that diverge from the trunk, and that kind of thing...)

rPraml:
@tonygermano thanks for feedback, I'm also unsure, which is correct or expected behaviour. I've made some tests at current master branch (1.7.14-RC1)
```javascript
// Test 1:
var list = []
list[0] = 'foo'
list[1] = 'foo'
list[0] === list[1]; // returns true
list[2] = 3.1
list[3] = 3.1
list[2] === list[3]; // returns true


// Test 2:
list[0] = 'averylongstring'
list[1] = 'averylongstring'
list[0] === list[1]; // returns true
list[2] = 3.1
list[3] = 3.1
list[2] === list[3]; // returns false in master branch
list[2] = list[3]
list[2] === list[3]; // returns true

// Test 3:
var list = []
list[0] = 'foo'
list[1] = 'f'
list[1] += 'oo'
list[0] === list[1]; // returns true

// Test 4:
list[0] = 'foo'
list[1] = 'f'
list[1] += 'oo'
list[0] === list[1]; // returns false in master branch
```

The problem with the 'foo' string is, that the same object instance is put into the java.list, so `list.get(0) == list.get(1)` on the java side. This is not true if the string is a result of a computation or if you use a double value.
Then we run into the `==` check in shallowEq
```java
            if (x instanceof Wrapper && y instanceof Wrapper) {
                return ((Wrapper) x).unwrap() == ((Wrapper) y).unwrap();
            }
```

What do you think is the best way to solve this issue - or should we accept it as a problem of liveConnect

tonygermano:
If you set the javaPrimitiveWrap to false on the Context's WrapFactory, all of your tests in https://github.com/mozilla/rhino/pull/1083#issuecomment-973928247 do what you expect without any code changes or breaking object comparisons.

It will end up returning the java Double and String objects as primitive numbers and strings instead so that the comparisons work as expected.

p-bakker:
Converted to draft, as currently failing the CI pipeline

PR #916, opened 2021-05-29
Try optional catch binding

tuchida:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/try...catch#the_exception_identifier
```js
  try {
    JSON.parse(text);
    return true;
  } catch {
    return false;
  }
```

The omission of the exception value is implemented, but there are another error. ref. #915
We need to either fix #915 or ignore `VerifyError`.

Closes #965

PR #1084, opened 2021-11-10
Try setting superblock locals from an initialised exception handler

### LINKED ISSUE #915: java.lang.VerifyError in language/statements/try of test262

carlosame:
The patch to `ClassFileWriter` contains two hunks: the first one is the real fix, the second unmaps dead exception handlers (that generally happen with nested `finally` clauses) which otherwise may trigger JRE verifier errors. That second hunk could be removed if the performance hit is considered undesirable.

Closes #915.

### DIFF
--- a/src/org/mozilla/classfile/ClassFileWriter.java
+++ b/src/org/mozilla/classfile/ClassFileWriter.java
@@ -1605,7 +1605,11 @@ public class ClassFileWriter {
                     int handlerPC = getLabelPC(ete.itsHandlerLabel);
                     SuperBlock handlerSB = getSuperBlockFromOffset(handlerPC);
                     locals = handlerSB.getLocals();
-                    break;
+                    if (handlerSB.isInitialized()) {
+                        break;
+                    } else {
+                        continue;
+                    }
                 }
                 if (eteStart > sbStart && eteStart < sb.getEnd()) {
                     int handlerPC = getLabelPC(ete.itsHandlerLabel);
@@ -1623,7 +1627,7 @@ public class ClassFileWriter {
             for (int i = 0; i < itsExceptionTableTop; i++) {
                 ExceptionTableEntry ete = itsExceptionTable[i];
                 int eteStart = getLabelPC(ete.itsStartLabel);
-                if (eteStart == sb.getStart()) {
+                if (eteStart == sb.getStart() || getLabelPC(ete.itsHandlerLabel) == sb.getStart()) {
                     for (int j = i + 1; j < itsExceptionTableTop; j++) {
                         itsExceptionTable[j - 1] = itsExceptionTable[j];
                     }
--- a/testsrc/test262.properties
+++ b/testsrc/test262.properties
@@ -6580,6 +6580,123 @@ language/statements/generators 225/259 (86.87%)
     yield-star-after-newline.js
     yield-star-before-newline.js
 
+language/statements/try 115/195 (58.97%)
+    dstr/ary-init-iter-close.js
+    dstr/ary-init-iter-get-err.js
+    dstr/ary-init-iter-get-err-array-prototype.js
+    dstr/ary-init-iter-no-close.js
+    dstr/ary-name-iter-val.js
+    dstr/ary-ptrn-elem-ary-elem-init.js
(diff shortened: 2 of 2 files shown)


src/org/mozilla/classfile/ClassFileWriter.java
```diff
@@ -1605,7 +1605,11 @@ private void killSuperBlock(SuperBlock sb) {
                     int handlerPC = getLabelPC(ete.itsHandlerLabel);
                     SuperBlock handlerSB = getSuperBlockFromOffset(handlerPC);
                     locals = handlerSB.getLocals();
-                    break;
+                    if (handlerSB.isInitialized()) {
+                        break;
+                    } else {
+                        continue;
```
tuchida, line 1611: Do you need to `continue`?
Since it is `if (sbStart > eteStart &&` and ` if (eteStart > sbStart &&`, the result is the same, but it feels more readable to write the same code as here.
https://github.com/mozilla/rhino/blob/f33a0c11f24f86fc6fff433f632a5013ca50e9af/src/org/mozilla/classfile/ClassFileWriter.java#L1617-L1620

tuchida (commented):
Thank you very much for fixing it. Once this has been merged, #916 will be ready for review.

src/org/mozilla/classfile/ClassFileWriter.java
carlosame: Thank you for your review. I try to be efficient in a hot code path (the comment immediately above says so, and I agree), and while perhaps a smart JIT compiler would put a jump right there, I'd rather put it myself and make no assumptions about the JIT. :-)

gbrail:
This is great -- thanks for tracking it down and fixing it!

PR #1085, opened 2021-11-12
Should ScriptableObject.sealObject also prevent modifications of parent and prototype

rPraml:
_Please consider this PR first as question, if I (mis)understand documentation or if it is really a bug_

My use case is, that I have a shared scope and I want to prevent that scope from any modifications.

So I thougt, it would be a good idea to call `scope.sealObject()`

I noticed, that no normal property is writeable - but I can change the `__parent__` and `__proto__` of such sealed objects.
(This is also what I try to fix in this PR)

I planned also to implement `sealObject` in NativeJavaObject, so that you can seal them too, to prevent from further modification. For example block `put` operations and especially `__parent__` and `__proto__` property (I'm aware that I can use UnmodifiableMap and so on, but this does not prevent modification of the special props)

Then I found in the documentation of [ScriptableObject](https://github.com/mozilla/rhino/blob/master/src/org/mozilla/javascript/ScriptableObject.java#L1942):
> It is possible to change the value of an existing property

This seems not to be true, at least what I have found in my tests. I dig further into documentation and found https://github.com/mozilla/rhino/issues/676 and https://262.ecma-international.org/11.0/#sec-object.seal which confirms that existing properties should be modifiable.

So there IS a difference between java `ScriptableObject.sealObject` and js `Object.seal` (the later one does not call the first one and objects sealed with `Object.seal` throw only an EcmaError in strict mode, while objects sealed with `ScriptableObject.sealObject` throw an EvaluatorException)

Maybe someone can explain, what's the intention of `sealObject`

### DIFF
--- a/src/org/mozilla/javascript/ScriptRuntime.java
+++ b/src/org/mozilla/javascript/ScriptRuntime.java
@@ -146,14 +146,10 @@ public class ScriptRuntime {
         scope.associateValue(LIBRARY_SCOPE_KEY, scope);
         (new ClassCache()).associate(scope);
 
-        BaseFunction.init(scope, sealed);
         NativeObject.init(scope, sealed);
+        BaseFunction.init(scope, sealed);
 
         Scriptable objectProto = ScriptableObject.getObjectPrototype(scope);
-
-        // Function.prototype.__proto__ should be Object.prototype
-        Scriptable functionProto = ScriptableObject.getClassPrototype(scope, "Function");
-        functionProto.setPrototype(objectProto);
 
         // Set the prototype of the object passed in if need be
         if (scope.getPrototype() == null) scope.setPrototype(objectProto);
--- a/src/org/mozilla/javascript/ScriptableObject.java
+++ b/src/org/mozilla/javascript/ScriptableObject.java
@@ -680,6 +680,7 @@ public abstract class ScriptableObject
     /** Sets the prototype of the object. */
     @Override
     public void setPrototype(Scriptable m) {
+        checkNotSealed("__proto__", 0);
         prototypeObject = m;
     }
 
@@ -692,6 +693,7 @@ public abstract class ScriptableObject
     /** Sets the parent (enclosing) scope of the object. */
     @Override
     public void setParentScope(Scriptable m) {
+        checkNotSealed("__parent__", 0);
         parentScopeObject = m;
     }
 
--- a/testsrc/org/mozilla/javascript/tests/SealedSharedScopeTest.java
+++ b/testsrc/org/mozilla/javascript/tests/SealedSharedScopeTest.java
@@ -9,7 +9,6 @@ import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
 
 import java.util.Locale;
-
 import org.junit.After;
 import org.junit.Before;
 import org.junit.Test;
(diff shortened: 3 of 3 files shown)


p-bakker:
I think the JavaScript Object.seal and ScriptableObject.sealObject aren't necessarily to do the same thing.

JavaScript's Object.seal probably got added long after ScriptableObject.sealObject. Also, besides Object.seal there's also the more restrictive Object.freeze

To my knowledge, the purpose of ScriptableObject.sealObject is to prevent modifications on shared scopes, in which case I would say that the __parent__ and __proto__ properties should also not be writable.

Then again: no experience here with shared scopes 🙂

Changing the behavior would be a potentially breaking change imho, so something to postpone for v2.0.0?

rPraml:
I've updated this PR.
My use case is still to protect a shared scope from any modification with`ScriptableObject.sealObject` (also `__parent__` and `__proto__`) This is what this PR would fix.

Nevertheless, it would be great if someone could explain the difference and intention of `ScriptableObject.sealObject` and `Object.seal`

p-bakker:
I think there's no one around anymore to detail the difference between ScriptableObject.sealObject and Object.seal, but I did find this in old release notes: https://p-bakker.github.io/rhino/releases/new_in_rhino_1.5r5.html#seal-and-changes-in-semantic-of-sealed-objects and https://p-bakker.github.io/rhino/releases/new_in_rhino_1.5r5.html#api-for-context-sealing

Just guessing here, but it looks like ScriptableObject.sealObject was added before Object.seal existed and that they ought to behave similarly?

Object.seal(...) however allows changing the values of existing properties, which is not desirable when 'sealing' scopes using ScriptableObject.sealObject(...) to make the scope sharable. so in this respect, ScriptableObject.sealObject(...) is more similar in behavior to Object.freeze(...) (which is like Object.freeze, but also prevents adding new properties).

Note that in browsers, Object.seal(...) makes the __proto__ property readonly, as modifying it could introduce new properties, which Object.seal(...) is to prevent.

Also note that `__parent__` is deprecated and non-existing in most JS engines, but as long at its still there in Rhino, I think when sealing (or freezing) an object it should be treated similarly as __proto__.

So, in summary: 
- I think I agree that ScriptableObject.sealObject(...) should also prevent modification of the __proto__/__parent__ property
- ScriptableObject.sealObject(...) behavior is more like Object.freeze than Object.seal
- Maybe the ScriptableObject.sealObject(...) implementation should be rewired to use the Object.freeze implementation? 
- `Then I found in the documentation of [ScriptableObject](https://github.com/mozilla/rhino/blob/master/src/org/mozilla/javascript/ScriptableObject.java#L1942): It is possible to change the value of an existing property` I think this is an anomaly in the JavaDocs, since that behavior seems to have been changes in 1.5r5

p-bakker:
> Full ack!
Do you mean 'full acknowledgement', as in agreed? 

> But one must note that Object.seal currently only works in strict mode. (AFAIR! I didn't check, if it prevents write access in non strict mode).
I'm not aware of that. If that is indeed the case, a case should be created about that I think, cause I dont think there is one

aardvark179:
`Object.seal(o)` will seal an object `o` whether strict mode is enabled or not. The confusion may come because if you do
```javascript
var o = {foo: 1};
Object.seal(o);
o.bar = 2;
```
You won't get an error. That's because, outside of strict mode, failing to set a property doesn't raise an error, but it also doesn't set the property.

PR #1094, opened 2021-11-24
Fix invalid single quotes in French messages

w3aponx:
When running in a French locale, some error messages are not very helpful as rogue single quotes prevent important pattern replacements elsewhere in the message.  An example:

> org.mozilla.javascript.EcmaError: TypeError: Impossible dappeler la méthode "{1}" de {0}

The single quote in _d'appeler_ causes the patterns to remain unchanged, even though the arguments were correctly provided.

Also fixed a tiny grammatical error on line 156.

### DIFF
--- a/src/org/mozilla/javascript/resources/Messages_fr.properties
+++ b/src/org/mozilla/javascript/resources/Messages_fr.properties
@@ -97,7 +97,7 @@ msg.eval.nonstring =\
     c''est la valeur en question qui est renvoy\u00E9e. \u00E9tait-ce votre intention?
 
 msg.eval.nonstring.strict =\
-    La fonction eval() ne peut \u00eatre appel\u00e9e avec une valeur autre qu'une cha\u00EEne primitive \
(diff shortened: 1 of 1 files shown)


p-bakker:
Hi,

First of all: tnx for your contribution!

The changes look fine to me, so once 1.7.14 is released, I think we can merge this

rbri:
@p-bakker do we really have to wait for the release here. I vote for merging this before the release.
(Lets assume someone from france found this while testing the release candidate)

p-bakker:
I didn't set the rule 🙂

I'm fine with merging it, but if merging this one, there are others as well

gbrail:
This seems fine to add to 1.7.14 before the final release. Do we have a label for that?

p-bakker:
@gbrail what kind of label are you referring to?

gbrail:
I guess I was wondering if it made sense to have a label for "must fix before 1.7.14 GA". But I think that there might only be a few such issues so it's likely overkill.

Anyway, this one is very simple and low risk so I'm going to merge it!

PR #1097, opened 2021-11-27
Print test result after testing.

MaxisTekfield:
I've added some code to `test` task in `build.gradle` in order to print test result after testing finishes.
The output message is as following:
```
Test Result: SUCCESS (119824 tests, 119816 success, 0 failure, 8 skipped)
```

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -100,6 +100,16 @@ test {
     // more resilient at the expense of performance.
     forkEvery = 1
     maxParallelForks = 64
+
+    testLogging {
+        afterSuite {
+            desc, result -> {
+                if (!desc.parent) {
+                    println "Test Result: ${result.resultType} (${result.testCount} tests, ${result.successfulTestCount} success, ${result.failedTestCount} failure, ${result.skippedTestCount} skipped)"
+                }
+            }
+        }
+    }
 }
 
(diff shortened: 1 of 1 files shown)


gbrail:
Sorry, going through PRs to try and get 1.7.14 done.

I actually don't think that we should do this -- I've become used to the philosophy that tests and builds generally shouldn't print out stuff if everything is OK, and the tests produce output files that are visible in most build systems where that sort of thing can be verified as well. I don't think that putting in additional output is really helpful.

MaxisTekfield:
I add this output in order to check total test cases and the brief result in a quick glance without opening the reports when running tests by command line.
Of course it helps nothing when running tests by CI or something.

p-bakker:
@gbrail @MaxisTekfield how do we want to proceed with this PR? Merge or close?

gbrail:
I'm sorry, but I would rather close this one.

p-bakker:
@MaxisTekfield do you have a really compelling reason to want this? Otherwise I agree with @gbrail and rather not merge this

p-bakker:
Am closing this one, as there's been no response from the author. Feel free to reopen if you still want to persue this and think it should be included

PR #1098, opened 2021-11-27
Fix issue #1041. Handles {0} quantifier with max of zero correctly.

### LINKED ISSUE #1041: RegExp does not handle quantifier with max of zero correctly
```js
js> var re = /ab{0}c/
js> re.test('abc')
true
```
Expected: `false`

MaxisTekfield:
I've fixed the issue #1041 .

Closes #1041

### DIFF
--- a/src/org/mozilla/javascript/regexp/NativeRegExp.java
+++ b/src/org/mozilla/javascript/regexp/NativeRegExp.java
@@ -2116,8 +2116,9 @@ public class NativeRegExp extends IdScriptableObject {
                                     pc += getOffset(program, pc);
                                     break switchStatement;
                                 }
-                                if (state.min == 0 && gData.cp == state.index) {
-                                    // matched an empty string, that'll get us nowhere
+                                if (state.min == 0 && (gData.cp == state.index || state.max == 0)) {
+                                    // matched an empty string or an {0} quantifier, that'll get us
+                                    // nowhere
                                     result = false;
                                     continuationPc = state.continuationPc;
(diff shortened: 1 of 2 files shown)


p-bakker:
Thnx for your contribution!

I see that you included a unit test: isn't this JavaScript behavior covered by a test in the Test262 suite, that was failing and now started passing?

MaxisTekfield:
> 

I'll check test262 later.

MaxisTekfield:
> Thnx for your contribution!
> 
> I see that you included a unit test: isn't this JavaScript behavior covered by a test in the Test262 suite, that was failing and now started passing?

I cannot find any corrsponding test in test262. Only `test262/implementation-contributed/v8/mjsunit/third_party/regexp-pcre/regexp-pcre.js` has some cases as follows:
```javascript
// ...
res[903] = /a{0}bc/;
// ...
assertToStringEquals("bc", res[903].exec("bc"), 2782);
// ...
```

It seems that we don't run any cases under `test262/implementation-contributed` at all.

p-bakker:
K, maybe such test exists in a more recent version of the Test262 suite, but unfortunately we cannot run the latest version, as Rhino misses a bunch of features newer versions of the test suite depends on.

So I guess a dedicated test is in order 🙂

And we're indeed not running the 'implementation contributed' tests: those are contributed to the test262 suite repo for eventual inclusion in the main test suite.

gbrail:
This works great. Thanks!

PR #1103, opened 2021-11-28
Provide console API in shell.

### LINKED ISSUE #1012: Support `console` API in shell and debugger
As I'm using `java -jar rhino-1.7.13.jar` to run rhino as a shell, or `java -cp rhino-1.7.13.jar org.mozilla.javascript.tools.debugger.Main` to run rhino debugger, I'd like to write `console.log(...)` or something like this to check values of variables.
The `console` API is very popular and widely used in other JS engines, eg nodejs, firefox, chrome. One of the very useful features is printf-like format, which `print` API does not support. Especially the `%o` and `%O` specifier can check objects very easily.
Expecting rhino can support `console` API in shell and debugger so that we can use same statements in all these popular js engines.

For reference:
- https://developer.mozilla.org/en-US/docs/Web/API/console
- https://developer.chrome.com/docs/devtools/console/api/
- https://nodejs.org/api/console.html
- https://nodejs.org/api/util.html#util_util_format_format_args

MaxisTekfield:
As #1012 says, I've been trying to implement the `console` API in shell.

PR #1104 is prerequisite as p-bakker said. After #1104 is merged. I'll rebase this PR.

Basic usage:
```
js> console.info('obj: %o', {a: 123, b: "abc"})
INFO obj: {"a":123,"b":"abc"}
js> console.error("Error message.");
ERROR Error message.
js> console.count('counter1')
INFO counter1: 1
js> console.count('counter1')
INFO counter1: 2
js> console.count('counter1')
INFO counter1: 3
js> console.countReset('counter1')
js> console.count('counter1')
INFO counter1: 1
js> console.time()
js> console.timeLog()
INFO default: 3804.4751ms
js> console.timeLog()
INFO default: 7525.4994ms
js> console.timeEnd()
INFO default: 15393.5158ms
js> console.timeEnd()
WARN Timer 'default' does not exist.
```

Closes #1012

### DIFF
--- a/src/org/mozilla/javascript/NativeConsole.java
+++ b/src/org/mozilla/javascript/NativeConsole.java
@@ -42,8 +42,8 @@ public class NativeConsole extends IdScriptableObject {
                 Context cx,
                 Scriptable scope,
                 Level level,
-                ScriptStackElement[] stack,
-                Object[] args);
+                Object[] args,
+                ScriptStackElement[] stack);
     }
 
     public static void init(Scriptable scope, boolean sealed, ConsolePrinter printer) {
@@ -149,25 +149,25 @@ public class NativeConsole extends IdScriptableObject {
                 {
                     ScriptStackElement[] stack =
                             new EvaluatorException("[object Object]").getScriptStack();
-                    printer.print(cx, scope, Level.TRACE, stack, args);
+                    printer.print(cx, scope, Level.TRACE, args, stack);
                     break;
                 }
 
             case Id_debug:
-                printer.print(cx, scope, Level.DEBUG, null, args);
+                printer.print(cx, scope, Level.DEBUG, args, null);
                 break;
 
             case Id_log:
             case Id_info:
-                printer.print(cx, scope, Level.INFO, null, args);
+                printer.print(cx, scope, Level.INFO, args, null);
                 break;
 
             case Id_warn:
-                printer.print(cx, scope, Level.WARN, null, args);
+                printer.print(cx, scope, Level.WARN, args, null);
                 break;
 
             case Id_error:
-                printer.print(cx, scope, Level.ERROR, null, args);
+                printer.print(cx, scope, Level.ERROR, args, null);
                 break;
 
             case Id_assert:
@@ -202,7 +202,7 @@ public class NativeConsole extends IdScriptableObject {
     }
 
     private void print(Context cx, Scriptable scope, Level level, String msg) {
-        printer.print(cx, scope, level, null, new String[] {msg});
+        printer.print(cx, scope, level, new String[] {msg}, null);
     }
 
     public static String format(Context cx, Scriptable scope, Object[] args) {
@@ -227,7 +227,7 @@ public class NativeConsole extends IdScriptableObject {
                 Object val = args[argIndex];
                 switch (placeHolder) {
                     case "%s":
-                        replaceArg = ScriptRuntime.toString(val);
+                        replaceArg = formatString(val);
                         break;
 
                     case "%d":
@@ -241,7 +241,7 @@ public class NativeConsole extends IdScriptableObject {
 
                     case "%o":
                     case "%O":
-                        replaceArg = formatObj(cx, scope, args, argIndex);
+                        replaceArg = formatObj(cx, scope, val);
                         break;
 
                     default:
@@ -258,12 +258,16 @@ public class NativeConsole extends IdScriptableObject {
         return buffer.toString();
     }
 
-    private static String formatFloat(Object val) {
-        if (val instanceof BigInteger || val instanceof Symbol) {
-            return ScriptRuntime.NaNobj.toString();
+    private static String formatString(Object val) {
+        if (val instanceof BigInteger) {
+            return ScriptRuntime.toString(val) + "n";
         }
 
-        return String.valueOf(ScriptRuntime.toNumber(val));
+        if (ScriptRuntime.isSymbol(val)) {
+            return val.toString();
+        }
+
+        return ScriptRuntime.toString(val);
     }
 
     private static String formatInt(Object val) {
@@ -271,7 +275,7 @@ public class NativeConsole extends IdScriptableObject {
             return ScriptRuntime.bigIntToString((BigInteger) val, 10) + "n";
         }
 
-        if (val instanceof Symbol) {
+        if (ScriptRuntime.isSymbol(val)) {
             return ScriptRuntime.NaNobj.toString();
         }
 
@@ -284,8 +288,15 @@ public class NativeConsole extends IdScriptableObject {
         return String.valueOf((long) number);
     }
 
-    private static String formatObj(Context cx, Scriptable scope, Object[] args, int argIndex) {
-        Object arg = args[argIndex];
+    private static String formatFloat(Object val) {
+        if (val instanceof BigInteger || ScriptRuntime.isSymbol(val)) {
+            return ScriptRuntime.NaNobj.toString();
+        }
+
+        return ScriptRuntime.numberToString(ScriptRuntime.toNumber(val), 10);
+    }
+
+    private static String formatObj(Context cx, Scriptable scope, Object arg) {
         if (arg == null) {
             return "null";
         }
--- a/src/org/mozilla/javascript/NativeConsole.java
+++ b/src/org/mozilla/javascript/NativeConsole.java
@@ -281,7 +281,7 @@ public class NativeConsole extends IdScriptableObject {
             return ScriptRuntime.toString(number);
         }
 
-        return String.valueOf((int) number);
+        return String.valueOf((long) number);
     }
 
     private static String formatObj(Context cx, Scriptable scope, Object[] args, int argIndex) {
--- a/src/org/mozilla/javascript/NativeConsole.java
+++ b/src/org/mozilla/javascript/NativeConsole.java
@@ -337,83 +337,53 @@ public class NativeConsole extends IdScriptableObject {
         return nano / 1000000D;
     }
 
-    // #string_id_map#
-
     @Override
     protected int findPrototypeId(String s) {
         int id;
-        // #generated# Last update: 2021-11-10 14:43:23 CST
-        L0:
-        {
-            id = 0;
-            String X = null;
-            int c;
-            L:
-            switch (s.length()) {
-                case 3:
-                    X = "log";
-                    id = Id_log;
-                    break L;
-                case 4:
-                    c = s.charAt(0);
-                    if (c == 'i') {
-                        X = "info";
-                        id = Id_info;
-                    } else if (c == 't') {
-                        X = "time";
-                        id = Id_time;
-                    } else if (c == 'w') {
-                        X = "warn";
-                        id = Id_warn;
-                    }
-                    break L;
-                case 5:
-                    switch (s.charAt(0)) {
-                        case 'c':
-                            X = "count";
-                            id = Id_count;
-                            break L;
-                        case 'd':
-                            X = "debug";
-                            id = Id_debug;
-                            break L;
-                        case 'e':
-                            X = "error";
-                            id = Id_error;
-                            break L;
-                        case 't':
-                            X = "trace";
-                            id = Id_trace;
-                            break L;
-                    }
-                    break L;
-                case 6:
-                    X = "assert";
-                    id = Id_assert;
-                    break L;
-                case 7:
-                    c = s.charAt(4);
-                    if (c == 'E') {
-                        X = "timeEnd";
-                        id = Id_timeEnd;
-                    } else if (c == 'L') {
-                        X = "timeLog";
-                        id = Id_timeLog;
-                    }
-                    break L;
-                case 8:
-                    X = "toSource";
-                    id = Id_toSource;
-                    break L;
-                case 10:
-                    X = "countReset";
-                    id = Id_countReset;
-                    break L;
-            }
-            if (X != null && X != s && !X.equals(s)) id = 0;
-            break L0;
+        switch (s) {
+            case "log":
+                id = Id_log;
+                break;
+            case "info":
+                id = Id_info;
+                break;
+            case "time":
+                id = Id_time;
+                break;
+            case "warn":
+                id = Id_warn;
+                break;
+            case "count":
+                id = Id_count;
+                break;
+            case "debug":
+                id = Id_debug;
+                break;
+            case "error":
+                id = Id_error;
+                break;
+            case "trace":
+                id = Id_trace;
+                break;
+            case "assert":
+                id = Id_assert;
+                break;
+            case "timeEnd":
+                id = Id_timeEnd;
+                break;
+            case "timeLog":
+                id = Id_timeLog;
+                break;
+            case "toSource":
+                id = Id_toSource;
+                break;
+            case "countReset":
+                id = Id_countReset;
+                break;
+            default:
+                id = 0;
+                break;
         }
-        // #/generated#
         return id;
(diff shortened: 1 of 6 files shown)


p-bakker:
Great work!

As for the reformatting of Globals.java: we tend to do reformatting in a separate commit/PR, so formatting and actual changes are separated. Any chance you can create a separate PR for that?

As for the console implementation: seems like the implementation is tied directly to the Shell via the print method, from which I assume the console implementation cannot be used by embeddings that don't use the Shell?

MaxisTekfield:
PR #1104 is created for reformatting `Global.java`.

I've thought about using `console` in embedding usage. I believe that is entirely possible.
But I am confusing which mechanism should be used to provide extension for implying `console` API in embedding usage. Java SPI or something else?

p-bakker:
As for how to enable the console in an embedded Rhino: the embedder can can the init method themselves

I think the more important part is how to allow the embedder to control how an 'output' is being handled

For the Shell is being dealt with by calling the println method. In an embedding it might have to be handled by calling the appropriate method(s) of the used log library

So i think maybe define some interface and require to pass an instance of a class implementing that interface to the init method of the NativeConsole?

That way what's being done with the output is decoupled from the console implementation

p-bakker:
Exactly what such interface should look like I'm not exactly sure: some implementations might want the raw data (for example an object passed as argument to console.log) while others would rather have the entire data as an already formatted string

And some might want to know the log level.

Btw: another option instead of an interface would be to provide a basic implementation based on System.out.println and just extend NativeConsole and override that implementation with whatever is desired

MaxisTekfield:
How about the following solution.

### First, when and where should I put the `NativeConsole` into scope?

I'm thinking these two ways:
1. Insert some code in `org.mozilla.javascript.ScriptRuntime#initSafeStandardObjects` - just like `NativeJSON` - everybody can use `console` by default.
2. Or (maybe not necessary), based on 1, add new feature switch to enable/disable the `NativeConsole` implementation. Default is disabled for compatibility reason. Someone may has already implemented his own `console`.

### Second, how to provide interface to let applications embedding rhino-runtime to implement their own output?

I found `org.mozilla.javascript.VMBridge#makeInstance` to be a good example. I can try following class names in order to find the implementation to do outputs.
1. org.mozilla.javascript.console.ConsoleOutput_custom  // applications can provide class with this name to implement their own outputs.
2. org.mozilla.javascript.tools.shell.ConsoleOutput  // the output of shell.
3. org.mozilla.javascript.console.DefaultConsoleOutput  // output by System.out and System.err.

p-bakker:
> First, when and where should I put the NativeConsole into scope?

I would say neither of the two options you mentioned. We can hadcode the console to be available in the Shell, but that's about it. For all other embeddings, it should be up to the embedder whether to include it or not and I think a feature flag is not really necesarry, an embedding has code already to initialize the embedding, like setting the language level etc. Adding a call there to initialize the console if they want is sufficient imho.

> Second, how to provide interface to let applications embedding rhino-runtime to implement their own output?

Why not just `NativeConsole.init(scope, sealed, new ConsoleImpl(){....}` or `new NativeConsole(scope, sealed) {...}`

tuchida:
Very nice PR!
Do you need to test? This is the [specification](https://console.spec.whatwg.org/), and this seems to be the [test](https://github.com/web-platform-tests/wpt/tree/master/console).

toolsrc/org/mozilla/javascript/tools/shell/NativeConsole.java
```diff
@@ -0,0 +1,455 @@
...
+        if (id > LAST_METHOD_ID) {
+            throw new IllegalStateException(String.valueOf(id));
+        }
+
+        String name;
+        int arity;
+        switch (id) {
+            case Id_toSource:
+                arity = 0;
+                name = "toSrouce";
```
p-bakker, line 76: typo?

MaxisTekfield: Sorry, my fault.

MaxisTekfield:
I have made some changes in order to make printer customizable in embedding usage.

> Do you need to test?
Yes, I need. I'll write them later.

gbrail:
I think it'd be great to have "console"! But I do think that we should only have it in Global (which makes it part of the shell) and:

1) Make it easy for any embedder to add it to any script, but not by default, even if they don't use "Global."
2) Make the choice of output location pluggable by supplying an OutputStream.

tonygermano:
As discussed in #1012 I think the plugable printer needs to take an `Object[]` (or `List<Object>`.)

If the first argument is a string, then NativeConsole can perform the substitutions, but any remaining arguments that are not used in the substitution or all of the arguments if the first is not a string should be passed to the printer to handle as appropriate.

Some printers may want to display debug level detail (possibly in a GUI like the Chrome and Firefox consoles) for nested objects rather than just performing a toString operation on them.

A Context object should also be passed to facilitate accessing the (nested) object properties.

That doesn't preclude a function that would return an implementation that would do that basic conversion.
```java
NativeConsole.ConsolePrinter createToStringConsolePrinter(OutputStream trace, OutputStream debug, OutputStream info, OutputStream warn, OutputStream error)
```

p-bakker:
> I think the plugable printer needs to take an Object[] (or List<Object>.)
Agree, also from the perspective of stringifying the whole thing being unneeded (and potentially expensive) if for example console.debug(...) is called, while the underlying output/log facitity might be configured to ignore everything lower than log level INFO

MaxisTekfield:
I've change args of print to Object[].

src/org/mozilla/javascript/NativeConsole.java
```diff
@@ -0,0 +1,437 @@
...
+    private double nano2Milli(Long nano) {
+        return nano / 1000000D;
+    }
+
+    // #string_id_map#
+
+    @Override
+    protected int findPrototypeId(String s) {
+        int id;
+        // #generated# Last update: 2021-11-10 14:43:23 CST
```
tuchida, line 345: ref. #848, #914
The latest code now uses String in Switch Case, and `idswitch` has been removed.

gbrail:
Thanks!

Can you follow the comments above and get rid of the now-obsolete "idswitch" stuff? You can look at any of the existing classes for an example.

Also, it'd be great to take a look at the man page (the file is rhino/man.1) and see if we can update it to mention that the "console" object is now supported.

src/org/mozilla/javascript/NativeConsole.java
```diff
@@ -0,0 +1,437 @@
...
+            return super.execIdCall(f, cx, scope, thisObj, args);
+        }
+
+        int methodId = f.methodId();
+        switch (methodId) {
+            case Id_toSource:
+                return "Console";
+
+            case Id_trace:
+                printer.print(cx, scope, Level.TRACE, args);
```
p-bakker, line 143: Shouldn't console.trace() automatically send the current stacktrace as the first entry in the args object[] to the printer?

p-bakker:
@MaxisTekfield any chance you could followup on the previous two comments on this PR?

MaxisTekfield:
> @MaxisTekfield any chance you could followup on the previous two comments on this PR?

Got it. These takes more time to make changes. I'll make them done in next few days.

src/org/mozilla/javascript/NativeConsole.java
tuchida: :+1: [ae1b6a9](https://github.com/mozilla/rhino/pull/1103/commits/ae1b6a9981bfecc668dda5d67fd1bbc10a74d659)

src/org/mozilla/javascript/NativeConsole.java
```diff
@@ -0,0 +1,405 @@
...
+                    replaceArg = "%";
+                    break;
+
+                case "%s":
+                    replaceArg = ScriptRuntime.toString(args, argIndex);
+                    break;
+
+                case "%d":
+                case "%i":
+                    double number = ScriptRuntime.toNumber(args, argIndex);
```
tuchida, line 221: BigInt will cause an error.
```js
console.log('%d', 10n)
// js: uncaught JavaScript runtime exception: TypeError: Cannot convert BigInt to a number
```
The specifications are as follows.
https://console.spec.whatwg.org/#formatting-specifiers
> %d or %i | Element which substitutes is converted to an integer | %parseInt%(element, 10)

The implementation are as follows.

V8 (Node.js v14.17.0)
```js
console.log('%d', 10n) // 10n
```
SpiderMonkey (Firefox/97.0)
```js
console.log('%d', 10n) // undefined
```

tuchida:
```js
console.log('%d', Symbol.iterator)
// js: uncaught JavaScript runtime exception: TypeError: The object is not a number
```
According to the specification, it seems that Symbol should be converted to NaN.
https://console.spec.whatwg.org/#formatter
> If Type(current) is Symbol, let converted be NaN

tuchida: Even `%o` or `%O` will cause an error.
```js
console.log('%o', { a: 0n }) 
// js: uncaught JavaScript runtime exception: TypeError: Do not know how to serialize a BigInt
```

MaxisTekfield: > ```js
> console.log('%d', 10n)
> // js: uncaught JavaScript runtime exception: TypeError: Cannot convert BigInt to a number
> ```

I got a syntax error. How to write a BigInt in rhino shell?

```
js> print(10n);
js: "<stdin>", line 10: missing ) after argument list
js: print(10n);
js: .........^
js: "<stdin>", line 10: Compilation produced 1 syntax errors.

js> console.log("%d", 10n);
js: "<stdin>", line 12: missing ) after argument list
js: console.log("%d", 10n);
js: .....................^
js: "<stdin>", line 12: Compilation produced 1 syntax errors.

js> 
```

src/org/mozilla/javascript/NativeConsole.java
MaxisTekfield: Done.

src/org/mozilla/javascript/NativeConsole.java
MaxisTekfield: How about the latest code which passes `ScriptStackElement[] stack` before `Object[] args`.

src/org/mozilla/javascript/NativeConsole.java
tuchida: Could you add `-version 200`.
```sh
$ java -jar buildGradle/libs/rhino-1.7.15-SNAPSHOT.jar -version 200
```

MaxisTekfield: > Even `%o` or `%O` will cause an error.
> 
> ```js
> console.log('%o', { a: 0n }) 
> // js: uncaught JavaScript runtime exception: TypeError: Do not know how to serialize a BigInt
> ```

I think this is not the problem of console. This TypeError is thrown at `org.mozilla.javascript.NativeJSON#str`

```java
        if (value instanceof Number) {
            if (value instanceof BigInteger) {
                throw ScriptRuntime.typeErrorById("msg.json.cant.serialize", "BigInt");
            }
            // ...
        }
```
This problem also occurs when using JSON.stringify
```
js> JSON.stringify({a: 10n});
js: uncaught JavaScript runtime exception: TypeError: Do not know how to serialize a BigInt
	at (unknown):9
```

tuchida: Yes, I do. It is a specification of ECMA262 that `JSON.stringify` will error on BigInt.
https://262.ecma-international.org/12.0/#sec-serializejsonproperty
> 10. 10. If Type(value) is BigInt, throw a TypeError exception.

Even in Node.js, using `%j` will result in an error.
```js
console.log('%j', 10n)
// Uncaught TypeError: Do not know how to serialize a BigInt

console.log('%o', 10n) // 10n
```
Other than that, `JSON.stringify` will cause an error with circular references.
```js
var o = {};
o.o = o;
JSON.stringify(o);
// js: uncaught JavaScript runtime exception: TypeError: Cyclic org.mozilla.javascript.NativeObject value not allowed.
```

If you definitely don't want to generate errors, you should use a method other than `JSON.stringify` to display the data. However, for now, I think it is enough to catch the exception and fall back to `ScriptRuntime.toString`.

src/org/mozilla/javascript/NativeConsole.java
```diff
@@ -216,61 +217,94 @@ public static String format(Context cx, Scriptable scope, Object[] args) {
...
+            return ScriptRuntime.NaNobj.toString();
+        }
+
+        double number = ScriptRuntime.toNumber(val);
+
+        if (Double.isInfinite(number) || Double.isNaN(number)) {
+            return ScriptRuntime.toString(number);
         }
 
+        return String.valueOf((int) number);
```
tuchida, line 284: `(long)` might be better here.

This PR
```js
// const INT32_MAX = 2147483647

console.log('%d', 2147483647)
// INFO 2147483647

console.log('%d', 2147483648)
// INFO 2147483647
```
Node.js
```js
console.log('%d', 2147483648)
// 2147483648
```
Firefox
```js
console.log('%d', 2147483648)
// -2147483648

console.log('%d', 2147483649)
// -2147483647
```

MaxisTekfield: Changed to `long`.

src/org/mozilla/javascript/NativeConsole.java
```diff
@@ -0,0 +1,448 @@
...
+        }
+
+        return Undefined.instance;
+    }
+
+    private void print(Context cx, Scriptable scope, Level level, String msg) {
+        printer.print(cx, scope, level, null, new String[] {msg});
+    }
+
+    public static String format(Context cx, Scriptable scope, Object[] args) {
```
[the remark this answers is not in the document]
tuchida: 13192a2d4da3a528b4b8e5e39320c11d9784d4f4 Thanks!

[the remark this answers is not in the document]
MaxisTekfield: It's hard to cover all those specifications, especially those who have different behaviors in different runtime engines.

p-bakker:
> First, can you move NativeConsole to the org.mozilla.javascript.tools package in the "toolsrc" tree?

Am not sure if this is a good idea: the console is (imho) something embedders and script-engine users might also want to use. Making them require to use the bundle that includes all the tools would not be so great I think

MaxisTekfield:
> > First, can you move NativeConsole to the org.mozilla.javascript.tools package in the "toolsrc" tree?
> 
> Am not sure if this is a good idea: the console is (imho) something embedders and script-engine users might also want to use. Making them require to use the bundle that includes all the tools would not be so great I think

Emm, I don't know what should I do.
I'd like to keep console in main src because I think it will be helpful for all those users - embedders and shell users.

tuchida:
Deno and Node.js seem to be able to use `console` without `import`.
https://doc.deno.land/deno/stable/~/console
https://nodejs.org/dist/latest-v17.x/docs/api/console.html

gbrail:
I would be concerned about this being in the main package if it did things like open files or execute processes like our "Global" object does. Since it doesn't, and if others disagree, then I'm fine with having it in the main package where it is in this PR.

So then the only remaining thing is to see if we can at least have a sanity test of each function in the API.

Thank you for all the work on this!

MaxisTekfield:
More tests added.

gbrail:
Thanks for all the work. Since this requires a merge and has many commits, I'm going to close this PR manually and merge the code into master.

p-bakker:
Tnx @MaxisTekfield for all your work!

Issue #1105, opened 2021-11-29
ScriptableObject.getGetterOrSetter changed incompatibly


gbrail:
During development of 1.7.14 we added a new "scope" parameter to ScriptableObject.getGetterOrSetter. This is an obscure function but a public one and it's used by some legacy code I found. We should add back the old function (using "this" as the scope) and deprecate the old one in favor of the new.

PR #1106, opened 2021-11-30
Restore old signature for getGetterOrSetter

gbrail:
This is technically a regression because we changed the signature
of a public method.

Fixes https://github.com/mozilla/rhino/issues/1105

### DIFF
--- a/src/org/mozilla/javascript/ScriptableObject.java
+++ b/src/org/mozilla/javascript/ScriptableObject.java
@@ -589,6 +589,7 @@ public abstract class ScriptableObject
      *
      * @param name Name of the object. If nonnull, index must be 0.
      * @param index Index of the object. If nonzero, name must be null.
+     * @param scope the current scope.
      * @param isSetter If true, return the setter, otherwise return the getter.
      * @exception IllegalArgumentException if both name and index are nonnull and nonzero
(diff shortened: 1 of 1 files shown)


src/org/mozilla/javascript/ScriptableObject.java
```diff
@@ -602,6 +603,22 @@ public Function getGetterOrSetter(String name, int index, Scriptable scope, bool
...
+     *
+     * @param name Name of the object. If nonnull, index must be 0.
+     * @param index Index of the object. If nonzero, name must be null.
+     * @param isSetter If true, return the setter, otherwise return the getter.
+     * @exception IllegalArgumentException if both name and index are nonnull and nonzero
+     *     respectively.
+     * @return Null if the property does not exist. Otherwise returns either the getter or the
+     *     setter for the property, depending on the value of isSetter (may be undefined if unset).
+     */
+    @Deprecated
```
rbri, line 617: what about adding also a javadoc comment like this

* @deprecated since xxx please use .... instead

and maybe some works about the motivation of the new method and the drawbacks when still using this one (if there are any)

gbrail:
Thanks, updated to address the comments!

PR #1115, opened 2021-12-07
ref. #1112 Fix Array.from when called with a sparse array

### LINKED ISSUE #1112: Possible issue with Array.from
Rhino:
```js
js> JSON.stringify(Array.from({length:3}).map(_ => 1))
[null,null,null]
```

Firefox and Chrome consoles return `[1,1,1]`

I haven't looked into the spec yet, but it appears that Array.from is returning a sparse array when it shouldn't be.

tuchida:
Closes #1112

https://262.ecma-international.org/12.0/#sec-array.from
> 12. Repeat, while k < len,
>   b. Let kValue be ? Get(arrayLike, Pk).
>   d. Else, let mappedValue be kValue.
>   e. Perform ? CreateDataPropertyOrThrow(A, Pk, mappedValue).

### DIFF
--- a/src/org/mozilla/javascript/NativeArray.java
+++ b/src/org/mozilla/javascript/NativeArray.java
@@ -812,13 +812,11 @@ public class NativeArray extends IdScriptableObject implements List {
         final long length = getLengthProperty(cx, items);
         final Scriptable result = callConstructorOrCreateArray(cx, scope, thisObj, length, true);
         for (long k = 0; k < length; k++) {
-            Object temp = getRawElem(items, k);
-            if (temp != Scriptable.NOT_FOUND) {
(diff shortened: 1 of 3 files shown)


gbrail:
This seems safe enough for 1.7.14. Any other opinions as to the risk?

gbrail:
An observation for a future enhancement -- it's ironic that you had to fix this to work with sparse arrays, because Array.from doesn't appear to be optimized for dense arrays either! A future enhancement could optimize the common case of copying from a dense-only NativeArray to a new NativeArray instance using System.arraycopy. The resulting implementation would be an order of magnitude faster I bet.

gbrail:
Thanks!

Issue #1114, opened 2021-12-07
Regression in __lookupSetter__


gbrail:
In the Kangax compat table tests, 1.7.14 newly fails with the following test function:

    function f() { 
         var a = { };
         var b = Object.create(a);
         b.foo = 1;
         a.__defineSetter__("foo", function () {})
         return b.__lookupSetter__("foo") === void undefined
    }  

The culprit is the slot map refactoring that I did in commit 4afe4dad397c4e26e0d16da227dfd944487666bc .

PR #1116, opened 2021-12-08
Fix regression in __lookupSetter__

gbrail:
This restores behavior that previously passed in the Kangax compat-table.

Fixes https://github.com/mozilla/rhino/issues/1114


src/org/mozilla/javascript/ScriptableObject.java
```diff
@@ -595,11 +595,15 @@ public void setGetterOrSetter(
      * @return Null if the property does not exist. Otherwise returns either the getter or the
```
rbri, line 595: @return Undefined.instance if....

rbri (approved):

PR #1117, opened 2021-12-08
Regression in   proto  

rbri:
Sorry for that
* for some unknown reasons we had some test ignored during the test run because of the missing parent class. Because of this the issue pops up that late (fixed)
* have more or less restored the old functionality and added the new only for function targets
* have changed two of the existing asserts - verified with the browser


gbrail:
Thanks!

I'm merging this manually so that it can have a more informational commit message.

Issue #1119, opened 2021-12-09
Cleanup the .zip artifact?


p-bakker:
This is what is inside the Rhino 1.7.13 folder in the .zip artifact for  v1.7.13:
![image](https://user-images.githubusercontent.com/4981149/145367481-38e57b64-455d-497c-a821-d127eb0ae046.png)

I think of that, the following should not be included:
- CODE_OF_CONDUCT.md
- README.md
- RELEASE-NOTES.md
- build.gradle
- docs/
- gradle/
- gradle.properties
- gradlew*
- release-steps.txt
- results.txt

And shouldn't the /lib folder not also include the `rhino-runtime-X.X.X.jar`, the `rhino-engine-X.X.X.jar` and the toolsrc/xmlimplsrc folders?

carlosame:
Until now, I haven't noticed that the Javadocs are part of the source Git repository, at `docs/javadoc/`. Shouldn't they be removed, and instead produce a javadoc jar on `./gradlew build`? (the javadocs are already being generated at `buildGradle/docs/javadoc/`)

p-bakker:
Yeah, not very clear that we're hosting the javadoc, but that's being worked on

Why wouldn't we host them and instead produce a jar?

p-bakker:
That was recently switched: the GitHub pages are now based on the docs folder in master

gbrail:
I personally find having a "docs" folder in master a much much easier way
to manage the (rather sparse) pages at mozilla.github.io/rhino. If there's
a better way to ensure that Javadocs are hosted somewhere, then I'm happy
to have some show us how to switch to that, but for now it's helpful IMO to
have Javadoc online somewhere and right now that's the easiest place to put
it!

The "zip" artifact is something that I don't know how people use -- it's
always been like that, and I've been told before that changing it breaks
various workflows, but in these days of people building with Gradle and
Maven I wonder what it's really for. Does anyone have an anecdote?

tonygermano:
> PErsonally also don't have a problem hosting the JavaDoc there, only wondering if we should not only host the latest version, but also (some) historic versions

+1 for hosting historic versions

tonygermano:
Is the zip artifact the same thing as if you were to go to the main project page and click the `Code` button (the same one that gives you the clone options, and select the "Download ZIP" option instead?

tonygermano:
[...]

PR #1124, opened 2021-12-09
Shell test

### LINKED ISSUE #956: LICENSE.txt missing from rhino-engine-x.x.x and rhino-runtime-x.x.x jars
Only seems included in the full rhino-x.x.x.jar

### LINKED ISSUE #1120: Reproducible jars with a LICENSE.txt file
Currently, every time that Rhino is built, different jar files are produced even for the same set of code. Which could be an annoyance if the packages were posted to additional Maven repositories (for example a wokflow that posted to Github packages), because the digests of those files would be different.

I'd suggest configuring the Gradle build for reproducible jar files, which would imply removing the `Built-Date`and `Built-Time` fields from `MANIFEST.MF`.

I'd also copy the `LICENSE.txt` file into the `META-INF` directory of each jar file (the license is currently being copied to the root directory of the main `rhino-<version>.jar` file, and not the others).

carlosame:
- Generate reproducible jar files.
- Copy `LICENSE.txt` into the `META-INF` directory of archive files.
- Update POM metadata descriptions.
- Compress Tar distribution file.

The only potentially controversial change that I see is that the `LICENSE.txt` file of the main jar archive is now inside the `META-INF` directory, instead of at the root directory. The `META-INF` directory is the place where licenses are generally put: I do it in my projects and I've seen other projects do the same; that's why I moved the file.

EDIT: most of this PR was already merged but the shell test remains pending.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -212,6 +212,9 @@ jar {
     from ('NOTICE.txt') {
         into 'META-INF'
     }
+    from ('NOTICE-tools.txt') {
+        into 'META-INF'
+    }
     excludes = ["org/mozilla/javascript/engine/**", "META-INF/services/**"]
     // Class ImplementationVersion uses 'Implementation-Title'
     manifest {
@@ -259,6 +262,9 @@ task sourceJar(type: Jar) {
         into 'META-INF'
     }
     from ('NOTICE.txt') {
+        into 'META-INF'
+    }
+    from ('NOTICE-tools.txt') {
         into 'META-INF'
     }
 }
--- a/NOTICE-tools.txt
+++ b/NOTICE-tools.txt
@@ -1,0 +1,37 @@
+Rhino is licensed subject to the terms of the Mozilla Public License, v. 2.0.
+See "LICENSE.txt" for the text of the license.
+
+The files in toolsrc/org/mozilla/javascript/tools/debugger/treetable (runtime
+package org.mozilla.javascript.tools.debugger.treetable) are available under
+the following license:
+
+----
+
+Copyright 1997, 1998 Sun Microsystems, Inc.  All Rights Reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions
+are met:
+  - Redistributions of source code must retain the above copyright
+    notice, this list of conditions and the following disclaimer.
+
+  - Redistributions in binary form must reproduce the above copyright
+    notice, this list of conditions and the following disclaimer in the
+    documentation and/or other materials provided with the distribution.
+
+  - Neither the name of Sun Microsystems nor the names of its
+    contributors may be used to endorse or promote products derived
+    from this software without specific prior written permission.
+
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS
+IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO,
+THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR
+PURPOSE ARE DISCLAIMED.  IN NO EVENT SHALL THE COPYRIGHT OWNER OR
+CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF
+LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
+NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
--- a/testsrc/org/mozilla/javascript/tools/shell/ShellTest.java
+++ b/testsrc/org/mozilla/javascript/tools/shell/ShellTest.java
@@ -1,0 +1,334 @@
+/* -*- Mode: java; tab-width: 8; indent-tabs-mode: nil; c-basic-offset: 4 -*-
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.javascript.tools.shell;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertTrue;
+
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.io.PrintStream;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+import org.junit.AfterClass;
+import org.junit.Before;
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+public class ShellTest {
+
+    static Global global;
+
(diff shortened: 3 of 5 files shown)


carlosame:
I added a commit with a fixed man page. Haven't checked the formatting results as my system has Windows and I do not have groff or similar installed (and the win32 version that I found carried dependencies from 15 years ago or so).

gbrail:
When you start the shell today, it prints out the Implementation-Version and the Build-Date:

$ java -jar buildGradle/libs/rhino-1.7.14-SNAPSHOT.jar
Rhino 1.7.14-SNAPSHOT 2021 12 10
js>

With this change, the build date would not appear. That's actually fine with me -- does anyone else think it'd be a problem?

And if you do, since we don't have a test for this, then we need to change the code in ImplementationVersion.java so it won't NPE if Build-Date isn't found. (And +100 to writing a test that starts the shell!)

carlosame:
> And +100 to writing a test that starts the shell!

Starting a shell is not a problem, the real issue is to quit the shell without exiting the JVM (which happened every time that I fed a `quit()` into the shell).

In fact the shell is not well prepared to be re-executed, so I include a few initial changes about that. And as the test cannot quit the shell cleanly, it interrupts the thread which isn't very elegant but does the job for now (further changes to the shell could be done in the future).

As always, I could remove any commit in this PR and leave it for later.

carlosame:
From the commit descriptions:

1) Gradle: include license in `META-INF` directory of jar files (#956)

2) Update the Rhino Shell man page, distribution fixes (#1119)
2.1 Ship the man page in Zip/Tar distributions.
2.2 Include the runtime and engine jars in the distributions.
2.3 Do not attempt to copy non-existing `build.properties` to distributions.
2.4 Compress Tar distribution file.
2.5 Update POM metadata descriptions.
2.6 Upgrade deprecated jar classifier configurations.
2.7 Make sure that Git handles the man page as text (`.gitattributes`).

3) Create reproducible jar files (#1120)
3.1 Remove Built-Date and Built-Time fields from MANIFEST files.
3.2 Do not depend on Built-Date in `ImplementationVersion` (reported by @gbrail).
3.3 Configure archive files to be reproducible.
3.4 Convert licensing files to Windows line endings before adding them to archives.
3.5 Explicitly set groovy files —used by new line ending conversion (LEC) utility— as text in `.gitattributes`.
3.6 On `clean`, clean also the new `buildSrc` directory where the LEC utility is located.

4) Shell: allow re-executions, add a test

Note: the LEC class is dual-licensed because is adapted from my BSD-licensed projects.

carlosame:
Added a commit with a `NOTICE-tools.txt` containing the license of the treetable used in the debugger, see https://github.com/mozilla/rhino/issues/956#issuecomment-991930317.

build.gradle
```diff
@@ -150,10 +150,30 @@ idea {
     }
 }
 
+tasks.register('lineEndingConvCopy', CRLFConvertCopy) {
```
gbrail, line 153: Can you help me understand why we need this? Is Windows-style CRLF the standard for these types of files in a JAR?

gbrail (commented):
Thanks! There's a lot of important and helpful stuff and I look forward to merging it very soon.

I don't really understand about the machinery to convert line endings. Do other Java / Gradle projects do things this way? Unless I misunderstand, we seem to be introducing some additional code and configuration so that we can guarantee Windows-style newlines in the LICENSE and NOTICE files in the JAR. I don't recall seeing that elsewhere. Is that something that we need to do?

build.gradle
carlosame: For the jar being reproducible, one must standardize on either Unix or Windows line endings for the text files inside, otherwise Unix and Windows machines would produce different jar files. I chose Windows because:

a) Market share of Windows is larger.
b) Unix line endings could cause problems when viewed with some Windows tools like Notepad, the other way around is less likely to be problematic.
c) Unix/Linux users generally know about line endings, Windows users often do not.

I do that conversion in _all_ of my projects; do not know who else does that. And now that I think about it, I need to convert the `MANIFEST.MF` file as well...

If you are uncomfortable with that conversion, I could remove it. Note that the original license text sources are left unchanged, otherwise there could be some Git glitches.

carlosame: > And now that I think about it, I need to convert the `MANIFEST.MF` file as well...

Manifest files are written with Manifest.write() which always use CRLF, no need to convert. What needs to be uniformized are the resource files, that could be switched to CRLF (or to Unix line endings) once they have been copied to `buildGradle/resources/main` during the build.

Do I proceed with the line ending conversion of the resources, or should I remove the conversion stuff completely?

gbrail: My first choice would be to not have the line ending conversion code in there, but ensure using .gitattributes or some other suitable mechanism that the line endings are consistent. 

I would also prefer to keep them Unix-style since the majority of the open source world lives in the Unix world (even on Windows). And this is, after all, a Mozilla project.

Is there a way that we can have this without the conversion code using .gitattributes?

carlosame: > Is there a way that we can have this without the conversion code using .gitattributes?

Yes, although a `git add --renormalize .` is recommended for people on Windows. I'll push it soon.

carlosame: The idea was to have the following `.gitattributes`:
```gitattributes
# Set default behavior, in case users don't have core.autocrlf set.
* text=auto eol=lf

.classpath text
.project text
*.html text eol=lf
*.htm text eol=lf
*.java text eol=lf
*.js text eol=lf
*.jstest text eol=lf
*.jsp text eol=lf
*.php text eol=lf
*.properties text eol=lf
*.sh text eol=lf
*.xml text eol=lf
*.md text eol=lf
*.bat text eol=crlf
*.gradle text eol=lf
*.pl text eol=lf
*.mk text eol=lf
*.tests text eol=lf
*.doctest text eol=lf
/man/*.1 text eol=lf
/gradlew text eol=lf

*.bmp binary
*.gif binary
*.jar binary
*.jpg binary
*.jpeg binary
*.png binary
```
Unfortunately, this breaks spotless. I'll push the commit without it, and then it can be addressed later (after 1.7.14 when spotless is applied to the remaining files).

By then, the following `.editorconfig` file could be added as well:
```ini
#
# https://editorconfig.org/
#
root = true

[*]
end_of_line = lf
charset = utf-8

[*.{java,js,gradle}]
indent_style = space
indent_size = 4

[*.bat]
end_of_line = crlf
charset = latin1
```

p-bakker:
This this branch would also close #918 and thus make #919 obsolete

carlosame:
> Merging this branch would I think also close #918 and thus make #919 obsolete

Well not exactly this one: I've deferred the line ending changes to a future PR which I could do but could instead be yours?

And given that the latest CI tests failed due to one of the shell timer tests (perhaps due to something that my new shell test left initialized in something static?), I'm thinking about splitting the `ShellTest` to a different PR. This branch has important Gradle and packaging changes (notably the shipped license information) that are clearly good for 1.7.14. Having or not having a shell test in 1.7.14 is equally good but more debatable.

carlosame:
Re-pushed 3 last commits. Changes:

1) Report timer JS error using Rhino's error reporting infrastructure.
2) On timer error, set the shell exit code to 3.
3) Use a `CountDownLatch` in the shell tests. Should be slightly faster and still behave well under CI.
4) Increased the timeout values in my version of the NestedTimers test. On some systems, timers are unreliable in timescales below 12-15ms, although this should not matter for Rhino's current Timer implementation.

gbrail:
Thank you a lot for all your hard work on this. I really want to get a release out the door by the end of the year, but I can't promise because I'm also taking time off. 

I would like to merge a version of this that has just the build fixes -- no shell test -- that seems like it'll take some more work! 

If you don't mind, I'm going to try and cherry-pick your changes to the build process into a new PR and upload that so that we can include it in the release. I'll send out the PR shortly.

carlosame:
> If you don't mind, I'm going to try and cherry-pick your changes to the build process into a new PR and upload that so that we can include it in the release.

Please do.

carlosame:
I've splitted the fix for #1130 to a new PR #1134.

build.gradle
```diff
@@ -150,10 +150,25 @@ idea {
...
     dependsOn compileJava
     archiveBaseName = 'rhino-runtime'
     from sourceSets.main.output
+    from ('LICENSE.txt') {
+        into 'META-INF'
+    }
+    from ('NOTICE.txt') {
+        into 'META-INF'
+    }
+    from ('NOTICE-tools.txt') {
```
p-bakker, line 169: Think `NOTICE-tools.txt` doesn't need to be included in the runtime jar, as that one doesn't contain the tool source, no?

carlosame: True, fixed.

PR #1127, opened 2021-12-10
ref. #995 Do not convert java.math.BigInteger.

### LINKED ISSUE #995: bigint values convert to java.lang.Double when being passed to java
```js
js> var big = BigInt('123')
js> var list = java.util.Collections.singletonList(big)
js> print(list.get(0).getClass())
class java.lang.Double
```
I expect they should be converted to java.math.BigInteger instead.

tuchida:
When passed `BigInt` to Java, it was converted to `java.lang.Double`. I fixed so that `BigInt` keeps `java.math.BigInteger`.

Closes #995

### DIFF
--- a/src/org/mozilla/javascript/NativeJavaObject.java
+++ b/src/org/mozilla/javascript/NativeJavaObject.java
@@ -13,6 +13,7 @@ import java.io.Serializable;
 import java.lang.reflect.Array;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
+import java.math.BigInteger;
 import java.util.Date;
 import java.util.Iterator;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


src/org/mozilla/javascript/NativeJavaObject.java
```diff
@@ -319,8 +322,10 @@ static int getConversionWeight(Object fromObj, Class<?> to) {
                     if (to == ScriptRuntime.StringClass) {
                         // native numbers are #1-8
                         return 9;
-                    } else if (to == ScriptRuntime.ObjectClass) {
+                    } else if (to == ScriptRuntime.BigIntegerClass) {
```
tuchida, line 325: I couldn't figure out how to write a test for `getConversionWeight`.

gbrail:
I think this is safe for 1.7.14. I'll test locally. Any other opinions on that?

p-bakker:
FWIW I think it is

testsrc/org/mozilla/javascript/tests/NativeJavaObjectTest.java
```diff
@@ -0,0 +1,58 @@
...
+import org.mozilla.javascript.Context;
+import org.mozilla.javascript.NativeJavaObject;
+import org.mozilla.javascript.Scriptable;
+
+public class NativeJavaObjectTest extends TestCase {
+
+    @Test
+    public void testCoerceType() {
+        Context cx = Context.enter();
+        try {
```
gbrail, line 23: It's not important to fix now, but for future reference, you can write:

`try (Context cx = Context.enter()) {}`

because Context is Closeable.

gbrail:
Thanks!

PR #1132, opened 2021-12-22
Date ctor has to respect the timezone in some cases

rbri:


gbrail:
This works great. Thanks! I'm rebasing it to slightly reduce the number of commits, and pushing, so I'll close this manually.

PR #1133, opened 2021-12-22
Improve the build process to include reproducible JARs

gbrail:


p-bakker:
@gbrail any reason why you excluded NOTICE-tools.txt that is included in https://github.com/mozilla/rhino/pull/1124

gbrail:
Sorry, I didn't see that in the PR -- but it's important, so I added it in a separate PR: https://github.com/mozilla/rhino/pull/1143

Issue #980, opened 2021-07-01
Publish a snapshot build somewhere


p-bakker:
Think it would be handy if we'd publish a snapshot build somewhere everytime we succesfully push to master, so users of Rhino could try out what's being done on master, without having to go through the motions of checking out the source and building the snapshot jar themselves

p-bakker:
Wondering if through GitHub Actions we could publish the snapshot jar as a [package](https://github.com/mozilla/rhino/packages) on our repo

tonygermano:
I think this is a good idea, and should be possible. We're already compiling with GitHub Actions to run the tests, but right now the only artifacts we are retaining are the reports. We'd probably have to start naming with build versions (timestamps?) to differentiate them instead of naming them all SNAPSHOT.

p-bakker:
My idea was to have every new snapshot overwrite the previous, so couldn't the naming be very straight forward on that case?

tonygermano:
That would be simpler from an implementation standpoint, but harder to know which snapshot version a user was running when reporting bugs or giving feedback.

It looks like as recently as 1.7.11 there were official release candidates before the final release.

https://docs.github.com/en/actions/reference/environment-variables#default-environment-variables
Perhaps we could incorporate `GITHUB_RUN_ID` or `GITHUB_RUN_NUMBER` into the filename?

`rhino-1.7.14-snap.${GITHUB_RUN_ID}.jar`?

It's been suggested in https://github.com/mozilla/rhino/issues/966#issuecomment-870926110 and https://github.com/mozilla/rhino/discussions/972#discussioncomment-942137 to move to semantic versioning. I think if that happens what I suggested above will have to change anyway as (nearly?) every PR would trigger a version bump and new release.

I just found this, which looks like a decent write-up on the topic https://css-tricks.com/how-to-automate-project-versioning-and-releases-with-continuous-deployment/

carlosame:
> Wondering if through GitHub Actions we could publish the snapshot jar as a package on our repo

Github packages require authentication to be used, see:

https://github.community/t/download-from-github-package-registry-without-authentication/14407

Github's Maven repository is currently useless.

botic:
Would be great to have a referencable snapshot release for our Ringo 3.0.0 release candidates. I'm building from source atm and 1.7.14 feels very solid so far.

gbrail:
[...]

PR #1135, opened 2021-12-23
Action to publish packages on Github (#980)

carlosame:
Publishes packages on release (important: the version should not be changed to '-SNAPSHOT' before creating the release in Github). Can also be run manually to publish snapshot packages (publishing a snapshot on each `master` push, as suggested in #980, seems too frequent).

The Github packages are not being signed, although that could be changed in the future if deemed appropriate (requires a private key being stored as a "secret" in Github).

Note: to automatically build on release, I'm using the `published` release type instead of `created` which is what is [being suggested by the documentation](https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows). The `created` type did not work for me (in another repository).

Once it is merged, I'd suggest testing it by manually publishing a snapshot package.

### DIFF
--- a/.github/workflows/package-publish.yml
+++ b/.github/workflows/package-publish.yml
@@ -1,0 +1,36 @@
+name: Github Packages deployment
+
+on:
+  # Publishes packages on release (the Rhino version should not be
+  # changed to '-SNAPSHOT' before creating the release in Github).
+  release:
+    types: [published]
+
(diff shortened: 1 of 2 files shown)


gbrail:
Thanks -- I did this a different way in https://github.com/mozilla/rhino/pull/1141, for both Maven and GitHub, but I based the GitHub work on this PR. Thanks!

PR #1141, opened 2022-01-04
Add rule to publish to Maven Central

gbrail:
Taking the lead from https://github.com/mozilla/rhino/pull/1135, but adding signing and testing it myself so that I can understand all this ;-)


gbrail:
In addition to this rule, I added another to publish to the Gradle package repositiory, based on the work by @carlosame. So this PR also closes https://github.com/mozilla/rhino/issues/980.

PR #1142, opened 2022-01-05
Add additional third-party license to the NOTICE

gbrail:


gbrail:
On second thought, I will indeed do it in a separate file, so that it doesn't pollute other JARs. Hold on...

Issue #1052, opened 2021-10-13
Support ES2022 `Object.hasOwn()`


p-bakker:
Proposal: https://github.com/tc39/proposal-accessible-object-hasownproperty
MDN: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwn

Issue #967, opened 2021-06-29
Resolve structural issue with exception handling in IRFactory


tuchida:
In the example below, one can see that the `IRFactory` throws a `ParserException` when in this case a `SyntaxException` ought to be thrown. 

Further down comment https://github.com/mozilla/rhino/issues/967#issuecomment-874616417 points to a fix that results in the proper Error being thrown, but proper location info is missing.

These issues ought to be fixed, possibly requiring the refactoring of the `IRFactory` as suggested below

https://github.com/mozilla/rhino/pull/853#issuecomment-813031347
If I evaluate `1=1`, I will expect only JavaScript error, but Java exception will be thrown.
```
$ java -jar buildGradle/libs/rhino-1.7.14-SNAPSHOT.jar
Rhino 1.7.14-SNAPSHOT 2021 06 29
js> 1=1
js: line 1: Invalid assignment left-hand side.
js: 
js: ^
Exception in thread "main" org.mozilla.javascript.Parser$ParserException
	at org.mozilla.javascript.Parser.reportError(Parser.java:329)
	at org.mozilla.javascript.Parser.reportError(Parser.java:315)
	at org.mozilla.javascript.Parser.reportError(Parser.java:310)
	at org.mozilla.javascript.IRFactory.createAssignment(IRFactory.java:2309)
	at org.mozilla.javascript.IRFactory.transformAssignment(IRFactory.java:448)
	at org.mozilla.javascript.IRFactory.transform(IRFactory.java:222)
	at org.mozilla.javascript.IRFactory.transformExprStmt(IRFactory.java:559)
	at org.mozilla.javascript.IRFactory.transform(IRFactory.java:219)
	at org.mozilla.javascript.IRFactory.transformScript(IRFactory.java:1133)
	at org.mozilla.javascript.IRFactory.transform(IRFactory.java:201)
	at org.mozilla.javascript.IRFactory.transformTree(IRFactory.java:121)
	at org.mozilla.javascript.Context.parse(Context.java:2437)
	at org.mozilla.javascript.Context.compileImpl(Context.java:2354)
	at org.mozilla.javascript.Context.compileString(Context.java:1368)
	at org.mozilla.javascript.Context.compileString(Context.java:1356)
	at org.mozilla.javascript.tools.shell.Main.processSource(Main.java:495)
	at org.mozilla.javascript.tools.shell.Main.processFiles(Main.java:181)
[...]

Issue #958, opened 2021-06-25
Support ES2015 Regex `u` flag (Unicode)


p-bakker:
See https://mathiasbynens.be/notes/es6-unicode-regex

p-bakker:
Got a whole bunch of tests from test262 failing, because the harness uses the `u` flag in some places

tuchida:
In order to run `Function.prototype.toString` tests, `assertNativeFunction` needs to work.
https://github.com/tc39/test262/blob/f94fc660cc3c59b1f2f9f122fc4d44b4434b935c/harness/nativeFunctionMatcher.js#L26

p-bakker:
Will try next week to resurrect my attempt to implement this and share my WIP

p-bakker:
Did make progress, but not yet as far as I hoped. Still at it though

p-bakker:
Got some Java implementation-related questions I'm hoping to get some answers on before going down the wrong rabbits-hole. 

Right now, inside the regex implementation there are 2 things going on that I'm contemplating on changing:
1. all logic operates on the chars. Chars however can only contain chars with a codePoint up to 65xxx. Unicode however is about also supporting codePoints beyond that limit. So I'm contemplating changing all the internal regex code from operating on chars to operating on ints
2. regexes are compiled to some 'bytecode', which is stored in a byte[], with some logic to store char values in two separate bytes. Unicode codePoints beyond 65xxx don't fit into two separate bytes anymore, so I either adjust the logic to support splitting higher codePoints into 3 separates bytes in the byte[] or I forego the byte[] completely and switch is to an int[], which allows any unicode codePoint to be stored in a single position

Anyone got insight what the changes from char > int and byte[] to int[] would do memory and performance-wise?

p-bakker:
And another questionThe logic needs the mappings defined in this file: https://www.unicode.org/Public/14.0.0/ucd/CaseFolding.txt

What's the better approach: 
1. include this file in the Rhino source and have a class read it at runtime and create a mapping object
2. write a build script that takes this file and generates a class file from it

tuchida:
> Anyone got insight what the changes from char > int and byte[] to int[] would do memory and performance-wise?

I am not familiar with Java performance, but when I implemented BigInt I used `./gradlew testBenchmark` to check it.
- #885

Perhaps you can check with `SunSpiderBenchmark.regexpDna`.

[...]

Issue #913, opened 2021-05-27
Support ES2015 Computed Property Names


p-bakker:
Case just to track this specific ES6 feature (and how hard can it be to implement :-) ) and also because some tests in Test262 are disabled because they use this feature (and thus we cannot run them)

```
var name = 'myPropertyName';

Object.keys({
   [x] = 'someValue';
})[0] === name; // to yield true
```

rPraml:
I've tried it here https://github.com/FOCONIS/rhino/commit/98ef8154bc056589daecbcc98c47d0b45f2b1f4a but I was lost in the ByteCode-generation for different optimization levels. Some tests will run, but some trigger a Kit.codeBug

tuchida:
I tried to implement this in the past, but I think it was difficult to evaluate them in the proper order.
https://github.com/tuchida/rhino/commit/0fef518cbecfb9fd5fdd9703216b0dec184975f9

```js
var obj = {
  [exp1]: exp2,
  "str1": exp3,
  [exp4]: exp5
};
```

p-bakker:
Great to see there are already 2 attempts underway, guess I shouldn't start a third :-)

Any thoughts on which one seems to be more feasible? Both seem to take quite a different approach. Just by quickly browsing through the changes (for just the interpreted impl.) in either branch, it seems @rPraml 's approach has a lot smaller footprint (but no idea though if it covers all angles)

FYI: think in both branches, this feature exclusion in Test262SuiteTest should also be removed: https://github.com/mozilla/rhino/blob/master/testsrc/org/mozilla/javascript/tests/Test262SuiteTest.java#L78

tuchida:
I don't have any plans to move forward with the implementation right now, as it is a long time ago on my branch.

PR #1157, opened 2022-01-16
add Object.hasOwn (#1052)

archethic:
add Object.hasOwn

### DIFF
--- a/src/org/mozilla/javascript/AbstractEcmaObjectOperations.java
+++ b/src/org/mozilla/javascript/AbstractEcmaObjectOperations.java
@@ -25,6 +25,32 @@ class AbstractEcmaObjectOperations {
     }
 
     /**
+     * Implementation of Abstract Object operation HasOwnProperty as defined by EcmaScript
+     *
+     * @param cx
+     * @param o
+     * @param pn
+     * @return boolean
+     * @see <a href="https://tc39.es/proposal-accessible-object-hasownproperty/"></a>
+     */
+    static boolean hasPropertyInObject(Context cx, Object o, Object pn) {
+        ScriptableObject obj = ScriptableObject.ensureScriptableObject(o);
+        boolean result;
+        if (pn instanceof Symbol) {
+            result = obj.has((Symbol) pn, obj);
+        } else {
+            ScriptRuntime.StringIdOrIndex s = ScriptRuntime.toStringIdOrIndex(cx, pn);
+            if (s.stringId == null) {
+                result = obj.has(s.index, obj);
+            } else {
+                result = obj.has(s.stringId, obj);
+            }
+        }
+
+        return result;
+    }
+
+    /**
      * Implementation of Abstract Object operation testIntegrityLevel as defined by EcmaScript
      *
      * @param cx
--- a/src/org/mozilla/javascript/AbstractEcmaObjectOperations.java
+++ b/src/org/mozilla/javascript/AbstractEcmaObjectOperations.java
@@ -29,17 +29,17 @@ class AbstractEcmaObjectOperations {
      *
      * @param cx
      * @param o
-     * @param pn
+     * @param property
      * @return boolean
-     * @see <a href="https://tc39.es/proposal-accessible-object-hasownproperty/"></a>
+     * @see <a href="https://262.ecma-international.org/12.0/#sec-hasownproperty"></a>
      */
-    static boolean hasPropertyInObject(Context cx, Object o, Object pn) {
+    static boolean hasOwnProperty(Context cx, Object o, Object property) {
         ScriptableObject obj = ScriptableObject.ensureScriptableObject(o);
         boolean result;
-        if (pn instanceof Symbol) {
-            result = obj.has((Symbol) pn, obj);
+        if (property instanceof Symbol) {
+            result = ScriptableObject.ensureSymbolScriptable(o).has((Symbol) property, obj);
         } else {
-            ScriptRuntime.StringIdOrIndex s = ScriptRuntime.toStringIdOrIndex(cx, pn);
+            ScriptRuntime.StringIdOrIndex s = ScriptRuntime.toStringIdOrIndex(cx, property);
             if (s.stringId == null) {
                 result = obj.has(s.index, obj);
             } else {
--- a/src/org/mozilla/javascript/NativeObject.java
+++ b/src/org/mozilla/javascript/NativeObject.java
@@ -51,6 +51,7 @@ public class NativeObject extends IdScriptableObject implements Map {
             addIdFunctionProperty(ctor, OBJECT_TAG, ConstructorId_entries, "entries", 1);
             addIdFunctionProperty(ctor, OBJECT_TAG, ConstructorId_fromEntries, "fromEntries", 1);
             addIdFunctionProperty(ctor, OBJECT_TAG, ConstructorId_values, "values", 1);
+            addIdFunctionProperty(ctor, OBJECT_TAG, ConstructorId_hasOwn, "hasOwn", 1);
         }
         addIdFunctionProperty(ctor, OBJECT_TAG, ConstructorId_keys, "keys", 1);
         addIdFunctionProperty(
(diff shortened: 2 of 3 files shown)


p-bakker:
Tnx for your contribution! I'm sure one of the committers will take a look at your PR as soon as they can.

One thing to note though is that your PR doesn't come with a UnitTest, which normally wouldn't be a problem, as the Test262 testsuite would contain a test for this feature. However, as Object.hasOwn is a brand new thing in EcmaScript and atm we're not yet running against the latest version of Test262 yet (as Rhino is missing a couple of features that are used in the Test262 test suite), there's no test covering this feature.

Not sure how we want to handle this, except for trying to implement the missing features in Rhino, so we can run the latest version of the Test262 test suite

testsrc/org/mozilla/javascript/tests/es2022/NativeObjectTest.java
```diff
@@ -0,0 +1,112 @@
...
+
+import static org.junit.Assert.assertEquals;
+
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+import org.mozilla.javascript.Context;
+import org.mozilla.javascript.ScriptableObject;
+
+public class NativeObjectTest {
```
tuchida, line 11: 1. Is there a test for not following the prototype chain?
  https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwn
> ```js
> Object.hasOwn(example, 'toString');         // returns false
> ```

2. How is `Symbol` handled in the spec?

tuchida:
The build is failing. Could you run [`spotlessApply`](https://github.com/mozilla/rhino#code-formatting)?
```sh
$ ./gradlew spotlessApply
```

src/org/mozilla/javascript/NativeObject.java
```diff
@@ -472,6 +473,24 @@ public Object execIdCall(
                     }
                     return cx.newArray(scope, ids);
                 }
+            case ConstructorId_hasOwn:
+                {
+                    Object arg = args.length < 1 ? Undefined.instance : args[0];
+                    ScriptableObject obj = ensureScriptableObject(arg);
+                    Object str = args.length < 2 ? Undefined.instance : args[1];
+                    boolean result;
```
tuchida, line 481: I read the spec.
https://tc39.es/proposal-accessible-object-hasownproperty/

`Object.hasOwn` calls the `HasOwnProperty` operation in the same way as `Object.prototype.hasOwnProperty`. Could you make this common? I think it is better to write it in `ScriptRuntime` or `ScriptableObject`.
https://github.com/mozilla/rhino/blob/a0a9fb97f0439a978581283277325e382557ac67/src/org/mozilla/javascript/NativeObject.java#L207-L219

p-bakker: Good point!

The correct location to put it however is org.mozilla.javascript.AbstractEcmaObjectOperations, as the spec defines the HasOwnProperty operation as one of the Abstract Operations on Objects (section 7.3.13).

Do cehck the notes at the top of the file as to how to add stuff: https://github.com/mozilla/rhino/blob/master/src/org/mozilla/javascript/AbstractEcmaObjectOperations.java#L8

archethic:
> The build is failing. Could you run [`spotlessApply`](https://github.com/mozilla/rhino#code-formatting)?
> 
> ```shell
> $ ./gradlew spotlessApply
> ```

I'm so sorry! I didn't check the code formatting properly. I did code formatting with the https://github.com/mozilla/rhino/pull/1157/commits/e073746112e665f38ab57e264002c820d7bcc153 commit.

src/org/mozilla/javascript/NativeObject.java
archethic: I changed 'hasOwnProperty' and 'hasOwn' to call the function 'hasOwnPropertyInObject' with the https://github.com/mozilla/rhino/pull/1157/commits/ec9b15047077626b59df3fc98407345f2787d433 commit. I'll have to find more codes to handle the same process here.

src/org/mozilla/javascript/AbstractEcmaObjectOperations.java
```diff
@@ -24,6 +24,32 @@
...
     }
 
+    /**
+     * Implementation of Abstract Object operation HasOwnProperty as defined by EcmaScript
+     *
+     * @param cx
+     * @param o
+     * @param pn
+     * @return boolean
+     * @see <a href="https://tc39.es/proposal-accessible-object-hasownproperty/"></a>
```
tuchida, line 34: This method is `HasOwnProperty` of abstract operation. This would be a more appropriate link.
https://262.ecma-international.org/12.0/#sec-hasownproperty

src/org/mozilla/javascript/AbstractEcmaObjectOperations.java
```diff
@@ -24,6 +24,32 @@
...
+     */
+    static boolean hasPropertyInObject(Context cx, Object o, Object pn) {
```
tuchida, line 36: I want the method name to be `hasOwnProperty`.

src/org/mozilla/javascript/AbstractEcmaObjectOperations.java
```diff
@@ -24,6 +24,32 @@
...
+        ScriptableObject obj = ScriptableObject.ensureScriptableObject(o);
+        boolean result;
+        if (pn instanceof Symbol) {
+            result = obj.has((Symbol) pn, obj);
```
tuchida, line 40: Don't you need `ensureSymbolScriptable`?
https://github.com/mozilla/rhino/blob/5a8d31100cf840d2d5746c25d10c0c7f52f8fad0/src/org/mozilla/javascript/NativeObject.java#L210

archethic: > Don't you need `ensureSymbolScriptable`?
> 
> https://github.com/mozilla/rhino/blob/5a8d31100cf840d2d5746c25d10c0c7f52f8fad0/src/org/mozilla/javascript/NativeObject.java#L210

When I tested, it went well without the method, but there was a problem that I didn't predict, so I'll fix it in the next commit! Thank you for pointing out.

src/org/mozilla/javascript/AbstractEcmaObjectOperations.java
```diff
@@ -24,6 +24,32 @@
         SEALED
     }
 
+    /**
+     * Implementation of Abstract Object operation HasOwnProperty as defined by EcmaScript
+     *
+     * @param cx
+     * @param o
+     * @param pn
```
p-bakker, line 32: I'd prefer a more descriptive name, for example 'property'

archethic: Thank you for pointing out! I refactor at https://github.com/mozilla/rhino/pull/1157/commits/de372ec5f245dbc205351d11212032d5b7f30ee6

src/org/mozilla/javascript/AbstractEcmaObjectOperations.java
archethic: Thank you for pointing out! I refactor in https://github.com/mozilla/rhino/pull/1157/commits/de372ec5f245dbc205351d11212032d5b7f30ee6

src/org/mozilla/javascript/AbstractEcmaObjectOperations.java
archethic: Thank you for pointing out! I refactor in https://github.com/mozilla/rhino/pull/1157/commits/de372ec5f245dbc205351d11212032d5b7f30ee6

tuchida (approved):
looking good!

p-bakker (approved):
Looking good to me as well

gbrail:
This seems good. Are you sure that there aren't tests in test262 that can be enabled with this change?

You can check for this automatically by doing this (described in testsrc/README.md)

./gradlew test --tests org.mozilla.javascript.tests.Test262SuiteTest --rerun-tas
ks -DupdateTest262properties

This is fine with me to merge as soon as we verify this. Thanks!

tuchida:
I couldn't find any.
```sh
$ find test262/test -type f -exec grep -e Object\\.hasOwn\\b {} + | wc -l
0
```

archethic:
https://github.com/tc39/test262/tree/f94fc660cc3c59b1f2f9f122fc4d44b4434b935c There seems to be no test for `hasOwn` based on the commit standard. But, in the recent commit of test262, there is a test of `hasOwn`, so can the test262 commit be uploaded a little more recently?

p-bakker:
> Are you sure that there aren't tests in test262 that can be enabled with this change?

Object.hasOwn got proposed and adopted after the version of the test262 suite we're running against got authored, so I don't expect any tests for it in the version we're running against

> so can the test262 commit be uploaded a little more recently?

Wish we could, but if we update to just the next commit in test262 after the one we're on, running the testsuite bombs out because as of that point the test262 suite depends on JavaScript features that Rhino doesn't implement yet, like #958, #913 and #678. The later two if these issues [might depend](https://github.com/mozilla/rhino/issues/967#issuecomment-1015070316) on #967, which is currently waiting for #1159, which in turn [is waiting for ](https://github.com/mozilla/rhino/pull/1159#issuecomment-1022770756) this pr to get merged, among others :-)

Hence those 3 issues are linked to the v1.7.15 milestone, as we should tackle them asap imho.

gbrail:
OK, I think that we are as far as we're going to get with test262, so I'm going to merge this then. Thanks!

PR #1159, opened 2022-01-18
Apply spotlessApply to all code

tuchida:
ref. #1024
```diff
diff --git a/build.gradle b/build.gradle
@@ -440,7 +440,6 @@ spotbugs {
 }
 
 spotless {
-    ratchetFrom 'code-formatting-required'
     java {
         googleJavaFormat().aosp()
     }
```
I ran `./gradlew spotlessApply` in this state.


p-bakker:
@gbrail Any thoughts on this?

p-bakker:
@gbrail @rbri thoughts? Can this be merged so we're done with all the reformatting and PR's can be cleaner?

gbrail:
I'd like to catch up with the current backlog of PRs over the next day or two, at which point I'll either try to merge this or create my own PR!

p-bakker:
Not sure if it should be tackled at the same time as this PR or not, but as part of #1120  it was [suggested](https://github.com/mozilla/rhino/issues/1120#issuecomment-1000332366) to also make line endings consistent, see a proposal [here](https://github.com/mozilla/rhino/pull/1124#discussion_r769095217).

I guess the reason for wanting to combine the two is that is means files only need to be touched once? But if we stick to unix line endings (contrary to the earlier proposed switch to Windows line endings) files wouldn't have to change as they already use unix line endings

gbrail:
I'd rather deal with the line endings separately (I'm still not sure that we need to change anything). Let's merge this!

PR #1163, opened 2022-01-24
Add proper import package instructions in the OSGi manifest

makusuko:
As Rhino now uses classes from javax-packages, add proper imports for them. Namely: javax.lang.model.SourceVersion and the various javax.script classes used for implementing the Java scripting support. This fixes the use of current Rhino version as an OSGi bundle.


gbrail:
Thanks for staying on top of this!

Issue #963, opened 2021-06-27
Fix all properties implemented as a data property where the spec requires an accessor property


p-bakker:
On several built-ins properties are implemented ad data properties, while the spec calls for accessor properties

Currently only the `size` property for Set and Map is implemented properly (through NativeMap/Set.init(...))

Affects:
- TypedArray.buffer/byteLength/byteOffset
- DataView.buffer/byteLength/byteOffset
- ArrayBuffer.byteLength
- Regex.flag/global/multiline/ignoreCase
- Symbol.description (see https://github.com/mozilla/rhino/issues/962)
- All @@species properties

All locations in the spec can be found by searching the spec for the text `is an accessor property`

rbri:
Will have a look....

PR #1183, opened 2022-02-15
convert RegEx properties

rbri:
Implements the Regex part of #963

### DIFF
--- a/src/org/mozilla/javascript/regexp/NativeRegExp.java
+++ b/src/org/mozilla/javascript/regexp/NativeRegExp.java
@@ -114,6 +114,7 @@ public class NativeRegExp extends IdScriptableObject {
     private static final SymbolKey GET_IGNORE_CASE = new SymbolKey("[Symbol.getIgnoreCase]");
     private static final SymbolKey GET_MULTILINE = new SymbolKey("[Symbol.getMultiline]");
     private static final SymbolKey GET_STICKY = new SymbolKey("[Symbol.getSticky]");
+    private static final SymbolKey GET_SOURCE = new SymbolKey("[Symbol.getSource]");
 
     public static void init(Context cx, Scriptable scope, boolean sealed) {
 
@@ -161,6 +162,12 @@ public class NativeRegExp extends IdScriptableObject {
         desc.put("configurable", desc, Boolean.TRUE);
         desc.put("get", desc, proto.get(GET_STICKY, proto));
         proto.defineOwnProperty(cx, "sticky", desc);
+
+        desc = (ScriptableObject) cx.newObject(scope);
+        desc.put("enumerable", desc, Boolean.FALSE);
+        desc.put("configurable", desc, Boolean.TRUE);
+        desc.put("get", desc, proto.get(GET_SOURCE, proto));
+        proto.defineOwnProperty(cx, "source", desc);
 
         if (sealed) {
             proto.sealObject();
@@ -2667,6 +2674,10 @@ public class NativeRegExp extends IdScriptableObject {
             initPrototypeMethod(REGEXP_TAG, id, GET_STICKY, "get sticky", 0);
             return;
         }
+        if (id == SymbolId_getSource) {
+            initPrototypeMethod(REGEXP_TAG, id, GET_SOURCE, "get source", 0);
+            return;
+        }
 
         String s;
         int arity;
@@ -2743,6 +2754,9 @@ public class NativeRegExp extends IdScriptableObject {
             case SymbolId_getSticky:
                 return realThis(thisObj, f).js_getSticky();
 
+            case SymbolId_getSource:
+                return realThis(thisObj, f).toString();
+
             case SymbolId_match:
                 return realThis(thisObj, f).execSub(cx, scope, args, MATCH);
 
@@ -2781,6 +2795,9 @@ public class NativeRegExp extends IdScriptableObject {
         }
         if (GET_STICKY.equals(k)) {
             return SymbolId_getSticky;
+        }
+        if (GET_SOURCE.equals(k)) {
+            return SymbolId_getSource;
         }
         return 0;
     }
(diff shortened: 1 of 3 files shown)


src/org/mozilla/javascript/regexp/NativeRegExp.java
```diff
@@ -2783,6 +2814,28 @@ protected int findPrototypeId(String s) {
         return id;
     }
 
+    private Object js_getFlags() {
```
p-bakker, line 2817: Is there any particular reason why these new methods are all prefixed with `js_`?

src/org/mozilla/javascript/regexp/NativeRegExp.java
```diff
@@ -2512,14 +2548,7 @@ private static void reportError(String messageId, String arg) {
...
 
-    private static final int Id_lastIndex = 1,
-            Id_source = 2,
-            Id_flags = 3,
-            Id_global = 4,
-            Id_ignoreCase = 5,
-            Id_multiline = 6,
-            Id_sticky = 7,
-            MAX_INSTANCE_ID = 7;
+    private static final int Id_lastIndex = 1, Id_source = 2, MAX_INSTANCE_ID = Id_source;
```
p-bakker, line 2551: Sry, looks like `source` also ought to be an accessor property now:
https://tc39.es/ecma262/#sec-get-regexp.prototype.source
https://tc39.es/ecma262/#sec-get-regexp.prototype.source

rbri: no problem, stay tuned :-)

src/org/mozilla/javascript/regexp/NativeRegExp.java
rbri: We used this pattern at other places also to mark methods providing the impl of a specific js method/accessor call

rbri: Yes maybe if all the other thinks are solved ;-)

p-bakker: https://github.com/mozilla/rhino/wiki/DRAFT:-Stuff-you'd-want-to-know-about-the-codebase

src/org/mozilla/javascript/regexp/NativeRegExp.java
```diff
@@ -109,6 +109,12 @@
 
     private static final int ANCHOR_BOL = -2;
 
+    private static final SymbolKey GET_FLAGS = new SymbolKey("[Symbol.getFlags]");
```
tuchida, line 112: Is this a `Symbol`? `Symbol` is written as `[ @@match ]`.
https://262.ecma-international.org/12.0/#sec-get-regexp.prototype.flags

src/org/mozilla/javascript/regexp/NativeRegExp.java
```diff
@@ -126,6 +132,36 @@ public static void init(Context cx, Scriptable scope, boolean sealed) {
 
         ctor.setImmunePrototypeProperty(proto);
 
+        ScriptableObject desc = (ScriptableObject) cx.newObject(scope);
```
tuchida, line 136: (I don't know much about this area, but...) Could you write simply using Lambda?
https://github.com/mozilla/rhino/blob/ae78dbfbbc50e205e71b19f1d44613ed462879b5/src/org/mozilla/javascript/NativePromise.java#L41-L76

testsrc/test262.properties
```diff
@@ -1407,47 +1403,25 @@ built-ins/RegExp 897/1464 (61.27%)
...
-    prototype/multiline/15.10.7.4-2.js
     prototype/multiline/cross-realm.js {unsupported: [cross-realm]}
-    prototype/multiline/length.js
-    prototype/multiline/name.js
-    prototype/multiline/S15.10.7.4_A9.js
     prototype/multiline/this-val-regexp-prototype.js
     prototype/source/cross-realm.js {unsupported: [cross-realm]}
-    prototype/source/length.js
-    prototype/source/name.js
     prototype/source/prop-desc.js
```
p-bakker, line 1416: Any idea why this one doesn't pass yet?

src/org/mozilla/javascript/regexp/NativeRegExp.java
rbri: Will check.....

testsrc/test262.properties
rbri: Strange will have a look

rbri:
This symbol seem to be too hacky, but the id scriptable object is a nightmare for this kind of properties. I have to digg deeper into the mud here. - will require some more days.

p-bakker:
@rbri just wondering how this PR is coming along, as I'm working on implementing the unicode flag and while analysing fialing tests related to the unicode property, some are due me currently having added the unicode property as a data property instead of an accessor property

testsrc/test262.properties
```diff
@@ -1407,47 +1403,25 @@ built-ins/RegExp 897/1464 (61.27%)
...
     prototype/source/this-val-regexp-prototype.js
     prototype/source/value-empty.js
     prototype/source/value-line-terminator.js
     prototype/source/value-u.js
     prototype/sticky/cross-realm.js {unsupported: [cross-realm]}
-    prototype/sticky/length.js
-    prototype/sticky/name.js
-    prototype/sticky/prop-desc.js
     prototype/sticky/this-val-regexp-prototype.js
```
p-bakker, line 1422: This test is dependant on the sticky property being implemented as an accessor property. However, it looks like your changes haven't caused it to pass yet.

Not sure if that is due to the accessor property impl or if something else, possibly unrelated, is going on

p-bakker:
@rbri just wondering how this PR is coming along

p-bakker:
@rbri any chance for a quick update on this one?

Is getting these types of properties properly implemented in IdScriptables maybe dead in the water completely and we should look at converting relevant implementations into lambda's instead?

rbri:
have migrated the test in pr #2455. Will close this as all of the tests are already passing.

PR #1184, opened 2022-02-17
fixed SourceVersion

archethic:
Check Android by creating a variable called `IS_ANDROID` in the `ScriptRuntime` class

++ Fixed `SourceVersion` bug.


p-bakker:
If this a stepping stone to another PR that will address the other Android challenges?

p-bakker:
Sure, but on is own it doesn't change anything, it just removed a tiny bit of code duplication.

I don't see why we wouldn't merge this, but if there would be another PR depending on this that would actually move the needle on Android support, I think it would be better

archethic:
There are many reasons for bugs in `Android`.

By changing the subject of pr slightly, I will solve the `SourceVersion` problem first.

p-bakker:
@carlosame given that you've spent time getting a proper source version check in place: what do you think about this approach?

PR #1185, opened 2022-02-18
Fix GitHub Languages for repo

### LINKED ISSUE #1129: Github classifies Rhino as a C++ project
I found it's because many of the old mozilla js tests have an incorrect emacs mode of C++ specified.

for example: https://github.com/mozilla/rhino/blob/f75125901ee1df681510200f63a6eb033342f488/testsrc/tests/lc3/ArrayMethods/byte-001.js#L1-L4) 

I think if all of them were fixed it would reclassify the project as javascript. Not sure if we'd want to exclude certain test directories from the calculation entirely so that Java is the predominant language.

Github uses [linguist](https://github.com/github/linguist) to calculate language usage in projects.

p-bakker:
Changes emacs linemode on all testsrc/tests/*.js files to javascript
(from C++/java)

New stats after this change (as can be seen on https://github.com/p-bakker/rhino):
- JavaScript 65.7%
- Java 33.0% 
- Other 1.3%

Closes #1129


rbri:
I think this has no risk to break anything - will merge this.

PR #1186, opened 2022-02-19
fix #780 fix Object.assign when undefined value and inextensible

### LINKED ISSUE #780: Object.assign() should copy undefined properties, I think
Hi! It looks like `Object.assign()` should copy undefined properties:

```
console.log(Object.assign({"a":undefined}, {"b":undefined}))
Object { a: undefined, b: undefined }
```

However, the Rhino code is skipping undefined values:

```
          case ConstructorId_assign:
          {
            if (args.length < 1) {
            }
            Scriptable targetObj = ScriptRuntime.toObject(cx, thisObj, args[0]);
              if ((args[i] == null) || Undefined.isUndefined(args[i])) {
                continue;
              }
              for (Object key : ids) {
                if (key instanceof String) {
                  Object val = sourceObj.get((String) key, sourceObj);
                  if ((val != Scriptable.NOT_FOUND) && !Undefined.isUndefined(val)) { // <=== HERE
                    targetObj.put((String) key, targetObj, val);
                  }
                  Object val = sourceObj.get(ii, sourceObj);
                  if ((val != Scriptable.NOT_FOUND) && !Undefined.isUndefined(val)) {  // <=== AND HERE
                    targetObj.put(ii, targetObj, val);
                  }
                }
              }
            }
            return targetObj;
          }
```

Is there a reason for those `!Undefined.isUndefined(val)` checks?

Thanks!

tuchida:
ref. #780
```js
Object.keys(Object.assign({a:undefined}, {b:undefined})).join()
// expected:<a[,b]> but was:<a[]>
```
I found another problem with it, so I fixed it while I was at it.
```js
var obj = Object.freeze({});
Object.assign(obj, { a: 1 });
// An error to be thrown, but nothing was thrown.
```

### DIFF
--- a/src/org/mozilla/javascript/ScriptableObject.java
+++ b/src/org/mozilla/javascript/ScriptableObject.java
@@ -1798,7 +1798,7 @@ public abstract class ScriptableObject
      * @param desc a property descriptor
      * @return true if this is a data descriptor.
      */
-    protected boolean isDataDescriptor(ScriptableObject desc) {
+    protected static boolean isDataDescriptor(ScriptableObject desc) {
         return hasProperty(desc, "value") || hasProperty(desc, "writable");
     }
 
@@ -1808,7 +1808,7 @@ public abstract class ScriptableObject
      * @param desc a property descriptor
      * @return true if this is an accessor descriptor.
      */
-    protected boolean isAccessorDescriptor(ScriptableObject desc) {
+    protected static boolean isAccessorDescriptor(ScriptableObject desc) {
         return hasProperty(desc, "get") || hasProperty(desc, "set");
     }
 
@@ -2365,7 +2365,7 @@ public abstract class ScriptableObject
         return Context.call(null, fun, scope, obj, args);
     }
 
-    private static Scriptable getBase(Scriptable obj, String name) {
+    static Scriptable getBase(Scriptable obj, String name) {
         do {
             if (obj.has(name, obj)) break;
             obj = obj.getPrototype();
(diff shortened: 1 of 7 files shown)


p-bakker:
Haven't looked at the changes, but haven't these fixes improved the test262 results?

tuchida:
The `undefined` test does not seem to be in test262.
The `Object.freeze` test also does not seem to be in the one used by Rhino. The latest version has it, but it requires the implementation of Computed Property Names.
- https://github.com/tc39/test262/pull/2945

p-bakker:
Right...

> The latest version has it, but it requires the implementation of Computed Property Names.

Not just Computed Property Names though. Also at least a proper const impl and unicode regex

src/org/mozilla/javascript/ScriptableObject.java
```diff
@@ -334,6 +334,38 @@ public void put(Symbol key, Scriptable start, Object value) {
         ensureSymbolScriptable(start).put(key, start, value);
     }
 
+    /**
+     * Set ( O, P, V, Throw = true)
+     *
+     * <p>https://262.ecma-international.org/12.0/#sec-set-o-p-v-throw
+     */
+    public static void putOrThrow(Context cx, Scriptable obj, String name, Object value) {
```
p-bakker, line 342: These two Abstract Object Operations ought to be placed in AbstractEcmaObjectOperations. While doing so, check the guidelines at the top of the file

tuchida: ref. [204cd5f](https://github.com/mozilla/rhino/pull/1186/commits/204cd5fb8ba8f2e5a4c17bf7902fee722cf65260)
I made to move to `AbstractEcmaObjectOperations`.

tuchida: I rebased it.

gbrail:
Yes -- this looks good. Thanks!

PR #1193, opened 2022-03-09
fix #934 Implement ES2017 Object.getOwnPropertyDescriptors

### LINKED ISSUE #934: Implement ES2017 Object.getOwnPropertyDescriptors
Proposal: https://github.com/tc39/proposal-object-getownpropertydescriptors
MDN: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/getOwnPropertyDescriptors

tuchida:
Closes #934
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/getOwnPropertyDescriptors

### DIFF
--- a/src/org/mozilla/javascript/NativeObject.java
+++ b/src/org/mozilla/javascript/NativeObject.java
@@ -64,6 +64,12 @@ public class NativeObject extends IdScriptableObject implements Map {
                 ConstructorId_getOwnPropertyDescriptor,
                 "getOwnPropertyDescriptor",
                 2);
+        addIdFunctionProperty(
+                ctor,
+                OBJECT_TAG,
+                ConstructorId_getOwnPropertyDescriptors,
+                "getOwnPropertyDescriptors",
+                1);
         addIdFunctionProperty(ctor, OBJECT_TAG, ConstructorId_defineProperty, "defineProperty", 3);
         addIdFunctionProperty(ctor, OBJECT_TAG, ConstructorId_isExtensible, "isExtensible", 1);
         addIdFunctionProperty(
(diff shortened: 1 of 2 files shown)


src/org/mozilla/javascript/NativeObject.java
```diff
@@ -507,6 +513,30 @@ public Object execIdCall(
                     Scriptable desc = obj.getOwnPropertyDescriptor(cx, nameArg);
                     return desc == null ? Undefined.instance : desc;
                 }
+            case ConstructorId_getOwnPropertyDescriptors:
```
p-bakker, line 516: EcmaScript spec says the getOwnPropertyDescriptors impl should use the [CreateDataPropertyOrThrow](https://262.ecma-international.org/12.0/#sec-createdatapropertyorthrow) Abstract Object operation

We don't have such method yet in AbstractEcmaObjectOperations yet. Would it make sense to refactor things (assuming that the logic of that Abstract Operation is already covered elsewhere) into the appropriate method in AbstractEcmaObjectOperations?

tuchida: Perhaps `CreateDataPropertyOrThrow` looks like the same behavior as `defineOwnProperty(cx, id, desc, checkValid = true)`.
https://github.com/mozilla/rhino/blob/c275c35f160ef7b035002c67c3540ba317964351/src/org/mozilla/javascript/ScriptableObject.java#L1569-L1570

I could use `defineOwnProperty` here, but I couldn't think of a test case. And, I think we will need `FromPropertyDescriptor` too. If any test case is violated by this PR, I will be fixed it.

src/org/mozilla/javascript/NativeObject.java
```diff
@@ -507,6 +513,30 @@ public Object execIdCall(
...
+                {
+                    Object arg = args.length < 1 ? Undefined.instance : args[0];
+                    // TODO(norris): There's a deeper issue here if
```
gbrail, line 519: A super minor thing but I don't think that we should add a new TODO comment when copying code, especially one that I bet is 10+ years old and assigned to someone who hasn't worked on this project for a very long time ;-)

tuchida: ref. https://github.com/mozilla/rhino/pull/1193/commits/2b80a8ab7d6a89039c464367e87b057cc99c3109 I removed. I will fixup in a few days.

gbrail:
This looks good -- thanks!

PR #1195, opened 2022-03-20
add test 

rbri:
that shows the fix https://github.com/mozilla/rhino/pull/1186 is not correct

this is related to #780


rbri:
#1294 is merged, we can close this

PR #1207, opened 2022-04-12
Various fixed for NativeConsole

rbri:
Have started my journey to switch from the HtmlUnit Console implementation to the new Rhino one. This fixes some cases found by the test suite and checked with real browsers.


rbri:
Because this is a new feature i think it is save to merge - will do more testing and maybe fixing with HtmlUnit

PR #1209, opened 2022-04-14
two json stringify fixes

rbri:


src/org/mozilla/javascript/NativeJSON.java
```diff
@@ -294,10 +294,17 @@ private static Object str(Object key, Scriptable holder, StringifyState state) {
...
         if (value instanceof Scriptable && hasProperty((Scriptable) value, "toJSON")) {
             Object toJSON = getProperty((Scriptable) value, "toJSON");
             if (toJSON instanceof Callable) {
-                value = callMethod(state.cx, (Scriptable) value, "toJSON", new Object[] {key});
+                value =
+                        callMethod(
+                                state.cx,
+                                (Scriptable) value,
+                                "toJSON",
+                                new Object[] {ScriptRuntime.toString(key)});
```
p-bakker, line 307: Perfromance-wise, would it make sense to do `key instanceof String ? key :  Integer.toString(key)` here, instead of  calling `ScriptRuntime.toString(key)`

src/org/mozilla/javascript/NativeJSON.java
```diff
@@ -294,10 +294,17 @@ private static Object str(Object key, Scriptable holder, StringifyState state) {
             value = getProperty(holder, ((Number) key).intValue());
         }
 
+        if (ScriptRuntime.isSymbol(value)) return Undefined.instance;
```
p-bakker, line 297: Is this correct, from an EcmaScript pov, that this is here, before the toJSON call or the call to the replacer (if provided)?

src/org/mozilla/javascript/NativeJSON.java
rbri: i think this is more or less what toString does already

p-bakker: It does, just wondering if my example code would be slightly more efficient, as `key` can only be of type string or int (judging the code above)

rbri: ahh, great, will update this

src/org/mozilla/javascript/NativeJSON.java
rbri: @p-bakker you are right here - update is on the way. Many thanks for the review.

src/org/mozilla/javascript/NativeJSON.java
```diff
@@ -288,23 +288,42 @@ private static Object str(Object key, Scriptable holder, StringifyState state) {
         Object value = null;
         Object unwrappedJavaValue = null;
 
+        //
```
p-bakker, line 291: ?

src/org/mozilla/javascript/NativeJSON.java
```diff
@@ -288,23 +288,42 @@ private static Object str(Object key, Scriptable holder, StringifyState state) {
...
+                value =
+                        callMethod(
+                                state.cx,
+                                (Scriptable) value,
+                                "toJSON",
+                                new Object[] {
+                                    keyString == null ? Integer.toString(keyInt) : keyString
+                                });
             }
         } else if (value instanceof BigInteger) {
```
p-bakker, line 314: Although you didnt change this code in this PR, I think the whole else/if can be removed (and thus deduplicating code), by putting the following in front of the `if (value instanceof Scriptable && ....`:
```
if (value instanceof BigInteger) {
   value = ScriptRuntime.toObject(state.cx, state.scope, value);
}
```

```

src/org/mozilla/javascript/NativeJSON.java
rbri: :-D

src/org/mozilla/javascript/NativeJSON.java
rbri: You are wrong here, the stuble difference is not to change value in case of bigint but no toJSON. Otherwise the replacer will be called with a NativeBigInt instead of a java BigInt (and this leads to failing tests).
Will have to check why NativeBigint does not overwrite getTypeof() - but this is a different story.

p-bakker: Right

p-bakker:
PR looks good to me

PR #1211, opened 2022-04-26
Fix typeof for native classes with shared global scope #1173

### LINKED ISSUE #1173: Rhino 1.7.14 breaks typeof for Java Classes
typeof on classes of imported packages is not working anymore.

```
importPackage(Packages.java.io);
importPackage(Packages.java.lang);

System.out.println("File: " + (typeof File));
```

outputs "File: function" till 1.7.13 and "File: undefined" in 1.7.14

Schmidor:
Closes #1173

### DIFF
--- a/src/org/mozilla/javascript/ScriptRuntime.java
+++ b/src/org/mozilla/javascript/ScriptRuntime.java
@@ -2869,6 +2869,9 @@ public class ScriptRuntime {
     /** The typeof operator that correctly handles the undefined case */
     public static String typeofName(Scriptable scope, String id) {
         Context cx = Context.getContext();
+        if (scope.getPrototype() instanceof ImporterTopLevel) {
+            scope.getPrototype().get(id, scope);
+        }
         Scriptable val = bind(cx, scope, id);
         if (val == null) return "undefined";
         return typeof(getObjectProp(val, id, cx));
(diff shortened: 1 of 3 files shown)


gbrail:
This fix seems right to me. Anyone else have an opinion?

rbri:
+ 1

p-bakker:
Does it make sense that this change is only applied to the `.getBase(Scriptable, String)` variant and not to the `Scriptable , int/Symbol` variants?

Schmidor:
Good point, I'm updating the other variants and checking

p-bakker:
Just curious: your PR changes what gets send into the `.has(key, start)` method for the start parameter. 

However, quickly looking I don't see any `Scriptable.has(key, start)` impl. in the Rhino codebase that actually uses the `start` parameter.

Any thoughts why the changes in your PR actually fix anything?

Schmidor:
For the perceived bug a change in  `getBase(Scriptable start, String name) ` would be enough, as it was in the previous commit.
There it is the ImporterTopLevel implementation of `has(String name, Scriptable start)` which uses the start-Parameter within ` getPackageProperty`.

But as you mentioned the other getBase variants, it is probably better to correct them to also to use `start` as parameter instead of the current obj, as it is documented in the `ScriptableObject.has` methods. Even it is currently not used.

p-bakker:
> There it is the ImporterTopLevel implementation of has(String name, Scriptable start)

Ah, missed that one

> But as you mentioned the other getBase variants, it is probably better to correct them to also to use start as parameter

Yeah, wasn't suggesting the fix was wrong in any way, was just wondering why the changes would fix anything, cause it looked to me that the start param was never used, but as you pointed out it is. Passing the start object (as you do now) instead of the current object is indeed the correct way, according to EcmaScript

So now, with the other variants adjusted as well, this PR looks good to me

gbrail:
This looks good to me too -- just testing for a quick merge. Thanks!

rPraml:
Looks good to me also

PR #1213, opened 2022-05-01
improve log output of Callable's

rbri:
NativeJSON.stringify outputs Callable's as null, convert to string to make the log output less confusing

### DIFF
--- a/src/org/mozilla/javascript/NativeConsole.java
+++ b/src/org/mozilla/javascript/NativeConsole.java
@@ -325,7 +325,24 @@ public class NativeConsole extends IdScriptableObject {
         }
 
         try {
-            Object stringify = NativeJSON.stringify(cx, scope, arg, null, null);
+            // NativeJSON.stringify outputs Callable's as null, convert to string
+            // to make the output less confusing
+            final Callable replacer =
+                    new Callable() {
+                        @Override
+                        public Object call(
+                                Context callCx,
+                                Scriptable callScope,
+                                Scriptable callThisObj,
+                                Object[] callArgs) {
+                            final Object value = callArgs[1];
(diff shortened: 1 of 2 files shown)


p-bakker:
You sure it should automatically stringify the function including the body? 

Not at my laptop, but I think Chrome just logs [Function] (for JavaScript functions, and [native function] for non-js functions. And if you want to log the entire function incl. the body, you have to call {{function}}.toString() yourself.

Not sure how your PR handles non-JS functions

Logging the entire function source might even be a bit of a security issue, depending on the context.

rbri:
At least better than logging null :-D
But a good point, will have a look at your suggested improvements.

src/org/mozilla/javascript/NativeConsole.java
```diff
@@ -325,7 +325,34 @@ private static String formatObj(Context cx, Scriptable scope, Object arg) {
...
+                                Object[] callArgs) {
+                            Object value = callArgs[1];
+                            while (value instanceof Delegator) {
+                                value = ((Delegator) value).getDelegee();
+                            }
+                            if (value instanceof BaseFunction) {
+                                StringBuilder sb = new StringBuilder();
+                                sb.append("function ")
+                                        .append(((BaseFunction) value).getFunctionName())
+                                        .append("() {}");
```
p-bakker, line 346: Wondering whether this should not be `"() {...}"` to make it clearer that the body content was stripped

p-bakker:
LGTM

rbri:
@p-bakker many thanks for all the reviews

gbrail:
This looks good to me. I'll work on merging it soon.

PR #1226, opened 2022-05-20
fix scope for defineOwnProperties

rbri:


src/org/mozilla/javascript/AbstractEcmaObjectOperations.java
```diff
@@ -54,19 +54,21 @@ static boolean hasOwnProperty(Context cx, Object o, Object property) {
...
      * @param cx
+     * @param scope the current scope.
      * @param o
      * @param level
      * @return boolean
      * @see <a
      *     href="https://262.ecma-international.org/11.0/#sec-testintegritylevel">TestIntegrityLevel</a>
      */
-    static boolean testIntegrityLevel(Context cx, Object o, INTEGRITY_LEVEL level) {
+    static boolean testIntegrityLevel(
```
p-bakker, line 64: Mmm, the idea of all the methods in this class is that they implement the abstract object operations as defined by EcmaScript (with the addition of Context as first param): testIntegretyLevel in the EcmaScript specification doesn't have a scope parameter

rbri: From my point of view scope is like context, we nee to add this for the params of many methods.

p-bakker: I think the (top level) scope should be derivable from the ScriptableObject.

I think the issue here is that the prototype of the function is a NativeObject instance on which the parentScope isn;t set accordingly, hence it being null in the ... method.

Adding `obj.setParentScope(this);` after the construction of the NativeObject instance in `BaseFunction.setupDefaultPrototype()` solves that. Then in `ScriptableObject.getOwnPropertyDescriptor(...)` the getParentSCope call starts working properly

Did a quick test with that and it seems to solve the problem

rbri: Sounds reasonable, will have a look and maybe prepare another PR,

rbri:
Based on @p-bakker  suggestion i made another one - #1227.

Issue #1077, opened 2021-10-24
Implement missing infrastructure required by Test262 test suite 


p-bakker:
See [Host defined Functions](https://github.com/tc39/test262/blob/main/INTERPRETING.md#host-defined-functions)

Not having this infrastucture means test that depend on it fail regardless of whether Rhino (correctly) supports the feature/behavior being tested 

For example: the Test262 test suite contains tests that test stuff cross realms. ATM those tests are disabled, I guess because they just fail because of this missing functionality, 'cause Rhino does (at least in basis) supports EcmaScript Realms I think. As far as I can determine, calling Context.initSafeObjects() gives you a new top level scope, which for all intents and purposes is the equivalent of what the specs refer to as a [Realm](https://tc39.es/ecma262/#realm)

As EcmaScript defines Realms, but doesn't define how they are created (because it depends on where JavaScript is being implemented), the Test262 test suite requires this additional infrastructure to get new realms without tying the tests to specific environments

So this case is about implemeting this additional infra structure in [Test262SuiteTest.java](https://github.com/mozilla/rhino/blob/master/testsrc/org/mozilla/javascript/tests/Test262SuiteTest.java)

In the process of implementing this, the exclusion of all tests tagged with `cross-realm` can be removed, see https://github.com/mozilla/rhino/blob/master/testsrc/org/mozilla/javascript/tests/Test262SuiteTest.java#L105

Note that this extra infra is not used just for cross-realm, so other tests might also start passing once implemented

Blocks #1078, #943

PR #1227, opened 2022-05-21
fix missing scope definition at some places

### LINKED ISSUE #934: Implement ES2017 Object.getOwnPropertyDescriptors

rbri:
next try to fix #934 (this replaces #1226)

### DIFF
--- a/src/org/mozilla/javascript/ArrowFunction.java
+++ b/src/org/mozilla/javascript/ArrowFunction.java
@@ -23,6 +23,7 @@ public class ArrowFunction extends BaseFunction {
 
         Function thrower = ScriptRuntime.typeErrorThrower(cx);
         NativeObject throwing = new NativeObject();
+        ScriptRuntime.setBuiltinProtoAndParent(throwing, scope, TopLevel.Builtins.Object);
         throwing.put("get", throwing, thrower);
         throwing.put("set", throwing, thrower);
         throwing.put("enumerable", throwing, Boolean.FALSE);
--- a/src/org/mozilla/javascript/BaseFunction.java
+++ b/src/org/mozilla/javascript/BaseFunction.java
@@ -509,7 +509,8 @@ public class BaseFunction extends IdScriptableObject implements Function {
             return prototypeProperty;
         }
         NativeObject obj = new NativeObject();
-        obj.defineProperty("constructor", this, DONTENUM);
+        obj.setParentScope(getParentScope());
+
         // put the prototype property into the object now, then in the
         // wacky case of a user defining a function Object(), we don't
         // get an infinite loop trying to find the prototype.
@@ -519,6 +520,8 @@ public class BaseFunction extends IdScriptableObject implements Function {
             // not the one we just made, it must remain grounded
             obj.setPrototype(proto);
         }
+
+        obj.defineProperty("constructor", this, DONTENUM);
         return obj;
     }
 
--- a/src/org/mozilla/javascript/BoundFunction.java
+++ b/src/org/mozilla/javascript/BoundFunction.java
@@ -39,6 +39,7 @@ public class BoundFunction extends BaseFunction {
 
         Function thrower = ScriptRuntime.typeErrorThrower(cx);
         NativeObject throwing = new NativeObject();
+        ScriptRuntime.setBuiltinProtoAndParent(throwing, scope, TopLevel.Builtins.Object);
         throwing.put("get", throwing, thrower);
         throwing.put("set", throwing, thrower);
         throwing.put("enumerable", throwing, Boolean.FALSE);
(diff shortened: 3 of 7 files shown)


p-bakker:
LGTM

tuchida:
Thanks! Looking good for now.

However, I felt that using `parentScope` here was not a good idea.
```js
var f = function() {};
f.prototype.__parent__; // null before fix, [object global] after
f.prototype.__parent__ = null;
Object.getOwnPropertyDescriptor(f.prototype, 'constructor').hasOwnProperty('value'); // TypeError
```
This Pull Request is correct and necessary, but it may be better not to use `parentScope` apart from that. Since `scope` is used to get `TopLevel.Builtins.Object`, it looks to me like `cx.topCallScope` is sufficient.

rbri:
>  Since scope is used to get TopLevel.Builtins.Object, it looks to me like cx.topCallScope is sufficient.

@tuchida Not sure i got your point - do you talk about 

    obj.setParentScope(getParentScope()); 

inside setupDefaultPrototype()?

tuchida:
This approach also seems to solve the problem.
```diff
diff --git a/src/org/mozilla/javascript/ScriptableObject.java b/src/org/mozilla/javascript/ScriptableObject.java
index 46e5aaeb..f1f45a24 100644
--- a/src/org/mozilla/javascript/ScriptableObject.java
+++ b/src/org/mozilla/javascript/ScriptableObject.java
@@ -2651,8 +2651,7 @@ public abstract class ScriptableObject
     protected ScriptableObject getOwnPropertyDescriptor(Context cx, Object id) {
         Slot slot = querySlot(cx, id);
         if (slot == null) return null;
-        Scriptable scope = getParentScope();
-        return slot.getPropertyDescriptor(cx, (scope == null ? this : scope));
+        return slot.getPropertyDescriptor(cx, cx.topCallScope);
     }
 
     protected Slot querySlot(Context cx, Object id) {
```

rbri:
@tuchida ok, is see, will test and add a commit

p-bakker:
If we're deriving the scope from the context, there's not much point in passing it into slot getPropertyas we're also passing in context, no?

But are we 100% sure that it is always correct to read the top level scope from the context and not derive it from the Scriptable?

rbri:
> If we're deriving the scope from the context, there's not much point in passing it into slot getPropertyas we're also passing in context, no?

Have seen this but i was not sure if we should stay with the signature for backward compatibility.

> But are we 100% sure that it is always correct to read the top level scope from the context and not derive it from the Scriptable?

Had the same question in mind but as long we have no test case that proves this i think both solutions are correct - so i did the commit because the code is simpler (from my point of view).

p-bakker:
> Had the same question in mind but as long we have no test case that proves this i think both solutions are correct - so I did the commit because the code is simpler (from my point of view).

The testcases might be in test262, but not enabled/failing as we're missing part of the required testing infrastructure, see https://github.com/mozilla/rhino/issues/1077

In the browser I can test the cross-realm scripting using iframes:
```
// Comparing the __proto__ of a property descriptor of an object from another realm with one from the current realm.
// Yields false, as the the frame ('myframe') has it's own instances of all the global JavaScript objects, like `Object`
let realm = window.myframe;
realm.Object.getOwnPropertyDescriptor(realm.Object, 'getOwnPropertyDescriptor').__proto__ === ({}).__proto__;

realm = window;
// Same check as before, but now all within the same realm, yields true
realm.Object.getOwnPropertyDescriptor(realm.Object, 'getOwnPropertyDescriptor').__proto__ === ({}).__proto__;
```
The above 'testcase' indicates (I think) that in EcmaScript the lookup of the instance of Object to use for creating for example property descriptors should be based not on the Object instance of the current realm, but on scope in which the object is created on which the getOwnPropertyDescriptor method is called.

The following code with Rhino behaves differently when using cx.topCallScope vs. using the scriptable as scope:
```
      Context ctx = Context.enter();
      ctx.setLanguageVersion(200);
      Scriptable scriptable = ctx.initStandardObjects();
      Scriptable realm = ctx.initStandardObjects();
      scriptable.put("realm", scriptable, realm);

      String expression = """
        realm.Object.getOwnPropertyDescriptor(realm.Object, 'getOwnPropertyDescriptor').__proto__ === ({}).__proto__;
      """;
      Object result = ctx.evaluateString(scriptable, expression, "<cmd>", 0, null);

      System.out.println(result);
```

p-bakker:
Created a (draft) PR for adding the missing infr, see #1229

p-bakker:
The codechange that @tuchida proposed in https://github.com/mozilla/rhino/pull/1227#issuecomment-1133900625 (using `return slot.getPropertyDescriptor(cx, cx.topCallScope);` wasn't the correct one, but I think the code in that area can be simplified: the getParentScope() call can be removed and instead just call `return slot.getPropertyDescriptor(cx, this);`

gbrail (approved):
LGTM for me as well. Thanks for tracking this down!

Issue #1079, opened 2021-10-26
Fix symbol impl to have the GlobalSymbolRegistry shared between Realms


p-bakker:
Current implementation keeps the `GlobalSymbolRegistry` in the topLevel scope.

However, EcmaScript defines that the GlobalSymbolRegistry needs to be shared by all Realms, which basically translates to: each topLevel scope must share the same GlobalSymbolRegistry

Note: need to see how such a global registery would work in combination with shared sealed scopes

~Depends on #1077 (for enabled the cross-realm tests)~

PR #1229, opened 2022-05-23
Make the $262 object available within the test262 tests

### LINKED ISSUE #1077: Implement missing infrastructure required by Test262 test suite

p-bakker:
Closes #1077

PR not yet complete: needs at least the test262.properties file regenerated

### DIFF
--- a/testsrc/org/mozilla/javascript/tests/Test262SuiteTest.java
+++ b/testsrc/org/mozilla/javascript/tests/Test262SuiteTest.java
@@ -44,6 +44,9 @@ import org.mozilla.javascript.EvaluatorException;
 import org.mozilla.javascript.RhinoException;
 import org.mozilla.javascript.Script;
 import org.mozilla.javascript.Scriptable;
+import org.mozilla.javascript.ScriptableObject;
+import org.mozilla.javascript.annotations.JSFunction;
+import org.mozilla.javascript.annotations.JSGetter;
 import org.mozilla.javascript.drivers.TestUtils;
 import org.mozilla.javascript.tools.SourceReader;
 import org.mozilla.javascript.tools.shell.ShellContextFactory;
@@ -102,7 +105,6 @@ public class Test262SuiteTest {
                             "class-fields-private",
                             "class-fields-public",
                             "computed-property-names",
-                            "cross-realm",
                             "default-arg",
                             "default-parameters",
                             "new.target",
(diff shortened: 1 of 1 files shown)


p-bakker:
Added this PR quickly as it might come in handly for #1227

testsrc/org/mozilla/javascript/tests/Test262SuiteTest.java
```diff
@@ -436,6 +438,50 @@ public static void tearDownClass() {
...
+    /** @see https://github.com/tc39/test262/blob/main/INTERPRETING.md#host-defined-functions */
+    private class $262 {
+        private ScriptableObject scope;
+
+        public $262(ScriptableObject scope) {
+            this.scope = scope;
+
+            scope.put("$262", scope, this);
+            scope.setAttributes("$262", ScriptableObject.DONTENUM);
+        };
```
rbri, line 450: ;

testsrc/org/mozilla/javascript/tests/Test262SuiteTest.java
```diff
@@ -436,6 +438,50 @@ public static void tearDownClass() {
...
+        public void detachArrayBuffer() {
+            throw new UnsupportedOperationException(
+                    "$262.detachArrayBuffer() method not yet implemented");
+        }
+
+        @JSGetter
+        public Object getAgent() {
+            throw new UnsupportedOperationException("$262.agent property not yet implemented");
+        }
+    };
```
rbri, line 483: ;

rbri: Did some tests but there is no effect, maybe we need more realm support

p-bakker: Lookin'... This is code I wrote quite a while ago and at the time adding the $262 stuff made some tests pass, but now they dont...

p-bakker: Pushed a fix for the $262 object not having much of a difference

Also updated the test262.properties now

rbri (commented):
LGTM

testsrc/org/mozilla/javascript/tests/Test262SuiteTest.java
```diff
@@ -436,6 +438,52 @@ public static void tearDownClass() {
...
+    public class $262 {
+        private ScriptableObject scope;
+
+        public $262() {}
+
+        public $262(ScriptableObject scope) {
+            this.scope = scope;
+
+            scope.put("$262", scope, this);
+            scope.setAttributes("$262", ScriptableObject.DONTENUM);
```
rbri, line 451: I think this should be done in the setup and mot in the ctor.

p-bakker: With `setup` you mean into the buildScope method?

That would entail duplicating the code between there and the $262.createRealm method, which i don't really like either

rbri: Ok, is see, my problem was as i saw this line

    new $262(scope);

i was really confused because i did not expect that this ctor call does some kind of registration in the scope. So from the first look this line looks really useless.

Maybe a factory method that has a hint in the name about the registration?

rbri:
Any idea why there is no real effect?

rbri:
Btw. i think this is again great step forward for Rhino - THANKS.

p-bakker:
There is an effect now no? Some previously failing tests are now passing.

I guess you mean why aren't more of the tests that depended on cross-realm passing now? Can't recall, but when I looked into it a while ago, they were failing due to other reasons. I filled for example #1079, because that is an EcmaScript incompatibility when using Symbols cross-realm

testsrc/org/mozilla/javascript/tests/Test262SuiteTest.java
p-bakker: How about now?

p-bakker:
This PR also needs a squash merge....

rbri (approved):
LGTM

rbri:
Yes I can :-)

Issue #1232, opened 2022-05-28
Encoding error with local .js files after openJDK 17.0.3


midgleyc:
In [https://bugs.openjdk.java.net/browse/JDK-8273655](https://bugs.openjdk.java.net/browse/JDK-8273655), a change was made so that ".js" files now are seen as having content type "text/javascript".

This leads to a change in the [UrlModuleSourceProvider](https://github.com/mozilla/rhino/blob/master/src/org/mozilla/javascript/commonjs/module/provider/UrlModuleSourceProvider.java#L171) where local ".js" files (or any ".js" files without a Content-Type encoding) are interpreted as having a character encoding of "8859_1" while previously they were interpreted as having a character encoding of "utf-8".

This has lead to a problem with KoLMafia's ability to parse javascript scripts, as discussed in [this thread](https://kolmafia.us/threads/java-bug-on-some-versions-of-java-weird-things-happen-to-strings-with-special-characters.27531/#post-168112).

I've put together a repo at [java-probe-content-type](https://github.com/midgleyc/java-probe-content-type) which shows the difference for local files.

I don't know what a good fix to this would look like (as assuming files without an explicit encoding are cp-1252 is as the spec says), but it might be a quicker fix if `getCharacterEncoding` were `protected` instead of `private static`. That way we could override it on our end to expect `utf-8` instead of `8859_1` without affecting other uses of this library.

p-bakker:
ONe way to solve this locally without any code changes is overiding the content-types.properties file used using System.setProperty("content.types.user.table","<path-to-file>");

midgleyc:
Filed #1233 to make `getCharacterEncoding` protected. Thinking more I think Rhino can fix the problem on its end: we only have the issue for local files, and you can use `url.getProtocol()` to determine whether the file is local or not (if local, the protocol is `file`). The `text/` adjustment should likely only be done if the file comes over the HTTP protocol, so 

```java
if (contentType != null && contentType.startsWith("text/")) {
    return "8859_1";
}
```

could be replaced with

```
String protocol = url.getProtocol();
boolean isHttp = protocol != null && (protocol.equals("http") || protocol.equals("https"));
if (isHttp && contentType != null && contentType.startsWith("text/")) {
    return "8859_1";
}
```

Does this seem reasonable?

p-bakker:
I haven't found any documentation that local files in CommonJS ought to be stored in utf-8. However, it seems to me that NodeJS expects this.

[...]

PR #1233, opened 2022-05-29
Make getCharacterEncoding in UrlModuleSourceProvider protected

midgleyc:
Allows subclasses to override the encoding used in case a local
javascript file is interpreted as being latin-1, when it should be
whatever the file is encoded with. Before openJDK 17.0.3 this was
automatically interpreted as utf-8.

Allows #1232 to be worked around by code calling Rhino.


gbrail:
Sorry we are so behind on these -- thanks for fixing this and it looks straighforward.

Could you please add a test case to ensure that future changes don't break this?

midgleyc:
I could add a test, but I've only changed an access modifier. Do you want a test that uses reflection to confirm the access modifier is protected or public? A test that creates a subclass and confirms that the method can be overridden?

gbrail:
I think that test is fine. Thanks!

PR #1234, opened 2022-06-01
Implement Map.put() for NativeObject

inad9300:
Untested code – could something as simple as this work? It sucks to see `implements Map` while not being able to rely on super basic `Map` methods.

### DIFF
--- a/src/org/mozilla/javascript/NativeObject.java
+++ b/src/org/mozilla/javascript/NativeObject.java
@@ -778,7 +778,13 @@ public class NativeObject extends IdScriptableObject implements Map {
 
     @Override
     public Object put(Object key, Object value) {
-        throw new UnsupportedOperationException();
+        Object previousValue = get(key);
+        if (key instanceof String) {
+            super.put((String) key, this, value);
+        } else if (key instanceof Number) {
+            super.put(((Number) key).intValue(), this, value);
+        }
+        return previousValue == NOT_FOUND ? null : previousValue;
     }
 
     @Override


p-bakker:
@rPraml given that you've done a lot of work in this area: any thoughts on this?

p-bakker:
See the notes on the original implementation of the Map interface: https://bugzilla.mozilla.org/show_bug.cgi?id=448816

p-bakker:
Besides the wrapping mentioned in the link in the previous comment, your PR doesn't always perform the put (only when the key is a String or Number), which is odd/unexpected/buggy/...

Additionally, all JavaScript logic when it comes to puts on Objects is bypassed, like checking whether the object is sealed/frozen, whether the property is writable etc.

So as is, your PR would break the contract of how puts ought to behave on JavaScript Objects, so I don't think it supporting Map.put is as straight-forward as you'd hoped.

I think the aforementioned issues could be solved, java.util.Map.put is allowed to throw certain exceptions, which could be mapped onto the scenario's where JavaScript is not allowing a put.

Having said all that, even if you'd fix all the aforementioned issues, its not guaranteed your PR will get excepted, as that's not up to me :-)

p-bakker:
JavaScript Objects are exposed to Java as TruffleMap in Graal(JS) apparently and TruffleMap implement java.util.Map is seems, so maybe for inspiration see how GraalJS handles puts on JavaScript Objects through the Map interface form Java

p-bakker:
> For better or worse, there isn't much future in Rhino, is my impression

Rhino is maintained by volunteers, so progress is dependant on who has time available to work on things and therefor things aren't moving as fast as one would hope/like

That said, there's a massive active userbase, so Rhino isn't going away any time soon.

tonygermano:
I believe it should be calling `ScriptRuntime.setObjectProp`, which requires a context. I'm fairly certain that the GraalVM java objects that represent javascript objects also do not work once the context has been closed.

Issue #1238, opened 2022-06-17
Parent-child relationship in TaggedTemplateLiteral


kuzjka:
Hi,

Our team is really happy to see string templates and tagged string templates support in Rhino 1.7.14.

However, I have noticed that descendants of `TaggedTemplateLiteral` node, which are _tag_ (target) and _string literal_, don't have parent relationship with `TaggedTemplateLiteral` node.

It looks a little illogical to me, and among other things, breaks `getAstRoot()`.

If parent is not set intentionally, please let me know.

I'll try to prepare a PR with tests and a fix soon.

Best regards,
Ihor.

rbri:
LGTM

PR #1239, opened 2022-06-17
fix: parent relationship in TaggedTemplateLiteral (#1238)

kuzjka:
This PR adds parent references to target and string template nodes of `TaggedTemplateLiteral` node (see #1238)

Since string templates are introduced in ES6, I placed tests into _org.mozilla.javascript.tests.es6_ package. Please let me know if there is a more suitable place for these tests.


rbri:
Many thanks for this pr. Please do a
gradlew spotlessApply
to fix the code style.

gbrail:
This looks good. Sorry about the long delay!

PR #1240, opened 2022-06-25
Add zh_CN messages resource

ylk2534246654:
Add zh_CN messages resource

### DIFF
--- a/src/org/mozilla/javascript/resources/Messages_zh_CN.properties
+++ b/src/org/mozilla/javascript/resources/Messages_zh_CN.properties
@@ -273,7 +273,7 @@ msg.fn.redecl =\
     类型错误：重新声明了函数 {0}。
 
 msg.let.decl.not.in.block =\
-    语法错误 let 语句不能直接在块中
+    语法错误：let 语句不能直接在块中
 
 msg.bad.object.init =\
     语法错误：无效的对象初始化程序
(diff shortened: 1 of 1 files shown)


rbri:
Thanks for the contribution.
I think we have some unit tests for the error message translation. Maybe you can enhance some tests.

gbrail:
This is awesome work and it'll be great to have. Before merge, do you think that we could have:

1) A basic unit test that verifies that, when the locale is set, messages at least come out from the right catalog. "MessageProviderLanguageTest.java" does this -- we just need to update it for Chinese.
2) I'd like one other Chinese speaker to verify that these messages are indeed reasonable Chinese translations and are OK to add in the source code.

I can do #2 if there is someone in the community who can do the first!

p-bakker:
@gbrail I've pasted the content of the modified file with the Chinese translations into Google Translate and the translations seem to hold up just fine

p-bakker:
@gbrail as mentioned above, I've validated the translations by pulling them through Google Translate and they seem ok, so I think that's proof enough that they are good, sorta resolving point 2 from your comment, agree?

For the sake of cleaning up or PR backlog, could this then be merged?

gbrail:
Agreed! I resolved the conflict and merged this manually. Thanks to @ylk2534246654 for the work on this long ago!

PR #1245, opened 2022-07-14
Update NativeRegExp.java

SuperMonster003:

PR #1260, opened 2022-09-16
fix the error message in case the quantifier maximum (second value) is smaller than the minimum (first value)

rbri:
This supersedes #1245


rbri:
Have tested this with

real Chrome - Error: Invalid regular expression: /b{2,1}/: numbers out of order in {} quantifier

real FF - Error: numbers out of order in {} quantifier.

Hope my version of the error msg is at least as good a the two.

They french version was created using DeepL.

PR #1266, opened 2022-10-17
Update README.md

wimjongman:
* Reversed the release table so that the newest release is on top 
* Made the current release more explicit
* Wrapped the release tables in a details block


gbrail:
Thanks for making it look better!

Issue #1092, opened 2021-11-22
Switch to a multiproject/modular source layout


carlosame:
As a step towards fully modularising Rhino (see https://github.com/mozilla/rhino/issues/1075#issuecomment-950907384), the source layout should be changed to a Gradle multiproject layout. I'd suggest adhering to the [Maven/Gradle layout conventions](https://docs.gradle.org/current/userguide/organizing_gradle_projects.html) that many projects follow, and the layout would then be something like:
```
org.mozilla.rhino/
                  src/
                      main/
                           java/
                                org/
                                    mozilla/
                           resources/
                                     org/
                                         mozilla/
                      test/
                           java/
                                org/
                                    mozilla/
                           resources/
                                     org/
                                         mozilla/
                      lc/
                           java/
                                com/
                                    netscape/
                                             javascript/
org.mozilla.rhino.engine/
                         src/
                             main/
                                  java/
                                       org/
                                           mozilla/
                                  resources/
                                            META-INF/
org.mozilla.rhino.runtime/
                          src/
                              main/
                                   java/
                                        org/
                                            mozilla/
                                   resources/
                                             org/
                                                 mozilla/
```
The "main" `rhino` jar would only contain the `org.mozilla.javascript.tools` package and would depend on the `rhino-runtime` jar, which is currently unusable in modular Java because it contains packages which are already present in the main `rhino` jar.

As to why I'm using full names like `org.mozilla.rhino.runtime` instead of just `runtime` or `rhino-runtime`: because the `javadoc` tool currently requires that convention, if one wants to produce a multi-module javadoc. BTW as an alternative one could use a smaller name, then rebuild the source tree elsewhere under `buildGradle` with the full names before building the Javadocs from that directory, it's just more complexity being added to the build.

The next step would be the modularisation itself, adding a `module-info.java` to each subproject (now module), also compiling classes that are specific to Java 11 and 18. There is more than one way to achieve that, and that could be discussed in #1075: one possibility would be to add two sourcesets:
```
org.mozilla.rhino.runtime/
                          src/
                              java11/
                                     java/
                                          org/
                                              mozilla/
                                     resources/
                                               org/
                                                   mozilla/
                              java18/
                                     java/
                                          org/
                                              mozilla/
                                     resources/
                                               org/
                                                   mozilla/
```
[...]

PR #1278, opened 2022-11-09
make rhino-engine an osgi bundle

stbischof:


p-bakker:
LGTM, but no osgi expert by far

It seems to bring `engineJar` more in line with `jar` and `runtimeJar`, but I wonder why `runtimeJar` would not also have ` "Automatic-Module-Name": "org.mozilla.rhino-runtime"`, similar to `engineJar` and `jar`

@carlosame any thoughts, since you're the one to last make changes in this area (in #873)?

build.gradle
```diff
@@ -201,6 +201,11 @@ task engineJar(type: Jar) {
                 "Implementation-Vendor": "Mozilla Foundation",
                 "Implementation-URL": "http://www.mozilla.org/rhino",
                 "Automatic-Module-Name": "org.mozilla.rhino.engine"
+                "Bundle-ManifestVersion": "2",
+                "Bundle-SymbolicName": "org.mozilla.rhino.engine",
+                "Bundle-Version": project.version.replaceAll("-.*", ""),
+                "Export-Package": "org.mozilla.javascript.engine",
+                "Import-Package": "javax.script,org.mozilla.javascript"
```
carlosame, line 208: I would not declare an `Import-Package`, and instead leave the default (automatic) behaviour.

carlosame:
> It seems to bring `engineJar` more in line with `jar` and `runtimeJar`, but I wonder why `runtimeJar` would not also have ` "Automatic-Module-Name": "org.mozilla.rhino-runtime"`, similar to `engineJar` and `jar`
> 
> @carlosame any thoughts, since you're the one to last make changes in this area (in #873)?

I commented on this in a few issues, like #1092. Essentially, `rhino-runtime` is modular-incompatible with the other jars, so it would be pointless to put a module name and thus invite people to use it in modular projects.

Quoting from one of my comments in the aforementioned issue (https://github.com/mozilla/rhino/issues/1092#issuecomment-995847859):

> "Split packages" problem with `rhino.jar` and `rhino-runtime.jar`. That's the reason why I did not add a module name to `rhino-runtime.jar`. The problem hasn't been observed out there because almost nobody uses `rhino-runtime.jar`, but if this project keeps telling people that "you should use `rhino-runtime.jar`" they will be hit (impact: unable to start the JVM). It depends on the current issue to be fixed.

build.gradle
stbischof: this is an OSGi import. If you do not declare it the bundle will not work/ throw an ClassNotFoundExcaption because bundle does not declare that is needs it.

carlosame: That's weird. I do not use OSGi myself, however from https://bnd.bndtools.org/heads/import_package.html:

> The `Import-Package` header lists the packages that are required by the contained packages. The default for this header is `*`, resulting in importing all referred packages. This header therefore rarely has to be specified.

stbischof: yep. bnd is a tool that calculates  die imports. if you say to bnd do `*` then it will habe as result in the manifest:

`Import-Package javax.script,org.mozilla.javascript`

@juergen-albert (OSGi Working Group Steering Committee)

could you please review this from the OSGi piunt of view?

juergen-albert: What @stbischof wrote is correct. bnd is a build tool, that can calculate the Imports for you with proper Version ranges. The same goes for the Export Packages. OSGi requires all the packages to be named separately and can not interpret the *.

bnd even has a dedicated gradle plugin that can automatically calculate a Manifest for your, not only putting in proper OSGi Headers, but can also automatically handle things in respect to Java Module System.

I'm not that firm with gradle, but you can find an example of the gradle plugin usage here: https://github.com/osgi/osgi-test/tree/main/examples/osgi-test-example-gradle

stbischof:
Hi we also should add the Capability.

```
Provide-Capability: osgi.service;objectClass:List<String>="javax.script.ScriptEngineFactory";effective:=active,osgi.serviceloader;osgi.serviceloader="javax.script.ScriptEngineFactory";register:="org.mozilla.javascript.engine.RhinoScriptEngineFactory"
Require-Capability: osgi.extender;filter:="(&(osgi.extender=osgi.serviceloader.registrar)(version>=1.0.0)(!(version>=2.0.0)))"
```

next Steps?

p-bakker:
Is been a while, but... In #1479 the entire codebase was reorganized into Java modules. 

The question arose whether that alleviates the need to make stuff an OSGI bundle. 

Curious what your thoughts on this are

stbischof:
I will look again here what is needed to update and rebase

p-bakker:
@stbischof how's the rebase/update coming along?

stbischof:
Could you ping me in tuesday again if i didnt it?

p-bakker:
I sure can!

stbischof:
Manifests are not written by hand anymore. they are to complex.

the tool [bnd](https://github.com/bndtools/bnd/tree/master/gradle-plugins) is doing the job.

The problem that i am facing is, that i cant write gradle.

this is ma try:
https://github.com/stbischof/rhino/commit/e3b6017cfa1174e76f1bd1260850634cfa9720ad

but it does not work.

Documentation: https://github.com/bndtools/bnd/tree/master/gradle-plugins

Maybe you can help me to fix the gradle build.

p-bakker:
Afraid my 'gradle' knowledge doesn't run that deep either

@gbrail given that you seem very fluent in 'gradle': got any pointers here maybe?

p-bakker:
Converted this to draft, as it still needs work

PR #1285, opened 2022-12-01
cleanup as follow up of pr 1284

rbri:
make setFunctionProtoAndParent context aware,
deprecate the old versions
cleanup promise because the setup is now correct done by the ctor


rbri:
because
* the tests are green
* no api changes done
* and HtmlUnit is waiting for this

i made the merge already....

@gbrail @p-bakker hope this is ok

PR #1286, opened 2022-12-02
Preserving cause on rethrown exceptions

rPraml:
We had an issue, that we loose the java stack trace, when we catch the error in javascript and rethrow it.

```javascript
try {
  // call java code
 someJavaObject.doSomething();
} catch (e) {
 // do some error handling and rethrow the error
 throw e;
}

```

In this case, you will get a JavaScriptException with only the message containing of the exception, that is thrown by `someJavaObject.doSomething();`
(So you might get a NPE, but you have no idea, which line of code caused the exception)

In this PR, we try to set the currently processing exception/NativeError and setting it as cause of the JavaScriptException.

Roland

### DIFF
--- a/src/org/mozilla/javascript/JavaScriptException.java
+++ b/src/org/mozilla/javascript/JavaScriptException.java
@@ -90,24 +90,18 @@ public class JavaScriptException extends RhinoException {
         }
     }
 
-    /**
-     * @return the value wrapped by this exception
-     */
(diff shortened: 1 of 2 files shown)


rbri:
LGTM - great improvement.

rbri:
@gbrail @p-bakker can i merge this?

rbri:
@rPraml thanks for this

PR #1287, opened 2022-12-03
fix ScriptException when bound functions are called inside Promise.then()

rbri:
see https://github.com/HtmlUnit/htmlunit/issues/522 for more


rbri:
@gbrail / @p-bakker any comments on this?

Issue #1051, opened 2021-10-13
Support ES2022 `.at()` on indexables (Array, String, etc)


p-bakker:
Proposal: https://github.com/tc39/proposal-relative-indexing-method
MDN: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/at

tuchida:
- ref. #1289

I look like this has already been implemented.

PR #1289, opened 2022-12-06
Es2022 at method

JohnBain:
Addresses issue #1051 , adding .at function to the indexed data types Array, TypedArray and String.


src/org/mozilla/javascript/NativeArray.java
```diff
@@ -1962,6 +1970,24 @@ private static Object js_copyWithin(
         return thisObj;
     }
 
+    private static Object js_at(Context cx, Scriptable scope, Scriptable thisObj, Object[] args) {
+        int k;
+        Scriptable o = ScriptRuntime.toObject(cx, scope, thisObj);
+        int len = (int) getLengthProperty(cx, o);
```
tuchida, line 1976: It seems bad to cast to int without inspecting length and index.
```js
var arrayLike = {
  0: 0,
  length: 0x7fffffff + 1  // Integer.MAX_VALUE + 1
};
console.log(Array.prototype.at.call(arrayLike, 0)); // expected 0, but actual undefined
```

JohnBain: Thanks for the feedback - I just redeployed and changed NativeArray js_at so that it no longer casts len. I also added argument checks to all functions. I added your example as a test and it passes.

Please let me know if it needs any additional changes.

src/org/mozilla/javascript/NativeArray.java
```diff
@@ -1962,6 +1970,19 @@ private static Object js_copyWithin(
...
+        Scriptable o = ScriptRuntime.toObject(cx, scope, thisObj);
+        Object targetArg = (args.length >= 1) ? args[0] : Undefined.instance;
+        long relativeIndex = (long) ScriptRuntime.toInteger(targetArg);
+        long len = getLengthProperty(cx, o);    
+        long k = (relativeIndex >= 0) ? relativeIndex : len + relativeIndex;
+        if ((k < 0) || (k >= len)) {
+            return Undefined.instance;
+        }
+
+        return ScriptableObject.getProperty(thisObj, (int) k);
```
tuchida, line 1983: Thank you for your fix.

I think it is better to use `getElem` or `getRawElem`.
https://github.com/mozilla/rhino/blob/c4c10f4f046fd72c9671636cb9a504f5f6b32953/src/org/mozilla/javascript/NativeArray.java#L990-L1001

And can you add a test for the ArrayLike?

p-bakker:
Some of the other NativeArray methods implementations check whether the instance being operated on is an instance of NativeArray and then whether the instance.denseOnly flag is set to true.

Not sure if such a branch in the code would be applicable for the .at() method as well. To my knowledge the .denseOnly flag is related to a performance optimize

src/org/mozilla/javascript/NativeArray.java
tuchida: With the current code, this is the result.
```js
var arrayLike = {
  0x80000000: 'Integer.MAX_VALUE + 1',
  length: 0x80000001  // Integer.MAX_VALUE + 2
};
console.log(Array.prototype.at.call(arrayLike, 0x80000000)); // expected 'Integer.MAX_VALUE + 1', but actual undefined
```

src/org/mozilla/javascript/typedarrays/NativeTypedArrayView.java
```diff
@@ -237,6 +238,19 @@ private void setRange(NativeArray a, int off) {
         }
     }
 
+    private static Object getElem(Context cx, Scriptable target, long index) {
```
tuchida, line 241: TypedArray cannot accept an ArrayLike, so it probably doesn't need `getElem` and `getRawElem`.
https://262.ecma-international.org/13.0/#sec-%typedarray%.prototype.at
> 2. 2. Perform ? [ValidateTypedArray](https://262.ecma-international.org/13.0/#sec-validatetypedarray)(O).

JohnBain: I pushed code removing the getElems functions and returning to getProperty. Please let me know what you think.

tuchida: I think it looks good. Thanks!

rbri (commented):
can you please move your tests to org.mozilla.javascript.tests.es2022

JohnBain:
> can you please move your tests to org.mozilla.javascript.tests.es2022

Done - https://github.com/mozilla/rhino/pull/1289/commits/8ea870b47b1d288767c9efe9036c88bba2dbd976

src/org/mozilla/javascript/NativeArray.java
```diff
@@ -1962,6 +1970,18 @@ private static Object js_copyWithin(
         return thisObj;
     }
 
+    private static Object js_at(Context cx, Scriptable scope, Scriptable thisObj, Object[] args) {
+        Scriptable o = ScriptRuntime.toObject(cx, scope, thisObj);
+        Object targetArg = (args.length >= 1) ? args[0] : Undefined.instance;
```
p-bakker, line 1975: ScriptRuntime has overloaded variants of toInteger that might be a better fit here, like `public static double toNumber(Object[] args, int index)`

src/org/mozilla/javascript/typedarrays/NativeTypedArrayView.java
```diff
@@ -254,6 +254,18 @@ private Object js_subarray(Context cx, Scriptable scope, int s, int e) {
                 new Object[] {arrayBuffer, Integer.valueOf(byteOff), Integer.valueOf(len)});
     }
 
+    private Object js_at(Context cx, Scriptable scope, Scriptable thisObj, Object[] args) {
+        Object targetArg = (args.length >= 1) ? args[0] : Undefined.instance;
+        int relativeIndex = (int) ScriptRuntime.toInteger(targetArg);
```
p-bakker, line 259: similar comment as the one above

gbrail:
This looks like a pretty complete and self-contained PR. We have two outstanding comments that could make the code a bit simpler (but I'm not actually 100% sure of that). Would you all be OK if we merged this now?

rbri:
@p-bakker will have a look at your comments in the next step....

rbri:
@p-bakker #1331 brings the code in sync with js_fill impl regarding parameter handling

rbri:
@JohnBain thanks for the pr

PR #1294, opened 2022-12-24
next try to fix issue #780

rbri:
test cases from https://github.com/mozilla/rhino/pull/1242
fix taken from p-bakker's sugestion


rbri:
Since this fixes a regression and the fix is simpler than the other proposed solutions i merge this one.
Thanks @p-bakker for this solution.

Issue #431, opened 2018-04-25
Expiry of cached commons-js script entity not updated if corresponding file resource not updated on filesystem


dhogan:
Line 281 of the UrlModuleSourceProvider currently reads:
_return lastModified == urlConnection.getLastModified();_
It should read
_return lastModified != urlConnection.getLastModified();_

This error prevents the expiry date of a cached, file-based commons-js module from being updated despite the absence of filesystem modifications. Thus the ModuleSourceProviderBase will not recognize that the module entity does not require revalidation (if block on line 37 of ModuleSourceProviderBase), resulting in filesystem resolution of the commons-js entity with every script reference.

travishaagen:
https://github.com/mozilla/rhino/pull/432

p-bakker:
Looks like this was closed by #432 a while back

Issue #198, opened 2015-06-15
Add ability to invalidate cache of require()


AlexTrotsenko:
1. `ModuleSourceProvider` has ability to return  [NOT_MODIFIED](https://github.com/mozilla/rhino/blob/Rhino1_7_6_RELEASE/src/org/mozilla/javascript/commonjs/module/provider/ModuleSourceProvider.java#L26) if source was not modified. 

2. `CachingModuleScriptProviderBase` has ability to obtain [cached ModuleScript](https://github.com/mozilla/rhino/blob/Rhino1_7_6_RELEASE/src/org/mozilla/javascript/commonjs/module/provider/CachingModuleScriptProviderBase.java#L73) if `NOT_MODIFIED` returned by `ModuleSourceProvider`. This ModuleScript that was already compiled previously.

The problem, that `Require` do not take into account `NOT_MODIFIED`, that is returned by `CachingModuleScriptProviderBase`

szegedi:
A single `Require` object will not reload modules, that is by design. Once a module is loaded in a single require instance, it's loaded, period. There's no way to reexecute it, and it would in fact not be correct to do so and come with a bunch of stability issues. (Imagine other modules depended on it, would they need to be reexecuted as well? In what order? If some code managed to hold on to old exports you'd have both old and new exports floating around, it'd be a nightmare to debug. There'd be a lot of problems with that approach.)

Using a single `ModuleSourceProvider` with multiple `Require` instances (one per global) is the right thing to do. Yes, each require will execute the script to create its exports object that belongs to that one `Require` (again, this is by design), but the scripts won't be recompiled each time. As long as the script isn't changed, the same compiled script will be re-executed by each `Require`. If executing a module takes 10 seconds, then you have one very expensive to initialize module.

p-bakker:
Sounds like everything is working as it should, leaving this open tagged with the `docs` label, so the documentation can be updted with some info about this before closing this issue

AlexTrotsenko:
Thanks for providing the information. 
This issue can be closed for sure given the time, which was passed from the time when the issue was created  :)

Issue #95, opened 2012-11-05
ES5 Object not accessible when used in commonjs module


dsolimando:
This code:

Object.defineProperties(obj, {
    "property1": { value: true,  writable: true },
    "property2": { value: "Hello", writable: false}
});

will fail with the following exception:

org.mozilla.javascript.EcmaError: ReferenceError: "Object" is not defined.

szegedi:
Works for me:
```java
import java.io.StringReader;
import java.net.URI;
import java.net.URISyntaxException;
import org.mozilla.javascript.Context;
import org.mozilla.javascript.Scriptable;
import org.mozilla.javascript.commonjs.module.Require;
import org.mozilla.javascript.commonjs.module.RequireBuilder;
import org.mozilla.javascript.commonjs.module.provider.ModuleSource;
import org.mozilla.javascript.commonjs.module.provider.ModuleSourceProvider;
import org.mozilla.javascript.commonjs.module.provider.StrongCachingModuleScriptProvider;

public class Main2 {
    public static void main(String[] args) {
        try(Context cx = Context.enter()) {
            cx.setOptimizationLevel(-1);
            Scriptable scope = cx.initStandardObjects();
            Require r = new RequireBuilder()
                .setModuleScriptProvider(new StrongCachingModuleScriptProvider(
                    new ModuleSourceProvider() {
                        @Override public ModuleSource loadSource(String moduleId, Scriptable paths,
                            Object validator) throws URISyntaxException {
                            var base = new URI("http://localhost/");
                            return loadSource(base.resolve(moduleId), base, validator);
                        }

                        @Override
                        public ModuleSource loadSource(URI uri, URI baseUri, Object validator) {
                            return new ModuleSource(
                                new StringReader(
                                    "var obj={}; Object.defineProperties(obj, {\n"
                                     + "\"property1\": { value: true, writable: true },\n"
                                     + "\"property2\": { value: \"Hello\", writable: false}\n"
                                     + "});"), null, uri, baseUri, validator);
                        }
                    })
                )
                .createRequire(cx, scope);
            r.install(scope);
            cx.evaluateString(scope, "require('whatever.js')", "", 1, null);
        }
    }
}
```
This programs exits without errors. Try to introduce a typo in "definePropertes" and you'll see it fails (so it does work without the typo.)

p-bakker:
Closing as cannot be reproduced. Feel free to reopen with a reproducible example

Issue #55, opened 2012-06-29
[CommonJS] Support for loading (nodejs) packages


facundocabrera:
Hello,

I'm trying to load a module's folder instead of a specific js file and I'm getting a error:

ReferenceError: "index" is not defined. (file:/****/modules/test#1).
1. The root of the modules is configured to be: file:/****/modules/
2. From a script (modules.js) I'm doing:
   
   var test = require("test");
   
   Where, test in a folder inside modules with an index.js file defining the module's startup.

Important:

```
If I define the require as:

    var test = require("test/index");

The code works perfect.
```

My Findings:

```
I'm using the class UrlModuleSourceProvider, and I guess I could check if the moduleId is a
directory, and then append the string "/index" to the moduleId in that situation, but I'm not 
sure if that is the best way to add the behaviour.
```

Could someone give me a hand?

Thanks!

p-bakker:
As @hns mentioned, loading (nodejs) packages isn't part of the CommonJS specification and thus not part of Rhino's CommonJS implementation, thus I marked this as a Feature

If packages support is required right now, either use https://github.com/micmath/Rhino-Require or fiddle with the (Url)ModuleSourceProvider as suggested in https://github.com/mozilla/rhino/issues/55#issuecomment-9448205 to achieve the result you want.

As (nodejs) package isn't standard CommonJS, I would not expect this to become part of the Rhino core, but maybe some add-on that can be pulled into your environment if needed

Also tagged this with the `docs` tag to make a note about this in the documentation

PR #1301, opened 2023-02-07
Fix the condition for isResourceChanged

szegedi:
I was experimenting with this code while building a bit of a showcase for my OSS work and uncovered this bug when loading code from an actual HTTP server and modifying it and testing reloading under modification.

### DIFF
--- a/src/org/mozilla/javascript/commonjs/module/provider/UrlModuleSourceProvider.java
+++ b/src/org/mozilla/javascript/commonjs/module/provider/UrlModuleSourceProvider.java
@@ -249,7 +249,7 @@ public class UrlModuleSourceProvider extends ModuleSourceProviderBase {
         private boolean isResourceChanged(URLConnection urlConnection) throws IOException {
             if (urlConnection instanceof HttpURLConnection) {
                 return ((HttpURLConnection) urlConnection).getResponseCode()
-                        == HttpURLConnection.HTTP_NOT_MODIFIED;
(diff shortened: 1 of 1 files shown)


p-bakker:
On the face it this seems an obvious bug and straight forward fix, so from that perspective I'd say: LGTM

However: could you maybe check the open commonjs issues, to see whether this PR fixes any of them or whether there are additional fixes to be done in this area? https://github.com/mozilla/rhino/issues?q=is%3Aopen+is%3Aissue+label%3ACommonJS

szegedi:
> However: could you maybe check the open commonjs issues

Sure, I can give them a read-over. Here's the results:

#431 is similar but it was already fixed (should be closed, I guess – I can't close it as I had trouble with Mozilla's authentication changes so I lost my maintainer privileges for now.)
#198 commented on it – works as expected, I tried to explain to the user that reloading of source code doesn't affect already loaded exports
#95 WFM, I added a working example in the comments.
#55 is a feature request, supporting node.js packages in addition to CommonJS modules.
#53 and #10 are some Windows issues; #10 seems to be working as expected since the module directory needs to be an URI.

p-bakker:
> Sure, I can give them a read-over. Here's the results:
Tnx a bunch!

szegedi:
> Tnx a bunch, updated all those issues accordingly

Awesome. Any chance you can merge this one?

p-bakker:
Can't, am not a committer :-/ 

@rbri @gbrail maybe?

rbri:
thanks Attila

PR #1302, opened 2023-02-07
Treat String, ConsString, Boolean, and Double as value types

szegedi:
I've been helping some folks that rely heavily in continuations comparisons in their product, and they had weird failures to compare continuations as equal. Turns out it's mostly due to differences in deserialized objects not having their constructor names etc. interned. The solution ultimately is to (correctly) consider JavaScript strings, booleans, and numbers to not have identity but treat them as purely value types, in line with the JS language semantics.

### DIFF
--- a/src/org/mozilla/javascript/EqualObjectGraphs.java
+++ b/src/org/mozilla/javascript/EqualObjectGraphs.java
@@ -63,9 +63,9 @@ final class EqualObjectGraphs {
             return true;
         } else if (o1 == null || o2 == null) {
             return false;
-        // String (and ConsStrings), Booleans, and Doubles are considered
-        // JavaScript primitive values and are thus compared by value and
-        // with no regard to their object identity.
+            // String (and ConsStrings), Booleans, and Doubles are considered
+            // JavaScript primitive values and are thus compared by value and
+            // with no regard to their object identity.
         } else if (o1 instanceof String) {
             if (o2 instanceof ConsString) {
                 return o1.equals(o2.toString());
--- a/src/org/mozilla/javascript/EqualObjectGraphs.java
+++ b/src/org/mozilla/javascript/EqualObjectGraphs.java
@@ -63,6 +63,21 @@ final class EqualObjectGraphs {
             return true;
         } else if (o1 == null || o2 == null) {
             return false;
+        // String (and ConsStrings), Booleans, and Doubles are considered
+        // JavaScript primitive values and are thus compared by value and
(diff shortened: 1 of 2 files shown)


src/org/mozilla/javascript/EqualObjectGraphs.java
```diff
@@ -63,6 +63,21 @@ boolean equalGraphs(Object o1, Object o2) {
...
+            if (o2 instanceof ConsString) {
+                return o1.equals(o2.toString());
+            }
+            return o1.equals(o2);
+        } else if (o1 instanceof ConsString) {
+            if (o2 instanceof String || o2 instanceof ConsString) {
+                return o1.toString().equals(o2.toString());
+            }
+            return false;
+        } else if (o1 instanceof Boolean || o1 instanceof Double) {
```
tonygermano, line 79: I think internally any `java.lang.Number` can represent a javascript primitive number, with the exception of BigInteger, which is a javascript bigint.

szegedi: Are you sure about that? Even otherwise integer literals will end up being represented as doubles in Rhino, e.g.
```java
import org.mozilla.javascript.Context;

public class TestNumbers {
        try (var cx = Context.enter()) {
            cx.setOptimizationLevel(9);
            var scope = cx.initStandardObjects();
            var result = cx.evaluateString(scope, "1", "", 1, null);
            System.out.println(result.getClass().getName());
        }
    }
}
```
prints:
```
java.lang.Double
```
(Notice I used the most aggressive optimization level.) I could add additional types that we _certainly_ know the engine uses internally, but I wouldn't blanket treat all `Number` subclasses especially since there can be user-defined subclasses too.

szegedi:
Okay, I modified the code to use by-value comparison for all Java primitive type wrappers.

src/org/mozilla/javascript/EqualObjectGraphs.java
```diff
@@ -37,6 +39,19 @@
 final class EqualObjectGraphs {
     private static final ThreadLocal<EqualObjectGraphs> instance = new ThreadLocal<>();
 
+    private static final Set<Class<?>> valueClasses =
+            Collections.unmodifiableSet(
+                    new HashSet<>(
+                            Arrays.asList(
```
szegedi, line 45: This is the closest we can get to a `Set.of` in a pre-9 Java; currently Rhino still compiles for 8.

rbri:
Thanks....

PR #1304, opened 2023-02-13
BUG: for X of javaList does not work properly in strict mode

rPraml:
When using "for X of javaList" objects are not properly wrapped. This causes errors in strict mode


rbri:
@rPraml thanks a lot

Issue #652, opened 2020-03-16
Rest Parameters Support


delchev:
Hi,

Has anyone already explored how we can support the "rest parameters" in functions?

```javascript
function sum(...theArgs) {
  return theArgs.reduce((previous, current) => {
    return previous + current;
  });
}

console.log(sum(1, 2, 3));
// expected output: 6

console.log(sum(1, 2, 3, 4));
// expected output: 10
```

> https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/rest_parameters

Any guidance will be very much appreciated!

Regards,
Nedelcho

p-bakker:
This case is blocking us from updating [test262](https://github.com/tc39/test262/tree/f94fc660cc3c59b1f2f9f122fc4d44b4434b935c) to their latest master, as the [harness code](https://github.com/tc39/test262/tree/f94fc660cc3c59b1f2f9f122fc4d44b4434b935c/harness) used by the tests in test262 make use of rest parameters, which crashes the entire test suite.

p-bakker:
Support for ...restParams on (regular, so not arrow) functions has been introduced in https://github.com/mozilla/rhino/pull/1451 and available in V1.7.15

anivar:
## Testing Results

I tested rest parameters in Rhino 1.8.1-SNAPSHOT with VERSION_ES6 (200):

```javascript
function sum(...theArgs) {
    console.log("Args:", theArgs);
    return theArgs.reduce((a, b) => a + b);
}

console.log("Rest parameters test:", sum(1, 2, 3, 4));
```

**Result:** ✅ Works perfectly! 
- Outputs `Args: [1,2,3,4]` 
- Outputs `Rest parameters test: 10`

**Code Evidence:** 
- Token.java has `DOTDOTDOT` token for spread/rest syntax
- Parser.java handles rest parameters in function definitions
- Arrow functions with rest parameters also work

Rest parameters are fully functional in current Rhino. This issue can be closed.

rbri:
@anivar as the one that did the ompl for ordinary functions, i will have a look at the arrow function part of this sory and hopefully close this. Thanks for the reminder and your investigation.

anivar:
@rbri  You're right.  arrow function rest parameters don't work, only regular function rest parameters do. The issue should stay open.

PR #1313, opened 2023-03-17
Support ES2019 Array.prototype.flat

### LINKED ISSUE #948: Support ES2019 Array.prototype.flat
See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat

midgleyc:
Closes #948.

I'm not sure whether the implementation of the various ids are correct: I just copied the surrounding code.

I took the harmony test from https://github.com/v8/v8/blob/main/test/mjsunit/harmony/array-flat.js, added `load("testsrc/assert.js");` to the start and `"success";` to the end, and then made adjustments as Rhino doesn't supported redeclared `const`s in blocks or the `Proxy` object.

The language version in the test I have set to ES6, but as `flat` is part of ES2019, perhaps there should be a new version? Perhaps not, and I should just remove the comment?

By [spec](https://tc39.es/ecma262/multipage/indexed-collections.html#sec-array.prototype.flat), we should throw a `TypeError` if the index `j` passed to `defineElem` is greater than 2<sup>53</sup> - 1. I have reused the message `msg.arraylength.too.big` here, but let me know if you'd like me to do something different.

### DIFF
--- a/src/org/mozilla/javascript/NativeArray.java
+++ b/src/org/mozilla/javascript/NativeArray.java
@@ -1971,7 +1971,7 @@ public class NativeArray extends IdScriptableObject implements List {
     private static Object js_flat(Context cx, Scriptable scope, Scriptable thisObj, Object[] args) {
         Scriptable o = ScriptRuntime.toObject(cx, scope, thisObj);
         double depth;
-        if (args.length < 1) {
+        if (args.length < 1 || Undefined.isUndefined(args[0])) {
             depth = 1;
         } else {
             depth = ScriptRuntime.toInteger(args[0]);
--- a/src/org/mozilla/javascript/NativeArray.java
+++ b/src/org/mozilla/javascript/NativeArray.java
@@ -151,7 +151,6 @@ public class NativeArray extends IdScriptableObject implements List {
         addIdFunctionProperty(ctor, ARRAY_TAG, ConstructorId_findIndex, "findIndex", 1);
         addIdFunctionProperty(ctor, ARRAY_TAG, ConstructorId_reduce, "reduce", 1);
         addIdFunctionProperty(ctor, ARRAY_TAG, ConstructorId_reduceRight, "reduceRight", 1);
-        addIdFunctionProperty(ctor, ARRAY_TAG, ConstructorId_flat, "flat", 0);
         addIdFunctionProperty(ctor, ARRAY_TAG, ConstructorId_isArray, "isArray", 1);
         addIdFunctionProperty(ctor, ARRAY_TAG, ConstructorId_of, "of", 0);
         addIdFunctionProperty(ctor, ARRAY_TAG, ConstructorId_from, "from", 1);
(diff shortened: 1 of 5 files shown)


src/org/mozilla/javascript/NativeArray.java
```diff
@@ -151,6 +151,7 @@ protected void fillConstructorProperties(IdFunctionObject ctor) {
         addIdFunctionProperty(ctor, ARRAY_TAG, ConstructorId_findIndex, "findIndex", 1);
         addIdFunctionProperty(ctor, ARRAY_TAG, ConstructorId_reduce, "reduce", 1);
         addIdFunctionProperty(ctor, ARRAY_TAG, ConstructorId_reduceRight, "reduceRight", 1);
+        addIdFunctionProperty(ctor, ARRAY_TAG, ConstructorId_flat, "flat", 0);
```
tuchida, line 154: This code adds the method to the constructor(`Array.flat`), not the prototype(`Array.prototype.flat`).
It is an old specification called Array generic methods.
https://lia.disi.unibo.it/materiale/JS/developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array.html#Array_generic_methods

I do not think this is necessary as it is not in latest ECMA262 specification.

tuchida (commented):
Thanks for your pull request.

Rhino uses Test262 to test ECMAScript specifications. Could you test the specifications by following these steps?
1. [Install Test262](https://github.com/mozilla/rhino/tree/master/testsrc#running-the-official-ecmascript-test-suite-test262)
2. Delete the following line
  https://github.com/mozilla/rhino/blob/d22447c4d82c5e975770c299ced589339c0ef809/testsrc/org/mozilla/javascript/tests/Test262SuiteTest.java#L92
3. [Updating the test262.properties file](https://github.com/mozilla/rhino/tree/master/testsrc#updating-the-test262properties-file)

midgleyc:
Thanks, I've updated.

The failing spec tests seem to be failing on something unrelated to `flat` itself -- I see the error 
> missing ) after formal parameters
so either I've failed to wire something up correctly, or there's something wrong with a different bit of Rhino.

src/org/mozilla/javascript/NativeArray.java
midgleyc: Right, I just added these because I saw other functions doing this -- e.g. `Array.reduceRight` doesn't exist either, just `Array.prototype.reduceRight`.

I've removed `ConstructorId_flat` entirely.

tuchida:
> > missing ) after formal parameters

ref. #652, #817
That is the problem that test262/harness/compareArray.js cannot be loaded because Rest Parameters are not implemented.
I am sorry, but some of the Test262 code does not work because of a specification that is not implemented in Rhino.

rbri:
Hi @midgleyc,
sorry for the silence here. This looks good to me and i like to merge.

Can you please do

* an rebase, and a regeneration of the test262.properties file (i have done a fix for the compareArray.js). This will show that even more tests are passing
* move the files ArrayFlatTest.java and array-flat.js into the testsrc/org/mozilla/javascript/tests/es6 folder (or if you like create a new es2019 folder

Thanks a lot for providing this

midgleyc:
Sure, no problem.

As you have merged this already (thanks for this!) I will make those changes in another PR.

rbri:
sorry, doing it right no. No need for changes.

rbri:
Thanks for the PR.

PR #1317, opened 2023-04-10
Test optimization levels

rbri:


rbri:
The idea is to make sure we get the same result when execution js stuff with different optimization levels.
(My next pr is exactly one case where we get different results at the moment)

rbri:
Ah now i see you point, but this is more a general problem with the test suite (i noted parts of the problem also while working on this).
But this is only about this special test cases - do you think they are running already with different optimization levels?

Maybe i will try to cleanup the test suit regarding this (step by step).

p-bakker:
Haven't double-checked it now, but from what I recall running `./gradlew test` does run both Test Suites at optLevel -1, 0 and 1. This is also what it says at https://github.com/mozilla/rhino/blob/master/testsrc/README.md#running-tests

p-bakker:
And I know for sure that the update test262.properties option only marks a particular test as passed in the .properties file if it passes all three tested optLevels

rbri:
ok, will have a look

rbri:
> And I know for sure that the update test262.properties option only marks a particular test as passed in the .properties file if it passes all three tested optLevels

Yes that's also my understanding, same for all test annotated @RhinoTest (but we have to at least sync the test levels used)

gbrail:
Sorry it's been such a long time -- some of the big frameworks like test262 do indeed run bunches of tests at multiple optimization levels but many more tests do things differently and it's inconsistent. If this PR makes more tests work at different optimization levels then it seems like a pretty good idea to fix!

PR #1318, opened 2023-04-10
remove some obsolete code and add another test

rbri:
Looks like this code was part of backward sync from HtmlUnit changes into Rhino.
But the code
* produces different results when running with/without optimization
* produces wrong results without optimization
* is no longer needed from HtmlUnit's point of view
* consumes some memory

see also https://github.com/HtmlUnit/htmlunit/issues/354

### DIFF
--- a/src/org/mozilla/javascript/CodeGenerator.java
+++ b/src/org/mozilla/javascript/CodeGenerator.java
@@ -8,15 +8,10 @@ package org.mozilla.javascript;
 
 import java.math.BigInteger;
 import java.util.List;
-import org.mozilla.javascript.ast.AstNode;
-import org.mozilla.javascript.ast.AstRoot;
-import org.mozilla.javascript.ast.Block;
 import org.mozilla.javascript.ast.FunctionNode;
 import org.mozilla.javascript.ast.Jump;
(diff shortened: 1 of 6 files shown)


p-bakker:
Hard time judging this PR, I agree that we shouldn't have different behavior depending on the optLevel used. 

However, it seems that the ECMAScript spec is not clearly defining how conditionally defined functions ought to behave (judging by https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/function#block-level_function_declaration), so which behavior do we want to follow?

The added test is running in non-strict (as in the test script doesn't have "use strict" in it). As the behavior in strict mode is clearly defined (function declarations inside blocks in strict-mode are local to the block, and hoisted within that block), shouldn't we have a test for strict-mode as well? (Not sure whether the MozillaSuiteTest already runs tests in both strict and non-strict mode, like the test262uite does)

And the code removed, isn't that code that is needed for a proper strict-mode implementation?

p-bakker:
Converted to draft as this PR needs work

rbri:
closing this, was something from the resunc of htmlunit core-js and this is more or less done. inbetween the whole thing got so many updates that this is no longer relevant.

PR #1324, opened 2023-05-27
PropertyDescriptor fixes

rbri:
* the PropertyDescriptor has an setter slot even if no setter function is available
* fix the order of the slots


rbri:
@p-bakker what do you think?

PR #1333, opened 2023-06-07
Update plugin versions

gbrail:
Not yet ready to build on Java 20, but this at least makes building on Java 15 easier.


gbrail:
Need to take a different tack on version compatibility.

PR #1336, opened 2023-06-08
Upgrade Gradle wrapper

gbrail:


gbrail:
Taking a different option here to make version compatibility work better.

PR #1340, opened 2023-06-09
docs(readme): remove broken MDN link

caugner:
MDN no longer hosts Mozilla project documentations, so the link results in an HTTP 404.


gbrail:
Thanks for fixing this!

PR #1344, opened 2023-06-16
Upgrade build system to support a minimum of Java 11 only

gbrail:
* Change language level to 11
* Change GitHub Actions to test Java 11 minimum
* Slightly upgrade Spotless version
* Fix environment so that reflection tests pass


gbrail:
I am going to close this and address it later. We're talking about a code reorganization after the 1.7.15 release, and it makes more sense to me to introduce this change then when we're changing a lot of other things anyway.

gbrail:
Closing and addressing later.

PR #1352, opened 2023-07-06
Improve overloaded method invocation when null is passed.

rPraml:
**Problem**

We often have to deal with overloaded methods like

```java
void setValue(String value)
void setValue(Object value)
```

When invoking such methods from rhino, everything is fine, as long as the value is not undefined or null

```javascript
v.value = 123; // will call setValue(Object)
v.value = "foo"; // will call setValue(String)
v.value = null; // will fail with ambiguous method call
```

Currently, there's no way to pass null to an overloaded value, as javascript cannot determine, which method should be called.

**Possible fix**

Adjusting the conversion weight for `JSTYPE_UNDEFINED` and `JSTYPE_NULL`:
Converting them to String should weight less than converting them to Object (this means, that `setValue(String)` will be preferred)

**Note and room for discussion**

I know, the choiche to favor String over Object is closely related to our use-case (and I'm OK if we can find a consent in which order such conversion should happen)

What I really want is to invoke overloaded methods with "null" values.
Unfortunately I cannot do a `setValue((String)null)` or `setValue((Object)null)` in javascript

It may be possible to construct some hacky null type for java classes (e.g. `String.null` `Integer.null` and so on)


p-bakker:
Can't you use the LiveConnect syntax from JavaScript to specify which overloaded variant you want to invoke like some instance['setValue(String)'](null)?

I think the LiveConnect spec also defined priorities for selecting which overloaded method to take based on the provided params if no specific overload is specified

rPraml:
Closing this PR. We must use the extended LiveConnect syntax

p-bakker:
Just for future reference: https://www-archive.mozilla.org/js/liveconnect/lc3_method_overloading.html

Issue #1076, opened 2021-10-22
Current `VMBridge` is not suitable for modular JDKs


carlosame:
`VMBridge` allows some customization by trying to load a `org.mozilla.javascript.VMBridge_custom`, however due to modularity access rules such a class would not be allowed to load in modular JDKs. Then `VMBridge_jdk18` is loaded, but that class is not 100% suitable for modular JDKs (for example it handles proxy objects in a pre-modular way which could lead to failures).

As a subset of what is currently done in PR #1072, a few changes could be made to `VMBridge`, respecting the current API signature but changing the way that it is loaded:

1) The `VMBridge_custom` would be located at `org.mozilla.javascript.bridge` which could be provided by a different module.
2) Provide a `VMBridge_jdk11` which inherits from a new `VMBridge_jdk8` (keeping a `VMBridge_jdk18` would be confusing) and uses the modular API.

I could modify #1072 to provide the above changes (and leave the protected access changes for a later PR).

p-bakker:
@rbri @gbrail could you comment on whether this is still something relevant? 

I tried following all the cases and (not merged) PRs related to the VMBridge, modular Java and Java 11 and could not make heads or tails out of it. 

@rbri you closed your PR in this area (#1070) in favor of #1072, but neither ever got merged

@gbrail you mentioned in https://github.com/mozilla/rhino/pull/1479#issuecomment-2155673632 that we might not need a VMBridge at all now...

I'm lost... 🤔 Should we fix something, remove something, do nothing?

PR #1353, opened 2023-07-06
EXPERIMENTAL: Remove SecurityManager

rPraml:
Hello, I want to discuss, if it is already time to remove the dependency to SecurityManager from Rhino.

There was already a discussion, where I was involved: https://github.com/mozilla/rhino/discussions/972#discussioncomment-1413364

and the last days I found some time to update our fork.

**Why should we remove the dependency to securityManager**
- because it is deprecated since Java 9
- because it is deprecated for removal since java 17 

**Why shouldn't we remove it**
- because there are JVMs that still have an active security-manager
- because people rely on a complex security policy, when executing javascript (at least we did this, but we removed it from our application code, because we realized that the concept is very hard to understand and there is no 100% security)

This PR is radical and removes all code pieces, that relate to SecurityManager.

(Note: I've made an earlier PR, where I moved all stuff to a "SecurityBridge". See: https://github.com/mozilla/rhino/pull/1068)

Maybe you can provide some feedback, what will be a good way to go here?


src/org/mozilla/javascript/ClassCache.java
```diff
@@ -132,7 +101,7 @@ public synchronized void setCachingEnabled(boolean enabled) {
     }
 
     /** @return a map from classes to associated JavaMembers objects */
-    Map<CacheKey, JavaMembers> getClassCacheMap() {
+    Map<Class, JavaMembers> getClassCacheMap() {
```
rPraml, line 104: No security context -> We can revert this change https://github.com/mozilla/rhino/pull/1019

src/org/mozilla/javascript/DefiningClassLoader.java
```diff
@@ -23,8 +23,7 @@ public Class<?> defineClass(String name, byte[] data) {
         // Use our own protection domain for the generated classes.
         // TODO: we might want to use a separate protection domain for classes
         // compiled from scripts, based on where the script was loaded from.
-        return super.defineClass(
-                name, data, 0, data.length, SecurityUtilities.getProtectionDomain(getClass()));
+        return super.defineClass(name, data, 0, data.length, getClass().getProtectionDomain());
```
rPraml, line 26: CHECKME: Would we need a protectionDomain at all?

src/org/mozilla/javascript/JavaAdapter.java
```diff
@@ -488,10 +488,7 @@ static Class<?> loadAdapterClass(String className, byte[] classBytes) {
         Class<?> domainClass = SecurityController.getStaticSecurityDomainClass();
         if (domainClass == CodeSource.class || domainClass == ProtectionDomain.class) {
             // use the calling script's security domain if available
-            ProtectionDomain protectionDomain = SecurityUtilities.getScriptProtectionDomain();
-            if (protectionDomain == null) {
-                protectionDomain = JavaAdapter.class.getProtectionDomain();
-            }
+            ProtectionDomain protectionDomain = JavaAdapter.class.getProtectionDomain();
```
rPraml, line 491: CHECKME: can this be simplified?

src/org/mozilla/javascript/JavaMembers.java
```diff
@@ -319,8 +318,7 @@ private void discoverAccessibleMethods(
                                 if (isPublic(mods) || isProtected(mods) || includePrivate) {
                                     MethodSignature sig = new MethodSignature(method);
                                     if (!map.containsKey(sig)) {
-                                        if (includePrivate && !method.isAccessible())
-                                            method.setAccessible(true);
+                                        VMBridge.instance.tryToMakeAccessible(method);
```
rPraml, line 321: CHECKME: Do we need a VMBridge for java 11/17?

rPraml (commented):
This PR contains some thoughts. Maybe more code can be removed/simplified (like SecureCaller)

gbrail:
I think that we should do this after the 1.7.15 release -- at that point I want to reorganize the source files (git seems to be smart enough to make that less of a nightmare), upgrade the build to require Java 11 minimum, and then this can be one of the first things that we address.

rPraml:
Hello @gbrail, @p-bakker 

I've resolved the merge conflicts here, but I will still consider this as experimental:

Maybe we shoud first decide, if we still need for VMBridge / VMBridge_jdk18, repectively JavaMembers / JavaMembers_jdk11

I don't know, if there are still use cases, where someone needs its own VMBridge, but trying to load `VMBridge_custom` from `org.mozilla.javascript` will not work in a modular java (see related issue #1076)
-> either we decide to remove custom VMBridges completely 
-> or load it with ServiceLoader

As minimum JVM version is now 11, there will be also no need for JavaMembers / JavaMembers_jdk11

p-bakker:
Converted as draft as still considered experimental by the author

p-bakker:
@rPraml in the mentioned [discussion](https://github.com/mozilla/rhino/discussions/972#discussioncomment-1413364) you said your company currently heavily relies on the SecurityManager to prevent untrusted JavaScript code from doing stuff that is not allowed. 

Have you come up with an alternative approach achieving the same that didn't rely on the SecurityManager?

rPraml:
@p-bakker 
The short story:

We do not longer rely on SecurityManager and I would agree, if it will be removed from rhino (in 2.0?)

Long story:

In the beginning we tried to allow the end user to edit our JavaScript code, where we have modelled a lot of our business logic.
This JavaScript code was (more or less) protected by ClassShutter and a whiteList of classes/packages that we have classified as "secure"

We found ways to bypass these mechansims (e. g. `trustedObject.getClass().getClassLoader().loadClass("non.allowed.Clazz")`) and began to experiment with a SecurtityManager for rhino.

Debugging and fixing one security issue brings the next to light... (it was a pain, to find a good balance, to enable features and yet offer enough security)

With the knowledge of JEP411, we abandoned the idea that we could ever provide a JavaScript user interface where the end user cannot compromise the system.

So in our current application, we have forbidden to view/edit JavaScript code to the end user

p-bakker:
Right, was afraid of that. I guess the only way to allow untrusted script to run 'securely' is to disable Java Interop completely and create custom host objects (Scriptable's) that expose whatever domain logic you want to expose

And 'securily' in brackets, because:
- a rogue script could consume a ton of resources, starving the system 
- unless you properly seal everything, including (hidden) prototypes etc (see Secure EcmaScript/SES) and execute in a new or user/tenant/... specific Realm, there scripts can interfere with eachother
- who knows what other attack vectors there might be...

p-bakker:
Wrt to removing the SecurityManager: while it has been delayed for removal, to my knowledge is not clear yet in which Java version it'll actually be removed, no? 

I propose to leave it in Rhino at least until after the next release, as the next release is sharing up to be a significant improvement on the EcmaScript-compatibility side (amongst other improvements) and I rather not lock Rhino users out of those goodies if they currently depend on the SecurityManager

What I think we could do though is mark the relevant methods as deprecated, include a note about it in the release notes and maybe write some documentation about security with Rhino in general and about moving away from the SecurityManager specifically

rPraml:
In https://openjdk.org/jeps/411 they said

> In feature releases after Java 18, we will degrade other Security Manager APIs so that they remain in place but with limited or no functionality. For example, we may revise AccessController::doPrivileged simply to run the given action, or revise System::getSecurityManager always to return null

But as far as I see, this hasn't been done yet. There are no relevant changes in
https://github.com/openjdk/jdk/commits/master/src/java.base/share/classes/java/lang/SecurityManager.java
https://github.com/openjdk/jdk/commits/master/src/java.base/share/classes/java/security/AccessController.java

There is a comment in https://bugs.openjdk.org/browse/JDK-8264713 where someone asked for a date.

The only thing I noticed, that some testcases print the warning
```
WARNING: A terminally deprecated method in java.lang.System has been called
WARNING: System::setSecurityManager has been called by org.mozilla.javascript.tests.SecurityControllerTest
```
on the console since some time.

Rhython:
`SecurityManager` will be Permanently Disable at JDK24 by [JEP486](https://openjdk.org/jeps/486). So I think it's time to accept this change.

gbrail:
The question is, how many of the old codebases relying on Rhino plan to
upgrade to a new version, but still rely on it?

That'd help us decide between either just removing all the
SecurityManager-related code, or putting in some mechanism to only call
those APIs with older Java versions, which will be ugly and might not work
but would allow more compatibility.

carlosame:
> `SecurityManager` will be Permanently Disable at JDK24 by [JEP486](https://openjdk.org/jeps/486). So I think it's time to accept this change.

Disabled but not removed.

> The question is, how many of the old codebases relying on Rhino plan to upgrade to a new version, but still rely on it?

Both [EchoSVG](https://github.com/css4j/echosvg) and Apache Batik rely on the Security Manager to run Rhino. In the case of EchoSVG, I do not think that applying this change on Rhino would have much impact (actually, the outcome could be positive because of the `doPrivileged` responsibility shift). My plan is to remove all the `SecurityManager` stuff from EchoSVG when the whole thing is removed from the JDK.

So as far as the EchoSVG project is concerned, it is fine to apply this PR.

I do not know about the impact on Apache Batik, given that their code and policy files do not work out-of-the-box, and users should write their own (I wonder if many do).

rbri:
as we are working toward 2.0 i think we have to do this now (and maybe something more).
Have done a quick test with my stuff and this works fine (no suprise because i do not use the security stuff at all).
So fine and appriciated for HtmlUnit to do this.

rPraml:
Hi everyone, I thought it was about time I checked in again. I know I used to be very active in this project, but unfortunately I simply don't have enough time at the moment. I hope you'll understand.

I also don't want to give the impression that I've turned my back on the project completely. However, my priorities have changed both in my personal life (I recently became a dad!) and at work, so I can't say when I'll be able to find the time to help out again.

I'm about to take a month of parental leave, and after that I’ll have to see how things go :)

rbri:
@rPraml at first congrats and all the best for the Praml family. Enjoy the time, there are more important things than Rhino's
And btw. my last comment was more for @gbrail and @aardvark179... maybe they can check if that fits into the plan/picture for 2.0

PR #1357, opened 2023-07-20
Symbol fixes

rbri:
* fix an problem with typeof when used for symbols
* take care of SymbolKey before casting - this fixes some problems when adding new symbols to java classes mapped into the js engine

### DIFF
--- a/src/org/mozilla/javascript/NativeSymbol.java
+++ b/src/org/mozilla/javascript/NativeSymbol.java
@@ -67,7 +67,7 @@ public class NativeSymbol extends IdScriptableObject implements Symbol {
         this.symbolData = null;
     }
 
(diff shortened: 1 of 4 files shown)


p-bakker:
Based on looking at the changed code and tests, I'm assuming there was an issue with the return value of the  typeof opertor when called on a Symbol retrieved through .getOwnSymbolProperty/ies?

But if explanation of what was broken/is now fixed in these type of PR's would be handy I think for future reference (and to include in the release notes when time comes).

Other than that: lgtm. No test262 tests that got unbroken by this?

rbri:
@p-bakker comment updated - thanks

gbrail:
Thanks -- I get what this is fixing, and I'll put together a more clear comment when I merge it.

PR #1362, opened 2023-08-01
Experiment with reorganizing into a modular but incompatible build

gbrail:
Restructure code to match current guidance from the Gradle docs on how to do multi-module projects, complete with proper module metadata.


p-bakker:
Hi @gbrail,

Seeing that you already closed this PR I assume you have no intent to pursue this (or at least not at this time)?

For future reference, could you add a little info in which way the resulting build is incompatible with the current?

gbrail:
Thanks for asking -- this should probably be a "discussion."

This particular change is "incompatible" in two ways.

* Since we're literally moving every single source file, in-flight PRs are
going to be hard to merge.
* I structured the library into four modular JARs (and included one "uber
jar" for compatibility). Users SHOULD still be able to use the uber-jar,
but most, especially those using modules elsewhere, should use the right
combination of the modular JARs for them

I'd like to revisit it, actually, and move in this direction.

p-bakker:
> Since we're literally moving every single source file, in-flight PRs are
going to be hard to merge

Disregarding very old PR's, we have about 15 pending PR that are relative recent. Think we should see which ones are still relevant and could be finished & merged before doing the restructuring

gbrail:
Yes -- I agree we should do some more PR merging before a .15 release and
before restructuring.

WIth that said, I merged some recent changes to master into my
"reorganized" branch yesterday and git did mostly the right thing (I had to
manually relocate one moved file and git actually suggested correctly where
I should relocate it). So the merging situation won't be as bad since git
does track where things moved to.

PR #1366, opened 2023-08-08
special handling for NativeError instances when generating the console output

rbri:
Without this, the output is generated by the json serializer ending up in something like this '{"fileName":"source.js","lineNumber":426}'; means the whole message is missing. This fix brings the output closer to the output of real browsers by writing the message and the stack trace.


gbrail:
Looks good to me. Thanks!

rbri:
Thanks

PR #1369, opened 2023-08-11
Fix regressions introduced to debugger

gbrail:
This PR consolidates a change made by @jonasPoehler with additional changes to fix debugger regressions introduced earlier.

Could regular users of the debugger please try it out? Some obviously-broken functionality seems to be working again compared
to previous releases but there may be more to test.


gbrail:
This includes fixes originally contributed to https://github.com/mozilla/rhino/pull/1367

jonasPoehler:
Hey @gbrail, thank you for looking into this. I did checkout your branch and did some quick testing and poking around. I can confirm, that the breakpoint-fix still works and I didn't find any other obvious problem. However I wouldn't know where to look for in the context of the `isCellEditable` method, so there is the potential for some bug lurking, but you get that all the time with manual testing.

PR #1374, opened 2023-08-15
feat: bump spotless plugin so it works without flags in JDK16+

midgleyc:
Running 5.12.1 gives an `IllegalAccessError`:
```
java.lang.IllegalAccessError: class com.google.googlejavaformat.java.JavaInput (in unnamed module @0x7f85e8d3) cannot access class com.sun.tools.javac.parser.Tokens$TokenKind (in module jdk.compiler) because module jdk.compiler does not export com.sun.tools.javac.parser to unnamed module @0x7f85e8d3
```
5.17.1 just works, and should still work on Java 8.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -7,7 +7,7 @@ plugins {
     id 'jacoco'
     id 'distribution'
     id 'checkstyle'
-    id 'com.diffplug.spotless' version "5.12.1"
+    id 'com.diffplug.spotless' version "5.17.1"
     id 'com.github.spotbugs' version "4.7.1"
 }
 


midgleyc:
Java 11 fails, nevermind that then.

gbrail:
Thanks for looking at this.

I think that I was able to get a version that worked for all versions by using the latest "spotless" plugin from Maven Central (so that it runs on newer VMs) but an older version of google-java-format (so that it doesn't break older VMs), like this:

googleJavaFormat('1.10.0').aosp()

Can you try that? We can use that for now if it works for you and then switch to a newer spotless when we move on from Java 11.

midgleyc:
no, no luck there. 1.10 fails on 11, and 1.7 fails on 8.

gbrail:
Thanks for the effort. After this upcoming release we'll do the upgrade and
then I think we can make it all work better.

PR #1377, opened 2023-08-18
take care of SymbolKey before casting

rbri:
again (see https://github.com/mozilla/rhino/pull/1357) take care of ScriptRuntime.isSymbol() == true; the object might be a NativeSymbol or a SymbolKey
(found while working on https://github.com/mozilla/rhino/pull/1332)


gbrail:
I think this is fine although I don't know that we have test coverage for it yet. Since there's more coming I will merge it anyway. Thanks!

PR #1379, opened 2023-08-19
feat: update github action versions to 3

midgleyc:
This fixes the deprecation warnings to do with Node 12 / Node 16.

Checking the release notes, I don't think any changes need to be made other than this.


gbrail:
Thanks for doing this!

PR #1380, opened 2023-08-20
Stop IRFactory from inheriting Parser

tuchida:
- ref. #1188

`IRFactory` no longer inherits from `Parser`. This eliminates Code reuse via inheritance.


rbri:
lets do this.....

gbrail:
Looks good to me too. Thanks!

Issue #917, opened 2021-05-31
Support ES2015 unicode escape sequence additions 


p-bakker:
ES6 added the `\u{...}` escape sequence format to allow escaping unicode values > FFFF

See https://exploringjs.com/es6/ch_unicode.html for example

tuchida:
I am implementing this and have noticed a few things.

- In Rhino, Unicode escaped identifier does not throw an error when it is the same as a reserved keyword, but this is a violation of the specification.
  https://github.com/mozilla/rhino/blob/c275c35f160ef7b035002c67c3540ba317964351/testsrc/org/mozilla/javascript/tests/ParserTest.java#L1122-L1126
  - Here is the cause.
  https://github.com/mozilla/rhino/blob/c275c35f160ef7b035002c67c3540ba317964351/src/org/mozilla/javascript/TokenStream.java#L755-L761
- In Rhino, an incomplete Unicode escape does not throw an error, but it is a violation of the specification.
  https://github.com/mozilla/rhino/blob/c275c35f160ef7b035002c67c3540ba317964351/testsrc/tests/ecma/LexicalConventions/7.7.4.js#L172
  - Here is the cause.
  https://github.com/mozilla/rhino/blob/c275c35f160ef7b035002c67c3540ba317964351/src/org/mozilla/javascript/TokenStream.java#L985

I can be modified as per specifications, but it would be breaking changes. It might be better to control it with some flag. These may be different issues.

p-bakker:
I've implemented this as well as part of trying to implement the 'u'/Unicode option for regexes and ran into the same problems, see 
https://github.com/mozilla/rhino/issues/958#issuecomment-1057366529

Also see that above comment for my thoughts on fixing this: enable the spec compliant behavior in a new ECMASCRIPT language version

p-bakker:
Also, the commit that implemented that specific behavior for keyword is [['+([0-9])?(.{+([0-9]),x}).x', 'master', 'next']](https://github.com/mozilla/rhino/commit/323f151a23216832ac01ff7e790b62cffc2ac0ac)

The commit references an issue 490, but I cannot find any reference to that, so no idea why exactly this was implemented this way

p-bakker:
Think I found what the '490' referred to: issue report against Closure Compiler on the archived Google Code project, see http://blickly.github.io/closure-compiler-issues/#490

tuchida:
- #1383

This has been supported.

PR #1383, opened 2023-08-24
Support unicode codepoint escape

tuchida:
ref. #917
```js
var \u{61} = 1 // var a = 1
```

## Breaking change
All errors occur only when version is 200.
### Incomplete unicode escape
ref. 495a74ed9fb3228a8affc382f50ff8e5c5198740
```js
"\u"; // invalid Unicode escape sequence
```
### Escaped reserved words
ref. 32c76f848fa294eaa4a4cc0365e621288bee759b
```js
var \u0069\u0066 = 1; // `var if = 1` is syntax error
```
### Characters that cannot be used as identifiers
ref. a1c57b501707cf674aeca427d5d561780535885e
```js
var \u000A = 1; // invalid Unicode escape sequence
```
In this Pull Request, the determination of characters that cannot be used as identifiers differs from the specification. The reason is that `isUnicodeIdentifierStart` and `isUnicodeIdentifierPart` are wrong in Java (old spec?). Java20 seems to be the same as the ECMA specification.
```java
java.lang.Character.isUnicodeIdentifierPart(0x2118)
// Java 11.0.20 => false
// Java 20.0.2 => true
```

### DIFF
--- a/src/org/mozilla/javascript/TokenStream.java
+++ b/src/org/mozilla/javascript/TokenStream.java
@@ -722,7 +722,8 @@ class TokenStream {
                 ungetChar(c);
 
                 String str = getStringFromBuffer();
-                if (!containsEscape) {
+                if (!containsEscape
+                        || parser.compilerEnv.getLanguageVersion() >= Context.VERSION_ES6) {
                     // OPT we shouldn't have to make a string (object!) to
                     // check if it's a keyword.
 
--- a/src/org/mozilla/javascript/TokenStream.java
+++ b/src/org/mozilla/javascript/TokenStream.java
@@ -576,6 +576,25 @@ class TokenStream {
         return id & 0xff;
     }
 
+    private static boolean isValidIdentifierName(String str) {
+        int i = 0;
+        for (int c : str.codePoints().toArray()) {
+            if (i++ == 0) {
+                if (c != '$' && c != '_' && !Character.isUnicodeIdentifierStart(c)) {
+                    return false;
+                }
+            } else {
+                if (c != '$'
+                        && c != '\u200c'
+                        && c != '\u200d'
+                        && !Character.isUnicodeIdentifierPart(c)) {
+                    return false;
+                }
+            }
+        }
+        return true;
+    }
+
     final String getSourceString() {
         return sourceString;
     }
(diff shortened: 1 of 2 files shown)


rbri:
Looks like there is a spotless violation ;-)

rbri:
Super cool progress - i like to see this merged!

p-bakker:
Nice progress indeed!

> In this Pull Request, the determination of characters that cannot be used as identifiers differs from the specification

As Java versions are tied to a specific version of Unicode, whereas the EcmaScript spec (at least as of ES2017/ES8) [says the EcmaScript implementation must conform to the latest Unicode standard](https://tc39.es/ecma262/2017/#sec-conformance).

And then the test262 suite is authored against the latest version of EcmaScript (but we aren't on the latest version of the 262 testsuite just yet)

As such, I think some differences can occur running Rhino/tests against different Java versions

tuchida:
This time I checked language version to avoid breaking compatibility as much as possible, but I would like to turn this off.
`TokenStream` requires a big change because unescaped Unicode support is not possible with char in Java.
https://github.com/tc39/test262/blob/f94fc660cc3c59b1f2f9f122fc4d44b4434b935c/test/language/identifiers/other_id_continue.js#L12

gbrail:
Although the conflict between the Unicode version supported by JavaScript and the one supported by Java seems problematic, in general this looks good and I like how many more tests it fixes. Do the rest of you feel it's ready to merge?

rbri:
+1 for merge

p-bakker:
Agree that we should not get too hung up on any issues due to the different Unicode versions supported by Java and required by the different language versions in Rhino, but would like some more insight into the impact/specifics of 'prevent breaking compatibility' and 'In this Pull Request, the determination of characters that cannot be used as identifiers differs from the specification' @tuchida mentioned, if only to make things clear in the eventual release notes and make note of it in the docs somewhere

Other than that, I think we should go ahead and merge, because it's great test262 progress

gbrail:
OK, I agree. We'll want to use the message on this PR in the eventual release notes.

gbrail:
I meant to squash that with a nice commit message, but didn't. I don't think that maintaining a pretty history is worth adding a bunch of revert commits or doing a force-push, so we'll have to use the original text in the PR description when writing up the release notes.

PR #1384, opened 2023-08-30
convert NativeMath and NativeJSON into a Lambda based ScriptableObject

rbri:
and add the toStringTag property

### DIFF
--- a/src/org/mozilla/javascript/NativeJSON.java
+++ b/src/org/mozilla/javascript/NativeJSON.java
@@ -23,22 +23,30 @@ import org.mozilla.javascript.xml.XMLObject;
  *
  * @author Matthew Crumley, Raphael Speyer
  */
-public final class NativeJSON extends IdScriptableObject {
+public final class NativeJSON extends ScriptableObject {
     private static final long serialVersionUID = -4567599697595654984L;
 
-    private static final Object JSON_TAG = "JSON";
+    private static final String JSON_TAG = "JSON";
 
     private static final int MAX_STRINGIFY_GAP_LENGTH = 10;
 
     static void init(Scriptable scope, boolean sealed) {
-        NativeJSON obj = new NativeJSON();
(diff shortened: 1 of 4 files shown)


p-bakker:
Given that it unbreaks a test in test262 and that IMHO the new code is MUCH easier to reason with, I'm all for this change

Are there any other upsides (performance?) or maybe some downsides to using a Lambda-based impl. vs. the IdScriptableObject approach?

rbri:
For me the impl of the toStringTag property based on lambda was simpler and it was just a test to see how hard it might be to do the same for RegEx. This is the next on my list (but after the vacation in october).

Regarding positive/negative side effects i guess @gbrail knows more.

But overall for me having this lambdas opens up some great improvements, its much simpler and more straight forward to implement the stuff in this way.

p-bakker:
> But overall for me having this lambdas opens up some great improvements, its much simpler and more straight forward to implement the stuff in this way.

Yeah, it looks sooo much cleaner :-)

p-bakker:
Any idea yet whether teh Lambda approach will make https://github.com/mozilla/rhino/issues/963 / https://github.com/mozilla/rhino/pull/1181#discussion_r805570782 easier?

gbrail:
Yes, I'm happy to see this too. I actually don't think that there are big downsides to the lambda approach -- I also found native JS stuff to be easier to implement that way as it maps much more directly to the native way that JavaScript works, and furthermore I've tested the performance and I don't recall seeing a significant difference.

One side note on all this -- it'd be easier for me if we start marking PRs "draft" if you are doing it just for discussion, or want input, and un-marking them "draft" when you want them to be seriously considered for merging.

For this particular case, I'm happy to merge this unless there are other concerns.

rbri:
i like to see it merged because it fixes two more test

gbrail:
I think this works and makes sense. Thanks!

PR #1385, opened 2023-09-01
Use a different method to determine if we are on Java 11

gbrail:
This should be more likely to work on Android.

Fix suggested by @carlosame and applied to the latest tree.

Can anyone with Android please try this out?


gbrail:
This passes on all the versions that we currently support and seems harmless, so I'm going to merge it.

zbynek:
Nice to see this merged. Is there a chance for a release including this fix anytime soon?

PR #1387, opened 2023-09-12
javascript 'Set' cannot handle wrapped java objects properly

rPraml:
When adding java objects to a javascript `Set()`, they are not deduplicated.

This is caused, because Java objects are wrapped each time in a new NativeObject. As NativeObject has no hashcode/equals they are inserted multiple times in a `Set`

This PR adds equals+hashcode to NativeObject and a testcase, where the same enum is added twice to a set and the expected size is 1.


gbrail:
Makes sense to me -- thanks!

Issue #1389, opened 2023-10-02
ES2015 Object own keys order issues


p-bakker:
Rhino doesn't conform to the rules laid out in ES2015 wrt the ordering of own keys in objects

Some quick examples of at least some of the issues:
```
var x = {'D': true, '-1': true}
x.A = true
x.Z = true
x.V = true
x.h = true

Object.keys(x) // yields [-1, D, A, Z, V, h], should be  [D, -1, A, Z, V, h] as -1 is not an array index key
JSON.stringify(x) // yields {"-1":true,"D":true,"A":true,"Z":true,"V":true,"h":true}, same issue as with Object.keys
JSON.parse(JSON.stringify(x)) // yields {true,D:true,A:true,Z:true,V:true,h:true}: -1 key completely missing so invalid JSON
```

See:
- https://mozilla.github.io/rhino/compat/engines.html#ES2015-misc-own-property-order
- https://dev.to/frehner/the-order-of-js-object-keys-458d

rbri:
At least for HtmlUnit i did some hacking in this way https://github.com/HtmlUnit/htmlunit-rhino-fork/commit/e6c4c79bf0a20a7bae6cbe121e9149691a7ac927

But i guess we can't fix this without breaking the backward compatibility.

p-bakker:
In what way does it break backwards compatibility? Just that the keys will be ordered differently (but properly)?

gbrail:
Thanks -- can we narrow this down to a specific list of issues?

It sounds like there is an issue with numeric keys that can't be used as
array indices -- Rhino is improperly treating them as numbers whereas they
should be left as strings. It's not clear yet whether this is a problem
when the keys are inserted or in the comparator that's used to compute the
(unnecessarily expensive, IMO) sort order required to implement the spec.

It sounds like there also might be an issue with JSON.stringify, as it
should never skip keys.

Are there other specific issues?

p-bakker:
I haven't done a full analysis, the same code above was just a quick test I did locally after I saw the many failures in the ES2015 own keys section in the Kangax Compat table

rbri:
more details about the HtmlUnit story are here https://github.com/HtmlUnit/htmlunit/pull/595

PR #1392, opened 2023-10-06
Make handling of object indices more in line with the spec

gbrail:
* In more cases, treat out-of-range numeric indices as strings
* For typed arrays, correctly implement index semantics

This addresses many (but maybe not all) of the issues identified in https://github.com/mozilla/rhino/issues/1389

### DIFF
--- a/src/org/mozilla/javascript/ScriptRuntime.java
+++ b/src/org/mozilla/javascript/ScriptRuntime.java
@@ -1476,6 +1476,10 @@ public class ScriptRuntime {
      * Return -1L if str is not an index, or the index value as lower 32 bits of the result. Note
      * that the result needs to be cast to an int in order to produce the actual index, which may be
      * negative.
+     *
+     * <p>Note that this method on its own does not actually produce an index that is useful for an
+     * actual Object or Array, because it may be larger than Integer.MAX_VALUE. Most callers should
+     * instead call toStringOrIndex, which calls this under the covers.
      */
     public static long indexFromString(String str) {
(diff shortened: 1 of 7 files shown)


p-bakker:
Looks like this PR fixes Object.keys(x) and JSON.stringify(x) testcases, but not the JSON.parse(...) test:
```sh
js> JSON.parse('{"D": true, "-1": true, "A": true}', console.log);
INFO -1 true
INFO D true
INFO A true
INFO {}
```

p-bakker:
Or its just that ScriptRuntime.indexFromString("-1") returns an odd value, namely 4294967295 instead of -1

p-bakker:
Nice! All tests in the compat table under (ES)6 > Misc > 'own property order' (except for the ones based on Reflect) now pass!

Amazed that this doesn't unbreak some Test262 tests, but it is what it is

So, LGTM!

rbri:
@gbrail sorry for being late at this party but while comparing your patch with my one i found one difference

org.mozilla.javascript.ScriptRuntime.getIndexObject(double)

    /**
     * If d is exact int value, return its value wrapped as Integer and othewise return d converted
     * to String.
     */
    static Object getIndexObject(double d) {
        int i = (int) d;
        if (i == d) {
            if (i < 0) {
                return Integer.toString(i);
            }
            return Integer.valueOf(i);
        }
        return toString(d);
    }

I have added the i<0 check to this method but you not. Is this missing or have you checked that this is not needed?

PR #1401, opened 2023-10-27
Modified `DoctestsTest` to include optimization level

andreabergia:
The current behavior is a bit misleading, because the test case name includes only the javascript source file, but not the optimization level. So, you get three tests with the exact same names, for levels -1, 0, and 9. This change simply adds the optimization level in the name of the test.


gbrail:
Thanks!

PR #1408, opened 2023-10-31
Turn forking off in gradle tests

gbrail:
This was on previously because some tests didn't always close their Rhino Context, but now it looks like they all do. This speeds up the tests a lot, save for the MozillaSuiteTest which continues to be a slow one.


gbrail:
This is a nice try but the problem is something more complicated -- we'll get 'em another time.

Issue #1412, opened 2023-11-03
ES2015 Generators don't work properly in non-interpreted mode


p-bakker:
Consider the following testcode `(function*(a) {console.log(a)})(11,22).next()`:
- interpreted: value of `a` parameter is `11`
- non-interpreted: value of `a` parameter is some function object

p-bakker:
@gbrail since you authored the implementation of ESGenerator ~4 years ago, any idea what might be going on here?

tonygermano:
There was an old javascript 1.7 generator implementation that I believe worked that way. My guess is that it's hitting that old code.

tonygermano:
Scratch that. I just reread what the issue was and the problem is with passing the parameter, not the result of next(). Is it still possible that it has something to do with the legacy implementation?

tonygermano:
I did some more testing. It assigns the same new generator function object to all of the parameters with the same function body as the original function. It appears to do that only when the generator function has no yield statement. 

```javascript
s> var param1, param2, g = function*(a,b) {param1 = a; param2 = b}, it = g(11,12)
js> typeof param1
undefined
js> it.next()
[object Object]
js> param1

function (a, b) {
    param1 = a;
    param2 = b;
}

js> param1 === param2
true
js> Object.prototype.toString.call(param2)
[object __GeneratorFunction]
js> param2 === g
false
js> oldResult = param2

function (a, b) {
    param1 = a;
    param2 = b;
}

js> param1 = undefined
js> oldResult(13,14).next()
[object Object]
js> param1

function (a, b) {
    param1 = a;
    param2 = b;
}

js> param1 === param2
true
js> param1 === oldResult
false
```

```javascript
js> var param1, param2, g = function*(a,b) {param1 = a; param2 = b;yield}, it = g(11,12)
js> it.next()
[object Object]
js> param1
11
js> param2
12
```

p-bakker:
Thnx for looking into it. Hope @gbrail can deduct what might be going wrong from all this input

gbrail:
I’ll try to look at this next week when I am back from vacation

p-bakker:
@gbrail any luck giving this a look?

PR #1416, opened 2023-11-04
Adds support for trailing commas in function parameters

p-bakker:
As the title says

BTW think this ought to be a squashed merge

### DIFF
--- a/src/org/mozilla/javascript/Parser.java
+++ b/src/org/mozilla/javascript/Parser.java
@@ -786,6 +786,10 @@ public class Parser {
         Set<String> paramNames = new HashSet<>();
         do {
             int tt = peekToken();
+            if (tt == Token.RP) {
+                fnNode.putIntProp(Node.TRAILING_COMMA, 1);
+                break;
+            }
             if (tt == Token.LB || tt == Token.LC) {
                 AstNode expr = destructuringPrimaryExpr();
                 markDestructuring(expr);
@@ -821,7 +825,7 @@ public class Parser {
                     fnNode.addParam(makeErrorNode());
                 }
             }
-        } while (matchToken(Token.COMMA, true) && peekToken() != Token.RP);
+        } while (matchToken(Token.COMMA, true));
 
         if (destructuring != null) {
             Node destructuringNode = new Node(Token.COMMA);
@@ -969,6 +973,9 @@ public class Parser {
         try {
             if (params instanceof ParenthesizedExpression) {
                 fnNode.setParens(0, params.getLength());
+                if (params.getIntProp(Node.TRAILING_COMMA, 0) == 1) {
+                    fnNode.putIntProp(Node.TRAILING_COMMA, 1);
+                }
                 AstNode p = ((ParenthesizedExpression) params).getExpression();
                 if (!(p instanceof EmptyExpression)) {
                     arrowFunctionParams(fnNode, p, destructuring, paramNames);
@@ -2295,7 +2302,7 @@ public class Parser {
             if (compilerEnv.isStrictMode() && !pn.hasSideEffects())
                 addStrictWarning("msg.no.side.effects", "", pos, nodeEnd(pn) - pos);
             if (peekToken() == Token.YIELD) reportError("msg.yield.parenthesized");
-            if (peekToken() == Token.RP && allowTrailingComma) {
+            if (allowTrailingComma && peekToken() == Token.RP) {
                 pn.putIntProp(Node.TRAILING_COMMA, 1);
                 return pn;
             }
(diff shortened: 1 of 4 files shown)


src/org/mozilla/javascript/Parser.java
```diff
@@ -2288,13 +2288,21 @@ else if (symDeclType == Token.LP) {
     }
 
     private AstNode expr() throws IOException {
+        return expr(false);
```
rbri, line 2291: do we really need this extra step?

testsrc/test262.properties
```diff
@@ -6151,8 +6145,8 @@ language/statements/generators 220/259 (84.94%)
     param-dflt-yield.js {unsupported: [default-parameters]}
     params-dflt-args-unmapped.js {unsupported: [default-parameters]}
     params-dflt-ref-arguments.js {unsupported: [default-parameters]}
-    params-trailing-comma-multiple.js
-    params-trailing-comma-single.js
+    params-trailing-comma-multiple.js non-interpreted
```
rbri, line 6148: Does this mean it fails in compiled mode?

src/org/mozilla/javascript/Parser.java
p-bakker: you mean eliminate `private AstNode expr()` and refactor all callsites to call `private AstNode expr(boolean)` instead?

testsrc/test262.properties
p-bakker: The testcase fails due to #1412 now, but the params-trailing-comma part is fixed

src/org/mozilla/javascript/Parser.java
rbri: yes, i guess eclipse can do this 4 you

tuchida:
Do you want `ast.toSource()` to change?
https://github.com/mozilla/rhino/blob/a3f43df06a8f62f7d9723b5f748075382887417d/src/org/mozilla/javascript/ast/FunctionNode.java#L392-L394

p-bakker:
> Do you want `ast.toSource()` to change?

I guess we should? Added it now, but done so by 'tagging' the FunctionNode with an int property, but I wonder if I should instead add some getter/setter combo on FunctionNode. Thoughts?

tuchida:
Thank you! Looking good!

> I guess we should? Added it now, but done so by 'tagging' the FunctionNode with an int property, but I wonder if I should instead add some getter/setter combo on FunctionNode. Thoughts?

As I read the code, `TRAILING_COMMA` flag is set to `FunctionNode` and `ParenthesizedExpression`. So I think the current code is fine.

p-bakker:
> As I read the code, `TRAILING_COMMA` flag is set to `FunctionNode` and `ParenthesizedExpression`. So I think the current code is fine.

Tnx. 

Just FYI: 'TRAILING_COMMA' is only set on ParenthesizedExpression to carry it forward to where the ParenthesizedExpression is used to construct the params of an Arrow function

src/org/mozilla/javascript/Parser.java
```diff
@@ -2287,14 +2294,18 @@ else if (symDeclType == Token.LP) {
         }
     }
 
-    private AstNode expr() throws IOException {
+    private AstNode expr(boolean allowTrailingComma) throws IOException {
```
gbrail, line 2297: This is a minor thing but I think it's worth discussing -- rather than add a boolean parameter to this function that is used very frequently in the codebase, why not add a new method called "exprWithTrailingComma?" That would prevent having to propagate the idea of a "trailing comma" to every single place in the codebase that calls "expr", but isolate the need to understand this to the few places where we actually support that syntax.

gbrail: This is a stylistic thing but I've seen cases where over time more and more boolean flags are added to a function and it makes the logic more complicated and the function signature gets hard to understand (like you have to go back to the target function every time to understand what each boolean does).

src/org/mozilla/javascript/Parser.java
```diff
@@ -3160,16 +3171,20 @@ private AstNode parenExpr() throws IOException {
             Comment jsdocNode = getAndResetJsDoc();
             int lineno = ts.lineno;
             int begin = ts.tokenBeg;
-            AstNode e = (peekToken() == Token.RP ? new EmptyExpression(begin) : expr());
+            AstNode e = (peekToken() == Token.RP ? new EmptyExpression(begin) : expr(true));
```
gbrail, line 3174: To my points above, if this is the only place where we pass "true," does it make more sense to add a new function instead?

src/org/mozilla/javascript/Parser.java
gbrail: Right, I see that we already had a stylistic conversation about this that went the other way -- my initial thought would be to do exactly what you did and have:

> private AstNode expr() {
>   return expr(false);
> }
> 
> private AstNode expr(boolean rarelyUsedFlag) {
> // and so on

But you already went through that. Given that we already had a conversation I'm OK with this the way that it is.

PR #1417, opened 2023-11-04
Support hashbang

### LINKED ISSUE #1274: Support ES2023 hashbang grammer
See https://github.com/tc39/proposal-hashbang

p-bakker:
Closes #1274

Replaces #1341

### DIFF
--- a/src/org/mozilla/javascript/TokenStream.java
+++ b/src/org/mozilla/javascript/TokenStream.java
@@ -1115,7 +1115,9 @@ class TokenStream {
                 String str = getStringFromBuffer();
                 this.string = (String) allStrings.intern(str);
                 return Token.STRING;
-            } else if (c == '#'
+            }
+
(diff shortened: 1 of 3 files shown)


src/org/mozilla/javascript/TokenStream.java
```diff
@@ -1115,6 +1115,13 @@ final int getToken() throws IOException {
                 String str = getStringFromBuffer();
                 this.string = (String) allStrings.intern(str);
                 return Token.STRING;
+            } else if (c == '#'
```
rbri, line 1118: i think we can have a plain if here like for all the other cases in the method if the block before was left with a return

rbri:
@p-bakker another minor step forward - great

overall it looks good, only a small comment, see above

gbrail:
Makes sense to me too -- I'll try it out. Thanks!

PR #1418, opened 2023-11-04
Disallow anonymous function declarations

p-bakker:
EcmaScript doesn't allow anonymous function declarations, only function expressions can be anonymous.

Rhino does allow anonymous function declarations, but this PR fixes it.

Test262 doesn't have any test for this (or at least not the test262 version we're currently on), so tests are missing

More importantly though, should this fix be tied to a specific language version? If so, everyone OK with >= ES6?

### DIFF
--- a/src/org/mozilla/javascript/Parser.java
+++ b/src/org/mozilla/javascript/Parser.java
@@ -888,11 +888,13 @@ public class Parser {
             syntheticType = FunctionNode.FUNCTION_EXPRESSION;
         }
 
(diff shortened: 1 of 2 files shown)


rbri:
>= ES6? if fine for my

But i think we need at least some test in our test suite for this

gbrail:
This seems OK to me but it should definitely be gated by a language version (ES6 works for me). I suspect that there are no tests because this is basically invalid syntax that I suspect the test authors didn't expect. We should write a few of our own tests for this just to exercise the code path.

p-bakker:
K, will add the language flag and tests

Issue #1427, opened 2023-12-06
Some OOM problem was found in rhino


Alex111998:
### description
When I test the latest version(1.7.14) of rhino by CIFuzz，OOM security issue was found when use the follow apis,may cause denial of service issues in applications when use unlimited:

### pom

```
<dependency>
        <groupId>org.mozilla</groupId>
        <artifactId>rhino</artifactId>
        <version>1.7.14</version>
</dependency>
```

### code

```
import org.mozilla.javascript.Kit;
import org.mozilla.javascript.ast.DoLoop;
import org.mozilla.javascript.ast.ElementGet;
import org.mozilla.javascript.ast.EmptyExpression;
import org.mozilla.javascript.ast.EmptyStatement;
import org.mozilla.javascript.ast.ForInLoop;
import org.mozilla.javascript.ast.ForLoop;
import org.mozilla.javascript.ast.FunctionCall;
import org.mozilla.javascript.ast.FunctionNode;
import org.mozilla.javascript.ast.GeneratorExpressionLoop;
import org.mozilla.javascript.ast.IfStatement;
import org.mozilla.javascript.ast.InfixExpression;
import org.mozilla.javascript.ast.KeywordLiteral;
import org.mozilla.javascript.ast.Label;
import org.mozilla.javascript.ast.LetNode;
import org.mozilla.javascript.ast.Name;
import org.mozilla.javascript.ast.NewExpression;

public class Rhino_OOM {

    public static void main(String[] args) {
        
        // api 1
        try {
            DoLoop doLoop = new DoLoop();
            String result = doLoop.toSource(Integer.MAX_VALUE);
        } catch (Exception e) {
        }

        // api 2
        try {
            ElementGet elementGet = new ElementGet();
            String result = elementGet.toSource(Integer.MAX_VALUE);
        } catch (Exception e) {
        }

        // api 3
        try {
            EmptyExpression emptyExpression = new EmptyExpression();
            String result = emptyExpression.toSource(Integer.MAX_VALUE);
        } catch (Exception e) {
        }

        // api 4
        try {
            EmptyStatement emptyStatement = new EmptyStatement();
            String result = emptyStatement.toSource(1093157160);
        } catch (Exception e) {
        }

        // api 5
        try {
            ForInLoop forInLoop = new ForInLoop();
            String result = forInLoop.toSource(Integer.MAX_VALUE);
        } catch (Exception e) {
        }

        // api 6
        try {
            ForLoop forLoop = new ForLoop();
[...]

PR #1428, opened 2023-12-09
limit the length of the string to be used for indentation

rbri:
use a cache of precalculated strings instead of generating new ones over and over again

this addresses #1427


gbrail:
Thanks -- makes sense and fixes a potential OOM to boot.

PR #1430, opened 2023-12-27
fix clz32 rounding errors

rbri:
use bit logic to avoid rounding errors  (based on https://github.com/HtmlUnit/htmlunit-rhino-fork/pull/10)

### DIFF
--- a/src/org/mozilla/javascript/NativeMath.java
+++ b/src/org/mozilla/javascript/NativeMath.java
@@ -185,7 +185,33 @@ final class NativeMath extends ScriptableObject {
         if (n == 0) {
             return Double32;
         }
-        return Double.valueOf(31 - Math.floor(Math.log(n >>> 0) * LOG2E));
+
+        int place = 0;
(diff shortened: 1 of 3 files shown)


gbrail:
Do you know if there's a test from test262 or anything that can verify this? For example, there is a "math-clz32.js" file in the V8 codebase, but it seems much different than this. I'm sure that this is right but I'd feel better if we had a reference.

rbri:
> there is a "math-clz32.js" file in the V8 codebase, but it seems much different than this

yes, found the same while preparing this

rbri:
> Do you know if there's a test from test262 or anything that can verify this?

Will have another look but i think i can't spend so much time on this and running all the samples in the browser proof that this is correct.

rbri:
@gbrail looks like 262 has only tests for the border cases of Math functions (see PR #1438). Maybe we have to roll our own.

gbrail:
OK - I think we're in good shape here then. Thanks!

PR #1432, opened 2023-12-28
BigIntLiteral.toSource includes suffix

JohnCokerC3:
I noticed that the `toSource` method for **BigIntLiteral** was losing the `n` suffix, which meant that parsing and toSource turned it into a **Number** instead.


gbrail:
Looks good to me. Thanks!

PR #1433, opened 2024-01-02
Non deterministic behaviour with vararg methods

rPraml:
**Summary**
If there are two methods, one with varargs and an other without, for example
- `String args(String arg1, String... args)`
- `String args(String arg1)`
rhino considers these two methods as `PREFERENCE_EQUAL` when it is invoked with `args('foo')` from javascript. This may result in a non deterministic behaviour as one of the two was taken.

**Details**
In the case described above, all methods are compared and it is tried to find the best matchin one. So after `preferSignature`we will run into this code part in
[NativeJavaMethod](https://github.com/mozilla/rhino/blob/Rhino1_7_14_Release/src/org/mozilla/javascript/NativeJavaMethod.java#L359)
```java
                            if (preference != PREFERENCE_EQUAL) Kit.codeBug();
                            // This should not happen in theory
                            // but on some JVMs, Class.getMethods will return all
                            // static methods of the class hierarchy, even if
                            // a derived class's parameters match exactly.
                            // We want to call the derived class's method.
                            if (bestFit.isStatic()
                                    && bestFit.getDeclaringClass()
                                            .isAssignableFrom(member.getDeclaringClass())) {
                                // ... 
                            } else {
                                if (debug)
                                    printDebug("Ignoring same signature member ", member, args); // CASE 2
                            }
```

In our application, where the bug occured, we run in the `Ignoring same signature member` code path and one (not really deterministic - because of HashMap) method was taken. So the code sometimes works and sometimes not.

It was also not easy to provide a deterministic failing test, that's why (and maybe to add more determinism) I've changed the map in `discoverAccessibleMethods` to a LinkedHashMap (and yet the test depends heavily on the order of the JVM's `getDeclaredMethods` implementation)

**Fix**
When there are two candidates, the no-vararg method is taken. This is the same behaviour, as it is in Java.

### DIFF
--- a/src/org/mozilla/javascript/JavaMembers.java
+++ b/src/org/mozilla/javascript/JavaMembers.java
@@ -21,6 +21,7 @@ import java.security.Permission;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 
@@ -308,7 +309,8 @@ class JavaMembers {
      */
     private Method[] discoverAccessibleMethods(
             Class<?> clazz, boolean includeProtected, boolean includePrivate) {
-        Map<MethodSignature, Method> map = new HashMap<>();
+        Map<MethodSignature, Method> map =
+                new LinkedHashMap<>(); // use linked hash map for deterministic discovery
         discoverAccessibleMethods(clazz, map, includeProtected, includePrivate);
         return map.values().toArray(new Method[0]);
     }
--- a/src/org/mozilla/javascript/NativeJavaMethod.java
+++ b/src/org/mozilla/javascript/NativeJavaMethod.java
@@ -496,6 +496,15 @@ public class NativeJavaMethod extends BaseFunction {
                 break;
             }
         }
+        if (totalPreference == PREFERENCE_EQUAL && vararg1 != vararg2) {
+            // It could happen that we have found two methods, that may fit
+            // In this case, we will take the no-vararg one, if possible
+            if (vararg1) {
+                totalPreference = PREFERENCE_SECOND_ARG;
+            } else {
+                totalPreference = PREFERENCE_FIRST_ARG;
+            }
+        }
         return totalPreference;
     }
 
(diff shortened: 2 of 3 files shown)


gbrail:
I think that this change makes sense and fixes a bug, and I'm fine with using LinkedHashMap here so that it's deterministic. However, I don't currently work on anything that uses all of the Java reflection stuff in Rhino, and I'd love an opinion by another regular user who works with Rhino this way to make sure that we're not missing anything.

Thanks!

rPraml:
@gbrail I've resolved merge commits here

p-bakker:
Never ran into this, but LGTM

@andreabergia @tonygermano care to comment on this one as well, given that both of you seem to use Java Interop in your Rhino integrations

andreabergia:
Looks good to me!

tonygermano:
I haven't reviewed the code, but the stated fix of making the behavior match what java does seems correct to me.

tonygermano:
Here's some information about how java resolves overloads. https://stackoverflow.com/a/27880923

tonygermano:
Looking at the java spec, I think it should be the very first check in the `preferSignature` method where if `vararg1 != vararg2` to automatically prefer the one that does not have variable arity.

This does 2 things,
1. It short circuits the function and skips a lot of other tests that end up not being required.
2. Without putting too much thought into it, I suspect there are cases that could result in `PREFERENCE_AMBIGUOUS` where one has variable arity and the other does not. This should still prefer the method which does not have variable arity.

gbrail:
Can you take a look at the comments by @tonygermano and let us know what you think? Thanks!

rPraml:
@tonygermano thanks for feedback

> I think it should be the very first check in the preferSignature method where if vararg1 != vararg2

I'm unsure. If I do the check at top of the method, I will always prefer `arg(String, String)` before `arg(String, int...)`.
in rhino `myObj.arg('foo', 1)` will call the wrong method.

Thanks for the link to java spec. I've extended the test cases, but I think there is still one failing case. when an array is passed

p-bakker:
Converted to draft, as it currently doesn't pass the CI build

PR #1434, opened 2024-01-05
Various fixes and implementations to NativeRegExp

rbri:
Slightly modified (improved) PR from @duonglaiquang. See https://github.com/HtmlUnit/htmlunit-rhino-fork/pull/12

----

### This PR does the following:

1. Handle the case where [RegExp.prototype.toString](https://tc39.es/ecma262/multipage/text-processing.html#sec-regexp.prototype.tostring) is called on `NativeObject` instead of `NativeRegExp` object (currently throwing `TypeError: Method "toString" called on incompatible object`)
  - These tests showcase the expected behavior of this method
    ```html
    <!DOCTYPE html>
    <html>
    <head>
    <script>
    console.log(RegExp.prototype.toString.call({})) // "/undefined/undefined"
    console.log(RegExp.prototype.toString.call({source: "Foo"})) // "/Foo/undefined"
    console.log(RegExp.prototype.toString.call({flags: "gy"})) // "/undefined/gy"
    console.log(RegExp.prototype.toString.call({source: "Foo", flags: "g"})) // "/Foo/g"
    console.log(RegExp.prototype.toString.call({source: "Foo", flags: "g", sticky: true})) // "/Foo/g"
    </script>
    </head>
    <body>
    </body>
    </html>
    ```
 
2. Implement [RegExp.dotAll](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/dotAll) flag

3. Implement [String.prototype.replaceAll](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replaceAll)

4. Fix an issue when `String.includes/startsWith/endsWith` throw `TypeError` when the first argument is a regex, even if [Symbol.match](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/match) of that regex has been set to false.

    > This function is also used to identify [if objects have the behavior of regular expressions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp#special_handling_for_regexes). For example, the methods [String.prototype.startsWith()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith), [String.prototype.endsWith()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/endsWith) and [String.prototype.includes()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes), check if their first argument is a regular expression and will throw a [TypeError](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypeError) if they are. Now, if the match symbol is set to false (or a [Falsy](https://developer.mozilla.org/en-US/docs/Glossary/Falsy) value except undefined), it indicates that the object is not intended to be used as a regular expression object.

  - These tests showcase the buggy behavior of the current implementation.
    ```html
    <!DOCTYPE html>
    <html>
    <head>
    <script>
    var regExp = /./;
    try {
      console.log("/./".includes(regExp))
    } catch (e) {
      console.log(e); // TypeError: First argument to String.prototype.includes must not be a regular expression
      regExp[Symbol.match] = false;
      console.log("/./".includes(regExp)) // expected: true, got TypeError
    }
    </script>
    </head>
    <body>
    </body>
    </html>
    ```

### DIFF
--- a/src/org/mozilla/javascript/NativeString.java
+++ b/src/org/mozilla/javascript/NativeString.java
@@ -469,10 +469,15 @@ final class NativeString extends IdScriptableObject {
                     String thisString =
                             ScriptRuntime.toString(requireObjectCoercible(cx, thisObj, f));
                     if (args.length > 0 && args[0] instanceof NativeRegExp) {
-                        throw ScriptRuntime.typeErrorById(
-                                "msg.first.arg.not.regexp",
-                                String.class.getSimpleName(),
-                                f.getFunctionName());
+                        if (ScriptableObject.isTrue(
+                                ScriptableObject.getProperty(
+                                        ScriptableObject.ensureScriptable(args[0]),
+                                        SymbolKey.MATCH))) {
+                            throw ScriptRuntime.typeErrorById(
+                                    "msg.first.arg.not.regexp",
+                                    String.class.getSimpleName(),
+                                    f.getFunctionName());
+                        }
                     }
 
                     int idx = js_indexOf(id, thisString, args);
--- a/src/org/mozilla/javascript/NativeString.java
+++ b/src/org/mozilla/javascript/NativeString.java
@@ -98,6 +98,7 @@ final class NativeString extends IdScriptableObject {
         addIdFunctionProperty(ctor, STRING_TAG, ConstructorId_match, "match", 2);
         addIdFunctionProperty(ctor, STRING_TAG, ConstructorId_search, "search", 2);
         addIdFunctionProperty(ctor, STRING_TAG, ConstructorId_replace, "replace", 2);
+        addIdFunctionProperty(ctor, STRING_TAG, ConstructorId_replaceAll, "replaceAll", 2);
         addIdFunctionProperty(ctor, STRING_TAG, ConstructorId_localeCompare, "localeCompare", 2);
         addIdFunctionProperty(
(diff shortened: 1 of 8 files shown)


gbrail:
Thanks for contributing this -- it's always great to make more tests pass.

Before we finish and merge this, is there a way to set up this PR so that the original author gets some credit? For example, could we prevail on him to push his PR to this repo in addition to HtmlUnit?

rbri:
> Before we finish and merge this, is there a way to set up this PR so that the original author gets some credit? For example, could we prevail on him to push his PR to this repo in addition to HtmlUnit?

@duonglaiquang any idea about this?

atnak:
@gbrail Thank you for your concern about the credit thing. I work with @duonglaiquang and I will respond.

We use HtmlUnit (and indirectly rhino) in our business operations and we periodically push fixes upstream because it also benefits us.  We're not in it for credit and these changes are not always created solely by the person pushing. @rbri doing these PRs for us is cool because it saves us a lot of time and dedication fixing code and test cases to meet rhino's / htmlunit's styles.

If credit is the only concern, we're cool without 👍

gbrail:
I get it about the credit -- thank you for your contribution!

PR #1435, opened 2024-01-10
Fixes for Symbol.iterator handling in NativeArray

rbri:
* Array.prototype[Symbol.iterator] points to the 'values' functions
* arguments[Symbol.iterator] points to Array.prototype[Symbol.iterator] (see https://github.com/HtmlUnit/htmlunit-rhino-fork/pull/14)

### DIFF
--- a/src/org/mozilla/javascript/Arguments.java
+++ b/src/org/mozilla/javascript/Arguments.java
@@ -6,8 +6,6 @@
 
 package org.mozilla.javascript;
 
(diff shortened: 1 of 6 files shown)


gbrail:
I have to say that it didn't look right to me at first, but I checked the spec and my browser and it sure is. So thanks!

rbri:
> I have to say that it didn't look right to me at first, but I checked the spec and my browser and it sure is.

:-D same here; JS is fun

Issue #1189, opened 2022-02-23
regexp hangs forever


dvega:
This code hangs forever (or is incredible slow). Tested in Rhino 1.7.14

```js
var aregex = /^[A-Z]+\.([A-Z]+)*$/;
aregex.test('HOTMAIL.COMMARIAJOSEPANCHOMORMARIAJOSEPANCHOMORA@HOTMAIL.COM')
```

Also the regexp code is non-interruptible. Please add `if (Thread.interrupted()) ...` to the regexp evaluation loop

p-bakker:
Hi, 

Its unfortunate that you ran into this. Tested this in Chrome and also quite slow there. Think this is just a limitation of many regex implementations out there, see https://v8.dev/blog/non-backtracking-regexp for an explanation of the problem/challenge.

As for your request for adding the `if (Thread.interrupted())` call: if you can provide a PR for it, it can be looked at

tuchida:
https://github.com/makenowjust-labs/recheck
I have not tried it, but it seems that time-consuming RegExp can be detected by such a library.

blutorange:
In general, JavaScript regexp are powerful enough that there will always be some that are very slow, even in browsers. For example:

```
regexp: (.*){1,32000}[bc]
input: aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
```

(If you try this in Chrome dev tools, be prepared to kill the browser tab)

We are currently still using Rhino to run JavaScript regexps, in order to perform server validation of some client side browser code.

Until now we started a separate thread and used the deprecated `Thread#stop` method to kill the thread if it did not complete within a certain timeout (as that is the only way to kill unresponsive threads).

Somewhere between Java 17 and 21, `Thead#stop` was changed not to work anymore and just throw a `UnsupportedOperationException`.

As such, a way to interrupt long running regexp would definitely be great. I'll try and clone this repo locally to see if the check for Thread.interrupted would work.

PR #1440, opened 2024-01-17
Make regexp execution loop interruptible #1189

blutorange:
#1189

Uses `Thread.currentThread.isInterrupted()` so that the interruption flag remains set to true, we only terminate the RegExp evaluation loop, but other (potentially third-party calling) code may still have to check for the interrupted flag to stop its execution as well.

I also added a test with a long-running regexp that fails without the interrupt check.

Perhaps it would have been better to throw an `InterruptedException`, but that would require changing the signature of many methods.

### DIFF
--- a/src/org/mozilla/javascript/regexp/NativeRegExp.java
+++ b/src/org/mozilla/javascript/regexp/NativeRegExp.java
@@ -1878,6 +1878,9 @@ public class NativeRegExp extends IdScriptableObject {
 
         for (; ; ) {
 
+            if (Thread.currentThread().isInterrupted()) {
+                throw new RuntimeException("Received interrupt while executing regexp");
+            }
             if (reopIsSimple(op)) {
                 int match = simpleMatch(gData, input, op, program, pc, end, true);
                 result = match >= 0;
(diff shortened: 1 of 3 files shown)


gbrail:
Can I suggest another way to implement this that could help solve your problem, but do it without adding code for people who don't need that extra check?

The Context class has a methods called "setInstructionObserverThreshold." When set, the interpreter or the generated bytecode will call a method on the ContextFactory (which you implement using inheritance) whenever N instructions are executed. You can have this method be implemented to check for thread interruption if that's what works best for you.

The thing you'd do in this case, then, would be to add a call to ScriptRuntime.addInstructionCount on every iteration of that top-level loop in the regex. (I'd suggest just calling each trip through that loop as, I don't know, let's say 10 "instructions," but it could be a different value.)

I think this would be more general than just checking for interruption in that loop, and means that you can use the same mechanism to check for long-running regexps as what we already have for checking for long-running code in general.

blutorange:
`ScriptRuntime.addInstructionCount` works just as well, and yeah, that seems like a solution that integrates better into the existing mechanism. I used `5` for the instruction count. A regexp like `/(.*){1,32000[bc]/.test("aaaaaaaaaaaaaaaaaaaaaaa");` takes a split second in browsers to execute (and a few seconds in rhino) and is enough to reach a quota of `Integer.MAX_VALUE`

gbrail:
Thanks for that -- it seems to be fine!

PR #1448, opened 2024-02-21
(Partial) implementation of `[Symbol.species]`

andreabergia:
This PR does:

- add `[Symbol.species]` to various built-in objects, as required by the spec:
  - `Set` (spec [24.2.2.2](https://tc39.es/ecma262/multipage/control-abstraction-objects.html#sec-get-promise-@@species))
  - `Map` (spec [24.1.2.3](https://tc39.es/ecma262/multipage/control-abstraction-objects.html#sec-get-promise-@@species))
  - `Regex` (spec [22.2.5.2](https://tc39.es/ecma262/multipage/control-abstraction-objects.html#sec-get-promise-@@species))
  - `Array` (spec [23.1.2.5](https://tc39.es/ecma262/multipage/control-abstraction-objects.html#sec-get-promise-@@species))
- fixes a bug in the existing implementation for `Promise` (spec [27.2.4.9](https://tc39.es/ecma262/multipage/control-abstraction-objects.html#sec-get-promise-@@species))
- changes the test262 runner so that it does not consider `Symbol.species` an ignored feature; this in turn increases the feature coverage because some more tests pass.

Notice that this PR does _not_ implement the behavior of the symbol (spec [7.3.22](https://tc39.es/ecma262/multipage/abstract-operations.html#sec-speciesconstructor)) - it just adds the symbol to some built-ins.

### DIFF
--- a/src/org/mozilla/javascript/NativeArray.java
+++ b/src/org/mozilla/javascript/NativeArray.java
@@ -46,9 +46,10 @@ public class NativeArray extends IdScriptableObject implements List {
     private static final Object ARRAY_TAG = "Array";
     private static final Long NEGATIVE_ONE = Long.valueOf(-1);
 
-    static void init(Scriptable scope, boolean sealed) {
+    static void init(Context cx, Scriptable scope, boolean sealed) {
         NativeArray obj = new NativeArray(0);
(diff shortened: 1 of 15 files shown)


rbri:
no complains, and because this is a step forward i will merge this soon

PR #1449, opened 2024-02-24
add @Override and some try-with-resources

rbri:


rbri:
i think this does not really harm and because there was no complaints by others i will merge this

PR #1453, opened 2024-02-28
When InterfaceAdapter is used, the wrong thisObj is used

rPraml:
I had an issue when I pass a JS object to Java class and an InterfaceAdapter is constructed:

Java interface
```java
public interface AdapterInterface {
        int m1(int i);
        int m2();
}
```

Corresponding impl. in javascript:
```javascript
function Obj() { this.myObj = {one: 1} }
Obj.prototype.m1 = function(i) { return i + this.myObj.one } // this.myObj is undefined
Obj.prototype.m2 = function() { return 7 }
```

When I pass this object in my AdapterClass, I get the error: `this.myObj is undefined` 
This PR provides a test and a possible fix

### DIFF
--- a/src/org/mozilla/javascript/InterfaceAdapter.java
+++ b/src/org/mozilla/javascript/InterfaceAdapter.java
@@ -152,9 +152,8 @@ public class InterfaceAdapter {
                 }
             }
         }
-        Scriptable thisObj = wf.wrapAsJavaObject(cx, topScope, thisObject, null);
 
-        Object result = function.call(cx, topScope, thisObj, args);
+        Object result = function.call(cx, topScope, (Scriptable) target, args);
         Class<?> javaResultType = method.getReturnType();
         if (javaResultType == Void.TYPE) {
             result = null;
--- a/testsrc/org/mozilla/javascript/tests/JavaAdapterInvokeTest.java
+++ b/testsrc/org/mozilla/javascript/tests/JavaAdapterInvokeTest.java
@@ -1,0 +1,111 @@
+package org.mozilla.javascript.tests;
+
+import org.junit.After;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+import org.mozilla.javascript.Context;
+import org.mozilla.javascript.Scriptable;
+
+public class JavaAdapterInvokeTest {
+    Context cx = null;
+    Scriptable topScope = null;
+
+    @Before
+    public void enterContext() {
+        cx = Context.enter();
+        cx.setOptimizationLevel(-1);
+        topScope = cx.initStandardObjects();
+    }
+
+    @After
+    public void exitContext() {
+        Context.exit();
+    }
+
+    public interface AdapterInterface {
+        int m1(int i);
+
+        int m2();
+    }
+
+    public static class AdapterClass {
+        private AdapterInterface adapter;
+
+        public AdapterClass(AdapterInterface adapter) {
+            this.adapter = adapter;
+        }
+
+        public int doIt(int i) {
+            return this.adapter.m1(i) + this.adapter.m2();
+        }
+    }
+
+    @Test
+    public void testInvoke() throws NoSuchMethodException {
+        String testCode =
+                "'use strict'\n"
+                        + "var impl = {"
+                        + "  m1: function(i) { return i + 1 },\n"
+                        + "  m2: function() { return 7 }\n"
+                        + "}\n"
+                        + "adapter = new Packages."
+                        + AdapterClass.class.getName()
+                        + "(impl)\n"
+                        + "adapter.doIt(42)";
+
+        Number result = (Number) cx.evaluateString(topScope, testCode, "", 1, null);
+        Assert.assertEquals(50, result.intValue());
+    }
+
+    @Test
+    public void testInvokeWithPrototype() throws NoSuchMethodException {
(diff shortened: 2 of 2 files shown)


src/org/mozilla/javascript/InterfaceAdapter.java
```diff
@@ -152,9 +152,8 @@ Object invokeImpl(
                 }
             }
         }
-        Scriptable thisObj = wf.wrapAsJavaObject(cx, topScope, thisObject, null);
```
rPraml, line 155: `thisObject` is an instance of `JavaAdapterInvokeTest.AdapterClass` and we want to call a method on `target`

testsrc/org/mozilla/javascript/tests/JavaAdapterInvokeTest.java
```diff
@@ -0,0 +1,111 @@
...
+import org.mozilla.javascript.Scriptable;
+
+public class JavaAdapterInvokeTest {
+    Context cx = null;
+    Scriptable topScope = null;
+
+    @Before
+    public void enterContext() {
+        cx = Context.enter();
+        cx.setOptimizationLevel(-1);
```
p-bakker, line 17: Dunno what our policy is, but should this test not also be run in non-interpreted mode?

testsrc/org/mozilla/javascript/tests/JavaAdapterInvokeTest.java
```diff
@@ -0,0 +1,111 @@
...
+
+        Number result = (Number) cx.evaluateString(topScope, testCode, "", 1, null);
+        Assert.assertEquals(50, result.intValue());
+    }
+
+    @Test
+    public void testInvokeJsOnly() throws NoSuchMethodException {
+        String testCode =
+                "'use strict'\n"
+                        + "function Obj() { this.myObj = {one: 1} }\n"
```
p-bakker, line 99: What about a testcase where .myObj is added on the prototype, instead of the `this` inside the constructor?

p-bakker:
Not a regular user of code that uses code constructs like the testcases, calling a java method passing as argument a JavaScript objects that implements a java interface, but the change looks ok to me

gbrail:
I'd love to see a few of the questions answered on the PR above and then see if it makes sense to merge this.

testsrc/org/mozilla/javascript/tests/JavaAdapterInvokeTest.java
rPraml: Didn't add a separate testcase, but added a second property "seven" on the prototype.
(So testcase will check both code paths at once - but FYI: the second path will also fail without the fix)

testsrc/org/mozilla/javascript/tests/JavaAdapterInvokeTest.java
rPraml: Usung `Utils.runWithAllOptimizationLevels` now

tests/src/test/java/org/mozilla/javascript/tests/JavaAdapterInvokeTest.java
```diff
@@ -0,0 +1,128 @@
...
+     * test fails before #1453
+     */
+    @Test
+    public void testInvokeWithPrototypeAndObjs() {
+        String testCode =
+                "'use strict'\n"
+                        + "function Obj() { this.myObj = {one: 1} }\n"
+                        + "Obj.prototype.seven = 7\n"
+                        + "Obj.prototype.m1 = function(i) { return i + this.myObj.one }\n"
+                        + "Obj.prototype.m2 = function() { return this.seven }\n"
```
rPraml, line 89: This test shows the problem. `this.myObj` and `this.seven` were undefined, when the js object is passed in an `InterfaceAdapter`
The other tests are all passing without the fix and are for completeness, not to break anything else

rPraml:
@gbrail I resolved the merge conflicts and updated the test with some documentation

rPraml:
Hmm. Seems that there might be a flaky test
```
ozillaSuiteTest > [3846, js=testsrc/tests/js1_5/Regress/regress-211590.js, opt=0] > runMozillaTest[3846, js=testsrc/tests/js1_5/Regress/regress-211590.js, opt=0] STANDARD_OUT
    In "testsrc/tests/js1_5/Regress/regress-211590.js":
     FAILED! Math.random should be random : Expected value 'between 48% and 52%', Actual value ' is 47.990' 
     FAILED! Math.random should be random : Expected value 'between 48% and 52%', Actual value ' is 47.990'
```
Will investigate later

p-bakker:
Converted this to a draft as it currently requires more work as the pipeline fails

Please remove the draft status once the pipeline passed and the PR can be considered for merging,.

This way it allows the committers to focus on ready-to-merge PRs

rPraml:
> Not a regular user of code that uses code constructs like the testcases, calling a java method passing as argument a JavaScript objects that implements a java interface, but the change looks ok to me

I've added a more realistic testcase where i use `var runnable = new MyRunnable()` and pass this object as constuctor argument for `java.lang.Thread`: 
https://github.com/mozilla/rhino/pull/1453/files#diff-05cf7412910fd42af1934af105265073d54b9fd36c8d433d53b3e51a4fae8650R106
(without the fix, the test fails)

PR #1456, opened 2024-03-13
NullPointerException fix when calling JS methods exposed as JavaProxys - JSR223

EduFrazao:
With a JavaScript like that
```javascript
    function run() {
         print('Some string');
    }
```

If I try to invoke it from a JavaProxy using this code, I got a NullPointerException
```java
   final ScriptEngine engine = engineManager.getEngineByName("rhino");
   engine.eval(myScript);
   Invocable inv = (Invocable) engine;
   Runnable runnable = inv.getInterface(Runnable.class);
   runnable.run();
```


rbri:
@EduFrazao can you please add an unit test to your PR

p-bakker:
Converted to draft as it currently is missing a testcase

PR #1460, opened 2024-03-29
an optimized version of Context#close()

rbri:
* implement an optimized version of Context#close()
* use try-with-resources for the context at more places


rbri:
A really small optimization that avoids at least one VM bride call and cleans up the code.
I got no complains for this, will merge this now.

PR #1462, opened 2024-04-04
fix: toString() on generator

0xe:
`toString()` on a generator misses the `*`. This change adds that to the encodedSource.

For example,
```javascript
  function* f() {
    yield 1;
  }; f.toString();
```

returns,

Before:
```
  function f() {
    yield 1;
  }
```

After:
```javascript
  function* f() {
    yield 1;
  }
```


rbri:
@0xe thanks for your PR

PR #1464, opened 2024-04-18
Fixed handling of unicode characters in the lexer

andreabergia:
The lexer was not handling correctly surrogate pairs. References:
- https://tc39.es/ecma262/#prod-IdentifierName
- https://en.wikipedia.org/wiki/UTF-16#Code_points_from_U+010000_to_U+10FFFF

### DIFF
--- a/src/org/mozilla/javascript/TokenStream.java
+++ b/src/org/mozilla/javascript/TokenStream.java
@@ -685,7 +685,7 @@ class TokenStream {
                     c = '\\';
                 }
             } else {
-                identifierStart = Character.isJavaIdentifierStart((char) c);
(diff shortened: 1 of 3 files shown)


rbri:
@andreabergia looks promising, can you please have a look at the failing tests...

andreabergia:
The problem is that the behavior of the Java unicode checks differs depending on the JDK - so the same code might say that a character is a valid unicode identifier on java 11 but not on java 8. Some tests of test262 are failing because of that, it seems.
I had originally run locally the test262 updater with JDK11. I've rerun it with JDK 1.8 and it looks like CI passes now. 🙂

rbri:
Thanks a lot @andreabergia, another step forward.

PR #1469, opened 2024-04-30
Fix unhandled promise rejection handler after a `.then`

### LINKED ISSUE #1461: Promise behaviour
Hi,

I start on my Linux console following:
```
java -jar rhino-1.7.14.jar -version 200
```
And redirect `console.log` to `print` so that the same code can be executed in firefox debugger and on rhino console:
```js
var console = { log: function(text) { print(text); } };
```
On Rhino I get this result for a simple Promise reject:
```javascript
new Promise((resolve, reject) => {console.log("Status: running"); reject("Failed"); }).
  then(function(result) { console.log(`Status: then(${result}) called`); }).
  catch(function(error) { console.log(`Status: catch(${error}) called`); });
  
Status: running
Status: catch(Failed) called
Unhandled rejected promise: Failed
[object Promise]
```

In firefox debugger (and chromium debugger) this result:
```javascript
new Promise((resolve, reject) => {console.log("Status: running"); reject("Failed"); }).
  then(function(result) { console.log(`Status: then(${result}) called`); }).
  catch(function(error) { console.log(`Status: catch(${error}) called`); });

Status: running [debugger eval code:1:43](chrome://devtools/content/webconsole/debugger%20eval%20code)
Status: catch(Failed) called [debugger eval code:3:35](chrome://devtools/content/webconsole/debugger%20eval%20code)
Promise { <state>: "fulfilled", <value>: undefined }
```

Rhino seems to not properly respect my `catch()`

Another bug is:
When I add a `setTimeout()` the `then()` is never called:

```
new Promise(resolve => {console.log("Status: running"); setTimeout(() => {resolve("Done");}, 2000); }).
  then(function(result) { console.log(`Status: then(${result}) called`); }).
  catch(function(error) { console.log(`Status: catch(${error}) called`) });
```

andreabergia:
If we have a promise `p1` that is rejected, and invoke `then` on it, we create a new promise `p2` (as it should be, per the spec) and mark `p1` as handled. However, if we call `.catch` on `p2`, we are _not_ marking `p1` as handled correctly since its status is "pending" and not "rejected". This patch fixes it and adds some tests.

Fixes #1461

### DIFF
--- a/src/org/mozilla/javascript/NativePromise.java
+++ b/src/org/mozilla/javascript/NativePromise.java
@@ -306,13 +306,17 @@ public class NativePromise extends ScriptableObject {
             cx.enqueueMicrotask(() -> fulfillReaction.invoke(cx, scope, result));
         } else {
             assert (state == State.REJECTED);
-            if (!handled) {
-                cx.getUnhandledPromiseTracker().promiseHandled(this);
-            }
+            markHandled(cx);
             cx.enqueueMicrotask(() -> rejectReaction.invoke(cx, scope, result));
         }
-        handled = true;
         return capability.promise;
+    }
+
+    private void markHandled(Context cx) {
+        if (!handled) {
+            cx.getUnhandledPromiseTracker().promiseHandled(this);
+            handled = true;
+        }
     }
 
     // Promise.prototype.catch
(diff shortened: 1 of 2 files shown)


rbri:
LGTM thanks a lot for this contribution

gbrail:
Thanks -- I agree that it looks good!

Issue #1465, opened 2024-04-19
Is Java 17 a requirement for building the library? 


why-not-try-calmer:
_updated_: I am now able to build using openjdk17. Is it the latest openjdk version supported for building the library? If so do you mind I PR this information to the README?

_deprecated_:
I am unable to build with
```
./gradlew jar
```
with
```
openjdk 19.0.2 2023-01-17
OpenJDK Runtime Environment (build 19.0.2+7-nixos)
OpenJDK 64-Bit Server VM (build 19.0.2+7-nixos, mixed mode, sharing)
```
Here is the log:
[error.txt](https://github.com/mozilla/rhino/files/15041232/error.txt)

Any hint warmly appreciated!

p-bakker:
It is not, Java 11 is the minimum

why-not-try-calmer:
Sure, sorry I forgot to close, I found the issue a long time ago

Issue #1339, opened 2023-06-08
Upgrade the minimum Java version


gbrail:
I have in the past been adamant that we maintain Java 8 compatibility in this project, but with all the changes going into Java today, that's getting increasingly difficult -- right now it's a lot of work to have a project that builds with both Java 8 and Java 20, for example. Large organizations seem to be deploying Java 17 pretty broadly, for example, so I think that policies that were holding people on Java 8 may be gone by now.

Given all that, what Java version does the community think that we should be targetting? I could see easily going up to Java 11, for example, but there are additional advantages to supporting even newer versions.

p-bakker:
Better move this to discussions? We already have somewhat of a discussion going there: https://github.com/mozilla/rhino/discussions/1169

p-bakker:
As of the nether of #1479 the minimum Java version for Rhino is 11

Issue #1075, opened 2021-10-22
Switch to Java 9 for compiling Rhino, while generating Java 8 compatible bytecode


p-bakker:
As mentioned by @gbrail in for example https://github.com/mozilla/rhino/pull/1040#issuecomment-943741626

p-bakker:
Also see this note: https://github.com/mozilla/rhino/issues/1149#issuecomment-1007682080

p-bakker:
As of today, with the merge of #1479, the minimum for Rhino is Java 11

PR #1479, opened 2024-05-24
Reorganize the entire source code tree into modern Java modules

gbrail:
This PR reorganizes the source tree into proper Java modules. We end up with the following modules:

* rhino-runtime: The core runtime, and all you need to run JavaScript
* rhino-tools: Adds the shell, debugger, and "global" object with its many utilities
* rhino-xml: Adds XML support
* rhino-engine: Adds the ScriptEngine implementation
* rhino: A "shadow JAR" that includes all the following for backward compatibility
* tests: Our existing test infrastructure, with all the legacy and test262 tests
* examples
* benchmarks

This PR moves literally every source file. Fortunately, git is pretty good at this -- I have been merging new changes for over a year and rarely get confused. However, if you are working on a large PR and adding many new files you will have to pay attention.

Build instructions:
* "./gradlew check" works just like it always did
* "./gradlew shadowJar" will build the backward-compatible jar in "./rhino/build/libs"
* "./gradlew jmh" runs ALL the benchmarks and you can edit build.gradle in the benchmarks directory for less

Benefits:
* Works natively with modular Java
* Following modern Gradle conventions
* Builds are more parallel
 
Side effects:
* Java 11 is now the minimum supported Java release
* I can't figure out how to make "spotless" (literally the Google formatter) work with different JVMs, so it only runs on 11
* We will continue to build using Java 11 but will test with newer versions

Opportunities:
* We could move unit tests that depend on only one module out of the "tests" hierarchy into individual modules

Notes:

Please try this out! I think it's a big step forward for maintainability. thanks!


rbri:
@gbrail thanks for all the work

My hope is still that we start a branch/fork that cuts all the old stuff to be able to move forward to get closer the real world. And therefor i think this can be a good starting point for that...

Outside of this i have no real experiences, how much work it will be to migrate the htmlunit-corejs project to stay in sync with this change. So i trust you...

p-bakker:
@gbrail From my perspective, moving to a layout that supports Java modules is a good thing. I'm not that well-versed in what that technically entails, so if you say these are the required changes, I'll take your word for it :-)

Does this PR fully resolve #1092?
Does this PR move the needle on any of theres open issues? #1465, #1339, #1277, #1120, #1269, #1076 or #1075

As as to @rbri's suggestion to use this as a branch/fork cut-off point that cuts all the old stuff to be able to move forward: imho opinion this restructuring doesn't require a new branch/fork. What's will be the impact for RHino users/embedders?

p-bakker:
> What's will be the impact for Rhino users/embedders?

Was reading up on some of the previous discussions about a v2 release that would break backwards compatibility in an effort to shed some historical ballast, which would then allow us to move forward faster 

@gbrail could you shed some light on if his PR contains any 'breaking' changes for users/embedders?

From the description you have, I only get that:
- the minimum Java version will now be Java 11
- the structure of the repo changes

Is there anything more that users/embedders will notice from this change?

gbrail:
Sorry, a few more answers to the questions:

Regarding incompatibility -- this PR introduces four distinct and "correct" Java modules that replace "rhino.jar". Projects using "rhino.jar" along with one of these four modules, like "rhino-engine", will likely see errors. So, projects that use Java modules elsewhere will want to stop using "rhino.jar" and use the individual JARs instead. The "rhino.jar" is still around so that people not using modular Java features don't necessarily have to change their build.

Yes, this PR will address https://github.com/mozilla/rhino/issues/1092. I chose not to put fully-qualified package names in the top of the directory tree, however, which means that I'm not sure how to create multi-module Javadoc. Creating a Javadoc for each module is pretty easy though, and I'm sure I can figure it out, and the pathnames are long enough without all that extra stuff IMO.

Other issues:

* https://github.com/mozilla/rhino/issues/1465 Yes, you will be able to build with Java 11, 17, and 21.
* https://github.com/mozilla/rhino/issues/1339 Yes, it upgrades the minimum Java version to 11.
* https://github.com/mozilla/rhino/issues/1277 We CAN add an OSGi manifest, but this PR introduces proper Java modules that meet all the rules of Java modularity. If people also want OSGi we can add it. Is that still necessary if we add modules?
* https://github.com/mozilla/rhino/issues/1120 There's no reason why we need Built-Date in the JAR. This PR doesn't necessarily introduce that so that it should do what this issue asks.
* https://github.com/mozilla/rhino/issues/1269 Once we merge this, if we want to use Java 11-specific APIs to better handle reflection, then we can do that
* https://github.com/mozilla/rhino/issues/1076 Once we make this Java 11 minimum, do we still need "VMBridge" at all? Perhaps we can just remove it.
* https://github.com/mozilla/rhino/issues/1075 I think this does what the issue says by reorganizing the source code so that the module layout and the source code layout are the same
 
Regarding Android compatibility -- I don't know. In the past the problem has been that we use APIs that are not present in some versions of Android. Without someone who works with Android regularly contributing to this project, I don't know to ensure that we know what this might affect. However, given that I believe this PR is still helpful in other ways, I think we should go ahead.

p-bakker:
@gbrail thnx for the response. 

Sounds like while embedders might have to make a few adjustments how to include phone in their project, it'll be fairly minor. 

So I'd day go for it!. If moving to Java 11 means Rhino won't be compatible with older Android version, I think that's just what it is. We can't let backwards compatibility with old platforms hold us back.

gbrail:
Closes https://github.com/mozilla/rhino/issues/1465.
Closes https://github.com/mozilla/rhino/issues/1339.
Closes https://github.com/mozilla/rhino/issues/1075.

p-bakker:
Nice 👍

p-bakker:
Don't get this part: `However, not all tools work with Java 11, such as "spotless", so Java 11 is required for regular developers.`

If the tools don't work on Java 11, why if Java 11 required for regular developers? 

Do you meant so say that Java 11 is the minimum Java version for users of Rhino, whereas developers working on developing Rhino itself require at least Java ??? To be able to use all the required tooling, like ???

p-bakker:
@gbrail Can it be that there's something wrong in the module setup? 

Test262SuiteTest.java for me gives an error: `The type org.mozilla.javascript.tools.SourceReader is not accessible` 

I see that class isn't exported in module org.mozilla.rhino.tools

p-bakker:
Also: I managed to get 2 folders auto-generated: `buildGradle` and `rhino.bin.test`.

As of yet I'm not sure how they came to be, but if those are part of the building/testing setup, should they maybe be added to .gitignore?

p-bakker:
@gbrail Any thoughts on the issues I'm having in tests with the non-accessibility of org.mozilla.javascript.tools.SourceReader and  org.mozilla.javascript.json.JsonParser

As build on GitLab don't fail, I'm sure it must be something in my setup, but I'm clueless as to what it might be...

gbrail:
Yes, it seems like IDEs are more sensitive to things than gradle and
actually the problems are real problems.
https://github.com/mozilla/rhino/pull/1505 fixes them!

p-bakker:
@gbrail #1505 indeed solves the compilation issues for me, many tnx!

PR #1483, opened 2024-06-06
simplify test code

rbri:


gbrail:
Hey -- this looks good, but it will need to be merged one way or another with https://github.com/mozilla/rhino/pull/1479, which is the PR that reorganizes the codebase -- as part of that, I moved some of the test cases so that some can be in the modules that they are associated with, rather than having all the tests in a separate module.

Assuming that the reorganization PR is going to happen, what makes sense for you? I could either:

1) Ask you to rebase this PR against the branch in https://github.com/mozilla/rhino/pull/1479 and see if you can make it work -- it will require some work on your part at first but I hope that after that it's easier
2) Or, I can merge https://github.com/mozilla/rhino/pull/1479 with the rest of the codebase first this weekend and then you can rebase on top of that
3) Or, I can abandon that change, but I have been working with it for a while and I think it's better.

rbri:
@gbrail this is still work in progress, will rebase if i'm ready

Issue #1482, opened 2024-06-06
Question on the prototype used in BaseFunction.hasInstance(Scriptable instance)


RuralHunter:
Look the code of [this line](https://github.com/mozilla/rhino/blob/c6a873f61fe7d32a560feca9d66318e2586ef169/src/org/mozilla/javascript/BaseFunction.java#L96), it uses the [[Prototype]] of the function object as the prototype to check. According to [this](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/instanceof), should we use the prototype property instead? That is, maybe change the code like this:
`Object protoProp = hasPrototypeProperty() ? getPrototypeProperty() : ScriptableObject.getProperty(this, "prototype");`

rbri:
hmmmm...


ScriptableObject.getProperty(this, "prototype"); already calls getPrototypeProperty() for BaseFunction.
![image](https://github.com/mozilla/rhino/assets/2544132/d6209987-5862-42ab-9a9f-48fd5a4135e6)

Do you have some test code that shows a problem here?
Did a quick test with 
![image](https://github.com/mozilla/rhino/assets/2544132/72180633-829b-47be-95dd-92038f77980d)
but works fine...

RuralHunter:
I have a function object throws exception of "msg.instanceof.bad.prototype" in hasInstance() check. The problem is the function has a 'prototype' slot set to undefined somehow. IdScriptableObject returns 'undefined' as the prototype [here](https://github.com/mozilla/rhino/blob/c6a873f61fe7d32a560feca9d66318e2586ef169/src/org/mozilla/javascript/IdScriptableObject.java#L369). I haven't figured out how the function is created yet since it's in a large obfuscated script.

rbri:
> have a function object
do you know the class of that object?

RuralHunter:
yeah I meant the function/class itself. It's defined like this:
```
(u = function e() {
                                    var r = e._v;
                                    return (0,
                                    e._u)(r[0], arguments, r[1], r[2], this)
                                }
                                )._v = [s, o, g];
```
But I haven't found out where the 'prototype' is set to undefined yet.

RuralHunter:
oh it's a InterpretedFunction if you mean that.

RuralHunter:
ok, I figured out the test case:
```
"use strict";
var f = function () {};
var a1="proto";
var p=a1+'type'; //this constructs a ConsString in rhino
print('before: '+f.prototype);
[...]

PR #1489, opened 2024-06-14
Object.defineProperty() has to take care of ConsString (issue #1482)

rbri:


rbri:
I think there is no problem with this, all tests are green and @gbrail suggested the same solution in #1482 - will merge this.

PR #1491, opened 2024-06-15
improved date localization

rbri:
- use DateTimeFormatter instead of SimpleDateFormat
- Date.prototype.toLocaleString now handles locales argument
- make the formatter use Context.timezone instead of system default
- ES6: use FormatStyle.SHORT for the formatter to make it behave more like real browsers when Intl.DateTimeFormat is not supported


gbrail:
@rbri any idea why the Java 21 test is failing?

FWIW if you rebase then the GitHub action will upload the reports again and we'll see what happened...

rbri:
made some git mistakes on my side - https://github.com/mozilla/rhino/pull/1535 is the new version of this one

PR #1498, opened 2024-06-20
NativeRegexp should only call ScriptRuntime.addInstructionCount when instruction counting is enabled (issue #1497)

### LINKED ISSUE #1497: 1.7.15: NativeRegexp calls ScriptRuntime.addInstructionCount when instruction counting is not enabled
org.mozilla.javascript.regexp.NativeRegExp at line 1891 does:

ScriptRuntime.addInstructionCount(cx, 5);

this will trigger a call to observeInstructionCount even if instructionThreshold is 0.  

To reproduce, create a Context class which overrides observeInstructionCount and use regexps.  observeInstructionCount  will be called even though setInstructionObserverThreshold has not been called

rbri:
and some code improvments suggested by idea

Closes #1497


rbri:
@gbrail / @p-bakker can i merge this?

gbrail:
Yes, looks good and I will merge it. Thanks!

PR #1501, opened 2024-06-25
Add Byte to list of java Numbers

tonygermano:
I think it was just an oversight that `Byte` was missing from the list. Originally, it applied to all java.lang.Numbers.


p-bakker:
Also see https://github.com/mozilla/rhino/discussions/996

tonygermano:
> Also see #996

I forgot I created that discussion post 3 years ago, but I've obviously been thinking about this for a long time!

gbrail:
This seems quite reasonable to me, but in most of the work I've done I haven't used the WrapFactory much.

Is there any way to add a test for this functionality so that we could see the effect of a regression here, even if it ends up being just a test of that instanceof check?

tonygermano:
Test added.

gbrail:
Great, and thanks!

Issue #1444, opened 2024-01-28
Cannot capture continuation from JavaScript code not called directly by executeScriptWithContinuations or callFunctionWithContinuations


821938089:
I've used executeScriptWithContinuations to call the js code, but it still throws this error.
In testing, I found that continuation cannot be captured in some cases.

```js
function test1 () {
// call a function that capture continuation
}
test1() // success
test1.call() // success
test1.apply(this) // success
(function test1 () {
// call a function that capture continuation
})(); // success

test1.bind(this)() // fail
test1.bind(this).call() // fail

test2 = () => {
// call a function that capture continuation
}
test2() // fail
test2.call() // fail
(() => {
// call a function that capture continuation
})(); // fail


eval('// call a function that capture continuation') // fail

eval(`
function test1 () {
// call a function that capture continuation
}
test1()
`) // fail

eval(`
function test1 () {
// call a function that capture continuation
}
`)() // success

eval(`
function test1 () {
// call a function that capture continuation
}
`).call() // success

eval(`
function test1 () {
// call a function that capture continuation
}
`).apply(this) // success

eval(`
function test1 () {
// call a function that capture continuation
}
`).bind(this)() // fail

eval(`
function test1 () {
// call a function that capture continuation
}
`).bind(this).call() // fail

```

example code that list fail case:

```java

import org.mozilla.javascript.*;

public class Main {

  public static class MyObj {
    public void capture() {
      try (var cx = Context.enter()) {
        throw cx.captureContinuation();
      }
    }
  }

  static Scriptable globalScope;

  public static void main(String[] args) {
    try (var cx = Context.enter()) {
      globalScope = cx.initStandardObjects();
      globalScope.put("myObj", globalScope, Context.javaToJS(new MyObj(), globalScope));
      cx.setOptimizationLevel(-1);

      System.out.println("capture.bind(this)()");
      executeScript(cx, "function capture(){myObj.capture()};capture.bind(this)()");

      System.out.println("capture.bind(this).call()");
      executeScript(cx, "function capture(){myObj.capture()};capture.bind(this).call()");

      System.out.println("[arrow function] capture()");
      executeScript(cx, "capture=()=>{myObj.capture()};capture()");

      System.out.println("[arrow function] capture.call()");
      executeScript(cx, "capture=()=>{myObj.capture()};capture.call()");

      System.out.println("[arrow function] immediate call");
      executeScript(cx, "(()=>{myObj.capture()})();");

[...]

PR #1510, opened 2024-06-30
Peel known functions in interpreter so they don't break continuations

szegedi:
Fixes https://github.com/mozilla/rhino/issues/1444

Avoids the interpreter from calling out from its current loop when it needs to invoke an arrow function, a lambda function, a bound function, `Function.apply`, `Function.call`, or a `NoSuchMethodHandleShim`. 

Also handles their combinations. Prior to this change, the best-effort for `Function.apply`, `Function.call`, and `NoSuchMethodHandleShim` only worked if their target was directly an `InterpretedFunction`. With this change, the interpreter will keep drilling down through (or, depending on how you like to visualize it, peel away) the functions (modifying the arguments as those functions would when they're invoked) until it finds an interpreted function to install on its loop stack. If the ultimately invoked function is not interpreted, it will call out to it as before (except it won't call out to the top-level `ArrowFunction`/`BoundFunction`/etc but rather it already did the work for unwrapping them, so it'll only correctly call the final irreducible function.

The net effect is that interpreter will not leave its loop even for combinations of functions such as `(a, b, c) =>

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/ArrowFunction.java
+++ b/rhino/src/main/java/org/mozilla/javascript/ArrowFunction.java
@@ -36,8 +36,7 @@ public class ArrowFunction extends BaseFunction {
 
     @Override
     public Object call(Context cx, Scriptable scope, Scriptable thisObj, Object[] args) {
-        Scriptable callThis = boundThis != null ? boundThis : ScriptRuntime.getTopCallScope(cx);
-        return targetFunction.call(cx, scope, callThis, args);
+        return targetFunction.call(cx, scope, getCallThis(cx), args);
     }
 
     @Override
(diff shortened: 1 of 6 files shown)


gbrail:
Looks good to me -- does anyone who is deep in to the continuation stuff have an opinion or want to try it out?

szegedi:
This doesn't really have to do anything with continuations per se. Rather, it's an improvement in the interpreter that increases the number of function invocation scenarios where the execution remains within a single invocation of the interpreter loop – which is a prerequisite for continuations working. So continuations working better is a consequence of this fix but continuations-related code itself was not touched.

I see @rbri  and @tuchida were somewhat active in this code in the recent-ish past, maybe they want to review it. 

FWIW, all 120k+ tests still pass and I added some new tests too. (I had _lots_ of test262 failures while I was developing this until I got it right, both in `/test262/test/built-ins/Function/prototype/{apply|call}` and in other tests heavily using `.call()` and `.apply()` so there seems to be relevant safety nets aplenty.)

gbrail:
I agree that this is not a super risky change, it's just such complicated code in general, and the code coverage is very good.

I see a few tiny holes in the interpreter test coverage. If you run

    ./gradlew testCodeCoverageReport

and look in ./tests/build/reports/jacoco/testCodeCoverageReport/html/index.html

you can see very good coverage but one or two places to improve. For example, it seems like we don't have tests where 'NativeContinuation.isContinuationConstructor' returns true, at least not in the "tests" module.

Would you mind taking a look and seeing if anything jumps out as an area where we might improve coverage?

gbrail:
Thank you for considering all that. Looks good to me!

Issue #1509, opened 2024-06-30
Parallelize Test262SuiteTest


szegedi:
It'd be great to modify `Test262SuiteTest` to execute tests in parallel. It's sad watching a M3 Pro CPU blaze through most of the tests in under a minute, and then keep waiting until the around 3m 39s mark for it to finish with the test262 suite on a single thread.

p-bakker:
Think this got done, so closing

PR #1514, opened 2024-07-02
Add a parallel test runner and use it for test262

gbrail:
Add a JUnit runner that is designed to ensure that our tests run in parallel. Default number of threads is 1/2 the number of CPUs.

This is an experiment to see if we can make tests run faster and should address https://github.com/mozilla/rhino/issues/1509

### DIFF
--- a/tests/src/test/java/org/mozilla/javascript/testing/ParallelRunner.java
+++ b/tests/src/test/java/org/mozilla/javascript/testing/ParallelRunner.java
@@ -1,0 +1,193 @@
+package org.mozilla.javascript.testing;
+
+import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.TimeUnit;
+import org.junit.runner.Description;
+import org.junit.runner.notification.Failure;
+import org.junit.runner.notification.RunListener;
+import org.junit.runner.notification.RunNotifier;
+import org.junit.runners.Parameterized;
+import org.junit.runners.model.RunnerScheduler;
+
+/**
+ * This class extends the standard Junit 4 "parameterized" test runner to run individual test cases
+ * in parallel threads, using a fixed-size thread pool. It should be a drop-in replacement. However,
+ * test runners must be thread-safe in order for this to work.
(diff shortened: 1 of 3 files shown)


p-bakker:
Tip: use tags GitHub recognizes to auto-link PRs to issues, like `Closes #1509`

See https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#about-linked-issues-and-pull-requests

gbrail:
@szegedi let me know if this speeds up things for you -- I see some improvement, but it's not the massive time saver that I hoped for.

It would also help to parallellize MozillaSuiteTest, but in my experience it made the flaky tests in that suite more flaky. Perhaps we need to retire some of the tests in that suite...

p-bakker:
We seen to be using JUnit4, while JUnit5 has been around for quite a while afaics and comes with features to run tests concurrently

Worth looking into our are we stuck on JUnit4 for some reason?

szegedi:
This alone didn't help the overall test execution time as now `MozillaSuiteTest` became the bottleneck. Setting `ParallelRunner` as the runner for that too brought the time down from 3:37 to 2:24 for me. That's quite a significant improvement. I also tried to bump up the number of threads to be equal to the number of processors as at the end of the run it's only the Mozilla suite that runs, but that didn't have a significant impact. Finally, I changed the runner to just use `ForkJoinPool.commonPool()` which ensured 100% CPU utilization during the tail end of Mozilla suite as well, but it again didn't have a significant impact on overall run time, although I can imagine that it might have on a machine with even more cores. It's also possible that Mozilla suite can be rewritten to be more concurrency friendly, although it seemed to be able to saturate all the cores just fine.

gbrail:
Thanks for trying it as for all this:

* I'll try with the commonPool as well
* I'm a bit worried about MozillaSuiteTest getting too flaky if it gets less CPU. Let's try it though.
* Finally, it looks like Junit 5 conversion would require renaming all of our test annotations. We can certainly do that -- maybe AI can help ;-) -- but it looks like a lot of work

gbrail:
I think that Junit 5 is a much better alternative -- didn't realize how easy it is in Gradle to support both versions at once.

I'm going to close this PR in favor of https://github.com/mozilla/rhino/pull/1515

PR #1515, opened 2024-07-03
Add support for Junit 5 and use it for test262

gbrail:
This lets the test 262 suite run in parallel.

It would be easy to parallelize the MozillaSuiteTest too, but there are many tests in that suite that are sensitive to memory and stack size and I fear that increasing the memory too much would cause a big resource drain. I will keep experimenting with that.


gbrail:
This is a better fix.

Closes https://github.com/mozilla/rhino/issues/1509https://github.com/mozilla/rhino/issues/1509

gbrail:
This is useful even though it doesn't solve all build time problems so I'm going to merge it.

Issue #1187, opened 2022-02-22
Error with destructuring array assignment(?)


ndeuma:
The following cannot be evaluated and leads to an exception with 1.7R3 and later
var a = [1, 2]; var b = 0; [a[b+1], a[b]] = [a[b], a[b+1]]; a[0]

Exception:
```
java.lang.IllegalStateException: FAILED ASSERTION: unexpected token: ADD
	at org.mozilla.javascript.Kit.codeBug(Kit.java:366)
	at org.mozilla.javascript.IRFactory.decompile(IRFactory.java:2498)
	at org.mozilla.javascript.IRFactory.decompileElementGet(IRFactory.java:2549)
	at org.mozilla.javascript.IRFactory.decompile(IRFactory.java:2492)
	at org.mozilla.javascript.IRFactory.decompileArrayLiteral(IRFactory.java:2509)
	at org.mozilla.javascript.IRFactory.decompile(IRFactory.java:2469)
	at org.mozilla.javascript.IRFactory.transformAssignment(IRFactory.java:441)
	at org.mozilla.javascript.IRFactory.transform(IRFactory.java:222)
	at org.mozilla.javascript.IRFactory.transformExprStmt(IRFactory.java:559)
	at org.mozilla.javascript.IRFactory.transform(IRFactory.java:219)
	at org.mozilla.javascript.IRFactory.transformScript(IRFactory.java:1133)
	at org.mozilla.javascript.IRFactory.transform(IRFactory.java:201)
	at org.mozilla.javascript.IRFactory.transformTree(IRFactory.java:121)
	at org.mozilla.javascript.Context.parse(Context.java:2484)
	at org.mozilla.javascript.Context.compileImpl(Context.java:2401)
	at org.mozilla.javascript.Context.compileString(Context.java:1369)
	at org.mozilla.javascript.Context.compileString(Context.java:1357)
	at org.mozilla.javascript.Context.evaluateString(Context.java:1135)
	at de.ndn.labs.RhinoBugTest.main(RhinoBugTest.java:16)
```

But it is evaluated with 1.7R2 (to "2"). I really don't understand why this works in a very old version and fails in all newer versions.

The only thing that works is removing the "+" in the array access:
var a = [1, 2]; var b = 0; var c = 1; [a[c], a[b]] = [a[b], a[c]]; a[0]

Minimal Java example here (and attached)
https://github.com/ndeuma/rhino-bugtest

[...]

Issue #989, opened 2021-07-05
`const [y, y] = []` throws a TypeError, instead of SyntaxError


p-bakker:
Followup of https://github.com/mozilla/rhino/issues/386

tuchida:
Perhaps this is the same problem.
```js
const y = 1;
const y = 2;
```
It should not be possible to achieve this without implementing TDZ.

p-bakker:
Is this really related to TDZ? IMHO this is just about redeclaration of a `const`, whereas TDZ is about accessing a `const` before it is declared, with let/const not being hoisted to the top of a function like vars, but only being defined as the point in the code where they are declared, so any code accessing them before they are declared is what TDZ is about (AFAIK)

tuchida:
Sorry, this may not be directly related.
However, if the redeclaration check is to be done in the parse phase instead of the execution phase, I think TDZ will have to be implemented at the same time.

p-bakker:
To my knowledge TDZ issues throw a ReferenceError and I think those are runtime errors, not parse-time errors.

And redeclaration of const/let throws a SyntaxError, which is a parse-time error I think

tuchida:
After applying these, I got the following error.
```
$ ./gradlew run -q --console=plain --args="-version 200"
Rhino 1.7.16-SNAPSHOT
js> const [y, y] = []
js: line 2: TypeError: redeclaration of const y.
js: 
js: ^
java.lang.IllegalStateException: FAILED ASSERTION
	at org.mozilla.javascript.Kit.codeBug(Kit.java:356)
	at org.mozilla.javascript.IRFactory.transformScript(IRFactory.java:950)
	at org.mozilla.javascript.IRFactory.transform(IRFactory.java:198)
	at org.mozilla.javascript.IRFactory.transformTree(IRFactory.java:128)
	at org.mozilla.javascript.Context.parse(Context.java:2526)
	at org.mozilla.javascript.Context.compileImpl(Context.java:2443)
	at org.mozilla.javascript.Context.compileString(Context.java:1405)
	at org.mozilla.javascript.Context.compileString(Context.java:1393)
	at org.mozilla.javascript.tools.shell.Main.processSource(Main.java:480)
	at org.mozilla.javascript.tools.shell.Main.processFiles(Main.java:180)
	at org.mozilla.javascript.tools.shell.Main$IProxy.run(Main.java:98)
	at org.mozilla.javascript.Context.call(Context.java:549)
[...]

Issue #406, opened 2018-03-11
ClassCastException when compiling malformed destructuring expression


rohanpadhye:
Input to Rhino 1.7.8:
```
(1 ? ({x: (y)}) : (490)) = 1
```
Expected: ReferenceError (Invalid left-hand side in assignment)
Instead, I get:
```
java.lang.ClassCastException: org.mozilla.javascript.Node cannot be cast to org.mozilla.javascript.ast.ObjectLiteral

	at org.mozilla.javascript.Parser.destructuringAssignmentHelper(Parser.java:3932)
	at org.mozilla.javascript.Parser.createDestructuringAssignment(Parser.java:3902)
	at org.mozilla.javascript.IRFactory.createAssignment(IRFactory.java:2235)
	at org.mozilla.javascript.IRFactory.transformAssignment(IRFactory.java:432)
	at org.mozilla.javascript.IRFactory.transform(IRFactory.java:212)
	at org.mozilla.javascript.IRFactory.transformExprStmt(IRFactory.java:516)
	at org.mozilla.javascript.IRFactory.transform(IRFactory.java:209)
	at org.mozilla.javascript.IRFactory.transformScript(IRFactory.java:1042)
	at org.mozilla.javascript.IRFactory.transform(IRFactory.java:192)
	at org.mozilla.javascript.IRFactory.transformTree(IRFactory.java:117)
	at org.mozilla.javascript.Context.compileImpl(Context.java:2540)
	at org.mozilla.javascript.Context.compileString(Context.java:1507)
	at org.mozilla.javascript.Context.compileString(Context.java:1496)
```

Found using [JQF](https://github.com/rohanpadhye/jqf).

rohanpadhye:
Hi, it's been more than one year since this was first reported. Can someone please confirm the bug?

tuchida:
This code will reproduce it.
```js
// Node{ type: Token.OBJECTLIT }
(1 ? {} : 490) = 1

// Node{ type: Token.ARRAYLIT }
(1 ? [] : 490) = 1
```
This will fix it.
```diff
diff --git a/rhino/src/main/java/org/mozilla/javascript/Parser.java b/rhino/src/main/java/org/mozilla/javascript/Parser.java
index 754dc43..a7df6dd 100644
--- a/rhino/src/main/java/org/mozilla/javascript/Parser.java
+++ b/rhino/src/main/java/org/mozilla/javascript/Parser.java
[...]

Issue #405, opened 2018-03-11
FAILED ASSERTION due to malformed destructuring syntax


rohanpadhye:
Rhino 1.7.8:
```
[x (y)] = 2, 3
```
Expected: syntax error (`EvaluatorException`)

Instead, I get:
```
java.lang.IllegalStateException: FAILED ASSERTION: unexpected token: CALL
	at org.mozilla.javascript.Kit.codeBug(Kit.java:433)
	at org.mozilla.javascript.IRFactory.decompile(IRFactory.java:2407)
	at org.mozilla.javascript.IRFactory.decompileArrayLiteral(IRFactory.java:2419)
	at org.mozilla.javascript.IRFactory.decompile(IRFactory.java:2381)
	at org.mozilla.javascript.IRFactory.transformAssignment(IRFactory.java:426)
	at org.mozilla.javascript.IRFactory.transform(IRFactory.java:212)
	at org.mozilla.javascript.IRFactory.transformExprStmt(IRFactory.java:516)
	at org.mozilla.javascript.IRFactory.transform(IRFactory.java:209)
	at org.mozilla.javascript.IRFactory.transformScript(IRFactory.java:1042)
	at org.mozilla.javascript.IRFactory.transform(IRFactory.java:192)
	at org.mozilla.javascript.IRFactory.transformTree(IRFactory.java:117)
	at org.mozilla.javascript.Context.compileImpl(Context.java:2540)
	at org.mozilla.javascript.Context.compileString(Context.java:1507)
	at org.mozilla.javascript.Context.compileString(Context.java:1496)
```

Found using [JQF](https://github.com/rohanpadhye/jqf).

rohanpadhye:
Hi, it's been more than one year since this was first reported. Can someone please confirm the bug?

p-bakker:
Such code should throw a SyntaxError (nowadays)

tuchida:
ref. #1519, #1520
Applying these will correct the problem.
```
$ ./gradlew run -q --console=plain --args="-version 200"
Rhino 1.7.16-SNAPSHOT
js> [x (y)] = 2, 3
js: line 2: Invalid assignment left-hand side.
js: 
js: ^
js: line 2: Compilation produced 1 syntax errors.
```

PR #1519, opened 2024-07-14
 Catch ParserException in IRFactory

### LINKED ISSUE #967: Resolve structural issue with exception handling in IRFactory

tuchida:
Closes #967
Errors in `IRFactory` have the following issues.

1. `Parser$ParserException` not caught.
   ```
   js> 1=1
   js: line 1: Invalid assignment left-hand side.
   js: 
   js: ^
   Exception in thread "main" org.mozilla.javascript.Parser$ParserException
   	at org.mozilla.javascript.Parser.reportError(Parser.java:340)
   	at org.mozilla.javascript.Parser.reportError(Parser.java:326)
   	at org.mozilla.javascript.Parser.reportError(Parser.java:321)
   	at org.mozilla.javascript.IRFactory.createAssignment(IRFactory.java:2325)
   ```
2. Wrong line position.
   ```
   js> var a = 1;
   var b = 2;
   const [c, c] = [3, 4];

   js> js> js: line 1: TypeError: redeclaration of const c.
   ```

`IRFactory` is re-parsing ASTs that have already been parsed by `Parser`, but the error thrown at that time was not handled properly.
So, I have fixed 0cd0e2b2bc662cbcf560d03a3cc9794792bb40d1 for 1. and 436d3f62f0ff4f92f614ef58fdbc7a1bc887ed8c for 2.

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/Context.java
+++ b/rhino/src/main/java/org/mozilla/javascript/Context.java
@@ -2520,7 +2520,7 @@ public class Context implements Closeable {
             }
         }
 
-        IRFactory irf = new IRFactory(compilerEnv, compilationErrorReporter);
+        IRFactory irf = new IRFactory(compilerEnv, sourceString, compilationErrorReporter);
         ScriptNode tree = irf.transformTree(ast);
         return tree;
     }
--- a/rhino/src/main/java/org/mozilla/javascript/IRFactory.java
+++ b/rhino/src/main/java/org/mozilla/javascript/IRFactory.java
@@ -6,6 +6,7 @@
 
 package org.mozilla.javascript;
 
+import java.util.ArrayDeque;
 import java.util.ArrayList;
 import java.util.List;
 import org.mozilla.javascript.ast.ArrayComprehension;
@@ -94,13 +95,16 @@ public final class IRFactory {
 
     private Parser parser;
     private Decompiler decompiler = new Decompiler();
+    private AstNodePosition astNodePos;
 
-    public IRFactory(CompilerEnvirons env) {
-        this(env, env.getErrorReporter());
+    public IRFactory(CompilerEnvirons env, String sourceString) {
+        this(env, sourceString, env.getErrorReporter());
     }
 
-    public IRFactory(CompilerEnvirons env, ErrorReporter errorReporter) {
+    public IRFactory(CompilerEnvirons env, String sourceString, ErrorReporter errorReporter) {
         parser = new Parser(env, errorReporter);
+        astNodePos = new AstNodePosition(sourceString);
+        parser.currentPos = astNodePos;
     }
 
     /**
@@ -118,11 +122,14 @@ public final class IRFactory {
         }
 
         ScriptNode script = null;
+        astNodePos.push(root);
         try {
             script = (ScriptNode) transform(root);
         } catch (Parser.ParserException e) {
             parser.reportErrorsIfExists(root.getLineno());
             return null;
+        } finally {
+            astNodePos.pop();
         }
 
         int sourceEndOffset = decompiler.getCurrentOffset();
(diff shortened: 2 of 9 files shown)


gbrail:
Thanks -- this looks really nice and I'm glad you did it.

I'm trying to also improve code coverage across the board, and from running "./gradlew testCodeCoverageReport" I can see that there are a few spots in IRFactory that we aren't testing. Specifically, there's some new exception-handling code around line 129 of IRFactory.java that doesn't seem to be hit by any of the tests. Would it be possible for you to come up with some sort of test code that would help us hit that spot? 

Specifically, from looking at the coverage report, the part below in "transformTree" is new code, it seems pretty important, and it seems to be missing coverage.

        ScriptNode script = null;
        astNodePos.push(root);
        try {
            script = (ScriptNode) transform(root);
        } catch (Parser.ParserException e) {
            // Doesn't seem to have any code coverage here...
            parser.reportErrorsIfExists(root.getLineno());
            return null;
        } finally {
            astNodePos.pop();
        }


Thanks!

tuchida:
@gbrail I added test to see if `runtimeError` was called ref. 6480350022227b0456bf13dee9ea7865f8736e66. However, the result of `./gradlew testCodeCoverageReport` did not change. What should I do?

gbrail:
Well there might be more for me to investigate, then, since I just replaced
those non-covered lines with an AssertionError and the tests failed. So
that tells me that the code coverage report could itself use some code
coverage ;-)

Given that, I'm happy to merge this as long as it doesn't conflict with
your other big PR for IRFactory!

p-bakker:
Hi @tuchida while rummaging though the backlog of issues, I ran into the following issues of which I wonder if this PR either resolved them or at least moves the needle somewhat: 
- #405 
- #406
- #989 
- #1187

tuchida:
@p-bakker #405 was fixed. Others remained the same or had other errors.

gbrail:
This is great and thanks for doing it!

Issue #1300, opened 2023-02-03
Support ES2019 Function.toString() revision


p-bakker:
See:
- https://github.com/tc39/Function-prototype-toString-revision
- https://tc39.es/Function-prototype-toString-revision/

Some examples:
Currently:
```
=> Math.pow
function pow() { [native code for Math.pow, arity=2] }
```
Expected:
```
=> Math.pow.toString()
function pow() { [native code] }
```

Currently:
```
=> (function(){}).bind({}).toString()
function () {
	[native code, arity=0]
}
```
Expected:
```
=> (function(){}).bind({}).toString()
function () { [native code] }
```

The spec revision also details that the .toString() of user-defined function inside JavaScript must return the source exactly as authored, which I'm not sure we're compliant with

p-bakker:
The part about .toString() returning the source code as authored might be fixed by #1188

tuchida:
It would be possible to handle this by removing the code that adds `arity=` to the result of `Function.prototype.toString`.
However, since #939(#326) is not implemented, the harness for test262 does not work.
```js
// nativeFunctionMatcher.js
  const eatWhitespace = () => {
    while (pos < source.length) {
      const c = source[pos]; // `c` does not change. There are three similar codes.
      if (isWhitespace(c) || isNewline(c)) {
        pos += 1;
        continue;
      }
```
This works by replacing `const` with `let`.

Besides native code, getter/setter test was failing.

p-bakker:
But besides the tests for this in test262 failing, code in Rhino needs to be fixed still to make it functionally work?

tuchida:
As for `arity=`, the following codes needs to be removed.
https://github.com/mozilla/rhino/blob/955af1565bd440b4954a7cbd3db4278b669302e7/rhino/src/main/java/org/mozilla/javascript/BaseFunction.java#L464-L466
https://github.com/mozilla/rhino/blob/955af1565bd440b4954a7cbd3db4278b669302e7/rhino/src/main/java/org/mozilla/javascript/IdFunctionObject.java#L115-L117

I have not looked too closely at the getter/setter.

tuchida:
This issue is not yet fixed.

p-bakker:
To clarify: #1520 fixed the .toString()  behaviour on user-authored JavaScript functions, but not yet on native (Java) methods

tuchida:
What should the Java method be?

https://262.ecma-international.org/15.0/index.html#sec-function.prototype.tostring
[...]

PR #1520, opened 2024-07-14
Fixed Function.prototype.toString as per the latest specification

tuchida:
ref. #1188
Spec. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/toString#getting_source_text_of_a_function
```js
(function() { /* comment */ if(1){}}).toString()
```
Expect
```
function() { /* comment */ if(1){}}
```
But
```
function () {
    if (1) {
    }
}
```
To fixed this, `Decompiler` has been removed.


p-bakker:
Any idea if it covers all function .toString() variations mentioned in #1300?

p-bakker:
@gbrail in https://github.com/mozilla/rhino/pull/1188#issuecomment-1666003584 you mentioned that you'd like to have backwards compatibility

Using GitHub's improved codesearch, I looked for Java files containing an import of the Decompiler and a call to the decompile function: https://github.com/search?q=org.mozilla.javascript.Decompiler+AND+decompile%28+language%3AJava&type=code&l=Java

This yields only a few results, most either of clones of Rhino or deprecated/abandoned projects (vjet, dltk) or projects that haven't had updates for years.

I'd say we bite the bullet and remove the Decompiler completely, even if this will go into 1.7.xx

rhino/src/main/java/org/mozilla/javascript/DecompilerFlag.java
```diff
@@ -0,0 +1,17 @@
+/* -*- Mode: java; tab-width: 8; indent-tabs-mode: nil; c-basic-offset: 4 -*-
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.javascript;
+
+public enum DecompilerFlag {
```
p-bakker, line 9: Does it make sense to call this DecompilerFlag if there's no decompiling going on anymore?

tuchida:
@p-bakker Thanks! I removed. ref. bff0cd1afbea9388bb6df992d82aa7f4af5e1a54

rhino/src/main/java/org/mozilla/javascript/DecompilerFlag.java
tuchida: Do you have any good names?

gbrail:
I'm fine with removing Decompiler -- we have a bunch of other big changes with this release and if we have to say, "instead of Decompiler, use toString()" then I think we will be OK.

gbrail:
Other than that, I see that this PR changes a lot of legacy test suites because the output changed. I had a quick look and I'm prett sure that none of the output changes are significant, so changing those test suites is OK with me, especially because we enable a bunch of test262 suite tests.

Are you fairly confident that these changes are benign?

Finally, you have two big changes to IRFactory in the queue. git will probably handle them OK but there's no guarantee, especially since the other one has a bunch of places where you wrap things with braces and change the indent level. Do you have a preference as to which one gets merged first?

tuchida:
> Are you fairly confident that these changes are benign?

In terms of being on spec, Yes. I do not know the impact of breaking changes.

> Do you have a preference as to which one gets merged first?

No, either is fine.

gbrail:
Thanks for working on this -- I agree it's a good PR. Now that I merged your other one, this one needs conficlit resolution, so can you please do that and push a new commit? Thanks!

tuchida:
@gbrail Thanks! I resolved the conflicts.

gbrail:
Thanks for doing this!

rbri:
@tuchida fantastic!!!! ❤️ 

Now i can remove my hack from the HtmlUnit fork of rhino. Thanks a lot.... hope to see more of you work gets merged soon

Issue #699, opened 2020-06-04
An issue about TypeArray.prototype.set


NWU-NISL:
According to ES6.0, when method `%TypedArray%.prototype.set(array,offset)` is invoked by `s` and the `length` of `array` plus ` offset` is larger than the `length` of `s`, it will throw a `RangeError` exception . But when the `length` of `array`is `undefined`, rhino didn't throw any exception. 

#### version

 `1.7.12 `

####  command

```
java -jar rhino/rhino-1.7.12.jar -debug -version 200 testcase.js
```

#### testcase

```
var NISLFuzzingFunc = function(){
	var s = new Uint8Array(0);
	print(s.length);
	var a = {};
	print(a.length);
	s.set(a,1);
};
NISLFuzzingFunc();
```

#### output

```
0
undefined
```

#### Expected output

```
0
undefined
RangeError: offset out of range
```

Contributor:[@YuanWangC](https://github.com/YuanWangC)

Issue #1411, opened 2023-11-03
ES2015 TypedArrays missing all methods they ought to share with regular Arrays, like map/indexOf/includes etc


p-bakker:
All TypedArray implementations ought to have a large chunk of methods that regular Arrays also have, like indexOf/includes/map/forEach etc. However, those all seem to be missing, see https://compat-table.github.io/compat-table/es6/ > built-ins > typed arrays

In addition to the missing methods that are defined in ES2015 (ES6):
- ES2016 specified the addition of the .includes() method on (Typed)Arrays
- ES2020 specifies two new TypedArray variants, see #1410
- ~ES2022 specifies the .at() method on (Typed)Arrays~ 
- ES2024 adds new methods via the `Array find from last`
- ES2024 adds new methods via the `Change Array by copy`, see #1307

p-bakker:
@andreabergia do you agree that all method are now implement (albeit having failing tests here and there), except for static stuff that ought to be implemented in the missing TypedArray prototype? 

If so, I think we can close this case, as we have individual cases for the other stuff, agree?

gbrail:
Neither here nor there, but I think that someone should take the time to
convert the NativeArray to lambdas (it's a big job) and then take a sharp
look at the architecture of the native arrays and see if that can be
improved once we've done that -- I did the native arrays a long time ago, I
recall that they were weird, and I recall doing a lot of weird things to
make them work that might not be necessary any more now that we have a
construct that maps more closely to the actual language.

p-bakker:
I think doing that in #1565, which involves creating a NativeTypedArray class containing most of the scripting API impl. And have that be the prototype for all typed array constructors would make sense

p-bakker:
@andreabergia what's your take on whether the scope of just adding the missing method is now done or not?

andreabergia:
> @andreabergia do you agree that all method are now implement (albeit having failing tests here and there), except for static stuff that ought to be implemented in the missing TypedArray prototype?
> 
> If so, I think we can close this case, as we have individual cases for the other stuff, agree?

Sorry for the late answer. 
I agree with you - everything should be implemented, except for the static `TypedArray.from` and `.of`. I couldn't find an issue for those, though.

p-bakker:
Everything mentioned in this case for done in various derived cases and PRs, so closing

Issue #1410, opened 2023-11-02
ES2020 support BigInt64Array & BigUint64Array Typed Arrays


p-bakker:
See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt64Array for example

Including accompanying DataView.prototype.getBigUint64/getBigInt64

Issue #1397, opened 2023-10-15
ES2024 resizable ArrayBuffer


p-bakker:
Proposal: https://github.com/tc39/proposal-resizablearraybuffer

camnwalter:
I was looking into this, and the changes themselves aren't difficult to add, it's just hard to run the tests. It's unfortunate because most of the 262 tests related to resizable array buffers use const inside for of loops as well as inside blocks, which we don't handle correctly.

p-bakker:
Create a PR on the test262 testsuite repo to replace the const for let's 

They in general are open to such PRs

rbri:
@p-bakker is your pr merged?

p-bakker:
I was just suggesting to create a PR, I haven't created one

rbri:
Hi @camnwalter,
still interested in this. Maybe we can work together here?

camnwalter:
Yeah sure :) Haven't looked too much into it since I opened up this issue, but it looks pretty straightforward to do. It's just a bit tricky because I don't want to make the changes in 262 without testing with rhino making sure it actually helps, and adding resizability was also a bit more complex than I thought. Could just be me misremembering though...

PR #1524, opened 2024-07-17
Implement various missing methods in `TypedArray`

andreabergia:
Since most methods were already implemented in `NativeArray`, I followed this approach:
- when reasonable, I extracted the implementation in `NativeArray` and moved it to a new class `ArrayLikeAbstractOperations`;
- otherwise, I copy&pasted the code from `NativeArray` into `NativeTypedArrayView` and simplified it.

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/AbstractEcmaObjectOperations.java
+++ b/rhino/src/main/java/org/mozilla/javascript/AbstractEcmaObjectOperations.java
@@ -18,7 +18,7 @@ package org.mozilla.javascript;
  *           will be put in the JavaDoc of each method that implements an Abstract Operations
  *     </ul>
  */
-class AbstractEcmaObjectOperations {
+public class AbstractEcmaObjectOperations {
     enum INTEGRITY_LEVEL {
         FROZEN,
         SEALED
@@ -161,7 +161,7 @@ class AbstractEcmaObjectOperations {
      *     constructor on "s" or if the "species" symbol is not set.
      * @see <a href="https://tc39.es/ecma262/#sec-speciesconstructor"></a>
      */
-    static Constructable speciesConstructor(
+    public static Constructable speciesConstructor(
             Context cx, Scriptable s, Constructable defaultConstructor) {
         /*
         The abstract operation SpeciesConstructor takes arguments O (an Object) and
--- a/rhino/src/main/java/org/mozilla/javascript/ArrayLikeAbstractOperations.java
+++ b/rhino/src/main/java/org/mozilla/javascript/ArrayLikeAbstractOperations.java
@@ -1,0 +1,304 @@
+package org.mozilla.javascript;
+
+import static org.mozilla.javascript.NativeArray.getLengthProperty;
+import static org.mozilla.javascript.ScriptRuntimeES6.requireObjectCoercible;
+import static org.mozilla.javascript.Scriptable.NOT_FOUND;
+
+import java.io.Serializable;
+import java.util.Comparator;
+import org.mozilla.javascript.regexp.NativeRegExp;
+
+/** Contains implementation of shared methods useful for arrays and typed arrays. */
+public class ArrayLikeAbstractOperations {
+    public enum IterativeOperation {
+        EVERY,
+        FILTER,
+        FOR_EACH,
+        MAP,
+        SOME,
+        FIND,
+        FIND_INDEX,
+    }
+
+    public enum ReduceOperation {
+        REDUCE,
+        REDUCE_RIGHT,
+    }
+
+    /** Implements the methods "every", "filter", "forEach", "map", and "some". */
+    public static Object iterativeMethod(
+            Context cx,
+            IdFunctionObject fun,
+            IterativeOperation operation,
+            Scriptable scope,
+            Scriptable thisObj,
+            Object[] args) {
+        Scriptable o = ScriptRuntime.toObject(cx, scope, thisObj);
+
+        if (IterativeOperation.FIND == operation || IterativeOperation.FIND_INDEX == operation) {
+            requireObjectCoercible(cx, o, fun);
+        }
+
+        long length = getLengthProperty(cx, o);
+        if (operation == IterativeOperation.MAP && length > Integer.MAX_VALUE) {
+            String msg = ScriptRuntime.getMessageById("msg.arraylength.bad");
+            throw ScriptRuntime.rangeError(msg);
+        }
+
+        Object callbackArg = args.length > 0 ? args[0] : Undefined.instance;
+
+        Function f = getCallbackArg(cx, callbackArg);
+        Scriptable parent = ScriptableObject.getTopLevelScope(f);
+        Scriptable thisArg;
+        if (args.length < 2 || args[1] == null || args[1] == Undefined.instance) {
+            thisArg = parent;
+        } else {
+            thisArg = ScriptRuntime.toObject(cx, scope, args[1]);
+        }
+
+        Scriptable array = null;
+        if (operation == IterativeOperation.FILTER || operation == IterativeOperation.MAP) {
+            int resultLength = operation == IterativeOperation.MAP ? (int) length : 0;
+            array = cx.newArray(scope, resultLength);
+        }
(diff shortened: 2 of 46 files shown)


p-bakker:
Great stuff!

Could you check whether this covers everything mentioned in #1411?
And also #1307 (or at least for TypedArrays?

And whether it maybe also fixes #699

andreabergia:
> Great stuff!
> 
> Could you check whether this covers everything mentioned in #1411? And also #1307 (or at least for TypedArrays?)

For #1411 it implements the "old" methods, i.e. not the ones added by ES2024. Therefore it also does not implement #1307 (because I felt that it's better to do a single implementation, for both normal and typed arrays).
I also have not added BigInt64Array/ BigUint64Array (#1410).

> And whether it maybe also fixes #699

It does not. I have only added new methods, not fixed existing ones.

p-bakker:
Note that there are a few more TypedArray-related issues, you might feel like taking a stab at :-)
- Support ES2023 Array find from last (.findLast/.findLastindex) #1241
- Support ES2023 Change Array by copy #1307
- ES2020 support BigInt64Array & BigUint64Array Typed Arrays #1410
- ES2024 resizable ArrayBuffer #1397

rbri:
Great!

+ 1 for merging this!!!!

p-bakker:
Besides makes a wack of tests from test262 pass, based on the diff of the test262.priperties file it seem it also breaks a bunch of tests that previously passed. 

What's up with those?

andreabergia:
> Besides makes a wack of tests from test262 pass, based on the diff of the test262.priperties file it seem it also breaks a bunch of tests that previously passed.

Thanks for pointing this out, I somehow missed it. There were three categories of failing tests:
- the tests in `built-ins/TypedArrayConstructors`, which are now failing for `BigInt` since those kind of typed arrays are not implemented. This is coherent with the previously existing methods (say `set` or `subarray`), which were already failing.
- the tests which use `detached`, such as `prototype/copyWithin/coerced-values-end-detached.js` - this is something that should be implemented in the test 262 harness and it is not yet. Previously the tests were passing because the invoked method did not exist; now that it does, since the detaching feature does not work, the tests fails. I don't think it's super important.
- the tests for species constructor, which made me realize that my implementation of `filter`, `map`, and `splice` was a bit incorrect. I fixed those. 🙂

p-bakker:
For the 2nd item,: you means something additional needs to be implemented [here](https://github.com/mozilla/rhino/blob/f152eb7366ac6f2b99ca05b4b67fe508b2b46a6d/tests/src/test/java/org/mozilla/javascript/tests/Test262SuiteTest.java#L433)? If so, could you create a new case about what needs to be implemented?

p-bakker:
The state of things in test262.properties seems ok now.

But I do see a recurring theme on failing tests for methods, like:
```
prototype/values/invoked-as-func.js
    prototype/values/invoked-as-method.js
    prototype/values/length.js
    prototype/values/name.js
    prototype/values/prop-desc.js
    prototype/values/this-is-not-object.js
    prototype/values/this-is-not-typedarray-instance.js
````
Any chance you could make a follow-up case for fixing whichever of those failing tests could be fixed by some TLC?

**Comment by andreabergia** [comment]
> For the 2nd item,: you means something additional needs to be implemented [here](https://github.com/mozilla/rhino/blob/f152eb7366ac6f2b99ca05b4b67fe508b2b46a6d/tests/src/test/java/org/mozilla/javascript/tests/Test262SuiteTest.java#L433)? If so, could you create a new case about what needs to be implemented?

Opened https://github.com/mozilla/rhino/issues/1527

> But I do see a recurring theme on failing tests for methods, like:
> 
> ```
> prototype/values/invoked-as-func.js
>     prototype/values/invoked-as-method.js
>     prototype/values/length.js
>     prototype/values/name.js
>     prototype/values/prop-desc.js
>     prototype/values/this-is-not-object.js
>     prototype/values/this-is-not-typedarray-instance.js
> ```
> 
> Any chance you could make a follow-up case for fixing whichever of those failing tests could be fixed by some TLC?

Created https://github.com/mozilla/rhino/issues/1528

**Comment by gbrail** [comment]
I think this is a good change in general, but it still looks like we fixed many test262 tests that were previously broken, but we still broke test262 tests that previously worked. I understand the issues above about "detached" support, but I'm not excited about merging this if it breaks tests that currently work. Is there anything that we can do here, or do others want to merge this and then fix the broken tests later?

**Comment by andreabergia** [comment]
> I think this is a good change in general, but it still looks like we fixed many test262 tests that were previously broken, but we still broke test262 tests that previously worked. I understand the issues above about "detached" support, but I'm not excited about merging this if it breaks tests that currently work. Is there anything that we can do here, or do others want to merge this and then fix the broken tests later?

I think the reason is clear by looking at the test case code, and not just at the `test262.properties` file.

---

The first category of tests are the ones about the detached, for example: `test/built-ins/TypedArray/prototype/copyWithin/coerced-values-end-detached.js`:

```js
testWithTypedArrayConstructors(function(TA) {
  var ta;
  function detachAndReturnIndex(){
      $DETACHBUFFER(ta.buffer);
      return 900;
  }

  var array = [];
  array.length = 10000; // big arrays are more likely to cause a crash if they are accessed after they are freed
  array.fill(7, 0);
  ta = new TA(array);
  assert.throws(TypeError, function(){
    ta.copyWithin(0, 100, {valueOf : detachAndReturnIndex});
  }, "should throw TypeError as array is detached");
});
```

Previously this test was passing because the function `copyWithin` did not exist, so trying to invoke it throw an exception. Now, the function exists, but since we don't have the "detached buffer" feature, the tests fails. However, this PR does not really break anything - the fact that the test was passing was, well, a lucky coincidence.

---

The other category of tests now broken are those in the directory `built-ins/TypedArrayConstructors`, where previously the tests like `test/built-ins/TypedArrayConstructors/prototype/lastIndexOf/bigint-inherited.js` were passing and now are failing. They are something like this:

```js
testWithTypedArrayConstructors(function(TA) {
  assert.sameValue(TA.prototype.hasOwnProperty("lastIndexOf"), false);
});
```

This fails because the functions are now defined. The `hasOwnProperty` should return false, but that was true also for existing functions such as `set` or `toString` - which were failing before this PR. All these tests should be fixed all at once, in a separate PR though, because there is a common problem shared between pre-existing and new functions.

p-bakker:
Does seem like there are a bunch of odd ones that are now seemingly broken

Besides the detached buffer ones for example:
- `prototype/map/speciesctor-get-species-custom-ctor-length-throws.js`
- within built-ins/TypedArrayConstructors

The detached buffer ones are also strange: if they never passed, why weren't they recorded in test262.properties  before? Does that mean something is amiss in our test262Suite?

So I agree this needs to be looked at before merging

p-bakker:
@andreabergia so our comments overlapped a bit, time-wise

What about tests like prototype/map/species/ctor-get-species-custom-ctor-length-throws.js? Why are those now reported as failing?

andreabergia:
> What about tests like prototype/map/species/ctor-get-species-custom-ctor-length-throws.js? Why are those now reported as failing?

Same as the others - there are three tests like these: map, filter, slice. All of these do something like the following:

```js
testWithTypedArrayConstructors(function(TA) {
  var sample = new TA(2);

  sample.constructor = {};
  sample.constructor[Symbol.species] = function() {
    return new TA();
  };

  assert.throws(TypeError, function() {
    sample.map(function() { return 0; });
  });
});
```

Before this PR, they are "passing" because the `map` method does not exist, and so we get a type error. Now they are "really" failing because my implementation of the `Symbol.species` is not 100% perfect, although it does handle quite a lot of cases.
So, again, the fact that they were passing before this PR is just misleading.

p-bakker:
@gbrail to answer your earlier question: I think the seemingly breaking tests aren't actual breakages. Instead, due to how those test are authored, they previously returned false positives

So i think this PR is good to go, from a test262 perspective

rbri:
agree with @p-bakker looks fine for me and another great step forward +1 for merge

and many thanks @andreabergia

gbrail:
Thanks!

I resolved the conflict and merged this manually. Great progress!

PR #1525, opened 2024-07-17
Adopt "errorprone" tool instead of Spotbugs

gbrail:
Compared to spotbugs, errorprone seems to run faster, finds a number of
useful things, is easier to configure, and is updated often.

* Add errorprone to the build for all non-test code
* Suppress or disable warnings that are too much for us to fix now
* Fix other errors and warnings
* Remove spotbugs (it and errorprone don't always agree)

Other notes:

* We always specify a character set for string conversions and output writing now. This is hard-coded to UTF-8 which I believe will be the right choice for everyone AFAIK.
* errorprone is very pedantic about parenthesis, but I think it will make the code more consistent in the long run so I left these warnings enabled
* errorprone is very unhappy about many aspects of our Javadoc and I decided to disable those checks. Someone else might find it rewarding to enable them and fix the problems that result.


gbrail:
This change improves build times a little bit, makes the code more consistent and modern, and found three bugs:

* A possible ConcurrentModificationException in the Java adapter
* Some dodgy concurrency practices in two places

Given that, I'm going to go ahead and merge it, although it may require be to do a bit more work on a few of the in-flight PRs.

rhino/src/main/java/org/mozilla/javascript/JavaMembers.java
```diff
@@ -328,8 +328,7 @@ private void discoverAccessibleMethods(
                                 if (isPublic(mods) || isProtected(mods) || includePrivate) {
                                     MethodSignature sig = new MethodSignature(method);
                                     if (!map.containsKey(sig)) {
-                                        if (includePrivate && !method.isAccessible())
-                                            method.setAccessible(true);
+                                        if (includePrivate) method.trySetAccessible();
```
rbri, line 331: Same question as below

PR #1531, opened 2024-07-19
Add all directories to test262.properties

tuchida:
I manually modified this file and added all directories in test262 to test262.properties.
```sh
tests/test262/test$ find . -type d | sort > ../../testsrc/test262.properties.tmp
```

### DIFF
--- a/tests/testsrc/test262.properties
+++ b/tests/testsrc/test262.properties
@@ -1,6 +1,8 @@
 # This is a configuration file for Test262SuiteTest.java. See ./README.md for more info about this file
 
-built-ins/Array 146/2670 (5.47%)
+~annexB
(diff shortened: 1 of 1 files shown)


tests/testsrc/test262.properties
```diff
@@ -408,32 +418,14 @@ built-ins/Date 39/707 (5.52%)
     value-to-primitive-result-non-string-prim.js
     value-to-primitive-result-string.js
 
-~built-ins/decodeURI this is really; slow disable for now
```
tuchida, line 411: The order has changed because of case-sensitive sorting.

gbrail:
Thanks. this is great!

PR #1535, opened 2024-07-22
improved date localization

rbri:
- use DateTimeFormatter instead of SimpleDateFormat
- Date.prototype.toLocaleString now handles locales argument
- make the formatter use Context.timezone instead of system default
- ES6: use FormatStyle.SHORT for the formatter to make it behave more like real browsers when Intl.DateTimeFormat is not supported

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/NativeDate.java
+++ b/rhino/src/main/java/org/mozilla/javascript/NativeDate.java
@@ -1452,7 +1452,9 @@ final class NativeDate extends IdScriptableObject {
         }
 
         final ZoneId zoneid = cx.getTimeZone().toZoneId();
-        return formatter.format(Instant.ofEpochMilli((long) t).atZone(zoneid));
+        final String formatted = formatter.format(Instant.ofEpochMilli((long) t).atZone(zoneid));
+        // jdk 21 uses a nnbsp in front of 'PM'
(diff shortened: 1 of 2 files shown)


rbri:
this is the new version of https://github.com/mozilla/rhino/pull/1491

Source code is unchanged, only applied to the latest code

rbri:
again fails with java 21 - will have a look

gbrail:
Thanks! I attempted to look at the test results myself, and I only had time
to figure out that it was a more complex problem than I realised.

rbri:
Think i found it... only i18n ;-D

rbri:
@gbrail green now hope it is ok for you and can be merged

gbrail:
Yes -- looks good, I will merge it.

It did introduce a few warnings from the (admittedly very pedantic) errorprone plugin. In a future PR someone should clean those up -- it just disagrees sometimes about where to use parenthesis.

Just because I'm trying to be a stickler, I'm going to add a commit to fix those warnings, and then I will merge this PR manually and close it.

One more thought -- I wonder how much simpler the whole NativeDate class would be if it used the new java.time package instead of doing its own calculations? (Plus we could use Lamba functions instead of the IdScriptableObject and it'd be even better!)

gbrail:
Merged manually. Thanks!

PR #1536, opened 2024-07-23
Remove Icode_SWAP as it was unused

szegedi:
While examining our code coverage tests I noticed that even after running the full ECMA and Mozilla test suites, the `Icode_SWAP` branch in the interpreter never executed. Turns out it is never emitted, so I removed it; it's dead code.


gbrail:
Thanks!

Issue #1477, opened 2024-05-18
NPE in MemberBox.java


p3k:
I am seeing a null pointer in [MemberBox.java](https://github.com/mozilla/rhino/blob/085d561953cd2ac0d28e8ff69b4b5742cd6074f5/src/org/mozilla/javascript/JavaMembers.java#L103) related to the changes in https://github.com/mozilla/rhino/commit/73a8b29d795af5c828d67334e75e46f4e3d2d183:

```java
// Wrapped java.lang.NullPointerException: Cannot read the array length because "args" is null
for (int i = 0; i < args.length; ++i) {
    if (args[i] instanceof Delegator) {
        args[i] = ((Delegator) args[i]).getDelegee();
    }
}
```

Tracing back the issue I see `Context.emptyArgs` is indeed null when calling `MemberBox.invoke()` in [JavaMembers.java](https://github.com/mozilla/rhino/blob/085d561953cd2ac0d28e8ff69b4b5742cd6074f5/src/org/mozilla/javascript/JavaMembers.java#L103):

```java
// Context.emptyArgs is null
rval = bp.getter.invoke(javaObject, Context.emptyArgs);
```

The weird thing is, though, that [`Context.emptyArgs` actually *should be* the same as `ScriptRuntime.emptyArgs`](https://github.com/mozilla/rhino/blob/085d561953cd2ac0d28e8ff69b4b5742cd6074f5/src/org/mozilla/javascript/Context.java#L356) (which is not null) – but it obviously isn’t:

```java
// This resolves the issue
rval = bp.getter.invoke(javaObject, ScriptRuntime.emptyArgs);
```

It could be there is some code in my project causing this, see https://github.com/antville/helma/pull/75#issuecomment-2118796627.

Unfortunately, I am not very experienced with Java, just inherited the Java/Rhino part of the project. Thus, I would be very thankful for any suggestions.

kleiteimc:
Came across the same error using Helma 1.7.2

p-bakker:
@rbri any thoughts on this one, as it's (supposedly) related to the changes made by you in https://github.com/mozilla/rhino/commit/73a8b29d795af5c828d67334e75e46f4e3d2d183?

rbri:
will have a look and provide a fix

p-bakker:
Resolved by #1538

PR #1538, opened 2024-07-26
add an additional null check

rbri:
This solves #1477.

i think an additional null check here brings us on the safe side and does not harm.


p-bakker:
LGTM

gbrail:
OK -- it'd be great to find a way to make the tests for MemberBox and other things more robust in the future though.

PR #1540, opened 2024-07-28
Update to latest version of test262

p-bakker:
Worked with the maintainers of test262 to eliminate anything in the harness that prevented us from running the latest version of test262, so now we can be on the latest version

A next step will be to:
- automatically add missing directories, if any, when running with the -DupdateTest262properties flag
- maybe create a GitLab CI Action that automatically create a PR to updates the revision of test262 that is used by the /tests/test22 submodule?

But I think this can be merged as is

### DIFF
--- a/tests/src/test/java/org/mozilla/javascript/tests/Test262SuiteTest.java
+++ b/tests/src/test/java/org/mozilla/javascript/tests/Test262SuiteTest.java
@@ -5,7 +5,6 @@
 package org.mozilla.javascript.tests;
 
 import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
 import static org.mozilla.javascript.drivers.TestUtils.JS_FILE_FILTER;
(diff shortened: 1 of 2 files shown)


p-bakker:
@gbrail The pipeline(s) for this PR fail, but the reason why cannot be deducted from the log: https://github.com/mozilla/rhino/actions/runs/10139556248/job/28033071384?pr=1540

It seems like there might be more info available given the message ` There were failing tests. See the report at: file:///home/runner/work/rhino/rhino/tests/build/reports/tests/test/index.html`, but this file isn't uploaded as a result.

Is that something that can be done? Having a hard time now figuring out why the pipeline fails

gbrail:
This is great! And it works for me and it looks like you figured out the problem with the automated tests.

I want to get rid of the Jacoco warning that pops up, so I'm going to merge this manually along with a fix for that as well. 

Thanks!

gbrail:
Thanks -- merged manually.

gbrail:
Just wanted to remind everyone to run "git submodule update" after they pull from GH because otherwise you may not have the version of the test262 tests that you should.

PR #1551, opened 2024-08-02
pass ScriptRuntime.emptyArgs in getApplyArguments()

rbri:
Sorry folks, but i have no test case for this - but i found this in a complex script scenario. Maybe someone has an idea how to test this.

What is this about: the parameter handling for the apply function replaces many args with emptyArgs. But this replacements does not check for emptyArgs itself - in this case the call throws an error. Therefore i added the check and the js code now runs fine.

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/ScriptRuntime.java
+++ b/rhino/src/main/java/org/mozilla/javascript/ScriptRuntime.java
@@ -2853,7 +2853,7 @@ public class ScriptRuntime {
     }
 
     static Object[] getApplyArguments(Context cx, Object arg1) {
-        if (arg1 == null || Undefined.isUndefined(arg1)) {
+        if (arg1 == null || Undefined.isUndefined(arg1) || arg1 == ScriptRuntime.emptyArgs) {
             return ScriptRuntime.emptyArgs;
         } else if (arg1 instanceof Scriptable && isArrayLike((Scriptable) arg1)) {
             return cx.getElements((Scriptable) arg1);


gbrail:
Sure, I get it, and we can add this check, but I feel like something very strange must be going wrong in your test case. It looks like that function is supposed to be called with the first argument -- if that new check is necessary it implies to me that someone is passing in an arguments array with a first argument that's itself an empty array, and that sounds wrong. So far you sure that there isn't a bug somewhere else first?

tonygermano:
I've been trying, and I can't figure out how to reproduce this. Even if I call apply with `org.mozilla.javascript.ScriptRuntime.emptyArgs` as the second arg, it's still fine because that ends up being wrapped in a NativeJavaArray, which is an instance of Scriptable and is array-like.

What is the error you are seeing when it happens?

rbri:
@tonygermano sorry but i also spend a lot of time trying to nail down this thing. Found this while trying to get a huge web page with all this f**** javascript that tracks you running in HtmlUnit. But finally i gave up and made this 'fix'. Sorry i do not have the web page and my local setup at hand. There are soo many reports i have to work on.

Maybe we can leave this and see if some day a similar scenario pops up....

anonyein:
> @tonygermano sorry but i also spend a lot of time trying to nail down this thing. Found this while trying to get a huge web page with all this f**** javascript that tracks you running in HtmlUnit. But finally i gave up and made this 'fix'. Sorry i do not have the web page and my local setup at hand. There are soo many reports i have to work on.
> 
> Maybe we can leave this and see if some day a similar scenario pops up....

[htmlunit-rhino-fork](https://github.com/HtmlUnit/htmlunit-rhino-fork)
I use your fork and quick good, only with some missing of parameters like https://github.com/HtmlUnit/htmlunit-rhino-fork/blob/master/rhino/src/main/java/org/mozilla/javascript/ImporterTopLevel.java#L143

p-bakker:
@rbri do you want this PR hanging around or can we close it? Can always reopen it when the cause it's found

rbri:
> @rbri do you want this PR hanging around or can we close it? Can always reopen it when the cause it's found

Do whatever you like - as always i have done this for core-js, my problem is gone without any negative effects for my cases (users). This was only to get rhino informed about my finding.

p-bakker:
Converted to a draft, as based on the discussion a testcase is needed for this PR to be considered and as of this moment providing one has proved to be a challenge

PR #1552, opened 2024-08-02
Fix problem with regex constructor

andreabergia:
The spec for ES6 says that we can create a new regex passing an existing one and a different set of flags, but that does not work. This PR fixes it (while keeping the old behavior for ES5 and below). We also now pass a few more test262 cases.

Spec link: https://tc39.es/ecma262/multipage/text-processing.html#sec-regexp-constructor

(If accepted, please squash this PR 🙂 )


gbrail:
Looks good -- but now that we just updated test262, could you please rebase your PR on top of the latest? (And also you'll want to "git submodule update" after doing that...) Thanks!

andreabergia:
Rebased on top of master and the latest test262 suite

gbrail:
Looks good -- thanks!

PR #1556, opened 2024-08-07
FEAT: Implement Symbol.unscopables

0xe:
Implements `Array.prototype[Symbol.unscopables]` from the [spec](https://tc39.es/ecma262/multipage/indexed-collections.html#sec-array.prototype-%symbol.unscopables%).

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/NativeArray.java
+++ b/rhino/src/main/java/org/mozilla/javascript/NativeArray.java
@@ -45,11 +45,30 @@ public class NativeArray extends IdScriptableObject implements List {
 
     private static final Object ARRAY_TAG = "Array";
     private static final Long NEGATIVE_ONE = Long.valueOf(-1);
+    private static final String[] UNSCOPABLES = {
+        "at",
+        "copyWithin",
+        "entries",
+        "fill",
+        "find",
+        "findIndex",
+        "findLast",
+        "findLastIndex",
+        "flat",
+        "flatMap",
+        "includes",
+        "keys",
+        "toReversed",
+        "toSorted",
+        "toSpliced",
+        "values"
+    };
 
     static void init(Context cx, Scriptable scope, boolean sealed) {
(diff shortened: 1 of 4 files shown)


rbri:
LGTM 👍

rbri:
do we need a feature check for ES6 here?

gbrail:
Looks good and thanks! Can you please re-base on top of the latest to pick up the changes to test262?

0xe:
> do we need a feature check for ES6 here?

I think we don't initialise Symbol support unless it's ES6. So evaluating any Symbol throws `ReferenceError` in non-es6 which seems to be expected behaviour. I was following what we are doing to `Symbol.species`. 

The [spec](https://tc39.es/ecma262/multipage/fundamental-objects.html#sec-symbol-objects) doesn't seem to talk about throwing an different error in older versions when the Symbols are referenced.

gbrail:
Thanks, this makes sense. Any idea why that one remaining test continues to fail -- is it something we can address in another PR?

prototype/Symbol.unscopables/change-array-by-copy.js

0xe:
> Thanks, this makes sense. Any idea why that one remaining test continues to fail -- is it something we can address in another PR?
> 
> prototype/Symbol.unscopables/change-array-by-copy.js

This test uses `for (const unscopable of [...])` loop which doesn't work in Rhino (we throw syntax error for this form). The same check works if rewritten using `for (var unscopable of [...])` like this:

```javascript
var unscopables = Array.prototype[Symbol.unscopables];

for (var unscopable of ["toReversed", "toSorted", "toSpliced"]) {
    verifyProperty(unscopables, unscopable, {
        value: true,
        writable: true,
        configurable: true
    })
};

assert(!Object.prototype.hasOwnProperty.call(unscopables, "with"), "does not have `with`");
```

gbrail:
OK, thanks for the explanation of the test -- looks good to me.

PR #1559, opened 2024-08-09
Rationalize benchmark setup and time units

gbrail:
* Make sure that all the benchmarks run using "gradle jmh"
* Make sure that each has a reasonable default time unit
* Add a way to change which benchmarks run via the command-line
* Add a README for the benchmarks


gbrail:
I don't think this one is going to be controversial so I'm going to check it in.

PR #1562, opened 2024-08-12
Remove obsolete "ObjArray" class

gbrail:
Depending on context, replace this venerable class with either ArrayList or ArrayDeque.

This is the first of a few PRs to get rid of some of the more obsolete things in the codebase.


rbri:
+1

p-bakker:
Although ObjArray was a public class, I've got no problem removing it

Tagged this PR with the `potentially breaking change' label, so we won't forget to add a note to the 1.7.16 release notes about is removal

gbrail:
Sure, I don't mind being super-pedantic and letting people know that we are deleting a public class. Realistically, I seriously doubt that anyone is using this class.

(Next, I plan to get rid of ObjToIntMap and UIntMap which I suspect will have similar issues!)

Anyway, this has been pretty uncontroversial so I'm going to merge it. Thanks!

Issue #507, opened 2019-01-05
Creation of many temporary arrays when calling Java methods in a compiled Javascript


Alexander-H-e:
When executing a compiled Javascript and calling a Java method with one parameter, an Object[1] array is created each time. The idea of using compiled methods is to be able to execute the same Javascript code many times fast. Having the garbage collector to remove these many Object[1] arrays slows down the process.

Here are two methods which could be optimized to make Rhino faster (by using significant less memory):

**org.mozilla.javascript.optimizer.OptRuntime:**
changing
```
public static Object call1(Callable fun, Scriptable thisObj, Object arg0, Context cx, Scriptable scope) {
    return fun.call(cx, scope, thisObj, new Object[]{arg0});
}
```
to
```
private static ThreadLocal<Object[]> paramCall1Cache=new ThreadLocal<Object[]>(); 
public static Object call1(Callable fun, Scriptable thisObj, Object arg0, Context cx, Scriptable scope) {
    Object[] obj=paramCall1Cache.get();
    if (obj==null) {obj=new Object[1]; paramCall1Cache.set(obj);}
    obj[0]=arg0;
    return fun.call(cx, scope, thisObj, obj);
}
```
and

**org.mozilla.javascript.ScriptRuntime:**
changing
```
public static Number wrapNumber(double x) {
    if (x != x) return ScriptRuntime.NaNobj;
    return new Double(x);
}
```
to
```
private final static Double[] fastBoxing=new Double[100];
static {for (int i=0;i<fastBoxing.length;i++) fastBoxing[i]=Double.valueOf(i);}
    
public static Number wrapNumber(double x) {
    if (x != x) return ScriptRuntime.NaNobj;
    final long xInt=(long)Math.floor(x);
    if (xInt-x==0 && xInt>=0 && xInt<fastBoxing.length) return fastBoxing[(int)xInt];
    return Double.valueOf(x);
}
```
decreases memory consumption in a test scenario in Flight Recorder by nearly 75% (5 GB down to 1.31 GB). Also org.mozilla.javascript.optimizer.OptRuntime.call2 could be changed in the same way as call1.

p-bakker:
This case seems (partially?) duplicated by #717, which has more detailed discussion on it. 

But this case makes an additional suggestion about org.mozilla.javascript.ScriptRuntimem.wrapNumber

Issue #350, opened 2017-10-23
Double/Integer in javascript body


luthionb:
I'm working with rhino library:
`<groupId>org.mozilla</groupId>
<artifactId>rhino</artifactId>
<version>1.7R4</version>`
To operate on numbers in javascript. Here I have a sample code:
`Context ctx = Context.enter();
Scriptable scope = ctx.initStandardObjects();
ctx.getWrapFactory().setJavaPrimitiveWrap(false);
String script = "var x = 2.0; x;";
Object result = ctx.evaluateString(scope, script, CMD, 1, null);
System.out.println(result.getClass());`
And what suprised me is the fact that rhino is returning Integer, as according to ECMAScript specification all "primitive" javascript Number types should be float with double precision. What's even more interesting is the fact that changing script to:
`String script = "var x = 1.0; x;";`
Will return a double. So results are not consistent if no decimal values are provided. 

And I think it's all due to implementation. Problem seems to be connected with class org.mozilla.javascript.optimizer.Codegen and method 
`pushNumberAsObject(ClassFileWriter cfw, double num)`
For 0.0, 1.0, -1.0 we are using some predefined statics. For other we are creating static constants with type defined by method
`String getStaticConstantWrapperType(double num)`
And this method for number 2.0 will always return "Ljava/lang/Integer" object. Is it correct behavior? What's the reason for keeping such strange behavior?

And if it is something that cannot be resolved, can I somehow give a bytecode node additional number property?

sainaen:
> And what suprised me is the fact that rhino is returning Integer, as according to ECMAScript specification all "primitive" javascript Number types should be float with double precision.

ECMAScript spec only governs effects/behavior observable *inside* of the ECMAScript code. As far as I know, most ES implementations use an optimistic assumption that all numbers are regular 32-bit int-s, until they have to fallback to doubles. I assume that in this case the difference you get is a results of similar optimization done by Rhino.

> And if it is something that cannot be resolved, can I somehow give a bytecode node additional number property?

I'm not sure I understand what do you mean here, but first, what is the problem you're trying to solve? If you want to always have doubles instead of int-s in the generated code, I don't think it's possible; even in the ES code some operations only work with integers.

gbrail:
The optimization level may have an effect on this -- there are certain integer optimizations that only take affect at optimization level 9.

luthionb:
Sorry for late response. Use case I'm trying to apply is pretty complicated. But let me simplify it. 

[...]

PR #1563, opened 2024-08-13
Aggressively try to use Integers for arithmetic

gbrail:
This PR changes some of the logic that does arithmetic using Number objects to try and do integer arithmetic when both operands are integers, rather than assuming that all arithmetic should be done using floating-point. 

This also changes a few fundamental constants -- for example, Rhino has always defined "zero" as "0.0" but this PR makes it truly "0". 

If Rhino embedders are assuming that all numbers coming back from Rhino are Double objects (rather than checking for Number and then getting either intValue() or doubleValue() as they need), then their code may break. I had to change a few test cases for this reason. If, on the other hand, embedders are working with Number, there should not be an issue.

Some math-heavy benchmarks, like some of the V8 and SunSpider benchmarks, run quite a bit faster with this change.

I would be interested to see if some people have code that breaks because of this, so I'm hoping that a few of you could try it out -- there are likely other optimizations along these lines that are possible if it works.

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/ArrayLikeAbstractOperations.java
+++ b/rhino/src/main/java/org/mozilla/javascript/ArrayLikeAbstractOperations.java
@@ -120,7 +120,8 @@ public class ArrayLikeAbstractOperations {
                     break;
                 case FIND_INDEX:
                 case FIND_LAST_INDEX:
-                    if (ScriptRuntime.toBoolean(result)) return ScriptRuntime.wrapNumber(i);
+                    if (ScriptRuntime.toBoolean(result))
+                        return ScriptRuntime.wrapNumber((double) i);
                     break;
             }
         }
--- a/rhino/src/main/java/org/mozilla/javascript/Interpreter.java
+++ b/rhino/src/main/java/org/mozilla/javascript/Interpreter.java
@@ -2150,13 +2150,11 @@ public final class Interpreter extends Icode implements Evaluator {
                                 }
                             case Icode_ZERO:
                                 ++stackTop;
-                                stack[stackTop] = DBL_MRK;
-                                sDbl[stackTop] = 0;
+                                stack[stackTop] = Integer.valueOf(0);
                                 continue Loop;
                             case Icode_ONE:
                                 ++stackTop;
-                                stack[stackTop] = DBL_MRK;
-                                sDbl[stackTop] = 1;
+                                stack[stackTop] = Integer.valueOf(1);
                                 continue Loop;
                             case Token.NULL:
                                 stack[++stackTop] = null;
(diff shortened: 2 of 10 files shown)


p-bakker:
Afraid I'm not able to try this out in the environment where I use Rhino, but I'm OK if the Rhino integration would break if it wrongfully expects Doubles, instead of Numbers...

Do wonder about whether this PR is in any way related to #350 or (parts of) #717 & #507, in the sense of whether those cases are (partly) resolved/changes by this PR or whether those cases give hints for possible further optimizations in the int/double area

Specifically for #350 I think this PR will just result in Java receiving an int more often, correct? I think #350 can just be closed as Won't Fix, as Java integration code should not expect Doubles, but Numbers. Agree?

Also a question: in https://github.com/mozilla/rhino/issues/350#issuecomment-339723149 you mentioned there being an integer optimization on optLevel 9. Is that (still?) really the case? I never found any code in Rhino that does anything different between optLevel 1 and 9

rbri:
sorry....

rbri:
@gbrail spend some time to give this a try by applying to core-js.
Looks like nothing really (see below) fails and the overall speed seems to be a bit better (the whole huge test suite runs on a virtual machine so the speed depends on the hoster).

I had one issue to fix - my Reflect/Proxy impl (see #1431) uses instanceof Double at two places in NativeReflect. After fixing that all my tests are running fine. Maybe other products extending Rhino in some ways and also depending on the assumption that every numbers a double internally. But for me that is fine, i guess i know what happens and a have a large test suite to check the compatibility.

Summary: i'm fine with this

drubanovich-soti:
@gbrail We just updated Rhino and this change broke one of our tests, which is not critical for us, but Rhino behavior seemed unexpected, so I would like to raise the concern for this specific case:
If a new JavaScript function added by Rhino has parameter or array type, then the actual argument will unexpectedly contain a mix  of Integers and Doubles. E.g. if this is the implementation of the function:
```       
        fun add(array: NativeArray): Double {
            var sum = 0.0
            for (number in array) {
                sum += number as Double
            }
            return sum
        }
```
Then the first invocation below will pass, but the second - will fail, as the NativeArray parameter contains unexpected mix of Integers and doubles:
```
    let sum1 = add([3, 2]); // Passes
    let sum2 = add([1, 2]); // Fails
```

gbrail:
Can you open an issue for this? Hard to keep up with comments on closed PRs.

I'm not sure what part of your example there is JavaScript and what parts are Kotlin or something else, but in general, a JavaScript function that would typically be expected to return a "number" might return different types of numbers, and internal optimizations may change that specific type, for example if we can do the work using only integers we will. This is probably worth an update to the docs (which need a big update anyway!) 

The best practice for calling JavaScript from non-JavaScript in Rhino is to not assume this, and to at least use the various functions of Number like intValue() and doubleValue() if you need a primitive. And in fact, if you are writing truly general-purpose code, you can't assume anything about what comes back from a function written in JS, and if you care you should use a method like ScriptRuntime.toNumber() which can handle most things.

Finally, maybe one of the folks on the project who works on the various reflection-based embeddings would have a different view, since I get the impression that's what you're trying to do here.

drubanovich-soti:
Thanks Greg, Yes you got it right and this answers all my questions. I'll open an issue as you asked.

PR #1564, opened 2024-08-14
Remove more obsolete classes: ObjToIntMap and UintMap

gbrail:
These classes can be easily replaced by HashMap and sometimes HashSet without a negative performance impact.

(We still have one custom hash table left in Rhino, in "EmbeddedSlotMap." Attempts to replace that particular class with HashMap make performance measurably worse.)


rbri:
@gbrail have done a test with this in HtmlUnit (core-js). Have seen no notable changed in terms of performance and all the tests are passing. From my point of view you can do this....

PR #1566, opened 2024-08-16
Ensure that generator functions pass parameters properly

gbrail:
Generators use a different calling convention and require "activation" in order to pick up the right parameters. This wasn't being triggered unless the generator function had a "yield" statement, resulting in some very strange results. Make sure that it always happens for ES6 generators.

Fixes https://github.com/mozilla/rhino/issues/1412


p-bakker:
Think it fixes #1412, not #1413, but otherwise LGTM!

Thnx for looking into it!

rbri:
Merge, merge, merge....

Issue #1307, opened 2023-03-04
Support ES2023 Change Array by copy


p-bakker:
Proposal: https://github.com/tc39/proposal-change-array-by-copy

pkkht:
Hello, 
Thanks for flagging this as "good first issue". Can I work on this please.

p-bakker:
Yes, we'd welcome a PR for this

Check [Contributing PRs](https://github.com/mozilla/rhino#contributing-prs) for some guidelines and pointers

As for tests: unfortunately we cannot run the latest version of the Test262 testsuite, which means the [Change Array by Copy tests](https://github.com/tc39/test262/commit/6f4601d095a3899d6102f2c320b671495cbe8757) are not included yet within Rhino.

So depending on your timeline, you'll need to include tests yourself or we might be able to switch to the latest version of Test262 before you're finished, see ([#958](https://github.com/mozilla/rhino/issues/958#issuecomment-868532333))

pkkht:
Cool, please assign the issue to me. I am reading the docs to get an understanding of the changes. I will post any question in the Google group.

p-bakker:
We don't typically assign issues to people, just submit a PR and everything will work out.

pkkht:
Thanks, I will be able to start work late next week.

p-bakker:
FYI: I just found out that all TypeArray implementations that Rhino supports are missing (almost) all methods that they ought to share with regular Arrays.

As such, I think there might be way more to it to also implement the new methods that are part of the 'Change Array by Copy' proposal to all Typed Arrays, so maybe use this case to just add the proposed methods to NativeArray and let the TypedArray part of the proposal be part of #1411

Issue #1565, opened 2024-08-14
Typed array prototype is not correct according to the spec


andreabergia:
This should return an object with all the various `TypedArray` methods defined on it (find, map, etc), but it returns `undefined` in rhino:

```js
Object.getPrototypeOf(Int8Array).prototype
```

The following should also be true, but are not:

```js
Object.getPrototypeOf(Int8Array).prototype !== undefined && 
  Object.getPrototypeOf(Int8Array).prototype == Object.getPrototypeOf(Uint8Array).prototype
```

```js
Object.getPrototypeOf(Int8Array.prototype) === Object.getPrototypeOf(Uint8Array).prototype
```

p-bakker:
Due to this, all TypedArray implementations are missing the static .from. and .of methods

p-bakker:
Strongly consider implementing the required NativeTypedArray class using Lambdas, see https://github.com/mozilla/rhino/issues/1411#issuecomment-2307664612

Issue #1528, opened 2024-07-18
Investigate failures on TypedArray test262 cases


andreabergia:
Most methods on TypedArray fail these test262 cases:

- invoked-as-func.js
- invoked-as-method.js
- length.js
- name.js
- prop-desc.js
- this-is-not-object.js
- this-is-not-typedarray-instance.js

Issue #1527, opened 2024-07-18
Implement `$262.detachArrayBuffer` for typed arrays


andreabergia:
Quite a few test cases for typed arrays fail because rhino lacks `$262.detachArrayBuffer`, which should be a implementation-defined way to detach an Array Buffer. For example, `prototype/copyWithin/coerced-values-end-detached.js`.

p-bakker:
[This](https://github.com/mozilla/rhino/blob/f152eb7366ac6f2b99ca05b4b67fe508b2b46a6d/tests/src/test/java/org/mozilla/javascript/tests/Test262SuiteTest.java#L433) should implement [this operation]( https://tc39.es/ecma262/#sec-detacharraybuffer) based on [this documentation](https://github.com/tc39/test262/blob/main/docs/rationale.md#host-defined-functions)

p-bakker:
Based on just looking at the NativeArrayBuffer it seems to me that is not just the test facility is missing, but that NativeArrayBuffer lacks the facilities to 'detach' as well

Could maybe be as straightforward as setting the internal (byte) buffer to null and adding some checks and throws

PR #1569, opened 2024-08-20
Implement ES2023 copy methods on Array and TypedArray

andreabergia:
This PR adds:
- `Array.prototype.toReversed`
- `Array.prototype.toSorted`
- `Array.prototype.toSpliced`
- `Array.prototype.with`
- `TypedArray.prototype.toReversed`
- `TypedArray.prototype.toSorted`
- `TypedArray.prototype.with`

(There is no `TypedArray.prototype.toSpliced` in the standard).
I tried to follow the spec very closely, even matching (most) variable names.

For arrays, it passes all test262 cases except those marked with `Reflect.construct`. Note that, to do so, I couldn't rely on some optimization, for example the `denseOnly` flag.

For typed arrays, in addition to the cases marked with `Reflect.construct`, there are some other test262 cases that don't pass:
- `prototype/xxx/metadata` - caused by https://github.com/mozilla/rhino/issues/1565 as far as I can tell. There's already https://github.com/mozilla/rhino/issues/1528 which tracks similar problems.
- `prototype/xxx/length-property-ignored.js` - do not pass because rhino does not allow to configure `TypedArray.prototype.length` - tracked by https://github.com/mozilla/rhino/issues/1572
- `prototype/toReversed/this-value-invalid.js` and `prototype/toSorted/this-value-invalid.js` - these try to call `$DETACHBUFFER`, which we do not support - see https://github.com/mozilla/rhino/issues/1527
- `prototype/toString/BigInt/detached-buffer.js` and `prototype/with/BigInt/early-type-coercion-bigint.js - tracked https://github.com/mozilla/rhino/issues/1410
- `prototype/with/index-validated-against-current-length.js` - we do not support `ArrayBuffer.prototype.resize`; https://github.com/mozilla/rhino/issues/1397 tracks this

Should fix https://github.com/mozilla/rhino/issues/1307

I ended up with a single PR that does both `Array` and `TypedArray` because I imagined they would share part of the implementation... but in the end, they are completely separate. Let me know if you would prefer two separate PRs.

Please squash this if accepted! 🙂

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/NativeArray.java
+++ b/rhino/src/main/java/org/mozilla/javascript/NativeArray.java
@@ -334,6 +334,22 @@ public class NativeArray extends IdScriptableObject implements List {
                 arity = 1;
                 s = "flatMap";
                 break;
+            case Id_toSorted:
+                arity = 1;
+                s = "toSorted";
+                break;
+            case Id_toReversed:
+                arity = 0;
+                s = "toReversed";
+                break;
+            case Id_toSpliced:
+                arity = 2;
+                s = "toSpliced";
+                break;
+            case Id_with:
+                arity = 2;
+                s = "with";
+                break;
             default:
                 throw new IllegalArgumentException(String.valueOf(id));
         }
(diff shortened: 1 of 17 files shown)


andreabergia:
> Is `prototype/xxx/length-property-ignored.js - do not pass because rhino does not allow to configure TypedArray.prototype.length` due to the length propertyDescriptor being setup with `configurable: false` or is something else going on? I assume we need a case for this one as well?

Yeah, I meant to do that, but I had a long meeting while I was updating the description and I missed it 😅 Here it is: https://github.com/mozilla/rhino/issues/1572

gbrail:
This looks pretty complete to me -- anyone else have an opinion? Otherwise I will merge it in the next few days.

gbrail:
This looks good, but I want to keep the code clean and fix one warning:

I added "ErrorProne" a while back and I'm trying to clear its warnings (and may consider making them errors).

Would you mind fixing this one minor one that shows up in the build with these new changes?

/home/runner/work/rhino/rhino/rhino/src/main/java/org/mozilla/javascript/NativeArray.java:2239: warning: [UnnecessaryParentheses] These grouping parentheses are unnecessary; it is unlikely the code will be misinterpreted without them
            int from = (int) (len) - k - 1;
                             ^
    (see https://errorprone.info/bugpattern/UnnecessaryParentheses)
  Did you mean 'int from = (int)  len - k - 1;'?

andreabergia:
> I added "ErrorProne" a while back and I'm trying to clear its warnings (and may consider making them errors).
> 
> Would you mind fixing this one minor one that shows up in the build with these new changes?

Sure, done.

PR #1574, opened 2024-08-22
Fix javadoc generation

rPraml:
This PR fixes Issues in javadoc generation, when using `gradlew publishToMavenLocal`


rhino/src/main/java/org/mozilla/javascript/Scriptable.java
```diff
@@ -49,6 +49,7 @@ public interface Scriptable {
      * <code>get</code> that takes an integer:
      *
      * <table>
+     * <caption style="display:none">Code examples</caption>
```
rPraml, line 52: Javadoc fails with `no summary or caption for table`

gbrail:
Thanks, this looks helpful to me!

PR #1576, opened 2024-08-22
Silenced all warnings

andreabergia:
Simply silences all warnings printed out in a build from master:

- two by marking them with `@SuppressWarnings("deprecation")`, because I don't think we can change the code without breaking anything:
- one by adding the cast suggested by ErrorProne. It does not change the behavior, since that cast was already being inserted implicitly by the compiler.


gbrail:
Thanks, looks good!

Issue #1376, opened 2023-08-17
Function interface vs. Constructable


rbri:
While working on the Proxy support i stumbled over some places where we check for instanceof Function and then casting to Function but only calling the construct method on the casted object. This is a problem if we are working with Lambda constructors because they are only implementing Constructable.

What do you think - should we change at least the places where Constructable is sufficient (guess ~5 palces)?

p-bakker:
Sounds like it would fix an issue (when using Lambda constructors) without there being any downside, no? 

If so, I'd say it makes sense to make such a change

gbrail:
I'm trying to make some of the standard objects use lambdas, and finding
that I have to change some stuff there. Let me make a PR and we can see if
that factors in to what you need to do. But yes we should do this!

rbri:
Ok, will wait for your stuff and see...

rbri:
@gbrail do you think about the Constructable interface as part of this. Just noted that Function extends Constructable. My feeling (at least for the depending lambda stuff) is that we are in trouble here. I guess only lambda constructor has to fulfill this interface and not Lambda function. Correct?

Or you are aware of an different way to figure out if a function is a ctor?

gbrail:
Can you do two things for me, and I'm sorry I haven't focused as much on this so far:

First, background: I added Constructable as a "single-function interface." That way both Constructable and Callable are interfaces that you can pass a lambda function to. That's why we have that.

Second, LambdaConstructor implements both Constructable and Callable, because constructors can be invoked either using "new"   or directly. It has a flag that you can set to determine whether both methods just call the constructor, or whether one or the other throws an exception.

Finally, I just merged new support into LambdaConstructor so that you can have the Constructable and Callable be different implementations -- I was trying to make the Date class into a lambda and I needed that, because the two forms of the Date constructor do different things.

So -- given all that, do you need this? And if you confirm this, then I'm happy to merge this, and I assume that you need it to complete the proxy implementation?

rbri:
> So -- given all that, do you need to merge https://github.com/mozilla/rhino/pull/1635? And if you confirm this, then I'm happy to merge it, and I assume that you need it to complete the proxy implementation?

No i do not need this for the proxy stuff - there was one place where i had to do the change from Function to Constructable but this is part of the other PR.

The motivation for #1635 was more about using the smallest possible interface - if we only have to call the construct method it is sufficient to check for the Constructable interface.

We can also drop #1635 and wait if we face problems.

[...]

Issue #1549, opened 2024-07-30
Object.getOwnPropertyNames/getOwnPropertyDescriptor/getOwnPropertyDescriptors/for-of loop dont with NativeJavaXxxx stuff


p-bakker:
```
=>Object.getOwnPropertyNames(java.math.BigInteger)
org.mozilla.javascript.EcmaError: TypeError: Expected argument of type object, but instead had type function (internal_anon#1)

=>Object.getOwnPropertyDescriptor(java.math.BigInteger, 'valueOf')
org.mozilla.javascript.EcmaError: TypeError: Expected argument of type object, but instead had type function (internal_anon#1)

=>for (var x of java.math.BigInteger) console.log(x)
org.mozilla.javascript.EcmaError: TypeError: [JavaClass java.math.BigInteger] is not iterable (internal_anon#1)
```

tuchida:
```js
// * NativeJavaClass
Object.getOwnPropertyNames(java.math.BigInteger)
// js: uncaught JavaScript runtime exception: TypeError: Expected argument of type object, but instead had type function
// 	at (unknown):2

// * NativeJavaObject
Object.getOwnPropertyNames(java.math.BigInteger.valueOf(0))
// js: uncaught JavaScript runtime exception: TypeError: Expected argument of type object, but instead had type object
// 	at (unknown):3

// * NativeJavaPackage
Object.getOwnPropertyNames(java.math)
// BigInteger
// 
// <- laoded Only?

// * NativeJavaMap
Object.getOwnPropertyNames(new java.util.HashMap({ a: 1 }))
// js: uncaught JavaScript runtime exception: TypeError: Expected argument of type object, but instead had type object
// 	at (unknown):3
//
// <- Is it 'a' to expect? or methods?

// * java.lang.String
Object.getOwnPropertyNames(java.lang.String.valueOf('abc'))
// js: uncaught JavaScript runtime exception: TypeError: Expected argument of type object, but instead had type object
// 	at (unknown):7
```

tonygermano:
I independently found this problem and mentioned it in https://github.com/mozilla/rhino/pull/1615#issuecomment-2351596095. The main problem is that `Object.getOwnPropertyNames` (and many other static methods on the built-in `Object`) requires the object being acted upon to be an instance of `ScriptableObject`, while most of the NativeJavaXxxx classes do not extend `ScriptableObject`.

[...]

PR #1577, opened 2024-08-22
Adding LambdaAccessorSlot

nabacg:
First time contributing, so let me know if I missed something important. 

I've noticed the new LambdaConstructor based pattern of registering scriptables is being used more and more ([NativePromise](https://code.devsnc.com/dev/rhino-ng/blob/9c3e2ee043617f27b38930d4b4bd6bc6f4787fda/src/org/mozilla/javascript/NativePromise.java#L78) or [NativeMath](https://code.devsnc.com/dev/rhino-ng/blob/8b248c1a9025f9dd6b5c8d9537232928f0d7ef01/src/org/mozilla/javascript/NativeMath.java#L30)) and I find it's very clear and elegant to define javascript object methods with java's Callable via Lambda Functions, but  I came across the limitation where if you want to define a property where it's Getter and/or Setter logic needs to access js object instance state ( needs a reference to `this` ) there isn't really a good way to do this on a `LambdaConstructor` object. NativePromise intentionally doesn't have any getter / setter properties, so it didn't run into this problem. Yes, it's possible it use [ScriptableObject.defineProperty](https://github.com/mozilla/rhino/blob/4a27e8eaf3e8d3ba690ed3841212a3d13de16887/rhino/src/main/java/org/mozilla/javascript/ScriptableObject.java#L1460), but it requires `Method` parameters which forces reflection and wouldn't it be so much nicer if we could define properties using Lambda functions like 
`Function<Scriptable, Object>` and `BiConsumer<Scriptable, Object>` and write code like this: 

```java 
constructor.definePrototypeProperty("status",
                        (thisObj) -> self(thisObj).getStatus(),
                        (thisObj, value) -> self(thisObj).setStatus(value), 
DONTENUM);

``` 

This PR is attempting to solve this problem by adding new type of Slot called OwnerAwareLambdaSlot, inspired by existing [LambdaSlot](https://github.com/mozilla/rhino/blob/b7ece52d83a0d558b28b5962c63e8eea31f0ecd8/rhino/src/main/java/org/mozilla/javascript/LambdaSlot.java#L17-L19).  It allows defining getter / setter properties using Lambda functions that accept Scriptable owner object ('this') as one of parameters, thus allowing to implement properties that require access to instance fields etc.

While I don't have specific use case for it in the code base (but if you can think of one, I'm happy to part take), I've built it for some internal use cases I sadly can't share, but I hope the new slot itself and the unit tests that demonstrate how to use it is also useful. Reading comments on some of the recent PRs makes me thing that more people would find it useful and hopefully would facilitate more refactoring into this new pattern:
- https://github.com/mozilla/rhino/pull/1561#issuecomment-2303719199
- https://github.com/mozilla/rhino/issues/1376#issuecomment-2302520051

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/LambdaAccessorSlot.java
+++ b/rhino/src/main/java/org/mozilla/javascript/LambdaAccessorSlot.java
@@ -14,15 +14,17 @@ import java.util.function.Function;
  * implementing properties that behave like standard JavaScript properties, but are implemented with
  * native functionality without the need for reflection.
  */
-public class OwnerAwareLambdaSlot extends Slot {
+public class LambdaAccessorSlot extends Slot {
     private transient Function<Scriptable, Object> getter;
     private transient BiConsumer<Scriptable, Object> setter;
+    private LambdaFunction getterFunction;
+    private LambdaFunction setterFunction;
 
-    OwnerAwareLambdaSlot(Object name, int index) {
+    LambdaAccessorSlot(Object name, int index) {
         super(name, index, 0);
     }
 
-    OwnerAwareLambdaSlot(Slot oldSlot) {
+    LambdaAccessorSlot(Slot oldSlot) {
         super(oldSlot);
     }
 
@@ -35,28 +37,31 @@ public class OwnerAwareLambdaSlot extends Slot {
     ScriptableObject getPropertyDescriptor(Context cx, Scriptable scope) {
         ScriptableObject desc = (ScriptableObject) cx.newObject(scope);
         if (getter != null) {
-            desc.defineProperty(
-                    "get",
-                    new LambdaFunction(
-                            scope,
-                            "get " + super.name,
-                            0,
-                            (cx1, scope1, thisObj, args) -> getter.apply(thisObj)),
-                    ScriptableObject.EMPTY);
+            if (getterFunction == null) {
+                this.getterFunction =
+                        new LambdaFunction(
+                                scope,
+                                "get " + super.name,
+                                0,
+                                (cx1, scope1, thisObj, args) -> getter.apply(thisObj));
+            }
+
+            desc.defineProperty("get", this.getterFunction, ScriptableObject.EMPTY);
         }
 
         if (setter != null) {
-            desc.defineProperty(
-                    "set",
-                    new LambdaFunction(
-                            scope,
-                            "set " + super.name,
-                            1,
-                            (cx1, scope1, thisObj, args) -> {
-                                setter.accept(thisObj, args[0]);
-                                return Undefined.instance;
-                            }),
-                    ScriptableObject.EMPTY);
+            if (setterFunction == null) {
+                this.setterFunction =
+                        new LambdaFunction(
+                                scope,
+                                "set " + super.name,
+                                1,
+                                (cx1, scope1, thisObj, args) -> {
+                                    setter.accept(thisObj, args[0]);
+                                    return Undefined.instance;
+                                });
+            }
+            desc.defineProperty("set", this.setterFunction, ScriptableObject.EMPTY);
         }
         desc.setCommonDescriptorProperties(getAttributes(), false);
         return desc;
--- a/rhino/src/main/java/org/mozilla/javascript/LambdaAccessorSlot.java
+++ b/rhino/src/main/java/org/mozilla/javascript/LambdaAccessorSlot.java
@@ -34,54 +34,61 @@ public class LambdaAccessorSlot extends Slot {
     }
 
     @Override
+    boolean isSetterSlot() {
+        return true;
+    }
+
+    @Override
     ScriptableObject getPropertyDescriptor(Context cx, Scriptable scope) {
         ScriptableObject desc = (ScriptableObject) cx.newObject(scope);
-        if (getter != null) {
-            if (getterFunction == null) {
-                this.getterFunction =
-                        new LambdaFunction(
-                                scope,
-                                "get " + super.name,
-                                0,
-                                (cx1, scope1, thisObj, args) -> getter.apply(thisObj));
-            }
 
+        int attr = getAttributes();
+        boolean es6 = cx.getLanguageVersion() >= Context.VERSION_ES6;
+        if (es6) {
+            if (getterFunction == null && setterFunction == null) {
+                desc.defineProperty(
+                        "writable",
+                        (attr & ScriptableObject.READONLY) == 0,
+                        ScriptableObject.EMPTY);
+            }
+        } else {
+            desc.setCommonDescriptorProperties(attr, getterFunction == null && setterFunction == null);
+        }
+
+        if (getterFunction != null) {
             desc.defineProperty("get", this.getterFunction, ScriptableObject.EMPTY);
         }
 
-        if (setter != null) {
-            if (setterFunction == null) {
-                this.setterFunction =
-                        new LambdaFunction(
-                                scope,
-                                "set " + super.name,
-                                1,
-                                (cx1, scope1, thisObj, args) -> {
-                                    setter.accept(thisObj, args[0]);
-                                    return Undefined.instance;
-                                });
-            }
+        if (setterFunction != null) {
             desc.defineProperty("set", this.setterFunction, ScriptableObject.EMPTY);
+        } else if (es6) {
+            desc.defineProperty("set", Undefined.instance, ScriptableObject.EMPTY);
         }
-        desc.setCommonDescriptorProperties(getAttributes(), false);
+
+        if (es6) {
+            desc.defineProperty(
+                    "enumerable", (attr & ScriptableObject.DONTENUM) == 0, ScriptableObject.EMPTY);
+            desc.defineProperty(
+                    "configurable",
+                    (attr & ScriptableObject.PERMANENT) == 0,
+                    ScriptableObject.EMPTY);
+        }
         return desc;
     }
 
     @Override
-    public boolean setValue(Object value, Scriptable scope, Scriptable owner, boolean isStrict) {
-        if (setter != null) {
-            setter.accept(owner, value);
+    public boolean setValue(Object value, Scriptable scope, Scriptable start, boolean isThrow) {
+        if (setter == null) {
+            if (getter != null) {
+                throwNoSetterException(start, value);
+                return true;
+            }
+        } else {
+            setter.accept(start, value);
             return true;
         }
 
-        if (isStrict) {
-            // in strict mode
-            throw ScriptRuntime.typeErrorById("msg.modify.readonly", name);
-        } else {
-            super.setValue(value, scope, owner, false);
-        }
-
-        return true;
+        return super.setValue(value, start, start, isThrow);
     }
 
     @Override
@@ -92,11 +99,30 @@ public class LambdaAccessorSlot extends Slot {
         return super.getValue(owner);
     }
 
-    public void setGetter(Function<Scriptable, Object> getter) {
+    public void setGetter(Scriptable scope, Function<Scriptable, Object> getter) {
         this.getter = getter;
+        if (getter != null) {
+            this.getterFunction =
+                    new LambdaFunction(
+                            scope,
+                            "get " + super.name,
+                            0,
+                            (cx1, scope1, thisObj, args) -> getter.apply(thisObj));
+        }
     }
 
-    public void setSetter(BiConsumer<Scriptable, Object> setter) {
+    public void setSetter(Scriptable scope, BiConsumer<Scriptable, Object> setter) {
         this.setter = setter;
+        if (setter != null) {
+            this.setterFunction =
+                    new LambdaFunction(
+                            scope,
+                            "set " + super.name,
+                            1,
+                            (cx1, scope1, thisObj, args) -> {
+                                setter.accept(thisObj, args[0]);
+                                return Undefined.instance;
+                            });
+        }
     }
 }
--- a/rhino/src/main/java/org/mozilla/javascript/LambdaAccessorSlot.java
+++ b/rhino/src/main/java/org/mozilla/javascript/LambdaAccessorSlot.java
@@ -52,7 +52,8 @@ public class LambdaAccessorSlot extends Slot {
                         ScriptableObject.EMPTY);
             }
         } else {
-            desc.setCommonDescriptorProperties(attr, getterFunction == null && setterFunction == null);
+            desc.setCommonDescriptorProperties(
+                    attr, getterFunction == null && setterFunction == null);
         }
 
         if (getterFunction != null) {
--- a/rhino/src/main/java/org/mozilla/javascript/OwnerAwareLambdaSlot.java
+++ b/rhino/src/main/java/org/mozilla/javascript/OwnerAwareLambdaSlot.java
@@ -84,8 +84,7 @@ public class OwnerAwareLambdaSlot extends Slot {
         if (getter != null) {
             return getter.apply(owner);
         }
-        Object v = super.getValue(owner);
-        return v == null ? Undefined.instance : v;
+        return super.getValue(owner);
     }
 
     public void setGetter(Function<Scriptable, Object> getter) {
--- a/rhino/src/main/java/org/mozilla/javascript/OwnerAwareLambdaSlot.java
+++ b/rhino/src/main/java/org/mozilla/javascript/OwnerAwareLambdaSlot.java
@@ -1,22 +1,23 @@
 package org.mozilla.javascript;
 
-import java.util.IdentityHashMap;
 import java.util.function.BiConsumer;
 import java.util.function.Function;
 
 /**
- * A specialized property accessor using lambda functions, similar to {@link LambdaSlot},
- * but allows defining properties with getter and setter lambdas that require access to the owner object ('this').
- * This enables the implementation of properties that can access instance fields of the owner.
+ * A specialized property accessor using lambda functions, similar to {@link LambdaSlot}, but allows
+ * defining properties with getter and setter lambdas that require access to the owner object
+ * ('this'). This enables the implementation of properties that can access instance fields of the
+ * owner.
  *
- * Unlike {@link LambdaSlot}, Lambda functions used to define getter and setter logic require the owner's
- * `Scriptable` object as one of the parameters.
- * This is particularly useful for implementing properties that behave like standard JavaScript properties,
- * but are implemented with native functionality without the need for reflection.
+ * <p>Unlike {@link LambdaSlot}, Lambda functions used to define getter and setter logic require the
+ * owner's `Scriptable` object as one of the parameters. This is particularly useful for
+ * implementing properties that behave like standard JavaScript properties, but are implemented with
+ * native functionality without the need for reflection.
  */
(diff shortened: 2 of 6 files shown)


rhino/src/main/java/org/mozilla/javascript/OwnerAwareLambdaSlot.java
```diff
@@ -0,0 +1,88 @@
+package org.mozilla.javascript;
+
+import java.util.IdentityHashMap;
+import java.util.function.BiConsumer;
+import java.util.function.Function;
+
+/**
+ * A specialized property accessor using lambda functions, similar to {@link LambdaSlot},
```
rPraml, line 8: As far as I understand, the OwnerAwareLambdaSlot is very similar to LambdaSlot. (but code flow is a bit different. see comments)

I made a quick and dirty test, If I replace 
```java
transient Supplier<Object> getter;
transient Consumer<Object> setter;
```
in LambdaSlot with your suggestion
```java
transient Function<Scriptable, Object> getter;
transient BiConsumer<Scriptable, Object> setter;
```
I would need to adjust the existing defineProperty this way.
```java
    public void defineProperty(
            String name, Supplier<Object> getter, Consumer<Object> setter, int attributes) {
        LambdaSlot slot = slotMap.compute(name, 0, ScriptableObject::ensureLambdaSlot);
        slot.setAttributes(attributes);
        slot.getter = (start) -> getter.get();
        slot.setter = (start, value) -> setter.accept(value);
    }
```

What I want to say: Your modified `OwnerAwareLambdaSlot` can cover the whole functionality of `LambdaSlot`

We only require to convert `Supplier<Object>` into a `Function<Scriptable, Object>` and  respectively `Consumer<Object>` into a `BiConsumer<Scriptable, Object>`

The question is, what is more important: Having two classes with nearly the same logic, and save some CPU cycles not compiling an additional lambda or having only one class which gurantees, that code does not diverge. (which already happen)

(!) I have no clear oppinion about this, so maybe wait, if there is an additional feedback

Maybe there is a third option. We could extend the existing `LambdaSlot` this way:

```java
transient Supplier<Object> getter;
transient Function<Scriptable, Object> ownerAwareGetter;
transient Consumer<Object> setter;
transient BiConsumer<Scriptable, Object> ownerAwareSetter;
```
This would only require 2-3 additional null checks, which is IMHO an acceptable tradeoff of additional complexity and CPU cycles

tests/src/test/java/org/mozilla/javascript/tests/OwnerAwareLambdaSlotTest.java
```diff
@@ -74,7 +74,7 @@ public void testCanUpdateValueUsingSetter() {
         Object setResult = cx.evaluateString(scope, "s.status = 'DONE';", "source", 1, null);
 
         Object newStatus = cx.evaluateString(scope, "s.status", "source", 1, null);
-        assertEquals( "NewStatus: DONE", newStatus);
+        assertEquals("NewStatus: DONE", newStatus);
```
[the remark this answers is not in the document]
nabacg: I'm sure I've run `./gradlew spotlessApply` before raising the PR, but now I see that I was running Java 17 and spotlessApply seems to do nothing?  I switched to Java 11 and that made the difference [d5cafc7](https://github.com/mozilla/rhino/pull/1577/commits/d5cafc7b4e55d6506e9c267f8d79220f1dbed63f).   

I guess this is mentioned [here](https://github.com/mozilla/rhino?tab=readme-ov-file#requirements), rookie mistake!

rhino/src/main/java/org/mozilla/javascript/OwnerAwareLambdaSlot.java
nabacg: Perfectly reasonable point to combine both into single class, although I also don't have strong opinion either way. I wasn't sure what the original intention behind `LambdaSlot` was, so it felt simpler to add another class.

I would welcome more feedback.

rhino/src/main/java/org/mozilla/javascript/OwnerAwareLambdaSlot.java
```diff
@@ -0,0 +1,88 @@
...
+        return desc;
+    }
+
+    @Override
+    public boolean setValue(Object value, Scriptable scope, Scriptable owner, boolean isStrict) {
+        if (setter != null) {
+            setter.accept(owner, value);
+            return true;
+        }
+
```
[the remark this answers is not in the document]
nabacg: Well I was trying to improve it slightly be implementing the can't set getter only property error (in strict mode only):  https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/Getter_only

if you don't like it, sure I can align them.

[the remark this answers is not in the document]
rPraml: > if you don't like it...

No, I didn't dislike it. I was also wondering, why LambdaSlot behaves different now.

As far as I understand, your intention in this PR is, to define also properties, where you can access 'thisObj' (very similar to the existing code, that uses the LambdaSlot) right?

IMHO both code paths should behave the same. So if OwnerAwareLambda slot throws an exception in strict mode, LambdaSlot should do the same. 

Or is there any difference between the two cases, that I didn't get, yet?

(Please note: I'm no maintainer, only contributor, so my comments might be totally wrong 🤣)

tests/src/test/java/org/mozilla/javascript/tests/OwnerAwareLambdaSlotTest.java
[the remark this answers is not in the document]
rPraml: This also happens to me in every new console window, when I've forgotten to change JAVA_HOME 🤣

gbrail:
This is a very complete PR! I was working on lambda stuff and had a need for the exact same thing, so I'm glad you did it! Now we need to talk about the details:

Today, there are a few varieties of ScriptableObject.defineProperty, and a few Slot subclasses.

1. Just takes a raw value and sets it directly, so that's not interesting here ;-). Uses a regular Slot.
2. One (actually "defineOwnProperties") takes a Scriptable descriptor and does just what it says in the spec, looking at attributes and supporting attaching Function objects to get and set the value. May use a Slot or AccessorSlot, which calls the getter and setter as if they were JavaScript functions.
4. One takes a Java class and can use reflection to create wrappers for various functions by calling the next method...
5. One takes a pair of Method objects and uses Java reflection to create wrappers, and creates an AccessorSlot
6. One takes a Getter and a Setter interface and is optimized for lambdas, and creates a LambdaSlot
7. One takes a single Callable and I'm not sure where we use it, even though I think I wrote it!

#6 behaves differently from #2 through #5 in that the LambdaSlot is invisible to the end user -- the property value is get and set as if it were just stored natively by Rhino. #2 to #5, on the other hand, allow the user to introspect the property and get at the JavaScript functions to "get" and "set" the value.

I'm saying this because what it looks like you did here is create something that looks to the user like the AccessorSlot was used, but it's more optimized for lambda functions. However, it's different from the existing LambdaSlot in that the resulting descriptor looks like the setter and getter were implemented as Java functions and not totally in a native way.

Also, I see that in the descriptor you're actually creating a new LambdaFunction on the fly every time -- I worry not only about performance but if that will pass the sniff test for all of test262, since sometimes we check that the same function is returned every time from a property, and here we generate a new function every time "getOwnPropertyDescriptor" is called.

I'd really like you to finish this (and I want to use it for stuff that I want to do), so can I suggest a few things:

1) Create the getter and setter "LambdaFunction" instance when the property is set -- once they're visible in in the property descriptor, people might call them, so we should at least only create them once, and then possibly even invoke that function through the Callable interface on actual get and set calls to make sure that we did it right. (I'm not totally sold on the last part, but I am on the first part.)
2) Instead of calling it "OwnerAwareLambdaSlot," we should instead call it something like "LambdaAccessorSlot".
3) If you look at the implementation of "defineOwnProperty," you'll see that it does additional validation, like making sure that the slot is writable or configurable or whatever. I know for now we'll only use this to define properties on prototypes, but in the interest of long-term success we might want to include some of that logic too.

Finally, I think that we ALSO will eventually need a version of LambdaSlot that works exactly like the current one, but passes "self." The interface would look just like this one but it would instead not show up to the user as an accessor property. I think that we could do that by changing the things that LambdaSlot calls and adding a wrapper for backward compatibility.

Thanks for reading a long comment!

p-bakker:
Reading the comment of @gbrail above, I wonder if it is in any way related to #1549 which is about some of the JavaScript APIs related to propertyDescriptors not working on Java objects

nabacg:
Thanks for a review @p-bakker, @gbrail and for comprehensive explanation of all the slot varieties! I just pushed changes for the easy stuff, renaming the slot and making sure I don't re-create LambdaFunction instances for each getOwnPropertyDescriptor call.  I'll need to think how I could incorporate validation like [checkPropertyChange](https://github.com/mozilla/rhino/blob/e0999319ad50b98f352bb8260f3bc39af379b94a/rhino/src/main/java/org/mozilla/javascript/ScriptableObject.java#L1743) into this, hopefully will have some time to work on it this weekend. 


@gbrail regarding the [topic](https://github.com/mozilla/rhino/pull/1577#discussion_r1727416868) of combining LambdaSlot with LambdaAccessorSlot, it sounds like you're saying they should stay as separate (due to one being hidden from introspection, etc) but LambdaSlot might need to evolve to support `self` as well. Did I get this right?

nabacg:
Finally, I'm happy to continue working on everything listed here, but does it all need to be addressed in a single PR? :)

p-bakker:
When I referred to #1549, I wasn't very clear on how I see it potentially being related :-)

I meant to just keep in the back of our minds the issues reported in #1549 and make sure whatever we do with slots either improves the issues reported in #1549 or otherwise at least won't make it harder to fix those issues in the future

gbrail:
Yes -- I think that "LambdaSlot" and "LambdaAccessorSlot" should stay separate. However, in the future I can imagine that we might need a version of LambdaSlot that passes the "this" value, but since we don't have that use case today, why don't we "YAGNI" and wait until we do -- whereas I know that we need this PR so that we can finish the implementation of Symbol. Thanks!

nabacg:
Ok, I pushed a few more commits, going through the short list @gbrail provided:

> Create the getter and setter "LambdaFunction" instance when the property is set -- once they're visible in in the property descriptor, people might call them, so we should at least only create them once, and then possibly even invoke that function through the Callable interface on actual get and set calls to make sure that we did it right. (I'm not totally sold on the last part, but I am on the first part.)

I didn't go as far as the last part, i.e. accessing property via Callable interface, but everything else is done. 

> Instead of calling it "OwnerAwareLambdaSlot," we should instead call it something like "LambdaAccessorSlot".

Renamed.  

>If you look at the implementation of "defineOwnProperty," you'll see that it does additional validation, like making sure that the slot is writable or configurable or whatever. I know for now we'll only use this to define properties on prototypes, but in the interest of long-term success we might want to include some of that logic too.

Changed new `defineProperty` to call both 
 - [checkPropertyDefinition](https://github.com/mozilla/rhino/pull/1577/files#diff-86fa44d9df4f5e093d256383808bde55181e04b4b193694f2ed1522cd7901e30R1762) 
 - [checkPropertyChange](https://github.com/mozilla/rhino/pull/1577/files#diff-86fa44d9df4f5e093d256383808bde55181e04b4b193694f2ed1522cd7901e30R1765)

in similar fashion as `defineOwnProperty` does and I added couple of [tests](https://github.com/mozilla/rhino/pull/1577/files#diff-1614b0516be905f8d504a3b0000dc35bb6ac575171175fe10305acb0e8c3aa26R287) to assert we go through those validations. 

I tried to bring `LambdaAccessorSlot` to look and behave similar to `AccessorSlot`, like throwing the same errors and setting those configurable, enumerable, etc properties [here](https://github.com/mozilla/rhino/pull/1577/files#diff-1b97f522ed50edebdf8504b9d46065246637563b0e6a370284c35b5e772ed954R69).  

Overall the feedback that `LambdaAccessorSlot` should behave more like `AccessorSlot` not `LambdaSlot` was super helpful.

rhino/src/main/java/org/mozilla/javascript/ScriptableObject.java
```diff
@@ -2695,6 +2781,17 @@ private static LambdaSlot ensureLambdaSlot(Object name, int index, Slot existing
         }
     }
 
+    private static LambdaAccessorSlot ensureLambdaAccessorSlot(
```
gbrail, line 2784: Looks like we're not using this method any more... FWIW, I was using this pattern as a simple way to switch slot classes in a clean way, and then doing the calculation after that, so the way I did this in other methods is a bit different from what you did, but it's all good.

But this is unused code so we should delete it!

gbrail (changes requested):
Looking good -- just an unused function crept in there. Thanks!

rhino/src/main/java/org/mozilla/javascript/ScriptableObject.java
nabacg: Thanks, I missed that one. It's gone.

nabacg: > I was using this pattern as a simple way to switch slot classes in a clean way, and then doing the calculation after that, so the way I did this in other methods is a bit different from what you did, but it's all good.

The version of`ensureLambdaAccessorSlot` I ended up using is closer to this [lambda function](https://github.com/mozilla/rhino/blob/7369f3fcd16d7925b9ca0d43e96100b1708cbdff/rhino/src/main/java/org/mozilla/javascript/ScriptableObject.java#L1614) in `defineOwnProperty`. It's not just (re)creating slot classes, but also doing validation, etc. If you think it's doing to much in a single function or could be named better, I'm happy to refactor this too @gbrail . 


Although maybe in next PR? :)

rhino/src/main/java/org/mozilla/javascript/ScriptableObject.java
```diff
@@ -1690,6 +1691,64 @@ public void defineProperty(
...
     }
 
+    /**
+     * Define a property on this object that is implemented using lambda functions accepting
+     * Scriptable `this` object as first parameter. Unlike with `defineProperty(String name,
+     * Supplier<Object> getter, Consumer<Object> setter, int attributes)` where getter and setter
+     * need to have access to target object instance, this allows for defining properties on
+     * LambdaConstructor prototype providing getter and setter logic with java instance methods. If
+     * a property with the same name already exists, then it will be replaced. This property will
+     * appear to the JavaScript user exactly like any other property -- unlike Function properties
```
gbrail, line 1701: I'm sorry I keep finding things, but now this comment is out of date-- as you've defined it, and unlike the other "defineProperty" that takes  Getter and a Setter, this property will not look like a normal value, but like a descriptor with a getter and setter, just as if they had been defined in JavaScript using Object.defineOwnProperty. Could you please update? I think this will be the last thing!

nabacg: Of course, it's not great to leave outdated comments! Especially in this case where it can already be pretty confusing which `defineProperty` should be used. I should have noticed this myself, thanks for being patient with me!

rhino/src/main/java/org/mozilla/javascript/LambdaAccessorSlot.java
```diff
@@ -0,0 +1,129 @@
...
+
+    @Override
+    ScriptableObject getPropertyDescriptor(Context cx, Scriptable scope) {
+        ScriptableObject desc = (ScriptableObject) cx.newObject(scope);
+
+        int attr = getAttributes();
+        boolean es6 = cx.getLanguageVersion() >= Context.VERSION_ES6;
+        if (es6) {
+            if (getterFunction == null && setterFunction == null) {
+                desc.defineProperty(
```
p-bakker, line 49: Not 100% sure, but doesn't the pd need a 'value' property as well if it won't have getter?

nabacg: Good question. I pretty much followed how [AccessorSlot](https://github.com/mozilla/rhino/blob/b7ece52d83a0d558b28b5962c63e8eea31f0ecd8/rhino/src/main/java/org/mozilla/javascript/AccessorSlot.java#L45) implements this. 

Having said that, trying below

``` js 

> const o = {};
undefined

>Object.defineProperty(o, "b", {
		enumerable: true,
		configurable: true
});
{b: undefined}


> Object.getOwnPropertyDescriptor(o, "b")
{value: undefined, writable: false, enumerable: true, configurable: true}
```

 in both Chrome and nodejs does return `value: undefined` so perhaps we should add it?

p-bakker:
Am pretty sure it ought to behave like you've seen in Chrome 

Rhino unfortunately has a bunch of issues In the area of property descriptors, but if we can prevent introducing a new one... 🙂

nabacg:
> Am pretty sure it ought to behave like you've seen in Chrome
> 
> Rhino unfortunately has a bunch of issues In the area of property descriptors, but if we can prevent introducing a new one... 🙂

If I'm going to make that change, it would probably make sense to change it in both `LambdaAccessorSlot` and `AccessorSlot`. I would appreciate @gbrail's thoughts, since it was his original suggestion that both classes should be aligned.

gbrail:
FWIW I think that changing how property descriptors in general is out of
the scope of this PR and if we want to change it, we should do it in
another PR.

p-bakker:
@gbrail you mean not fix the bug in the new slot impl. that got copied over from another slot implementation?

Or that this PR shouldn't be extended to fix the other pd implementation issues?

gbrail:
I think it's fine if this PR extends the pattern that's already in place for the existing "AccessorSlot," and that we merge this and track any inconsistencies that we think there are in the existing implementation, rather than have two ways to do the same basic thing that produce different results.

Also someone will have to explain to me the difference between having a property present but with a value of "undefined" and a value that is not present, but it's JavaScript so I'm sure it's in there somewhere!

nabacg:
I pushed one more commit updating tests to `Utils.runWithAllOptimizationLevels`  as suggested by @rPraml in other PRs. Is there anything else outstanding for this PR?

p-bakker:
> Also someone will have to explain to me the difference between having a property present but with a value of "undefined" and a value that is not present, but it's JavaScript so I'm sure it's in there somewhere!

`somePDInstance.value` will return undefined in both cases, but these will display different behavior:
- somePDInstance.hasOwnProperty/Object.hasOwn: false vs. true
- Object.getPropertyDescriptor: undefined vs. Object
- Object.getPropertyDescriptors/Object.keys/Object.entries: missing entry for value property PD vs including a PD for the value property
- forOf/forIn loops won't include the missing value property

I'm sure I missed a few scenarios :-)

p-bakker:
@nabacg Could you make a follow-up case regarding the missing value property with all relevant info you have and tagit with the `PropertyDescriptors` label?

nabacg:
Issue #1601, I don't think I have permissions to assign a label.

p-bakker:
>  I don't think I have permissions to assign a label.

Ah, didn't know that was restricted, added them now

Thx for this follow-up!

gbrail (approved):

gbrail:
Thanks for everyone's work on this!

PR #1578, opened 2024-08-22
Fixed a bug with interpreter peeling introduced in #1510

andreabergia:
I found a bug with the interpreter peeling introduced in #1510.

The code has an `if (fun instanceof LambdaFunction)` that does not work for `LambdaConstructor`. These are a subclass of `LambdaFunction`, so the `instanceof` check passes, but they do not have the `target` from the base class filled, so we get an error.

Added also an unit test.


tests/src/test/java/org/mozilla/javascript/tests/InterpreterFunctionPeelingNonRegressionTest.java
```diff
@@ -0,0 +1,86 @@
...
+                                    new Person(ScriptRuntime.toString(args[0])));
+            scope.put("makePerson", scope, makePerson);
+
+            String source =
+                    "function testLambdaFunction() {\n"
+                            + "   return makePerson('Andrea');\n"
+                            + "}\n"
+                            + "testLambdaFunction().name";
+            Object value = cx.evaluateString(scope, source, "test", 1, null);
+            assertEquals("Andrea", value);
```
rbri, line 36: Ok, its about the interpreter, but i think we should run as many tests as possible with all optimization levels.

gbrail: Agreed, not much of a risk to use one of the methods to "runWithAllOptimizationLevels".

gbrail:
Thanks for catching this -- I ran in to the very same issue when trying to "lambdaize" some other classes. Could you please address the comments on running the test at many levels, in case this "white box test" ends up being important in the future?

tests/src/test/java/org/mozilla/javascript/tests/InterpreterFunctionPeelingNonRegressionTest.java
rbri: There are also some assert... methods in the Utils class, maybe you can also use one of these to make this simpler
eg. assertWithAllOptimizationLevels

andreabergia:
Actually I ended up modifying the pre-existing `LambdaFunctionTest` so that it runs in all modes. That reproduces the bug, and it is a lot better than adding a new test.

tests/src/test/java/org/mozilla/javascript/tests/InterpreterFunctionPeelingNonRegressionTest.java
rbri: Thanks

gbrail:
Great solution. Thanks!

PR #1580, opened 2024-08-22
make sure to use only the RegExpProxy interface

rbri:
make sure to use only the RegExpProxy interface instead of some hardcoded instanceof

This is the successsor of PR  #1419


gbrail:
Thanks! It looks like that new "archunit" plugin needs an SLF4J provider to print its log messages, so I'm going to add that and commit this manually, hold on...

gbrail:
Merged manually. Thanks!

PR #1581, opened 2024-08-23
Implement ES2024 `groupBy`

### LINKED ISSUE #1545: ES2024 Array Grouping
https://github.com/tc39/proposal-array-grouping

camnwalter:
Resolves #1545

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/AbstractEcmaObjectOperations.java
+++ b/rhino/src/main/java/org/mozilla/javascript/AbstractEcmaObjectOperations.java
@@ -1,5 +1,10 @@
 package org.mozilla.javascript;
 
+import java.util.ArrayList;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+
 /**
  * Abstract Object Operations as defined by EcmaScript
  *
@@ -22,6 +27,11 @@ public class AbstractEcmaObjectOperations {
     enum INTEGRITY_LEVEL {
         FROZEN,
         SEALED
+    }
+
+    enum KEY_COERCION {
+        PROPERTY,
+        COLLECTION,
     }
 
     /**
@@ -231,5 +241,66 @@ public class AbstractEcmaObjectOperations {
         } else {
             base.put(p, o, v);
         }
+    }
+
+    /**
+     * Implement the ECMAScript abstract operation "GroupBy" defined in section 7.3.35 of ECMA262.
+     *
+     * @param cx
+     * @param scope
+     * @param items
+     * @param callback
+     * @param keyCoercion
+     * @see <a href="https://tc39.es/ecma262/#sec-groupby"></a>
+     */
+    static Map<Object, List<Object>> groupBy(
+            Context cx,
+            Scriptable scope,
+            IdFunctionObject f,
+            Object items,
+            Object callback,
+            KEY_COERCION keyCoercion) {
+        if (cx.getLanguageVersion() >= Context.VERSION_ES6) {
+            ScriptRuntimeES6.requireObjectCoercible(cx, items, f);
+        }
+        if (!(callback instanceof Callable)) {
+            throw ScriptRuntime.typeErrorById(
+                    "msg.isnt.function", callback, ScriptRuntime.typeof(callback));
+        }
+
+        // LinkedHashMap used to preserve key creation order
+        Map<Object, List<Object>> groups = new LinkedHashMap<>();
+        final Object iterator = ScriptRuntime.callIterator(items, cx, scope);
+        try (IteratorLikeIterable it = new IteratorLikeIterable(cx, scope, iterator)) {
+            double i = 0;
+            for (Object o : it) {
+                if (i > NativeNumber.MAX_SAFE_INTEGER) {
+                    it.close();
+                    throw ScriptRuntime.typeError("Too many values to iterate");
+                }
+
+                Object[] args = {o, i};
+                Object key =
+                        ((Callable) callback).call(cx, scope, Undefined.SCRIPTABLE_UNDEFINED, args);
+                if (keyCoercion == KEY_COERCION.PROPERTY) {
+                    if (!ScriptRuntime.isSymbol(key)) {
+                        key = ScriptRuntime.toString(key);
+                    }
+                } else {
+                    assert keyCoercion == KEY_COERCION.COLLECTION;
+                    if ((key instanceof Number)
+                            && ((Number) key).doubleValue() == ScriptRuntime.negativeZero) {
+                        key = ScriptRuntime.zeroObj;
+                    }
+                }
+
+                List<Object> group = groups.computeIfAbsent(key, (k) -> new ArrayList<>());
+                group.add(o);
+
+                i++;
+            }
+        }
+
+        return groups;
     }
 }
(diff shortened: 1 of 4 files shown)


rhino/src/main/java/org/mozilla/javascript/AbstractEcmaObjectOperations.java
```diff
@@ -232,4 +242,65 @@ static void put(Context cx, Scriptable o, int p, Object v, boolean isThrow) {
...
+        if (!(callback instanceof Callable)) {
+            throw ScriptRuntime.typeErrorById(
+                    "msg.isnt.function", callback, ScriptRuntime.typeof(callback));
+        }
+
+        // LinkedHashMap used to preserve key creation order
+        Map<Object, List<Object>> groups = new LinkedHashMap<>();
+        final Object iterator = ScriptRuntime.callIterator(items, cx, scope);
+        try (IteratorLikeIterable it = new IteratorLikeIterable(cx, scope, iterator)) {
+            double k = 0;
```
gbrail, line 275: Note to self for the future -- I wonder if there's a way that some day we can take this code and others like it and optimize it to use integers except in the rare case where we have more than 2^31 elements in an array. (And BTW I'm not so sure that we can handle more than 2^31 elements in an array anyway...)

rhino/src/main/java/org/mozilla/javascript/AbstractEcmaObjectOperations.java
```diff
@@ -232,4 +242,65 @@ static void put(Context cx, Scriptable o, int p, Object v, boolean isThrow) {
...
+                } else {
+                    assert keyCoercion == KEY_COERCION.COLLECTION;
+                    if ((key instanceof Number)
+                            && ((Number) key).doubleValue() == ScriptRuntime.negativeZero) {
+                        key = ScriptRuntime.zeroObj;
+                    }
+                }
+
+                groups.putIfAbsent(key, new ArrayList<>());
+                groups.get(key).add(o);
```
gbrail, line 298: You can (and should) do this with only one trip through the hash table -- use "computeIfAbsent", which will return either the new ArrayList or the existing value, and then you can add directly to that, like this (give or take some syntax):

a = groups.computeIfAbsent(key, (k) -> new ArrayList<>());
a.add(o)

That saves you a hash lookup on every iteration which will be significant.

rhino/src/main/java/org/mozilla/javascript/NativeObject.java
```diff
@@ -687,6 +689,38 @@ && isEnumerable(stringId, sourceObj)) {
...
+                    Map<Object, List<Object>> groups =
+                            AbstractEcmaObjectOperations.groupBy(
+                                    cx,
+                                    scope,
+                                    f,
+                                    items,
+                                    callback,
+                                    AbstractEcmaObjectOperations.KEY_COERCION.PROPERTY);
+
+                    ScriptableObject obj = new NativeObject();
```
gbrail, line 706: I like to just use "cx.newObject" here but I doubt that it makes a difference.

gbrail (changes requested):
Thanks! I saw things that could be better but this is a great submission and I'm looking forward to having it in the product!

rhino/src/main/java/org/mozilla/javascript/AbstractEcmaObjectOperations.java
```diff
@@ -232,4 +242,65 @@ static void put(Context cx, Scriptable o, int p, Object v, boolean isThrow) {
...
+        if (cx.getLanguageVersion() >= Context.VERSION_ES6) {
+            ScriptRuntimeES6.requireObjectCoercible(cx, items, f);
+        }
+        if (!(callback instanceof Callable)) {
+            throw ScriptRuntime.typeErrorById(
+                    "msg.isnt.function", callback, ScriptRuntime.typeof(callback));
+        }
+
+        // LinkedHashMap used to preserve key creation order
+        Map<Object, List<Object>> groups = new LinkedHashMap<>();
```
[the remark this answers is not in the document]
camnwalter: The spec says it is a list of records and appends a new group each time. I just used a map as it seemed a little easier to implement that way. Haven't researched other engines yet but I'd assume they follow it.

https://tc39.es/ecma262/#sec-add-value-to-keyed-group

rhino/src/main/java/org/mozilla/javascript/NativeObject.java
```diff
@@ -687,6 +689,37 @@ && isEnumerable(stringId, sourceObj)) {
...
+                                    callback,
+                                    AbstractEcmaObjectOperations.KEY_COERCION.PROPERTY);
+
+                    NativeObject obj = (NativeObject) cx.newObject(scope);
+                    obj.setPrototype(null);
+
+                    for (Map.Entry<Object, List<Object>> entry : groups.entrySet()) {
+                        Scriptable elements = cx.newArray(scope, entry.getValue().toArray());
+
+                        ScriptableObject desc = (ScriptableObject) cx.newObject(scope);
```
gbrail, line 712: One last nit -- you can (and should please) replace lines 712-717 with:

obj.defineProperty(entry.getKey(), elements, ScriptableObject.PERMANENT);

This is not only less code but it's more efficient, because we're not creating a new object and messing around in a hash table to figure out what attributes to set.

gbrail (changes requested):
Just one last final thing, but I'm seeing an opportunity to make this code very clean and tight. Thanks!

rhino/src/main/java/org/mozilla/javascript/NativeObject.java
camnwalter: The only issue with that is that the entry's key is an Object, which `defineProperty` doesn't allow for the first parameter.

gbrail: OK -- good point -- let's merge this!

gbrail (approved):

rhino/src/main/java/org/mozilla/javascript/AbstractEcmaObjectOperations.java
```diff
@@ -232,4 +242,65 @@ static void put(Context cx, Scriptable o, int p, Object v, boolean isThrow) {
...
+
+    /**
+     * Implement the ECMAScript abstract operation "GroupBy" defined in section 7.3.35 of ECMA262.
+     *
+     * @param cx
+     * @param scope
+     * @param items
+     * @param callback
+     * @param keyCoercion
+     * @see <a href="https://tc39.es/ecma262/#sec-groupby"></a>
```
p-bakker, line 254: Sorry, just seeing this now: the link to the spec doesn't reference a specific version of the spec, which it should 😐

rhino/src/main/java/org/mozilla/javascript/AbstractEcmaObjectOperations.java
```diff
@@ -232,4 +242,65 @@ static void put(Context cx, Scriptable o, int p, Object v, boolean isThrow) {
             base.put(p, o, v);
         }
     }
+
+    /**
+     * Implement the ECMAScript abstract operation "GroupBy" defined in section 7.3.35 of ECMA262.
```
p-bakker, line 247: And no need to reference the section number here, as they change from version to version

camnwalter:
I'll make a separate pr addressing those couple changes if you want

p-bakker:
Thnx, sry, should've spotted be it earlier

PR #1582, opened 2024-08-23
Add a benchmark to measure Rhino startup time

gbrail:
I'm a bit worried about changes I'm making with lambdas and how that will impact how long it will take Rhino to initialize all the standard objects, so let's create a microbenchmark for that.


gbrail:
In case anyone is keeping score, yes, the changes to lambda objects will slow down startup time because they create big old hash tables when "initStandardObjects" is called, although they perform as well or better later. So far we're talking about going from 17 microseconds to 22 or 23, which is not a big deal unless people start to initialize Rhino a lot, but we should keep an eye out and possibly increase usage of the "lazy" feature that we already use for these things...

Issue #1584, opened 2024-08-24
Several line break related issues when running the test suite on windows


rbri:
For other reasons i do most of the work on this lib on windows. There are several tests that failing due to problems with line breaks.

* tests\test262\test\built-ins\Function\prototype\toString\line-terminator-normalisation-LF.js
this fails because the checkout created cr lf line as line terminator and the test expects the original lf to be preserved

We can add this line to the .gitattributes in folder tests\test262

    line-terminator-normalisation-LF.js text eol=lf

but i'm not sure if this is the right way to do that

@p-bakker @gbrail what do you think?

p-bakker:
Great! See #919 and #918 for some, possibly related, prior art

rbri:
For me this is  done. Will open a new one if something more comes up

PR #1585, opened 2024-08-24
force LF as newLine character to let out test suite pass on windows also

rbri:
This solves all the issues i have on windows when running our test suite. There is only one test from the test262 suite that still fails and something unicode related. Will have a look at this later and maybe prepare a separate PR.

This also fixes https://github.com/mozilla/rhino/issues/918 - did a short test and it looks like spotless leaves the LF untouched.

@gbrail having this merged will be a great help - i already started with the Proxy & Reflect update and therefor having a working test suite is essential

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -5,8 +5,9 @@
 .project text
 *.html text
 *.htm text
-*.java text
-*.js text
+*.java text eol=lf
+*.js text eol=lf
+*.jstest text eol=lf
 *.jsp text
 *.php text
 *.properties text


p-bakker:
I assume this will choose #1584?

gbrail:
Yes, this is OK and also works for me (except the Unicode thing which I appreciate you looking at. Thanks!)

rbri:
> (except the Unicode thing which I appreciate you looking at. Thanks!)

Will do, this is the next step...

PR #1590, opened 2024-08-29
Use UTF-8 decoding for java source files

tonygermano:
This is the default on Linux and Mac, but not on Windows. This should standardize across platforms. I believe this will also allow us to revert https://github.com/mozilla/rhino/pull/1588.

Without ensuring that UTF-8 encoding is always expected, any characters in the java source with a code point above 127 will be read incorrectly by Windows. While we could escape all of those characters, one must remember to do it, and escaping those characters harms readability.


gbrail:
Yes, I think this makes sense and is a better way to ensure our Windows builds work. Thanks!

Issue #1279, opened 2022-11-10
Update Shell to use Context.VERSION_ES6 by default (instead of Context.VERSION_1_8)


p-bakker:
Any objections?

MaxisTekfield:
Sounds good.

TheRohinth:
Hi @p-bakker , Greetings for the day.

I look forward for this enhancement of using Context.VERSION_ES6 by default.
When will this be completed ? 
Will this enhancement be delivered in the next version of rhino ?

As I need to use Promises in Rhino-1.7.14
1. I can't just add rhino in dependencies task
2. I need to execute rhino with -version 200 in order to make it support promises
3. but I couldn't configure the -version flag for the same in build.gradle.

I tried to create a task in build.gradle which has the type "JavaExec" that executes the rhino-1.7.14.jar with 
"-version 200" argument. But it did not help.

Do you think this approach would help me ?
If yes, Could you help me with this approach ? 
(if this works out, it will be helpful for me until Context.VERSION_ES6 is made default)

Thanks in advance!

Issue #1060, opened 2021-10-14
Introduce new Context.JAVA_INTEROP flag and move relevant functionality behind this flag


p-bakker:
So we can have a clean Context.VERSIoN.ECMASCRIPT that doesn't allow Java Interop by default

Issue #1057, opened 2021-10-14
Introduce new Context.ANNEX_B flag and move relevant functionality behind this flag


p-bakker:
As Rhino started out as a port of SpiderMonkey, which was the JavaScript Engine for FireFox, some stuff got implemented that later on got moved in the Annex B of the EcmaScript sepcification, which should only be implemented by browsers and not by non-browser-based JavaScript engines, like Rhino

This case is to introduce the Context.ANNEX_B flag and move all functionality/behavior that is not part of the core EcmaScript specification, but is part of Annex B behind this new flag.

In v2.0.0 this flag should be disabled by default. We could opt to implement this flag alreadt in v1.7.15 and have it set to true by default (if Context.VERSION_ECMASCRIPT is not being used)

Issue #1056, opened 2021-10-14
Identify all non-EcmaScript behavior and extentions to disable/remove/make optional in v2.0.0


p-bakker:
In v2.0.0 we want to, by default, run Rhino with Context.VERSION.ECMASCRIPT (see #1055), which should give the user a clean, EcmaScript compliant JavaScript engine, with, by default, no (api) extensions

In order to achieve that goal, all non-standard behavior and API extensions need to be identifier.

To aid in this, one of the goals for v1.7.15 is to be able to run the latest version of the Test262 suite, so we can get good feedback on how standard-compliant Rhino is. Note that standard-compliant doesn't necessarily mean feature completeness.

Test262 however doesn't not report on non-compliant API extensions, so those need to be identified through other means.

See the [Next steps for Rhino after the v1.7.14 release](https://github.com/mozilla/rhino/discussions/972) discussion for already identifier things

## (incomplete) list of incompatibilites
- [ ] cases tagged with [non-Ecma](https://github.com/mozilla/rhino/issues?q=is%3Aissue+is%3Aopen+label%3A%22non-Ecma%22+)
- [ ] [Nashorn Compatibility Script](https://wiki.openjdk.java.net/display/Nashorn/Rhino+Migration+Guide#:~:text=the%20JavaScript%20method.-,Compatibility%20script,-There%20are%20few)
- [ ] [MDN New in JavaScript](https://github.com/mdn/archived-content/tree/main/files/en-us/archive/web/javascript/new_in_javascript)
- [ ] [Expression Closures](https://github.com/mdn/archived-content/tree/main/files/en-us/archive/web/javascript/expression_closures)
- [ ] [Conditional catch clause](https://esdiscuss.org/topic/conditional-catch-clause)

PR #1594, opened 2024-08-29
Update Shell to use Context.VERSION_ES6 by default

tonygermano:
resolves https://github.com/mozilla/rhino/issues/1279

### DIFF
--- a/rhino-tools/src/main/java/org/mozilla/javascript/tools/shell/ShellContextFactory.java
+++ b/rhino-tools/src/main/java/org/mozilla/javascript/tools/shell/ShellContextFactory.java
@@ -13,7 +13,7 @@ import org.mozilla.javascript.ErrorReporter;
 public class ShellContextFactory extends ContextFactory {
     private boolean strictMode;
     private boolean warningAsError;
-    private int languageVersion = Context.VERSION_1_8;
+    private int languageVersion = Context.VERSION_ES6;
     private int optimizationLevel;
     private boolean generatingDebug;
     private boolean allowReservedKeywords = true;
--- a/tests/src/test/java/org/mozilla/javascript/tests/MozillaSuiteTest.java
+++ b/tests/src/test/java/org/mozilla/javascript/tests/MozillaSuiteTest.java
@@ -21,6 +21,7 @@ import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.junit.runners.Parameterized;
 import org.junit.runners.Parameterized.Parameters;
+import org.mozilla.javascript.Context;
 import org.mozilla.javascript.drivers.JsTestsBase;
 import org.mozilla.javascript.drivers.ShellTest;
 import org.mozilla.javascript.drivers.TestUtils;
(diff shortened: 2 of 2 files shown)


tonygermano:
Of course it couldn't have been that easy.

gbrail:
Sorry, I didn't think it would be!

tonygermano:
MozillaSuiteTest uses ShellContextFactory (as do some other tests,) but did not explicitly set the version to 1.8, so I'm pushing a change to do that now. Of the handful of failing tests that I looked at, they all used legacy features that are not part of the current javacsript spec (java 1.7 generators and let expressions.)

I just did a search, and there are quite a few other tests that reference Context.VERSION_1_8. At some point should we try to update what we can to run with VERSION_ES6 or just leave the old tests at the old language versions and expect newer tests and test262 to cover everything in ES6+?

tonygermano:
I think this one is ready now after my last commit if someone wants to review.

gbrail:
Doing a sanity test myself, but I think that this looks good.

@p-bakker do you think that we can have some way to flag PRs / commits that will need extra attention in the release notes?

And is this a sufficient reason why the next release should be called "1.8.0"?

And should we explore making "VERSION_ES6" the default version for everything?

gbrail:
I think that this is good. I'm going to leave this open for a few days in case any active users want to try it out themselves.

p-bakker:
Tagged the PR with the 'Potential breaking change' label, so we can look for those when making the release notes and include a warning

iMHO this change doesn't warrant a minor version update, as I doubt there's much usage of the shell these days (based on Google search results the shell used to be used to run all kinds of JS tooling that have long since moved to NodeJS or other runtimes). And if there's usage left, I'd think they either aren't likely to upgrade to a newer Rhino version or will use ES6

As for making ES6 the default: personally I rather make a new language version called EcmaScript, which by default doesn't include all non-standard stuff and make that the default, see #1055 (and related #1056, #1057 & #1060)

p-bakker:
My thought was to create an EcmaScript language version that just had plain EcmaScript, nothing else and then there would be flags to opt-in on additional things, like Java Interop, E4X, Continuations etc.

Whether or not the Shell would have all those flags turned on our of (by default) would be a decision to be made, but I can see it being useful to have them on by default. 

But, this is all just my thoughts, as a group we'd have to come to a consensus how to move forward on things like this 🙂

tonygermano:
I looked at #1055 before I made my last comment, and I like your idea as it applies to embedding, but I do think the shell should at least have java interop turned on by default. Personally, I'd like e4x, too, but if I have to enable that with a command-line option I could live with it (though I also don't see the harm in having it enabled by default in the shell, as it has always been.)  And of course there are the shell-only objects that currently get added which would not be part of plain ecmascript.

But until we have all of that implemented, my hope would be that we could move forward with this simple PR with the understanding that it will likely get changed again down the road.

p-bakker:
> But until we have all of that implemented, my hope would be that we could move forward with this simple PR

Oh, I definitely agree: update to use ES6 now and if we'll get to the other cases at some point, week see them how to proceed with the language version/setup in the shell as well

gbrail:
Yes, this makes sense to me now - thanks!

PR #1595, opened 2024-08-30
Revert "make tests more robust"

gbrail:
Since we changed the build configuration to force UTF-8, we don't need this change any more. Tested it myself on Windows.

Reverts mozilla/rhino#1588


gbrail:
@rbri can you double check that things are still OK on your end with this PR merged?

rbri:
@gbrail double checked +1 for the revert of my change.

Issue #938, opened 2021-06-21
Support ES2020 Nullish Coalescing Operator


p-bakker:
Proposal: https://github.com/tc39/proposal-nullish-coalescing
MDN: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator

p-bakker:
See https://github.com/mozilla/rhino/pull/838/files for a previous operator addition

PR #1597, opened 2024-08-30
Implement logical assigment operators `||=` and `&&=`

### LINKED ISSUE #936: Support ES2021 Nullish Assignment Operator
Proposal: https://github.com/tc39/proposal-logical-assignment
MDN:
1. DONE https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_OR_assignment
2. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_nullish_assignment
3. DONE https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_AND_assignment

andreabergia:
Turned out to be pretty simple 🙂 

I had to re-number all the entries in `Token` because the number for the new tokens has to land between `FIRST_ASSIGN` and `LAST_ASSIGN`

Also, we get two regression in test262 (`lgcl-and-assignment-operator-non-simple-lhs.js` and `lgcl-or-assignment-operator-non-simple-lhs.js`), where the tests does something like this:

```js
// It is a Syntax Error if AssignmentTargetType of LeftHandSideExpression is not simple.
function test() {}
test() ||= 1;
```

Before, the code was throwing a `SyntaxError` because it did not recognize the `||=` operator. Now, it does, and it does not handle correctly the fact that there is a non-simple expression on the left-hand side. However, this behavior is shared with all other assignments operator (e.g., `test() += 32` does not throw a `SyntaxError` either), so I don't think this is a real "regression".

Partially resolves #936

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/IRFactory.java
+++ b/rhino/src/main/java/org/mozilla/javascript/IRFactory.java
@@ -2066,11 +2066,17 @@ public final class IRFactory {
             case Token.ASSIGN_BITOR:
                 assignOp = Token.BITOR;
                 break;
+            case Token.ASSIGN_LOGICAL_OR:
+                assignOp = Token.OR;
+                break;
             case Token.ASSIGN_BITXOR:
                 assignOp = Token.BITXOR;
                 break;
             case Token.ASSIGN_BITAND:
                 assignOp = Token.BITAND;
+                break;
+            case Token.ASSIGN_LOGICAL_AND:
+                assignOp = Token.AND;
                 break;
             case Token.ASSIGN_LSH:
                 assignOp = Token.LSH;
--- a/rhino/src/main/java/org/mozilla/javascript/Node.java
+++ b/rhino/src/main/java/org/mozilla/javascript/Node.java
@@ -932,8 +932,10 @@ public class Node implements Iterable<Node> {
             case Token.ASSIGN_DIV:
             case Token.ASSIGN_MOD:
             case Token.ASSIGN_BITOR:
+            case Token.ASSIGN_LOGICAL_OR:
             case Token.ASSIGN_BITXOR:
             case Token.ASSIGN_BITAND:
+            case Token.ASSIGN_LOGICAL_AND:
             case Token.ASSIGN_LSH:
             case Token.ASSIGN_RSH:
             case Token.ASSIGN_URSH:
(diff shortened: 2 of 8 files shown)


p-bakker:
Great stuff!

What about ??= through? That is also part of the same proposal, see #938 

As for the 'regression': Can you create a case for that (assuming you don't (directly) go off and provide a PR for that one as well :-)

p-bakker:
BEsides the 'regression': seems not all `lgcl-or/and-assignment-opearator-*` are passing.

I just looked at [language/expressions/logical-assignment/lgcl-or-assignment-operator-lhs-before-rhs.js](https://github.com/tc39/test262/blob/d62fa93c8f9ce5e687c0bbaa5d2b59670ab2ff60/test/language/expressions/logical-assignment/lgcl-or-assignment-operator-lhs-before-rhs.js) for example

Aren't those fixable?

andreabergia:
> What about ??= through? That is also part of the same proposal, see #938

I know, but IMHO that is better implemented after https://github.com/mozilla/rhino/pull/1591 is merged. So, I'd prefer opening a new PR after this and that one are merged.


> BEsides the 'regression': seems not all `lgcl-or/and-assignment-opearator-*` are passing.
> 
> I just looked at [language/expressions/logical-assignment/lgcl-or-assignment-operator-lhs-before-rhs.js](https://github.com/tc39/test262/blob/d62fa93c8f9ce5e687c0bbaa5d2b59670ab2ff60/test/language/expressions/logical-assignment/lgcl-or-assignment-operator-lhs-before-rhs.js) for example

All of these problems happen also for existing operators, like `+=`. They should be tracked in proper issues (I'll try to open some), but they are not strictly related to the new operators that this PR adds.

andreabergia:
Opened https://github.com/mozilla/rhino/issues/1598 https://github.com/mozilla/rhino/issues/1599 and https://github.com/mozilla/rhino/issues/1600 which should track the test262 failures for the new operators

gbrail:
I'm OK with merging this and opening new issues and PRs for the other features if others think this is a good PR. (looks good to me, BTW). Out of time today but it won't be long. Thanks!

p-bakker:
This PR looks good to me as well: have no clue how the limited set of changes actually made things work 🤔, but the relevant tests in test262 pass where they can, so apparently they work 🙂 so I'm all for merging this!

gbrail:
Thanks for doing this!

andreabergia:
> This PR looks good to me as well: have no clue how the limited set of changes actually made things work 🤔, but the relevant tests in test262 pass where they can, so apparently they work 🙂 so I'm all for merging this!

Well, very simply: we already have a lot of logic for operators such as `+=` or `|=`. It was super easy to extend it to `||=` because it has the same semantic.

PR #1605, opened 2024-08-31
Fix publishing scripts

gbrail:
Going to see if we can publish snapshots to GitHub. (Maven Central no longer supports them.)


p-bakker:
This related to #980 I assume?

Issue #477, opened 2018-09-21
getLineNumbers misses the first line


OLSander:
The line numbers returned by DebuggableScript#getLineNumbers do not include the first line of a script if the first line is a debuggable line.

Reproducible with 1.7.9:
```
public class Main {

	public static void main(String[] args) {
		printLineNumbers("var a = 123;");
		printLineNumbers("\nvar a = 123;");
	}

	private static void printLineNumbers(String source) {
		ContextFactory.getGlobal().call(new ContextAction() {

			@Override
			public Object run(Context context) {
				context.setGeneratingDebug(true);
				context.setOptimizationLevel(-1);
				Script script = context.compileString(source, null, 0, null);
				int[] lineNumbers = Context.getDebuggableView(script).getLineNumbers();
				System.out.println(Arrays.toString(lineNumbers));
				return null;
			}
		});
	}
}
```

Expected output:
```
[0]
[1]
```

Actual output:
```
[]
[1]
```

I believe this is related to the "lineno != lineNumber" condition in CodeGenerator:
```
    private void updateLineNumber(Node node)
    {
        int lineno = node.getLineno();
        if (lineno != lineNumber && lineno >= 0) {
            if (itsData.firstLinePC < 0) {
                itsData.firstLinePC = lineno;
            }
            lineNumber = lineno;
            addIcode(Icode_LINE);
            addUint16(lineno & 0xFFFF);
        }
    }
```

.. which fails specifically for the first line since the initial value of lineNumber is zero.

Apologies if this is a known issue, I failed to find any similar reports.

gbrail:
This looks like a pretty straightforward thing to put in a PR for someone who is looking for an issue to work on!

camnwalter:
It looks like if we just set the initial value https://github.com/mozilla/rhino/blob/82416a171354427422f0bcca8c3206b4a5ae4105/rhino/src/main/java/org/mozilla/javascript/CodeGenerator.java#L41 to -1 then it should work?

This breaks some of the `StackTraceTest`s `failureStackTraceMozilla` and `failureStackTraceMozillaLf`, but I think those might be wrong. According to the comment above `StackStyle.MOZILLA`,
```
    /**
     * This stack trace style comes from the old Mozilla code: <code>
     * functionName()@fileName:lineNumber</code>
     */
```
[...]

PR #1609, opened 2024-09-04
Set initial line number to -1 and fix StackTraceTests

camnwalter:
This fixes https://github.com/mozilla/rhino/issues/477. I also changed the tests to match the comments in [`StackStyle`](https://github.com/mozilla/rhino/blob/master/rhino/src/main/java/org/mozilla/javascript/StackStyle.java), so they all show the line number in them.


gbrail:
Cool. Thanks!

Issue #1608, opened 2024-09-02
Issue comparing bigint to string


tonygermano:
```javascript
9007199254740992n < "9007199254740993"
```
Expected: true
Actual: false

tonygermano:
I have a PR coming for this, and I found there is a test262 test which already covers it.

rbri:
@p-bakker have added some tests in #1681 but it looks like this is already solved on the head. I think we can close this.

p-bakker:
Closed as already fixed

PR #1611, opened 2024-09-05
Implement Symbol.prototype and other related Symbol and type coersion changes

tonygermano:
There are quite a few changes, but they are spread out in multiple commits to hopefully make them manageable and many are dependent on the previous commits. If I need to break this into more than one PR, let me know.

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/AbstractEcmaObjectOperations.java
+++ b/rhino/src/main/java/org/mozilla/javascript/AbstractEcmaObjectOperations.java
@@ -46,7 +46,7 @@ public class AbstractEcmaObjectOperations {
     static boolean hasOwnProperty(Context cx, Object o, Object property) {
         Scriptable obj = ScriptableObject.ensureScriptable(o);
         boolean result;
-        if (property instanceof Symbol) {
+        if (ScriptRuntime.isSymbol(property)) {
             result = ScriptableObject.ensureSymbolScriptable(o).has((Symbol) property, obj);
         } else {
             ScriptRuntime.StringIdOrIndex s = ScriptRuntime.toStringIdOrIndex(property);
--- a/rhino/src/main/java/org/mozilla/javascript/NativeArray.java
+++ b/rhino/src/main/java/org/mozilla/javascript/NativeArray.java
@@ -1949,7 +1949,7 @@ public class NativeArray extends IdScriptableObject implements List {
         Object compareTo = args.length > 0 ? args[0] : Undefined.instance;
 
         Scriptable o = ScriptRuntime.toObject(cx, scope, thisObj);
-        long len = ScriptRuntime.toLength(new Object[] {getProperty(thisObj, "length")}, 0);
+        long len = getLengthProperty(cx, o);
         if (len == 0) return Boolean.FALSE;
 
         long k;
--- a/rhino/src/main/java/org/mozilla/javascript/NativeDate.java
+++ b/rhino/src/main/java/org/mozilla/javascript/NativeDate.java
@@ -255,6 +255,10 @@ final class NativeDate extends IdScriptableObject {
                 arity = 1;
                 s = "toJSON";
                 break;
+            case SymbolId_toPrimitive:
+                initPrototypeMethod(
+                        DATE_TAG, id, SymbolKey.TO_PRIMITIVE, "[Symbol.toPrimitive]", 1);
+                return;
             default:
                 throw new IllegalArgumentException(String.valueOf(id));
         }
@@ -322,6 +326,26 @@ final class NativeDate extends IdScriptableObject {
                                 ScriptRuntime.toString(result));
                     }
                     return result;
+                }
+            case SymbolId_toPrimitive:
+                {
+                    Scriptable o = ScriptRuntime.toObject(cx, scope, thisObj);
+                    final Object arg0 = args.length > 0 ? args[0] : Undefined.instance;
+                    final Optional<String> hint =
+                        Optional.ofNullable(arg0 instanceof CharSequence ? arg0.toString() : null);
+                    final Class<?> tryFirst = hint
+                        .map(h -> {
+                            if (h.equals("string") || h.equals("default")) {
+                                return ScriptRuntime.StringClass;
+                            } else if (h.equals("number")) {
+                                return ScriptRuntime.NumberClass;
+                            }
+                            return null;
+                        })
+                        .orElseThrow(() -> ScriptRuntime.typeErrorById(
+                            "msg.invalid.toprimitive.hint",
+                            ScriptRuntime.toString(arg0)));
+                    return ScriptableObject.getDefaultValue(o, tryFirst);
                 }
         }
 
@@ -1898,6 +1922,14 @@ final class NativeDate extends IdScriptableObject {
         return id;
     }
 
+    @Override
+    protected int findPrototypeId(Symbol key) {
+        if (SymbolKey.TO_PRIMITIVE.equals(key)) {
+            return SymbolId_toPrimitive;
+        }
+        return 0;
+    }
+
     private static final int ConstructorId_now = -3,
             ConstructorId_parse = -2,
             ConstructorId_UTC = -1,
@@ -1948,7 +1980,8 @@ final class NativeDate extends IdScriptableObject {
             Id_setYear = 45,
             Id_toISOString = 46,
             Id_toJSON = 47,
-            MAX_PROTOTYPE_ID = Id_toJSON;
+            SymbolId_toPrimitive = 48,
+            MAX_PROTOTYPE_ID = SymbolId_toPrimitive;
 
     private static final int Id_toGMTString = Id_toUTCString; // Alias, see Ecma B.2.6
 
--- a/rhino/src/main/java/org/mozilla/javascript/NativeDate.java
+++ b/rhino/src/main/java/org/mozilla/javascript/NativeDate.java
@@ -298,7 +298,8 @@ final class NativeDate extends IdScriptableObject {
                     final String toISOString = "toISOString";
 
                     Scriptable o = ScriptRuntime.toObject(cx, scope, thisObj);
-                    Object tv = ScriptRuntime.toPrimitive(o, Optional.of(ScriptRuntime.NumberClass));
+                    Object tv =
+                            ScriptRuntime.toPrimitive(o, Optional.of(ScriptRuntime.NumberClass));
                     if (tv instanceof Number) {
                         double d = ((Number) tv).doubleValue();
                         if (Double.isNaN(d) || Double.isInfinite(d)) {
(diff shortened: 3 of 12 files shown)


tonygermano:
fixes #1608 

in commit https://github.com/mozilla/rhino/pull/1611/commits/e17cc255a0e31355dab4aaee20af739eabd8d584

tonygermano:
CI is still running, but it looks like I may have some test failures to work through. I only ran test262 locally, and there was only 1 new failure added with 100+ removed from the ignore list. The new failure was for a throw assertion, and it was only passing before because it was throwing the wrong error.

tonygermano:
@gbrail @p-bakker @rbri 
I need some feedback.
1. If language version >= ES6, should we assume that all Scriptables also implement SymbolScriptable?
2. If we are not making that assumption, should `ScriptableObject.getProperty(Scriptable, Symbol)` return `Scriptable.NOT_FOUND` rather than throwing an exception when a `Scriptable` is not also a `SymbolScriptable`? (`ensureSymbolScriptable(obj)` throws the exception.)
https://github.com/mozilla/rhino/blob/3577b8355713ba7d59ef2cf7b8f2bb9194dd8e3f/rhino/src/main/java/org/mozilla/javascript/ScriptableObject.java#L2128-L2138

https://github.com/mozilla/rhino/blob/master/rhino/src/test/java/org/mozilla/javascript/tests/IterableTest.java
This test specifically tests passing a Scriptable which is not a SymbolScriptable to a for..of loop using VERSION_ES6, and I don't know if it should throw an exception because it tried to look up a symbol property on an object which does not support symbols, or if it should just treat it like the symbol property is not defined on that object.

Having `ScriptableObject.getProperty(Scriptable, Symbol)` return Scriptable.NOT_FOUND for non-SymbolScriptables, also allows it to still go up the prototype chain in case a parent object has the symbol property defined.

p-bakker:
I would say getting a Symbol property on a non-SymbolScriptable should not throw

As for setting a Symbol property on a non-SymbolScriptable: in non-strict mode I think just ignoring the put wood be fine, but in strict mode? Not sure, cannot think of anything similar to that in EcmaScript ottomh

p-bakker:
Seems EcmaScript leaves room for non-standard host objects to be implemented as the implementation sees fit

But most common is the behavior I wrote above + throwing a TypeError in strict mode

p-bakker:
Note that there also this PR that being worked on actively and which touches the Symbol implementation: #1579

tonygermano:
Thanks for pointing out the other PR. I skimmed it, and I don't think it will conflict with anything that I'm doing in this PR.

tonygermano:
> I would say getting a Symbol property on a non-SymbolScriptable should not throw
> 
> As for setting a Symbol property on a non-SymbolScriptable: in non-strict mode I think just ignoring the put wood be fine, but in strict mode? Not sure, cannot think of anything similar to that in EcmaScript ottomh

I agree with everything you said. Regarding setting a symbol property on a non-SymbolScriptable, that would match the behavior of setting a property on a sealed or frozen (non-extensible) object running in non-strict vs strict mode.

p-bakker:
@tonygermano any progress on this? 

Noticed you started multiple PRs in a short timeframe, not sure how interdependent they are, but would be nice if (parts of) at least this PR could move forward: the bit about toString taking into account Symbol.toPrimitive() unbreaks quite a few test 🙂

rhino/src/main/java/org/mozilla/javascript/ScriptRuntime.java
```diff
@@ -3926,18 +3926,32 @@ public static boolean compare(Object val1, Object val2, int op) {
...
-            }
-            if (val2 instanceof Scriptable) {
-                val2 = ((Scriptable) val2).getDefaultValue(NumberClass);
-            }
+            val1 = toPrimitive(val1, Optional.of(NumberClass));
+            val2 = toPrimitive(val2, Optional.of(NumberClass));
             if (val1 instanceof CharSequence && val2 instanceof CharSequence) {
                 return compareTo(val1.toString(), val2.toString(), op);
             }
+            if (val1 instanceof BigInteger && val2 instanceof CharSequence) {
```
rbri, line 3935: Maybe we can have one 'val1 instanceof CharSequence' base if with another if inside....

rbri:
@tonygermano i fear this will be another endless story to get this merged. Therefore i lik to suggest to split this up. First step might be a pr that has only the first two commits from this story.
What do you think?

If you like i can prepare such a pr.

rbri:
short info - working on carving out the nect part

p-bakker:
@rbri have you cherry-picked everything from this PR into separate PRs now, so this one can be closed? Or is there still something left in this one that warrants leaving it open?

rbri:
@p-bakker i do it one step after another - currently step 2 is waiting for review/merge (pr #1674). I plan at least two more steps (each requires the one one before i fear).

But finally i might not take all from this - i do not like to pick the 'Add asyncIterator and matchAll well-known Symbols' commit because i guess the related functions are not yet implemented and having only the symbols available makes not so much sense. But i have to have a closer look at this after step 4.

p-bakker:
Thx for the update

Agree with not including the Symbols: we have cases for matchAll and async support already and better to add the Symbols when those get implemented

rbri:
@tonygermano looks like you are back (https://github.com/mozilla/rhino/issues/2153) - I'd be really happy if you would support rhino again.

And if yes, i like to align a bit with you about this PR. Maybe we can find a way to get the last pices sorted carved out and get this close. What do you think?

aardvark179:
On the question of Scriptable and SymbolScriptable I agree that put shouldn’t throw in non-strict mode, and throw in strict mode. I think we could do this with `default` implementations on Scriptable.

rhino/src/main/java/org/mozilla/javascript/AbstractEcmaObjectOperations.java
```diff
@@ -46,7 +46,7 @@ enum KEY_COERCION {
     static boolean hasOwnProperty(Context cx, Object o, Object property) {
         Scriptable obj = ScriptableObject.ensureScriptable(o);
         boolean result;
-        if (property instanceof Symbol) {
```
rbri, line 49: this is not correct - see https://github.com/mozilla/rhino/pull/2158 for a test that proves this

rbri: to be more precise - instanceof is correct

rhino/src/main/java/org/mozilla/javascript/NativeArray.java
```diff
@@ -1949,7 +1949,7 @@ private static Boolean js_includes(
         Object compareTo = args.length > 0 ? args[0] : Undefined.instance;
 
         Scriptable o = ScriptRuntime.toObject(cx, scope, thisObj);
-        long len = ScriptRuntime.toLength(new Object[] {getProperty(thisObj, "length")}, 0);
+        long len = getLengthProperty(cx, o);
```
rbri, line 1952: already implemented

rhino/src/main/java/org/mozilla/javascript/NativeDate.java
rbri, line 1: all the changes for NativeDate are implemented

rhino/src/main/java/org/mozilla/javascript/NativeObject.java
```diff
@@ -361,7 +361,7 @@ public Object execIdCall(
                                 Integer.toString(args.length));
                     }
                     Scriptable proto = (args[1] == null) ? null : ensureScriptable(args[1]);
-                    if (proto instanceof Symbol) {
+                    if (ScriptRuntime.isSymbol(proto)) {
```
rbri, line 364: this change is part of https://github.com/mozilla/rhino/pull/1611

rhino/src/main/java/org/mozilla/javascript/NativeObject.java
```diff
@@ -440,7 +440,7 @@ public Object execIdCall(
                             (key, value) -> {
                                 if (key instanceof Integer) {
                                     obj.put((Integer) key, obj, value);
-                                } else if (key instanceof Symbol
```
rbri, line 443: this is correct - see https://github.com/mozilla/rhino/pull/2158 for a test that proves this

rhino/src/main/java/org/mozilla/javascript/Symbol.java
```diff
@@ -7,7 +7,7 @@
 package org.mozilla.javascript;
 
 /**
- * A Symbol is a JavaScript objecy that obeys the special properties of the Symbol prototype. This
+ * A Symbol is a JavaScript object that obeys the special properties of the Symbol prototype. This
```
rbri, line 10: done

tests/src/test/java/org/mozilla/javascript/tests/Test262SuiteTest.java
```diff
@@ -112,7 +112,6 @@ public class Test262SuiteTest {
                             "regexp-unicode-property-escapes",
                             "super",
                             "String.prototype.matchAll",
-                            "Symbol.matchAll",
```
rbri, line 115: done

rhino/src/main/java/org/mozilla/javascript/ScriptRuntime.java
```diff
@@ -436,15 +436,10 @@ public static double toNumber(Object val) {
...
-            if (val instanceof Symbol) throw typeErrorById("msg.not.a.number");
-            if (val instanceof Scriptable) {
-                val = ((Scriptable) val).getDefaultValue(NumberClass);
-                if ((val instanceof Scriptable) && !isSymbol(val))
-                    throw errorWithClassName("msg.primitive.expected", val);
-                continue;
-            }
-            warnAboutNonJSObject(val);
-            return NaN;
+            if (isSymbol(val)) throw typeErrorById("msg.not.a.number");
```
rbri, line 439: done

rhino/src/main/java/org/mozilla/javascript/ScriptRuntime.java
```diff
@@ -713,56 +708,45 @@ public static double toNumber(String s) {
 
     /** Convert the value to a BigInt. */
     public static BigInteger toBigInt(Object val) {
```
rbri, line 710: done

rhino/src/main/java/org/mozilla/javascript/ScriptRuntime.java
```diff
@@ -841,6 +825,7 @@ public static BigInteger toBigInt(String s) {
      * <p>See ECMA 7.1.3 (v11.0).
      */
     public static Number toNumeric(Object val) {
+        val = toPrimitive(val, Optional.of(NumberClass));
```
rbri, line 828: done

rhino/src/main/java/org/mozilla/javascript/ScriptRuntime.java
```diff
@@ -1027,24 +1012,22 @@ public static String toString(Object val) {
                 return val.toString();
             }
             if (val instanceof BigInteger) {
-                return val.toString();
+                return ((BigInteger) val).toString(10);
```
rbri, line 1015: done

rhino/src/main/java/org/mozilla/javascript/ScriptRuntime.java
```diff
@@ -1027,24 +1012,22 @@ public static String toString(Object val) {
...
             }
             if (val instanceof Number) {
                 // XXX should we just teach NativeNumber.stringValue()
                 // about Numbers?
                 return numberToString(((Number) val).doubleValue(), 10);
             }
-            if (val instanceof Symbol) {
```
rbri, line 1037: done

rhino/src/main/java/org/mozilla/javascript/ScriptRuntime.java
```diff
@@ -2983,55 +2966,58 @@ public static boolean isObject(Object value) {
     // implement the '~' operator inline in the caller
     // as "~toInt32(val)"
 
-    public static Object add(Object val1, Object val2, Context cx) {
```
rbri, line 2986: done.

rbri:
Have done a review and wrote some more tests. Thes means, all the code is merged so far.
Next step is to look at at the ScriptableObject stuff.

PR #1622, opened 2024-09-12
Extend LambdaConstructor for more flexibility

gbrail:
This lets us create constructors that behave differently when invoked
using "new" than when called directly. The native "Date" class is an
example. The default behavior, which uses the same behavior in either
case but can automatically throw an exception if only one form is
supported, still works.

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/LambdaConstructor.java
+++ b/rhino/src/main/java/org/mozilla/javascript/LambdaConstructor.java
@@ -66,12 +66,30 @@ public class LambdaConstructor extends LambdaFunction {
         this.flags = flags;
     }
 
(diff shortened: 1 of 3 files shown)


gbrail:
@p-bakker and @rbri I think that you had some questions about lambda constructors for the Proxy implementation. I tried to lambda-ize the Date class, and found that I needed the changes in this PR. Let me know if you find it useful.

gbrail:
This doesn't break anything but I think it'll be useful in upcoming changes. If anyone has a chance to take a look, that'd be great, because I'd love to move forward with it.

gbrail:
I really want people to use this so I'm merging it. Thanks!

Issue #1614, opened 2024-09-07
Cannot use Symbol as key in computed key accessor property:  `var x = {get [Symbol()](){ /* some code */ }}`


p-bakker:
See https://github.com/mozilla/rhino/pull/1610#issuecomment-2330680331

I locally fixed it by patching ScriptRuntime.newObjectLiteral(...) by adding a check for Symbol and following a different codepath:
![image](https://github.com/user-attachments/assets/16e35e8c-1e26-4e20-ac61-4c3758fb829f)

In the `id instanceof Symbol` branch, I'm calling defineOwnProperty, whereas for all other type of keys it uses setGetterOrSetter, which is tailored to handling just String and int keys and directly modifies the slot map

Using defineOwnProperty however requires to instantiate a descriptor ScriptableObject, which then follows a different path to modify the slot map.

Thoughts on this approach? Feels a bit messy to me, so like to get some other opinions @gbrail @rbri @andreabergia

p-bakker:
Another approach seems to be changing the type of the `name` param of setGetterOrSetter to Object instead of String: most usage of the `name` param are calls to methods that expect an Object anyway

rbri:
@p-bakker sorry, I'm on vacation now, have my brain switched off 😅

PR #1626, opened 2024-09-12
Update test262 suite + include resizable-arraybuffer in unsupported feature list

p-bakker:
test262 update brings tests for AggregateError constructor w/o mandatory 1st Iterable argument and tests for #1614


gbrail:
This is great -- thanks!

PR #1627, opened 2024-09-13
Try a different strategy for Java versions

gbrail:
Always build using Java 11, and rely on Gradle to make that happen. Support tests running on different versions via an environment variable.

This is an experiment to see what GitHub Actions does.

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -23,19 +23,21 @@ jobs:
       run: git submodule update --init --single-branch
     - name: Set up Java
       uses: actions/setup-java@2dfa2011c5b2a0f1489bf9e433881c92c1631f88 # v4.3.0
+      # By putting JVMs in reverse order, the last one is the one that
+      # will go in the path.
       with:
         java-version: |
-          21
-          17
           11
+          17
+          21
         distribution: 'adopt'
-    - name: Build and test Java 11
+    - name: Build and test Java 21
       run: >-
         ./gradlew check
-    - name: Upload results Java 11
+    - name: Upload results Java 21
       uses: actions/upload-artifact@50769540e7f4bd5e21e526ee35c689e35e0d6874 # v4.4.0
       with:
-        name: reports-java-11
+        name: reports-java-21
         path: '*/build/reports'
     - name: Build and test Java 17
       env:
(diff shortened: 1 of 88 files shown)


p-bakker:
Not been able yet to try this out in my setup to might have a m moment over the weekend

gbrail:
FWIW this works fine in the CI/CD pipeline, and it makes the overall CI/CD
pipeline run about two minutes faster. (Out of nearly 45 minutes.)

The main impact of this is that if you have multiple JVMs on your machine,
Gradle will always pick Java 11 to build and test unless you ask for a
different JVM for testing.

The downside is that if you don't have Java 11, I believe that Gradle will
not build.

p-bakker:
Little later, but got around to it now.

The altered setup results in all projects being setup by Gradle to use Java 11 now in Eclipse, including the test project.

So the test project will compile with Java 11, but when runnign the tests, it'll run with another Java version if  RHINO_TEST_JAVA_VERSION is set, but only if the tests are executed through Gradle. 

Personally I setup (JUnit) Test Configurations in Eclipse manually and when I do that I specify the Java version to use anyway, so that works as well.

So I think this PR works as intended :-)

Only question I have left is whether the usage of java > toolchain requires anything else from devs checking out Rhino? For me it worked out of the box, but I've done some stuff with toolchains in the past, so maybe I've already got some stuff configured on my machine and therefor it worked straight away

p-bakker:
Another Q: when running the Test262 Test Suite with the `-DupdateTest262properties` parameter, should it maybe also block if not running on Java 11, because if run with a newer Java version, it may mark tests as passing, while they wouldn't on Java 11, for example tests in the area of `language/identfiiers/...`

gbrail:
OK -- this is an entirely different approach now, inspired by this article, which points out, among other things, that running Gradle and compiling on the latest JDK is a good idea:

https://jakewharton.com/gradle-toolchains-are-rarely-a-good-idea/

The idea is this:

* Build and compile with whatever Java version the user has, whether it's from the path or from JAVA_HOME
* Set the "--release" flag on the compiler so that we build for the right bytecode version
* Use the RHINO_TEST_JAVA_VERSION to test with each JVM

To make this work, since "google-java-format" (used by Spotless) is still sensitive, I had to upgrade it to the latest version, which does not run on Java 11. And the new version requires that we reformat some stuff, so the yak is being shaved here.

So the upside of all this is the following:

* You can build Rhino with whatever Java you have
* However, to properly build and test it you need at least Java 17
* We will test with older Java versions to ensure compatibility before release

gbrail:
The latest version will fail to run the tests if "updateTest262properties" is set and we're not testing on Java 11. Given that, I'd like to go ahead and merge this as letting people use Java 21-based build tools is a nice improvement.

p-bakker:
Unfortunately haven't gotten around to testing it, but just looking at the changes it looks good to me!

gbrail:
Merging this -- it means that we can now all build with newer Java versions but still should use Java 11 before updating test262 -- if that causes confusion we may have to address it in the test262 infrastructure.

PR #1628, opened 2024-09-13
Add well formed JSON.stringify, String.prototype.isWellFormed, and String.prototype.toWellFormed

### LINKED ISSUE #982: Support ES2019 well formed JSON.stringify
Proposal: https://github.com/tc39/proposal-well-formed-stringify
MDN: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#well-formed_json.stringify
Also see: https://exploringjs.com/es2018-es2019/ch_well-formed-stringify.html

camnwalter:
Closes #982

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/NativeJSON.java
+++ b/rhino/src/main/java/org/mozilla/javascript/NativeJSON.java
@@ -549,7 +549,9 @@ public final class NativeJSON extends ScriptableObject {
                     product.append("\\t");
                     break;
                 default:
-                    if (isLeadingSurrogate(c) && i < length - 1 && isTrailingSurrogate(string.charAt(i + 1))) {
+                    if (isLeadingSurrogate(c)
+                            && i < length - 1
+                            && isTrailingSurrogate(string.charAt(i + 1))) {
                         // do nothing as the next case will add both surrogates
                         break;
                     } else if (isTrailingSurrogate(c) && isLeadingSurrogate(prev)) {
@@ -569,11 +571,11 @@ public final class NativeJSON extends ScriptableObject {
         return product.toString();
     }
 
-    private static boolean isLeadingSurrogate(char c) {
+    static boolean isLeadingSurrogate(char c) {
         return c >= 0xD800 && c <= 0xDBFF;
     }
 
-    private static boolean isTrailingSurrogate(char c) {
+    static boolean isTrailingSurrogate(char c) {
         return c >= 0xDC00 && c <= 0xDFFF;
     }
 
--- a/rhino/src/main/java/org/mozilla/javascript/NativeJSON.java
+++ b/rhino/src/main/java/org/mozilla/javascript/NativeJSON.java
@@ -523,6 +523,7 @@ public final class NativeJSON extends ScriptableObject {
                 new StringBuilder(string.length() + 2); // two extra chars for " on either side
         product.append('"');
         int length = string.length();
+        char prev = 0;
         for (int i = 0; i < length; i++) {
             char c = string.charAt(i);
             switch (c) {
@@ -548,7 +549,12 @@ public final class NativeJSON extends ScriptableObject {
                     product.append("\\t");
                     break;
                 default:
-                    if (c < ' ') {
+                    if (isLeadingSurrogate(c) && i < length - 1 && isTrailingSurrogate(string.charAt(i + 1))) {
+                        // do nothing as the next case will add both surrogates
+                        break;
+                    } else if (isTrailingSurrogate(c) && isLeadingSurrogate(prev)) {
+                        product.append(prev).append(c);
+                    } else if (c < ' ' || isLeadingSurrogate(c) || isTrailingSurrogate(c)) {
                         product.append("\\u");
                         String hex = String.format("%04x", Integer.valueOf(c));
                         product.append(hex);
@@ -557,9 +563,18 @@ public final class NativeJSON extends ScriptableObject {
                     }
                     break;
             }
+            prev = c;
         }
         product.append('"');
         return product.toString();
+    }
+
+    private static boolean isLeadingSurrogate(char c) {
+        return c >= 0xD800 && c <= 0xDBFF;
+    }
+
+    private static boolean isTrailingSurrogate(char c) {
+        return c >= 0xDC00 && c <= 0xDFFF;
     }
 
     private static Object javaToJSON(Object value, StringifyState state) {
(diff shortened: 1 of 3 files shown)


tests/testsrc/test262.properties
```diff
@@ -5253,7 +5254,7 @@ language/expressions/new 41/59 (69.49%)
 
 ~language/expressions/new.target
 
-language/expressions/object 865/1169 (73.99%)
+language/expressions/object 863/1169 (73.82%)
```
p-bakker, line 5257: Any idea what's going on here? The stats for updated, but I don't see any tests being unbroken in this section

tests/testsrc/test262.properties
```diff
@@ -2189,7 +2188,8 @@ built-ins/String 140/1182 (11.84%)
     prototype/indexOf/searchstring-tostring-errors.js
     prototype/indexOf/searchstring-tostring-toprimitive.js
     prototype/indexOf/searchstring-tostring-wrapped-values.js
-    prototype/isWellFormed 8/8 (100.0%)
+    prototype/isWellFormed/not-a-constructor.js {unsupported: [Reflect.construct]}
+    prototype/isWellFormed/to-string-primitive.js
```
p-bakker, line 2192: Due to missing Symbol.toPrimitive I presume

camnwalter: It actually fails with a syntax error beforehand because it uses for const with destructuring. Meant to mention that before but forgot.

tests/testsrc/test262.properties
camnwalter: Not sure, could a previous PR have fixed something but forgot to run spotlessApply?

tests/testsrc/test262.properties
p-bakker: Ah, another one of our challenges 🫣

tests/testsrc/test262.properties
p-bakker: Well, the last thing that got merged was another one of your PRs 🤔

camnwalter: Ok no idea then 😅

rbri: I have the impression i have seen something similar for other pr's also, but i might be wrong

rhino/src/main/java/org/mozilla/javascript/NativeString.java
```diff
@@ -772,6 +780,56 @@ else if (Normalizer.Form.NFC.name().equals(formStr))
 
                         return str.substring(k, k + 1);
                     }
+                case Id_isWellFormed:
+                    {
+                        String str = ScriptRuntime.toString(requireObjectCoercible(cx, thisObj, f));
```
gbrail, line 785: This is minor but it will likely help with performance in a bunch of cases -- how about using "toCharSequence" here instead of "toString"? That would mean that concatenated strings here don't need to be re-concatenated, and would work fine since it looks like you're just iterating character-by-character anyway.

gbrail:
See above, I'm curious if you could use CharSequence here instead for performance, otherwise this looks good and thanks!

rhino/src/main/java/org/mozilla/javascript/NativeJSON.java
```diff
@@ -557,11 +565,20 @@ private static String quote(String string) {
                     }
                     break;
             }
+            prev = c;
         }
         product.append('"');
         return product.toString();
     }
 
+    static boolean isLeadingSurrogate(char c) {
```
rbri, line 574: Are these checks specific to JSON or more general? If they are not specific to JSON i vote for haing a more central place for these methods. Not sure if we already have one (sorry i only have my cell phone at hand at the moment)

rhino/src/main/java/org/mozilla/javascript/NativeString.java
```diff
@@ -772,6 +780,56 @@ else if (Normalizer.Form.NFC.name().equals(formStr))
...
+                                foundLeadingSurrogate = false;
+                            } else if (foundLeadingSurrogate) {
+                                return false;
+                            }
+                        }
+                        return !foundLeadingSurrogate;
+                    }
+                case Id_toWellFormed:
+                    {
+                        String str = ScriptRuntime.toString(requireObjectCoercible(cx, thisObj, f));
```
rbri: Maybe this kind of string processing can be optimized even more. Have a first loop that simply traverses the string until you reach a char that has to be replaced. Then create the string builder an use the already processed chars as initial input. Now you can process like you did.
This ends up with more code but avoidas the string buider and the reconstruction of the string in all standard cases. I assume that most of the json to be normalized does not contain surrogates.

rhino/src/main/java/org/mozilla/javascript/NativeJSON.java
camnwalter: No they aren't. Not sure where they should go though...

rhino/src/main/java/org/mozilla/javascript/NativeString.java
camnwalter: That's a great point, thanks! I can try this out and see how it goes.

rhino/src/main/java/org/mozilla/javascript/NativeJSON.java
rbri: @gbrail @p-bakker @tonygermano any suggestion?

rhino/src/main/java/org/mozilla/javascript/NativeString.java
```diff
@@ -772,6 +782,80 @@ else if (Normalizer.Form.NFC.name().equals(formStr))
...
+
+                        int len = str.length();
+                        char prev = 0;
+                        int firstSurrogateIndex = -1;
+                        for (int i = 0; i < len; i++) {
+                            char c = str.charAt(i);
+
+                            if (NativeJSON.isLeadingSurrogate(prev)
+                                    && NativeJSON.isTrailingSurrogate(c)) {
+                                surrogates.put(Integer.valueOf(i - 1), Boolean.TRUE);
```
camnwalter, line 828: I used the Boxed types for the map because I saw a commit from a few years ago that reduced the amount of auto boxing.

tests/testsrc/test262.properties
tonygermano: You probably need to rebase this branch. `language/expressions/object 863/1169 (73.82%)` is the value currently in the master branch. It's not being added with this PR.

camnwalter: Will do, thanks

gbrail:
This looks good and I'm going to merge it. Thanks!

PR #1635, opened 2024-09-20
Use the Constructable interface instead of Function if appropriate

rbri:
see #1376 for some details


gbrail:
Yes -- I see what this does and I agree it's a good idea. Thanks!

Issue #756, opened 2020-08-06
Add support for default values inside destructuring assignments


Osiris-Team:
```java

            // default values inside destructuring assignments,
            // like 'var [a = 10] = b' or 'var {a: b = 10} = c',
            // are not supported
            reportError("msg.destruct.default.vals");

```

Osiris-Team:
Wow thats great!

Osiris-Team:
Seems like just a matter of years now.

gbrail:
@Osiris-Team the fastest way to get that part of the standard implemented into Rhino is to submit a PR. There are a few of us here who are willing and ready to help you with that.

p-bakker:
To expand on @gbrail comment: Rhino is an all-volunteer effort and as such hasn't got anyone working on it fulltime.

Cases are picked up based on own need/interest or general desire to move things forward.

We have recently made the effort to at least start creating cases for all the things where Rhino is not compatible with the latest EcmaScript spec and assigning those issues into milestones. But this is more to gain insight and hopefully inspire people to contribute, than a plan to start implementing all these things asap.

So, as @gbrail said: we welcome contributions!

p-bakker:
#1640 has moved the needle on this one significantly, if not completely resolved it 

Waiting on analysis whether there's anything remaining (see https://github.com/mozilla/rhino/pull/1640#issuecomment-2365992097) before potentially closing this as done

anivar:
## Testing Results

I tested destructuring assignments in Rhino 1.8.1-SNAPSHOT with VERSION_ES6 (200):

```javascript
const obj = {a: 1, b: 2};
const {a, b} = obj;
console.log("Destructured:", a, b);
```

**Result:** ✅ Works perfectly! Outputs `Destructured: 1 2`

**Code Evidence:** Parser.java has extensive destructuring support:
- `destructuringAssignExpr()` method exists
- `destructuring` and `destructuringDefault` maps handle parameter destructuring  
- Token support is implemented

This issue can be closed as destructuring assignments are fully functional in current Rhino.

PR #1481, opened 2024-05-28
FEAT: Support default arguments in functions

0xe:
Adds support for default values in function parameters and default values in destructuring assignment.

What's not implemented yet, that I plan to implement subsequently:

- Destructuring with defaults in for loops (eg: for (let {x = 2} = {}; ; ))
- Reading [Symbol.iterator] from default array/object
- Nested Literals

(Added ignored tests for these in DefaultParametersTest)

PR #1640, opened 2024-09-21
FEAT: Support default parameters

0xe:
Adds support for default values in function parameters and default values in destructuring assignment.

What's not implemented yet, that I plan to implement subsequently:

- Destructuring with defaults in for loops (eg: for (let {x = 2} = {}; ; ))
- Reading [Symbol.iterator] from default array/object
- Nested Literals

(Added ignored tests for these in DefaultParametersTest)

(same changes as https://github.com/mozilla/rhino/pull/1481, but squashed and on top of master)

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/Arguments.java
+++ b/rhino/src/main/java/org/mozilla/javascript/Arguments.java
@@ -120,6 +120,12 @@ final class Arguments extends IdScriptableObject {
             return false;
         }
         NativeFunction f = activation.function;
+
+        // Check if default arguments are present
+        if (f == null || f.hasDefaultParameters()) {
+            return false;
+        }
+
         int definedCount = f.getParamCount();
         if (index < definedCount) {
             // Check if argument is not hidden by later argument with the same
--- a/rhino/src/main/java/org/mozilla/javascript/BaseFunction.java
+++ b/rhino/src/main/java/org/mozilla/javascript/BaseFunction.java
@@ -70,6 +70,11 @@ public class BaseFunction extends IdScriptableObject implements Function {
         return isGeneratorFunction;
     }
 
+    // Generated code will override this
+    protected boolean hasDefaultParameters() {
+        return false;
+    }
+
     /**
      * Gets the value returned by calling the typeof operator on this object.
      *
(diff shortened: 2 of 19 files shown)


gbrail:
Thanks -- this has been outstanding for a long time, and it makes a lot more things work so I'm going to merge it.

I notice that there are test262 tests that actually pass but are marked as failing, so I'm going to fix that and manually merge. Thanks for working on this!

gbrail:
Merged manually with some test262 fixes -- looking forward to more!

0xe:
> Merged manually with some test262 fixes -- looking forward to more!

Thank you!

p-bakker:
@0xe would you say we can close #756 now as well? 

Unless you'll open a PR for the three things you mentioned (that aren't supported yet) very shortly, cloud you maybe create a follow-up issue, so we don't lose track of it?

Just checking: this PR didn't move the needle on rest/spread syntax anywhere, correct?

0xe:
> @0xe would you say we can close #756 now as well?
> 
> Unless you'll open a PR for the three things you mentioned (that aren't supported yet) very shortly, cloud you maybe create a follow-up issue, so we don't lose track of it?
> 
> Just checking: this PR didn't move the needle on rest/spread syntax anywhere, correct?

Yeah, I will create issues for the TODO items from this PR tonight. 

There should be no changes to rest/spread from this PR.

p-bakker:
Any chance you can have a look what, if anything, is still not working wrt default values in destructuring assignments?

I would think support/issues between destructuring assignments and default function parameter values to be quite similar, but I might be wrong. 

For now I'm holding of on closing #756 a bit until we have some (more) insight into this

Osiris-Team:
@0xe Wow amazing, 1k lines changed. Thanks for your work! I remember opening the issue 4 years ago as a complete Java noob. Time flies.

0xe:
Added an issue for known unimplemented things from this PR: https://github.com/mozilla/rhino/issues/1641

0xe:
> Any chance you can have a look what, if anything, is still not working wrt default values in destructuring assignments?
> 
> I would think support/issues between destructuring assignments and default function parameter values to be quite similar, but I might be wrong.
> 
> For now I'm holding of on closing #756 a bit until we have some (more) insight into this

Makes sense. I will take a look at what's pending with destructuring assignments and update #756.

From my quick check in the shell, following expressions evaluate correctly:

```javascript
js> let [a, b] = [1, 2];
js> a
1
js> b
2
js> let [c=3, d=4] = [];
js> c
3
js> d
4
js> let {e = 5, f = 6} = {};
js> e
5
js> f
6
js> let {g = 7, h = 8} = {g: 9, h: 10}
js> g
9
js> h
10
``` 

One thing that should be addressed in https://github.com/mozilla/rhino/issues/1641 is reading `[Symbol.iterator]` when destructuring default values from objects.

PR #1642, opened 2024-09-22
FIX: test for nested literals in DefaultParametersTest

0xe:


gbrail:
Makes sense to me. Thanks!

PR #1644, opened 2024-09-23
Optimize map accesses in several places

rPraml:
In this PR I will try to optimize map access on several places and pick up @gbrail's comment: https://github.com/mozilla/rhino/pull/1575#discussion_r1730067620

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/Hashtable.java
+++ b/rhino/src/main/java/org/mozilla/javascript/Hashtable.java
@@ -123,20 +123,25 @@ public class Hashtable implements Serializable, Iterable<Hashtable.Entry> {
     public void put(Object key, Object value) {
         final Entry nv = new Entry(key, value);
 
-        if (!map.containsKey(nv)) {
-            // New value -- insert to end of doubly-linked list
-            map.put(nv, nv);
-            if (first == null) {
-                first = last = nv;
-            } else {
-                last.next = nv;
-                nv.prev = last;
-                last = nv;
-            }
-        } else {
-            // Update the existing value and keep it in the same place in the list
-            map.get(nv).value = value;
-        }
+        map.compute(
+                nv,
+                (k, existing) -> {
+                    if (existing == null) {
+                        // New value -- insert to end of doubly-linked list
+                        if (first == null) {
+                            first = last = nv;
+                        } else {
+                            last.next = nv;
+                            nv.prev = last;
+                            last = nv;
+                        }
+                        return nv;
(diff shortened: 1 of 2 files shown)


rhino/src/main/java/org/mozilla/javascript/Hashtable.java
```diff
@@ -123,20 +123,25 @@ public int size() {
...
-            } else {
-                last.next = nv;
-                nv.prev = last;
-                last = nv;
-            }
-        } else {
-            // Update the existing value and keep it in the same place in the list
-            map.get(nv).value = value;
-        }
+        map.compute(
```
rPraml, line 126: we will save one hash computation, but have the price of an additional invocation of a lambda expression:
I did this short test:
```java
	@Test
	public void speed() {
		for (long i = 0L; i < 10_000_000L; i++) {
			ht.put("one", 1);
			ht.put("two", 2);
			ht.put("three", 3);
			ht.put("four", 4);
			ht.put("five", 5);
			ht.put("six", 6);
			ht.put("seven", 7);
			ht.put("eight", 8);
			ht.put("nine", 9);
			ht.put("ten", 10);

		}
	}
```
it runs with the old code in apporx 8,5sec and with the new code approx 5,5 sec

rhino/src/main/java/org/mozilla/javascript/JavaMembers.java
```diff
@@ -595,23 +599,21 @@ private void reflect(
                     NativeJavaMethod setters = null;
                     String setterName = "set".concat(nameComponent);
 
-                    if (ht.containsKey(setterName)) {
```
rPraml, line 598: I think we can safely remove `containsKey` here, as we call `get` in the next line, which is checked for null by the `instanceOf` test.

rhino/src/main/java/org/mozilla/javascript/JavaMembers.java
```diff
@@ -687,13 +689,11 @@ private Field[] getAccessibleFields(boolean includeProtected, boolean includePri
     private static MemberBox findGetter(
             boolean isStatic, Map<String, Object> ht, String prefix, String propertyName) {
         String getterName = prefix.concat(propertyName);
-        if (ht.containsKey(getterName)) {
-            // Check that the getter is a method.
-            Object member = ht.get(getterName);
```
rPraml, line 692: same here. Do not hit map twice

gbrail:
Yes, this is a good change. Thanks!

PR #1645, opened 2024-09-24
Begin to use invokedynamic in the bytecode

gbrail:
This begins to use invokedynamic instructions in Rhino's bytecode.

It replaces the operations that call the ScriptRuntime operations for
common property operations like setting and getting properties and
elements, and common context operations like setting and getting variables,
with invokedynamic instructions.

It also adds components that will wire up those invokedynamic instructions
to the appropriate ScriptRuntime operations using the Dynalink package.

The result should be that Rhino behaves exactly the same and performs the
same as well.

However, once this is implemented we can begin to create additional dynalink
"linkers" that do specific things to optimize performance based on what is
happening at runtime.

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/optimizer/Signatures.java
+++ b/rhino/src/main/java/org/mozilla/javascript/optimizer/Signatures.java
@@ -1,0 +1,143 @@
+package org.mozilla.javascript.optimizer;
+
+/**
+ * This class defines the method signatures for the properties used by the invokedynamic
+ * instructions in the bytecode. This helps us identify what each bytecode operation means, and what
+ * the method signature should be.
+ */
+interface Signatures {
+    /**
+     * PROP:GET:{name}: Looks up the object property named "name". Falls back to
+     * ScriptRuntime.getObjectPropNoWarn.
+     */
+    String PROP_GET =
+            "(Ljava/lang/Object;"
+                    + "Lorg/mozilla/javascript/Context;"
+                    + "Lorg/mozilla/javascript/Scriptable;"
+                    + ")Ljava/lang/Object;";
+
+    /**
+     * PROP:GETNOWARN:{name}: Looks up the object property named "name" and does not warn if it does
+     * not exist. Falls back to ScriptRuntime.getObjectProp.
+     */
+    String PROP_GET_NOWARN =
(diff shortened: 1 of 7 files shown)


gbrail:
Unless there are no objections, I'd like to merge this. Just so that you all understand the scope:

Downsides:
1. It adds a dependency on the jdk.dynalink module. That should not be a problem for anyone I can think of as long as we make sure that this code is never activated on an Android platform (and I think that it will be OK)
2. It adds an extra layer of abstraction to the bytecode which requires understanding more stuff to understand what it does. I am going to experiment more with the structure of the code. I fear in the current state it will be hard to understand.

Upsides:
1. The bytecode is smaller and IMO easier to read.
2. I have an optimization waiting in the wings that will make constants faster. However it only works for "real constants," and sadly existing JavaScript benchmarks tend not to actually use "const." (I also tried a more aggressive optimization that assumed things were const until they aren't, but that needs more work.)
3. I think that further optimizations are possible for the various math operations.
4. I have spent lots and lots of time trying to make property and variable accesses more efficient, and so far nothing makes a significant difference, but I will keep working on it.

p-bakker:
No objections here! 

Wrt to the downsides: dunno how feasible it is, but some documentation on the bytecode generation may help? Got the idea that only a few people understand how that works right now. I for one don't at all 🙂

rhino/src/main/java/org/mozilla/javascript/optimizer/Signatures.java
```diff
@@ -92,17 +94,17 @@ interface Signatures {
...
 
     /**
      * NAME:GETWITHTHIS:{name}: Looks up a name in the scope like NAME:GET, and sets "this" in the
      * "last stored scriptable." Falls back to ScriptRuntime.getNameFunctionAndThis.
      */
     String NAME_GET_THIS =
-            "(Lorg/mozilla/javascript/Context;"
-                    + "Lorg/mozilla/javascript/Scriptable;"
+            "(Lorg/mozilla/javascript/Scriptable;"
+                    + "Lorg/mozilla/javascript/Context;"
```
rbri, line 107: maybe a comment somewhere why the order is important (at least because we usually have the context first) might be helpful to avoid changing the order back in the future

gbrail: I updated the comments. Thanks!

PR #1646, opened 2024-09-24
Always upload the logs!

gbrail:


p-bakker:
I think this one resolved #1602, no?

Issue #1639, opened 2024-09-21
Use enums more freely


gbrail:
We have a lot of places, especially in parser-related things, where we have a pattern of integer identifiers that have to remain unique, like this:

https://github.com/mozilla/rhino/blob/b08de525f3daefa9800a509a5e5a5b27848da7ff/rhino/src/main/java/org/mozilla/javascript/Token.java#L32

Now that people are adding lots of new language features (thanks!) this is becoming a maintenance nightmare, because all it takes is two or three people modifying the same list of integers to create a merge conflict.

Fortunately, since the 1990s Java has added proper enums. 

It would be wonderful if someone could take on the task of replacing these enum-like structures in Rhino with proper Java enums.

p-bakker:
This would constitute a breaking change, for people that interact with the parser directly, no?

Do note that roughly, these are the current language features that Rhino doesn't yet support that involve new syntax: 
- rest & spread #1217, #968, #756, #652 (some rest support introduced in #1451)
- modules #945 
- class #835
- async & await #395 
- optional chaining operator #937 (PR in progress)
- nullish assignment operator #936
- super (in Objects) #1299 

And stuff in the TC39 proposals pipeline (stage [2.7](https://github.com/tc39/proposals?tab=readme-ov-file#stage-27) & [3)](https://github.com/tc39/proposals?tab=readme-ov-file#stage-3):
- [decorators](https://github.com/tc39/proposal-decorators)

All this to say: yes we had quite a few language features added recently with super being added for some additional operators, but likely going forward well have less conflicting PRs. Not saying we shouldn't do this, but given this resulting in a breaking change, I think we should consider not doing this before a V2.x.x

andreabergia:
IMHO the benefits would overweight the cost. If someone has opened PRs yes, it might take a bit of refactoring, but the maintenance of some of these existing patterns (for example, the `Token`, which I've had to touch for some recent PR) is a lot worse.

rbri:
fine for me, please do it

rPraml:
I've tried to do this, but I see serveral show stoppers:
- There is `Token` and there is `ICode` - it would be necessary to keep all codes either in one enum or do a huge refactoring
- There is `currentFlaggedToken` which has additional bits and would require refactoring
[...]

PR #1648, opened 2024-09-24
Compute Token and Icode constants

rPraml:
This is an alternate approach for solving the problems mentioned here #1639

and should avoid large renumbering as you can see here: https://github.com/mozilla/rhino/pull/1593/files#diff-c1caeb9d13c1dceec5f4603f3836f3e227350ee1a57679238ce9a3a434601b60

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/Token.java
+++ b/rhino/src/main/java/org/mozilla/javascript/Token.java
@@ -32,205 +32,206 @@ public class Token {
     public static final int
             // start enum
             ERROR = -1, // well-known as the only code < EOF
-            EOF = 0, // end of file token - (not EOF_CHAR)
-            EOL = 1, // end of line
+            EOF = ERROR + 1, // end of file token - (not EOF_CHAR)
+            EOL = EOF + 1, // end of line
 
             // Interpreter reuses the following as bytecodes
-            FIRST_BYTECODE_TOKEN = 2,
-            ENTERWITH = 2,
-            LEAVEWITH = 3,
-            RETURN = 4,
-            GOTO = 5,
-            IFEQ = 6,
-            IFNE = 7,
-            SETNAME = 8,
-            BITOR = 9,
-            BITXOR = 10,
-            BITAND = 11,
-            EQ = 12,
-            NE = 13,
-            LT = 14,
-            LE = 15,
-            GT = 16,
-            GE = 17,
-            LSH = 18,
-            RSH = 19,
-            URSH = 20,
-            ADD = 21,
-            SUB = 22,
-            MUL = 23,
-            DIV = 24,
-            MOD = 25,
-            NOT = 26,
-            BITNOT = 27,
-            POS = 28,
-            NEG = 29,
-            NEW = 30,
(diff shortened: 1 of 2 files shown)


rhino/src/main/java/org/mozilla/javascript/Token.java
```diff
@@ -32,205 +32,206 @@ public static enum CommentType {
...
+            NULL = STRING + 1,
+            THIS = NULL + 1,
+            FALSE = THIS + 1,
+            TRUE = FALSE + 1,
+            SHEQ = TRUE + 1, // shallow equality (===)
+            SHNE = SHEQ + 1, // shallow inequality (!==)
+            REGEXP = SHNE + 1,
+            BINDNAME = REGEXP + 1,
+            THROW = BINDNAME + 1,
+            RETHROW = THROW + 1, // rethrow caught exception: catch (e if ) use it
```
rPraml, line 89: When inserting new opcodes, only the line before and after the insertion point has to be changed

gbrail:
This is awfully clever -- I have not seen this before in Java and I'm wondering why not! I'll try it tonight.

gbrail:
Does anyone else have a concern about this -- I'm having trouble seeing the downside and would like to merge it.

p-bakker:
LGTM (we use the same approach in JS running in Rhino works like a charm 🙂)

rbri:
> Does anyone else have a concern about this -- I'm having trouble seeing the downside and would like to merge it.

+1

We use this pattern since years for JDBC index based access of result sets (sometimes using the index is a bit faster)

rPraml:
>  I'm having trouble seeing the downside and would like to merge it.

The only downside is IMHO, you must understand the concept and add new constants the correct way (which should be more straight forward as before)
The produced class file should be the same, as the compiler resolves the constants at compile time.

p-bakker:
One slightly other approach that we also use and that is imho easier to understand and less likely to go wrong when adding new codes:
```
MIN_ICODE = 0,
Icode_DELNAME = MIN_ICODE--,

// Stack: ... value1 -> ... value1 value1
Icode_DUP = MIN_ICODE--,

....

// Call to GetTemplateLiteralCallSite
Icode_TEMPLATE_LITERAL_CALLSITE = MIN_ICODE--,
Icode_LITERAL_KEYS = MIN_ICODE--,
Icode_LITERAL_KEY_SET = MIN_ICODE--;
```

rPraml:
> One slightly different approachwe also use and that is imho easier to understand and less likely to go wrong when adding new codes (at the cost of declaring another (private) variable):
> 
> ```
> private static int count = 0,
> 
> static final int
>     // delete operator used on a name
>     Icode_DELNAME = count--,
> ...
> ```
Unfortunately, this does not work here, as the result isn't computed at compile time, but on class instantiation. So these are no compile-time 'constants', that can be used in switch statements or annotations.

p-bakker:
K, similar stuff in JavaScript dus do the trick, but Java ain't JavaScript or vise versa 🙂

rbri:
@rPraml thanks a lot for this

PR #1652, opened 2024-09-25
regenerate test262.properties, seems to be a bit outdated

rbri:


gbrail:
Exactly what did you do to make that happen:

    ./gradlew test --updateTest262properties=true

on both Windows and Linux, just now, produced 8000 lines of changes again, even on the master branch. I must be doing something wrong -- what command do you run?

rbri:
@gbrail nothing special - jdk 11,  and --rerun-tasks -DupdateTest262properties

What kind of diffs do you see?

PR #1654, opened 2024-09-26
Pass correct parameters in splitScope

rPraml:
When I tried to replace Tokens by Enum (#1639) I noticed, that the constructor `Scope(int position)` was invoked with the tokenType.
I would suggest, invoke the correct ctor here (but maybe someone with more experience should take a look at this)
I suspect it has no impact, but it is wrong in my opinion even if no tests fail.


rbri:
I think you are correct +1 for merging this.

gbrail:
Thanks, looks good!

PR #1656, opened 2024-09-28
feat: implement Symbol.toStringTag

naijun0403:
implement Symbol.toStringTag

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/ScriptRuntime.java
+++ b/rhino/src/main/java/org/mozilla/javascript/ScriptRuntime.java
@@ -1053,11 +1053,9 @@ public class ScriptRuntime {
         if (Undefined.isUndefined(obj)) return "[object Undefined]";
 
         String tag = null;
-        if (ScriptableObject.hasProperty(obj, SymbolKey.TO_STRING_TAG)) {
-            Object tagValue = ScriptableObject.getProperty(obj, SymbolKey.TO_STRING_TAG);
-            if (tagValue instanceof String) {
-                tag = (String) tagValue;
-            }
+        Object tagValue = ScriptableObject.getProperty(obj, SymbolKey.TO_STRING_TAG);
+        if (tagValue != Scriptable.NOT_FOUND && tagValue instanceof CharSequence) {
+            tag = tagValue.toString();
         }
 
         if (tag == null) {
--- a/rhino/src/main/java/org/mozilla/javascript/ScriptRuntime.java
+++ b/rhino/src/main/java/org/mozilla/javascript/ScriptRuntime.java
@@ -1052,17 +1052,12 @@ public class ScriptRuntime {
         if (obj == null) return "[object Null]";
         if (Undefined.isUndefined(obj)) return "[object Undefined]";
 
-        String tag = null;
         Object tagValue = ScriptableObject.getProperty(obj, SymbolKey.TO_STRING_TAG);
         if (tagValue != Scriptable.NOT_FOUND && tagValue instanceof CharSequence) {
-            tag = tagValue.toString();
+            return "[object " + tagValue.toString() + "]";
         }
 
-        if (tag == null) {
-            tag = obj.getClassName();
-        }
-
-        return "[object " + tag + "]";
+        return "[object " + obj.getClassName() + "]";
     }
 
     public static String toString(Object[] args, int index) {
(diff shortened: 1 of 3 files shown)


rhino/src/main/java/org/mozilla/javascript/ScriptRuntime.java
```diff
@@ -1051,7 +1051,20 @@ public static String toString(Object val) {
     static String defaultObjectToString(Scriptable obj) {
         if (obj == null) return "[object Null]";
         if (Undefined.isUndefined(obj)) return "[object Undefined]";
-        return "[object " + obj.getClassName() + ']';
+
+        String tag = null;
+        if (ScriptableObject.hasProperty(obj, SymbolKey.TO_STRING_TAG)) {
+            Object tagValue = ScriptableObject.getProperty(obj, SymbolKey.TO_STRING_TAG);
```
rbri, line 1057: Maybe you can use only ScriptableObject.getProperty and then check for Scriptable.NOT_FOUND.
This will save one property lookup.

rhino/src/main/java/org/mozilla/javascript/ScriptRuntime.java
```diff
@@ -1051,7 +1051,20 @@ public static String toString(Object val) {
...
+            if (tagValue instanceof String) {
```
rbri, line 1058: please check for CharSequence here to support at lease ConsString also

rhino/src/main/java/org/mozilla/javascript/ScriptRuntime.java
naijun0403: Thanks for your feedback! Now, check Scriptable.NOT_FOUND.

rhino/src/main/java/org/mozilla/javascript/ScriptRuntime.java
naijun0403: Thanks for your feedback! Now, check CharSequence

rhino/src/main/java/org/mozilla/javascript/ScriptRuntime.java
```diff
@@ -1053,11 +1053,9 @@ static String defaultObjectToString(Scriptable obj) {
         if (Undefined.isUndefined(obj)) return "[object Undefined]";
 
         String tag = null;
-        if (ScriptableObject.hasProperty(obj, SymbolKey.TO_STRING_TAG)) {
-            Object tagValue = ScriptableObject.getProperty(obj, SymbolKey.TO_STRING_TAG);
-            if (tagValue instanceof String) {
-                tag = (String) tagValue;
-            }
+        Object tagValue = ScriptableObject.getProperty(obj, SymbolKey.TO_STRING_TAG);
+        if (tagValue != Scriptable.NOT_FOUND && tagValue instanceof CharSequence) {
```
rbri, line 1057: cool, not_found is not a char sequence, the means you can let out this check also.
But i guess this makes to code less readable - maybe you have to add then a small comment about the missing not_found check

rhino/src/main/java/org/mozilla/javascript/ScriptRuntime.java
```diff
@@ -1051,7 +1051,18 @@ public static String toString(Object val) {
     static String defaultObjectToString(Scriptable obj) {
         if (obj == null) return "[object Null]";
         if (Undefined.isUndefined(obj)) return "[object Undefined]";
-        return "[object " + obj.getClassName() + ']';
+
+        String tag = null;
+        Object tagValue = ScriptableObject.getProperty(obj, SymbolKey.TO_STRING_TAG);
+        if (tagValue != Scriptable.NOT_FOUND && tagValue instanceof CharSequence) {
+            tag = tagValue.toString();
```
rbri, line 1058: what about doing a return here

    return "[object " + tagValue.toString() + "]";

Then you do not need to have the tag var at all and some more if's are gone for the price to have two static strings twice (but the compiler will take care of that).

naijun0403: Thanks for your feedback! As you advised, now returning earlier.

rhino/src/main/java/org/mozilla/javascript/ScriptRuntime.java
naijun0403: Thanks for your feedback! That looks better. As suggested, I removed the explicit check for NOT_FOUND and added a related comment.

rbri:
@naijun0403 thanks a lot, this looks now good for me, hopefully @gbrail will merge this soon

rbri:
@naijun0403 is is ok for you to squash the commits?

gbrail:
Looks good -- thanks!

PR #1637, opened 2024-09-20
Proxy and Reflect implementation

rbri:
Next try on the long story.... see #268

This is the follow up of #1431.

Closes #268

PR #1660, opened 2024-09-29
Reflect and Proxy implementation

rbri:
this is an recreation of #1637 from scratch

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/AbstractEcmaObjectOperations.java
+++ b/rhino/src/main/java/org/mozilla/javascript/AbstractEcmaObjectOperations.java
@@ -1,9 +1,12 @@
 package org.mozilla.javascript;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
+import java.util.function.Predicate;
 
 /**
  * Abstract Object Operations as defined by EcmaScript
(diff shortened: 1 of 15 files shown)


gbrail:
Once again, how do we move forward on this? I don't want to let it just sit. I think that we have two options:

1. Agree that this is in good enough shape and merge it into the codebase, possibly with a few new issues so that we can track additional work.
2. Create a new feature flag on Context, perhaps something like "ENABLE_EXPERIMENTAL_FEATURES" that guards this implementation until we all agree it's feature-complete.

@rbri , @p-bakker , what do you think?

rbri:
I do not like the idea of a feature flag because it is sooo demotivating.
With this pr
* Proxy is at 76/311 (24.44%)
* Reflect is at 13/153 (8.5%)

In the past we let things in
* Promise 392/631 (62.12%)
* BigInt 13/75 (17.33%) just reaching this level at the last week with a bunch of commits
* rest-parameters 5/11 (45.45%)

And also some old stuff like
* language/statements/let 91/145 (62.76%)
* language/expressions/arrow-function 167/343 (48.69%)
* ....
is far away from that.

I have the feeling doing all this discussion about the this PR is not really fair compared to the stuff we already have.

This is just the personal point of view, but i guess that most of the work is done based on a strong motivation by the peoples that doing this work. And together with all the technical things we have to have the motivation of the people in mind to go forward here (like we did in the last months).

----

But i will try to handle this serious - i will try to spend some time to document the problems in separate issues (but my time is really limited for the next months) and also i will rebase this like i did in the last weeks.

gbrail:
So far, this has come down to a conflict between contributors, as to whether or not we merge this. I think that this PR adds an important set of new capability to Rhino and that we should have it.

Unless someone has a strong objection, I'm asking @rbri to do one more conflict resolution, and then we should take it.

And after that, I'm going to propose that we call it a day and ship Rhino 1.8.0!

rbri:
@rbri will do the rebase after all the fantastic stuff from @andreabergia and friends is merged

rbri:
@gbrail rebase is done

gbrail:
Thanks for all your work on this!

PR #1662, opened 2024-09-30
updating v8-symbols.js

rbri:
As map/set/weakmap are supported by Rhino, we can enable one more test;
In contrast to the original test, weakmap does not support symbols as keys (have tested that with real browsers). Therfore I had to comment some lines.


gbrail:
Looks good -- thanks!

rbri:
Thanks @gbrail  - i guess you got the original file from the chrome folks - maybe we have to update that stuff to a recent version

PR #1664, opened 2024-09-30
The $262 object does not use LiveConnect any more

rPraml:
I noticed, that the $262 object, which defines the [host functions](https://github.com/tc39/test262/blob/main/INTERPRETING.md#host-defined-functions) was wrapped into a NativeJavaObject.

In other words, the `@JsMethod` annotations were useless, instead, the methods were exposed by LiveConnect.

In this PR, I've changed the $262 to a subclass of ScriptableObject, so that the annotations take place and we do not longer depend on LC when running the 262 tests.

### DIFF
--- a/tests/src/test/java/org/mozilla/javascript/tests/Test262SuiteTest.java
+++ b/tests/src/test/java/org/mozilla/javascript/tests/Test262SuiteTest.java
@@ -15,7 +15,6 @@ import java.io.FileReader;
 import java.io.FileWriter;
 import java.io.IOException;
 import java.io.Reader;
-import java.lang.reflect.InvocationTargetException;
 import java.math.BigDecimal;
 import java.math.RoundingMode;
 import java.nio.file.Path;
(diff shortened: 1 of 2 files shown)


gbrail:
The initialization makes sense to me if it's using the reflection based attributes there. You could also just create a ScriptableObject and attach LambdaFunctions to it if you'd like to do it a different way -- the latter is a bit faster once it's set up. There's a method in ScriptableObject that does it quickly but it's protected -- but it could be public.

tests/src/test/java/org/mozilla/javascript/tests/Test262SuiteTest.java
```diff
@@ -436,6 +435,43 @@ public static void tearDownClass() {
...
+        $262(Scriptable scope, Scriptable prototype) {
+            super(scope, prototype);
+        }
+
+        static $262 init(Context cx, Scriptable scope) {
+            $262 proto = new $262();
+            proto.setPrototype(getObjectPrototype(scope));
+            proto.setParentScope(scope);
+
+            proto.defineProperty("toSource", "Math", DONTENUM | READONLY | PERMANENT);
```
rbri, line 451: this looks strange to me

rbri:
Great - the lambda thing is much better - at least from my point of view - thanks a lot

tests/src/test/java/org/mozilla/javascript/tests/Test262SuiteTest.java
rPraml: Copy-paste error

gbrail:
Thanks -- looks good!

PR #1668, opened 2024-10-02
Removed obsolete IOException

rPraml:
This PR removes a never thrown IO-Exception

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/Context.java
+++ b/rhino/src/main/java/org/mozilla/javascript/Context.java
@@ -1414,21 +1414,16 @@ public class Context implements Closeable {
             String sourceName,
             int lineno,
             Object securityDomain) {
-        try {
-            return (Script)
-                    compileImpl(
-                            null,
-                            source,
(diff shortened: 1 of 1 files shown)


rhino/src/main/java/org/mozilla/javascript/Context.java
```diff
@@ -2513,8 +2501,7 @@ private ScriptNode parse(
             int lineno,
             CompilerEnvirons compilerEnv,
             ErrorReporter compilationErrorReporter,
-            boolean returnFunction)
-            throws IOException {
+            boolean returnFunction) {
```
rPraml, line 2504: I assume, that in former time a Reader was passed instead of `String sourceString` and that's why the IOException was still there. It can be removed and in all depending calls also.

Note: This might be a breaking change, if someone has implemented his own context with overwritten `compileImpl`

rbri:
fine for me i do not need backward compatibility here

gbrail:
This is the kind of incompatibility I can live with. Thanks!

PR #1669, opened 2024-10-02
Performance optimizations in initSafeStandardObjects

rPraml:
This PR optimizes the performance, especially in `initSafeStandardObjects`  by factor 2

The problem is that initSafeStandardObjects creates a dozen NativeError prototypes. Each prototype is initialized with an empty exception as default stackProvider.
This means, you construct ~10 exceptions (with stack trace etc.) on every initSafeStandardObjects.

This PR changes the initialization of prototypes in a way, that the stack is not generated for the prototypes (but still for new Error instances)

This speeds up the initialization about factor 2


gbrail:
This one cuts startup time almost in half, so I like it! Thanks!

PR #1670, opened 2024-10-02
Optimize performance of SlotMap a bit

rPraml:
This is a small improvement, but I imagine it is measurable

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/EmbeddedSlotMap.java
+++ b/rhino/src/main/java/org/mozilla/javascript/EmbeddedSlotMap.java
@@ -244,7 +244,6 @@ public class EmbeddedSlotMap implements SlotMap {
         for (Slot slot : oldSlots) {
             while (slot != null) {
                 Slot nextSlot = slot.next;
-                slot.next = null;
                 addKnownAbsentSlot(newSlots, slot);
                 slot = nextSlot;
             }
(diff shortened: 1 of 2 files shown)


rhino/src/main/java/org/mozilla/javascript/EmbeddedSlotMap.java
```diff
@@ -257,9 +256,8 @@ private static void copyTable(Slot[] oldSlots, Slot[] newSlots) {
      */
     private static void addKnownAbsentSlot(Slot[] addSlots, Slot slot) {
         final int insertPos = getSlotIndex(addSlots.length, slot.indexOrHash);
-        Slot old = addSlots[insertPos];
+        slot.next = addSlots[insertPos];
```
rPraml, line 259: Both changes in embeddedSlotMap save ~ 5 bytecode instructions: https://godbolt.org/z/1EjhsWs7E

rhino/src/main/java/org/mozilla/javascript/Slot.java
```diff
@@ -82,7 +82,7 @@ int getAttributes() {
         return attributes;
     }
 
-    synchronized void setAttributes(int value) {
+    void setAttributes(int value) {
```
rPraml, line 85: I don't understand why you should use "synchronized" here. `attributes` is not a long or double and write should be atomic: https://docs.oracle.com/javase/specs/jls/se9/html/jls-17.html#jls-17.7

(even if it is a long/double, it would be atomic on 64bit jvms)

gbrail:
I will check this myself later, but you can also verify the effects
yourself, by running:

BENCHMARK=SlotMap ./gradlew jmh

(or the equivalent on Windows)

gbrail:
FWIW this didn't result in any measurable changes (Java does a ton of optimization of bytecode including making uncontended synchronization very cheap). But if you all think it makes sense -- less bytecode is always good -- I'm happy to merge it.

Issue #1658, opened 2024-09-28
Rationalize use of optimization levels


gbrail:
Rhino supports an optimization level in the Context which goes from -1 to 9. From inspecting the code, I have seen that we use this as follows:

- -1: At level -1 (or lower) we disable classfile generation and run in interpreted mode. (This is the only way to run on Android.)
- 0: We force the optimization level to 0 when "generating debug" is set. (I didn't remember that and I'd like to understand more why we do this.)
- 1: At level 1 and higher we support larger sparse native arrays (200,000 elements versus 10,000)
- 1: At level 1 and higher we activate the "direct call" optimization, which tries to create direct function calls in the generated bytecode, rather than looking up a function in the global context and calling it via the Callable interface. This dramatically speeds up the common case when a script defines a number of top-level functions, and then calls them from elsewhere in the same script.
- 1: At level 1 and higher, when there are a string of math optimizations, we replace the boxed Number on the stack with a native Java double, and replace calls to the math operations in ScriptRuntime with native math instructions (like DADD, DMUL, etc). This speeds up math-heavy code, and can even speed up a simple "for" loop.

**Proposal**: Should we combine all optimization levels into one, so that we have only two modes -- interpreted and compiled?

I think that this is low-risk because the optimizations enabled at level 1 above have been in the codebase for a decade or more and seem pretty stable by now. The result would be less complexity to test, and all the tests would run 1/3 faster. Also, JavaScript engines and Java runtimes these days tend not to have a lot of choices of optimization level.

I can think of two reasons against this:

1. At some point we reduced the optimization level when "generating debug." We'll have to go back and figure out why we did that and what "generating debug" does any more (since we generate stack traces in all cases anyway).
2. We may add additional optimizations in the future, particularly around invokedynamic instructions. Should we make those active at a higher level as well?

Here is some data on the effects of the current optimizations, using the SunSpider benchmark suite. You'll see the differences -- sometimes dramatic -- between the levels.

Optimization level 9:

```
Benchmark                                                    Mode  Cnt      Score       Error  Units
SunSpiderBenchmark.AccessBinaryTreesState.accessBinaryTrees  avgt    5   4554.025 ┬▒  1372.394  us/op
SunSpiderBenchmark.AccessFannAccessNsieveState.accessNsieve  avgt    5   3757.903 ┬▒   675.956  us/op
[...]

PR #1671, opened 2024-10-02
Optimize performance in test262

rPraml:
This PR targets two issues I found with a profiler, while running the 262 test suite:

- The method `addTestFiles` has a nested loop with a complexity of O(n^2). I've replaced this with a hashmap, that improves startup time from ~3 sec down to <1 sec
- The getSourcePositionFromStack uses the more efficient StackWalker now

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/Context.java
+++ b/rhino/src/main/java/org/mozilla/javascript/Context.java
@@ -25,6 +25,7 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Locale;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
 import java.util.TimeZone;
 import java.util.function.UnaryOperator;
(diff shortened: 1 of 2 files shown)


rbri (approved):
LGTM

rbri:
NativeJavaListTest > length FAILED
    org.junit.ComparisonFailure at NativeJavaListTest.java:161

RegexpTest > unsupportedFlagCtor FAILED
    org.junit.ComparisonFailure at RegexpTest.java:71

rPraml:
Fixed the condition in the "filter"method. Now ShellTimerTest fails. Waiting fpr #1672, then I'll do a rebase

rbri:
@rPraml #1672 is merged go ahead :-)

rhino/src/main/java/org/mozilla/javascript/Context.java
```diff
@@ -2568,23 +2569,29 @@ static String getSourcePositionFromStack(int[] linep) {
...
-        }
 
-        return null;
+        return getSourcePositionFromJavaStack(linep);
+    }
+
+    /** Returns the current filename in the java stack. */
+    @SuppressWarnings("AndroidJdkLibsChecker")
+    // Android uses interpreter, so we should not get here.
+    static String getSourcePositionFromJavaStack(int[] linep) {
```
rPraml, line 2579: Rebase is done. Maybe someone can confirm, that android always uses interpreter, so we should not get here ;)

rbri: see @gbrail's comment on #1658

rbri:
@rPraml thanks for that, faster tests are always a good thing

PR #1661, opened 2024-09-29
improved ScriptRuntime.toPrimitive() (taken from #1611 done by @tonygermano)

rbri:
**this is an excerpt from PR https://github.com/mozilla/rhino/pull/1611 done by @tonygermano (commit https://github.com/mozilla/rhino/commit/9664e7753b4645c716b61ff75b4d777b4f458922, https://github.com/mozilla/rhino/commit/e17cc255a0e31355dab4aaee20af739eabd8d584)**

* more standard compliant implementation of ScriptRuntime.toPrimitive();
  * have not used the Optional for more backward compatibility
  * have changed the code a bit to be more in sync with the spec (explicit if stmt's)
* improved support for Symbol/BigInt in equals and compare

I did this to make it possible to merge this improvement soon - at least this is my hope. Will extract more smaller commits if this is merged.

PR #1674, opened 2024-10-03
Fixed ScriptRuntime.toPrimitive() and use it for toXXX() (taken from #1611 done by @tonygermano)

rbri:
This is another carve out from #1661.

It contains
* some major fixes for the toPrimitive() implementation itself and 
* the usage of toPrimitive() in ScriptRuntime#toString(), ScriptRuntime#toNumeric(), ScriptRuntime#toBigInt(), and ScriptRuntime#toNumber().

This fixes again a bunch of Test262 cases and should be still backward compatible.

I left the single steps as single commits to (hopefully) make the review simpler. But I think we should squash this on merge.

@gbrail @tonygermano @rPraml @p-bakker please have a look, it was a bit hard to do the backward compatibility in toPrimitive() - i'm only 99% sure that all this is correct...

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/NativeArray.java
+++ b/rhino/src/main/java/org/mozilla/javascript/NativeArray.java
@@ -1947,10 +1947,9 @@ public class NativeArray extends IdScriptableObject implements List {
     */
     private static Boolean js_includes(
             Context cx, Scriptable scope, Scriptable thisObj, Object[] args) {
-        Object compareTo = args.length > 0 ? args[0] : Undefined.instance;
 
         Scriptable o = ScriptRuntime.toObject(cx, scope, thisObj);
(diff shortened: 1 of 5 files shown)


rbri:
@gbrail will rebase my P&R branch after this is merged because i have to sort out some conflicts regarding to symbol handling

rPraml:
I did a quick review and this looks good to me

p-bakker:
Will this one resolve #1608?

rbri:
> Will this one resolve #1608?

I guess yes - will test this later today and report back.

gbrail:
The code looks good to me and it moves us closer enough to our goal so that I think we should merge it -- LMK if you all have any objections.

rbri:
Thanks, new one will arrive soon.

PR #1675, opened 2024-10-04
Fixed printing tokens - added two missing cases

andreabergia:
Teeny tiny PR 🙂


rPraml:
Maybe a small test `for (i = 0; i <= LAST_TOKEN; i++) typeToName(i)` would be also great here to prevent future mistakes ;)

rbri:
thanks @andreabergia and @rPraml

PR #1676, opened 2024-10-04
Skip importing same class twice (fixes #1463)

### LINKED ISSUE #1463: importClass usage that worked in 1.7.12 now fails in 1.7.14
Hi,

we recently upgraded from version 1.7.12 to 1.7.14 and now one of our test fails.
The test uses importClass with a class in our JUnit tests and run the same script multiple times. We get this exception in 1.7.14:

org.mozilla.javascript.EvaluatorException: Cannot import "JSLongTest" since a property by that name is already defined. (TestJSScriptCode2#2)
	at org.mozilla.javascript.DefaultErrorReporter.runtimeError(DefaultErrorReporter.java:79)
	at org.mozilla.javascript.Context.reportRuntimeError(Context.java:907)
	at org.mozilla.javascript.Context.reportRuntimeError(Context.java:963)
	at org.mozilla.javascript.Context.reportRuntimeErrorById(Context.java:914)
	at org.mozilla.javascript.ImporterTopLevel.importClass(ImporterTopLevel.java:216)
	at org.mozilla.javascript.ImporterTopLevel.js_importClass(ImporterTopLevel.java:176)
	at org.mozilla.javascript.ImporterTopLevel.execIdCall(ImporterTopLevel.java:257)
	at org.mozilla.javascript.IdFunctionObject.call(IdFunctionObject.java:100)
	at org.mozilla.javascript.optimizer.OptRuntime.callName(OptRuntime.java:59)
	at org.mozilla.javascript.gen.TestJSScriptCode2_1._c_script_0(TestJSScriptCode2:2)
	at org.mozilla.javascript.gen.TestJSScriptCode2_1.call(TestJSScriptCode2)
	at org.mozilla.javascript.ContextFactory.doTopCall(ContextFactory.java:380)
	at org.mozilla.javascript.ScriptRuntime.doTopCall(ScriptRuntime.java:3868)
	at org.mozilla.javascript.gen.TestJSScriptCode2_1.call(TestJSScriptCode2)
	at org.mozilla.javascript.gen.TestJSScriptCode2_1.exec(TestJSScriptCode2)

I tried to debug  org.mozilla.javascript.ImporterTopLevel.importClass and compared the sources. There is now a scope object that is used instead of this, and there are 2 differences that I see:

In 1.7.12, the NativeJavaClass cl is always the same Java object. And we always have a new ImporterTopLevel object in each run, so cl is not found:

    private void importClass(NativeJavaClass cl)
    {
        String s = cl.getClassObject().getName();
        String n = s.substring(s.lastIndexOf('.')+1);
        Object val = get(n, this);
        if (**val != NOT_FOUND** && val != cl) {
            throw Context.reportRuntimeError1("msg.prop.defined", n);
        }
        //defineProperty(n, cl, DONTENUM);
        put(n, this, cl);
    }

In 1.7.14, the NativeJavaClass cl is a different Java object (but equal) in each run. And the ImporterTopLevel object is the same on the second run, so the object is found, but not the same object, so the error occurs:

    private static void importClass(Scriptable scope, NativeJavaClass cl) {
        String s = cl.getClassObject().getName();
        String n = s.substring(s.lastIndexOf('.') + 1);
        Object val = scope.get(n, scope);
        if (val != NOT_FOUND && **val != cl**) {
            **throw Context.reportRuntimeErrorById("msg.prop.defined", n);**
        }
        // defineProperty(n, cl, DONTENUM);
        scope.put(n, scope, cl);
    }

The code would probably work as expected, if the cl object would be the same Java object as it is in 1.7.12.

Any idea why this happens in 1.7.14?

Regards,
Andreas Mueller

rPraml:


rPraml:
See #1463 - I would like to have a testcase for this situation ;)

gbrail:
So, do you have an idea how to test? Or do we feel that existing tests cover this well enough?

rPraml:
I found one case, where I can trigger that error:
```javascript
importPackage(java.util);
UUID.randomUUID(); // calls getPkgProperty("UUID", global, false)
importClass(java.util.UUID);
UUID.randomUUID();
```

While the first access calls `getPkgProperty("UUID", global, false)`, the second one calls calls `getPkgProperty("UUID",nativeJavaPackage, true)`
We could skip this, if we would follow the parent chain here:
```java
synchronized Object getPkgProperty(String name, Scriptable start, boolean createPkg) {
        Object cached = super.get(name, start);
        if (cached != NOT_FOUND) return cached;
        // TODO: check start.getParentScope()
```
I doubt, that the above use case is the same case as @Easy65 had, as this code fails also in 1.7.11.
I do not have much objections, to use `equals` instead of `==`, but it would be nice, to understand what was the breaking change. (yes #826 is very suspicious)


importPackage(java.util);\n"
                                + "UUID.randomUUID();\n" // calls getPkgProperty("UUID", global,
                                // false)
                                + "importClass(java.util.UUID);\n"
                                + "UUID.randomUUID()

rPraml:
Sorry, I've run out of ideas on how to write a test case that shows the regression between 1.7.11 and the current version. I tried different things, shared scopes etc., but either the test fails under both 1.7.11 and current version or passes under both versions.

If no further input comes in, then I would merge this, as it is an improvement.

Easy65:
Hi, I could finally provide a test, see my latest comment in the issue.
Regards,
Andreas

rPraml:
Thanks for the test. I'll try to convert this to a unittest the next days

tests/src/test/java/org/mozilla/javascript/tests/SealedSharedScopeTest.java
```diff
@@ -40,12 +41,10 @@ public void setUp() throws Exception {
         }
 
         ctx = Context.enter();
-        scope1 = ctx.newObject(sharedScope);
+        scope1 = new NativeObject();
```
rPraml, line 44: This was done too complex

gbrail:
Thanks for doing this! I used this commit along with a few more fixes in a different PR:

https://github.com/mozilla/rhino/pull/1770

PR #1677, opened 2024-10-04
Move the whole LiveConnect initializion in "initStandardObjects"

rPraml:
This PR moves the initialization of all live-connect related objects in "initStandardObjects".

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/Context.java
+++ b/rhino/src/main/java/org/mozilla/javascript/Context.java
@@ -1749,7 +1749,8 @@ public class Context implements Closeable {
         if (value instanceof String
                 || value instanceof Number
                 || value instanceof Boolean
-                || value instanceof Scriptable) {
+                || value instanceof Scriptable
+                || value instanceof Undefined) {
             return value;
         } else if (value instanceof Character) {
             return String.valueOf(((Character) value).charValue());
--- a/rhino/src/main/java/org/mozilla/javascript/MemberBox.java
+++ b/rhino/src/main/java/org/mozilla/javascript/MemberBox.java
@@ -89,26 +89,6 @@ final class MemberBox implements Serializable {
         return memberObject.getDeclaringClass();
     }
 
-    String toJavaDeclaration() {
-        StringBuilder sb = new StringBuilder();
-        if (isMethod()) {
-            Method method = method();
-            sb.append(method.getReturnType());
-            sb.append(' ');
-            sb.append(method.getName());
-        } else {
-            Constructor<?> ctor = ctor();
-            String name = ctor.getDeclaringClass().getName();
-            int lastDot = name.lastIndexOf('.');
-            if (lastDot >= 0) {
-                name = name.substring(lastDot + 1);
-            }
-            sb.append(name);
-        }
-        sb.append(JavaMembers.liveConnectSignature(argTypes));
-        return sb.toString();
-    }
-
     @Override
     public String toString() {
         return memberObject.toString();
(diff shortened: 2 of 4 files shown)


rhino/src/main/java/org/mozilla/javascript/MemberBox.java
```diff
@@ -89,26 +89,6 @@ Class<?> getDeclaringClass() {
...
-            Constructor<?> ctor = ctor();
-            String name = ctor.getDeclaringClass().getName();
-            int lastDot = name.lastIndexOf('.');
-            if (lastDot >= 0) {
-                name = name.substring(lastDot + 1);
-            }
-            sb.append(name);
-        }
-        sb.append(JavaMembers.liveConnectSignature(argTypes));
-        return sb.toString();
```
rPraml, line 109: this is a liveconnect only thing, so I moved it away from here

rhino/src/main/java/org/mozilla/javascript/ScriptRuntime.java
```diff
@@ -4402,26 +4413,27 @@ public static Scriptable newCatchScope(
...
-                Object wrap = cx.getWrapFactory().wrap(cx, scope, re, null);
-                ScriptableObject.defineProperty(
-                        errorObject,
-                        "rhinoException",
-                        wrap,
-                        ScriptableObject.PERMANENT
-                                | ScriptableObject.READONLY
-                                | ScriptableObject.DONTENUM);
+            if (ScriptRuntime.isLiveConnectEnabled(scope)) {
+                if (javaException != null && isVisible(cx, javaException)) {
```
rPraml, line 4417: do not declare these additional properties, when the scope does not support liveConnect

rhino/src/main/java/org/mozilla/javascript/Context.java
```diff
@@ -1749,7 +1749,8 @@ public static Object javaToJS(Object value, Scriptable scope, Context cx) {
         if (value instanceof String
                 || value instanceof Number
                 || value instanceof Boolean
-                || value instanceof Scriptable) {
+                || value instanceof Scriptable
+                || value instanceof Undefined) {
```
rPraml, line 1753: do not pass undefined to wrapperFactory

gbrail:
I haven't read this yet, but I was wondering in general if it'd make sense to use the "ServiceLoader" Java facility to load optional classes during initialization. I was considering trying to do it for the XML implementation, and I could see how it'd also be helpful here, even if it's in the same module, and certainly in the future if it moves to a different module. It would let us drop in additional standard objects by just adding to the classpath, rather than messing around with reflection and stuff.

https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/ServiceLoader.html

p-bakker:
@gbrail like that idea, as I foresee more optional add-ons in the future for Rhino, like a WASM implementation depending on a 3-rd party lib, EcmaScript intl-402 stuff that needs ICU4J etc. 

The ability to develop those outside the core Rhino engine and some auto-loading mechanism of they are on the classpath would make for a good facility I think

rPraml:
Thanks for feedback. Yes that would be the long term plan:
The wrapfactory (or standardobjects) should be discovered by ServiceLoader.

But first I would like to sort out some things.

You may take a look in my experimental branch #1655

Maybe the xml stuff could also be a good candidate to move it to a module first

rhino/src/main/java/org/mozilla/javascript/Context.java
p-bakker: Can you elaborate on the 'why' of this?

rPraml: `Undefined` is already a js compatible object, so I think, it does not need a `javaToJs` conversion.

Background: I plan to make the wrapperFactory optional (no wrapperFactory == no liveConnect) and I found some failing tests, that have passed `undefined` here, although they do not use use liveConnect

> If value is an instance of String, Number, Boolean, Function or Scriptable, it is returned
> as it and will be treated as the corresponding JavaScript type of string, number, boolean,
> function and object.

I will line up the code with the documentation. (Function mentioned in doc, but not handled in code) and maybe extract this in a separate commit.

PR #1679, opened 2024-10-04
NativeJavaClass: Add class literal support

### LINKED ISSUE #757: please support java class access, eg: java.lang.Object.class

camnwalter:
First time making tests so let me know if there is some case that isn't covered or whatever else I should change.

Closes #757


p-bakker:
Looks like at least one of the testcases you added fails

camnwalter:
I think that should be fixed now. I guess it was running it multiple times and so the static count variable was higher than expected

gbrail:
The code looks OK and I see the issue, but will people understand what we changed here? I think that at the very least the PR or commit should describe what we did and what people can do now -- since I'm not sure where the LiveConnect "spec" or docs are, I'm not sure how people will understand what is or isn't possible unless we update our docs.

camnwalter:
Renamed the PR and added a more detailed description on what was changed. Not sure where else to put this.

gbrail:
IMO, this is basically adding a new feature to LiveConnect, which looks useful to me, but I could use some more opinions.

What do regular users of the Java interop think about this?

p-bakker:
Did you see the discussion on #757?

This makes sense to me, to simplify access to the class

rPraml:
Hello, we use LC a lot. But until now I can't remember that we were missing this feature. If a Java method expected a class as parameter, you could just call `myJavaFunc(java.lang.String)`

Nevertheless, I can imagine use cases where it would be necessary.
But I also think we should just rename this "__javaObject__" to "class" because 
- it is relatively undocumented
- and therefore unintuitive.
- There is no test.
- it is probably hardly used ([Google search](https://www.google.com/search?q=%22__javaObject__%22+rhino) does not really find something, expect these GH issues or rhino source code.)

So, I would have nothing against simply replacing `__javaObject__` with the more intuitive `class`. But perhaps others can give their opinion on whether we can afford this breaking change or if we should impmement the same functionality twice?

BTW: It was added 17 years ago with this commit
https://github.com/mozilla/rhino/commit/d6233135d460a52d34663e2a57d53c0409e5a4f0
and then changed from `__javaClass__` to `__javaObject__`
https://github.com/mozilla/rhino/commit/c782d036e849f94f655b52ea5fc54c80c9b7c738

camnwalter:
> Nevertheless, I can imagine use cases where it would be necessary.

Yeah the main reason I see it being useful is for reflection.

gbrail:
This seems useful but given that so much of this stuff isn't documented we'll I'm kind of waiting to see if we have more opinions.

I am trying to resurrect some of the old docs, once I do that perhaps you can all take a look and see if that's a good place to update.

p-bakker:
@gbrail which old docs are you referring to? Thought we did resurrect all (relevant) docs from the Mozilla website into rhino.github.io already

Issue #1600, opened 2024-08-30
In assignments, property keys should be evaluated only after checking the object is not null


andreabergia:
This should throw a `TypeError` because we are accessing `null`, but instead it is throwing a `MyError`:

```js
function MyError() {}
MyError.prototype.toString = () => 'MyError';

var base = null;
var prop = {
    toString: function() {
        throw new MyError();
    }
  };
 
base[prop] = 42;
```

andreabergia:
I have some ideas, but this requires checking the spec carefully (and probably what other engines do). But I definitely won't get to work on this in the next couple of weeks.

gbrail:
We'll get to this later on. Thanks!

PR #1684, opened 2024-10-07
Fix: make optional chaining expression work

andreabergia:
I.e. makes this work:

```js
o = {a: true};
o?.['a']
```

Note that the spec says that the expression on the right-hand side should _not_ be evaluated if the left-hand side object is null or undefined.

Follow-up from https://github.com/mozilla/rhino/pull/1593

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/optimizer/BodyCodegen.java
+++ b/rhino/src/main/java/org/mozilla/javascript/optimizer/BodyCodegen.java
@@ -1370,6 +1370,50 @@ class BodyCodegen {
                 }
                 break;
 
+            case Token.GETELEM_OPTIONAL:
+                {
+                    generateExpression(child, node); // object
+
+                    int getExpr = cfw.acquireLabel();
+                    int putUndefined = cfw.acquireLabel();
+                    int after = cfw.acquireLabel();
+
+                    // if (object == null) goto putUndefined
+                    cfw.add(ByteCode.DUP);
+                    ++maxStack;
+                    cfw.add(ByteCode.IFNULL, putUndefined);
+
+                    // if (!Undefined.isUndefined(object)) goto getExpr
+                    cfw.add(ByteCode.DUP); // No need to increase maxStack again!
+                    cfw.addInvoke(
+                            ByteCode.INVOKESTATIC,
+                            "org.mozilla.javascript.Undefined",
+                            "isUndefined",
+                            "(Ljava/lang/Object;)Z");
+                    cfw.add(ByteCode.IFEQ, getExpr);
+
+                    cfw.markLabel(putUndefined);
+                    cfw.add(ByteCode.POP); // Remove object
+                    cfw.add(
+                            ByteCode.GETSTATIC,
+                            "org/mozilla/javascript/Undefined",
+                            "instance",
+                            "Ljava/lang/Object;");
+                    cfw.add(ByteCode.GOTO, after);
+
+                    cfw.markLabel(getExpr);
+                    generateExpression(child.getNext(), node); // id
+                    cfw.addALoad(contextLocal);
+                    cfw.addALoad(variableObjectLocal);
+                    if (node.getIntProp(Node.ISNUMBER_PROP, -1) != -1) {
+                        addDynamicInvoke("PROP:GETINDEX", Signatures.PROP_GET_INDEX);
+                    } else {
+                        addDynamicInvoke("PROP:GETELEMENT", Signatures.PROP_GET_ELEMENT);
+                    }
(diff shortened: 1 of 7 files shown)


rbri:
Great!

p-bakker:
I guess you're saying that the previously merged optional chaining operator support didn't properly supported computed properties? 

Wrt to your 'note': think you filled #1600 for that already, no?

andreabergia:
> I guess you're saying that the previously merged optional chaining operator support didn't properly supported computed properties?

It didn't support them at all - it didn't even parse the syntax 🙂 

> Wrt to your 'note': think you filled #1600 for that already, no?

Yeah, but I have done something smaller in scope in this PR: I've just made this _new_ implementation work correctly per the spec. I haven't fixed the general problem because, honestly, I can't imagine how to do it without having a big performance impact.

rhino/src/main/java/org/mozilla/javascript/optimizer/BodyCodegen.java
```diff
@@ -1370,6 +1370,50 @@ private void generateExpression(Node node, Node parent) {
...
+                            "instance",
+                            "Ljava/lang/Object;");
+                    cfw.add(ByteCode.GOTO, after);
+
+                    cfw.markLabel(getExpr);
+                    generateExpression(child.getNext(), node); // id
+                    cfw.addALoad(contextLocal);
+                    cfw.addALoad(variableObjectLocal);
+                    if (node.getIntProp(Node.ISNUMBER_PROP, -1) != -1) {
+                        addDynamicInvoke("PROP:GETINDEX", Signatures.PROP_GET_INDEX);
```
gbrail, line 1409: Awesome -- using the new instructions!

I do see from the coverage report that we don't get to this specific case, and given that we just changed it, I'd feel better if we did -- I have a suggestion below.

rhino/src/test/java/org/mozilla/javascript/tests/OptionalChainingOperatorTests.java
```diff
@@ -126,4 +126,27 @@ public void testOptionalChainingOperator() {
                     return null;
                 });
     }
+
+    @Test
+    public void expressionsInOptionalChaining() {
+        Utils.assertWithAllOptimizationLevelsES6(true, "o = {a: true}; o?.['a']");
+        Utils.assertWithAllOptimizationLevelsES6(Undefined.instance, "o = null; o?.['a']");
+        Utils.assertWithAllOptimizationLevelsES6(Undefined.instance, "o = undefined; o?.['a']");
```
gbrail, line 134: Could you add a version of this that specifically uses an index (like "o?.[1]")? There is a whole code path in the bytecode that I see we're not going down and I'd feel better if we were.

gbrail:
I see one place where I think that the code coverage could be improved. Can you please take a look? Otherwise this looks good!

rhino/src/main/java/org/mozilla/javascript/optimizer/BodyCodegen.java
andreabergia: I copy&pasted the code from the existing `GETELEM` path [here](https://github.com/mozilla/rhino/blob/master/rhino/src/main/java/org/mozilla/javascript/optimizer/BodyCodegen.java#L1366):

```java
            case Token.GETELEM:
                generateExpression(child, node); // object
                generateExpression(child.getNext(), node); // id
                cfw.addALoad(contextLocal);
                cfw.addALoad(variableObjectLocal);
                if (node.getIntProp(Node.ISNUMBER_PROP, -1) != -1) {
                    addDynamicInvoke("PROP:GETINDEX", Signatures.PROP_GET_INDEX);
                } else {
                    addDynamicInvoke("PROP:GETELEMENT", Signatures.PROP_GET_ELEMENT);
                }
                break;
```

However, after trying to add a simple test case for this, I've actually realized that the _original_ code I copied does not work either. I.e. if I do this:

```java
        Utils.assertWithAllOptimizationLevelsES6(true, "o = {[0]: true}; o[0]");
```

then the code path for `PROP:GETINDEX` is not followed. So, this looks like a problem in https://github.com/mozilla/rhino/pull/1645

rhino/src/test/java/org/mozilla/javascript/tests/OptionalChainingOperatorTests.java
andreabergia: Pushed it, but see my comment above.

andreabergia:
Going to mark this as draft after the problems I've summarized in #1688

andreabergia:
Closed in favour of https://github.com/mozilla/rhino/pull/1694

PR #1685, opened 2024-10-07
Use to primitive for plus operation and date (taken from #1611 done by @tonygermano)

rbri:
This is another carve out from https://github.com/mozilla/rhino/pull/1661.

Please squash this on merge.

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/NativeDate.java
+++ b/rhino/src/main/java/org/mozilla/javascript/NativeDate.java
@@ -342,7 +342,6 @@ final class NativeDate extends IdScriptableObject {
                         throw ScriptRuntime.typeErrorById(
                                 "msg.invalid.toprimitive.hint", ScriptRuntime.toString(arg0));
                     }
-                    ;
 
                     return ScriptableObject.getDefaultValue(o, typeHint);
                 }
--- a/rhino/src/main/java/org/mozilla/javascript/NativeDate.java
+++ b/rhino/src/main/java/org/mozilla/javascript/NativeDate.java
@@ -255,6 +255,10 @@ final class NativeDate extends IdScriptableObject {
                 arity = 1;
                 s = "toJSON";
                 break;
+            case SymbolId_toPrimitive:
+                initPrototypeMethod(
+                        DATE_TAG, id, SymbolKey.TO_PRIMITIVE, "[Symbol.toPrimitive]", 1);
+                return;
             default:
                 throw new IllegalArgumentException(String.valueOf(id));
         }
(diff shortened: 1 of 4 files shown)


rbri:
From my point of view this is ready to be squash-merged... what do you think?

rhino/src/main/java/org/mozilla/javascript/NativeDate.java
```diff
@@ -323,6 +327,25 @@ public Object execIdCall(
...
+                    if ("string".equals(hint) || "default".equals(hint)) {
+                        typeHint = ScriptRuntime.StringClass;
+                    } else if ("number".equals(hint)) {
+                        typeHint = ScriptRuntime.NumberClass;
+                    }
+                    if (typeHint == null) {
+                        throw ScriptRuntime.typeErrorById(
+                                "msg.invalid.toprimitive.hint", ScriptRuntime.toString(arg0));
+                    }
+                    ;
```
gbrail, line 345: Minor, but do you know why there is a semicolon here?

tests/testsrc/test262.properties
```diff
@@ -697,18 +697,7 @@ built-ins/Date 90/770 (11.69%)
...
-    prototype/Symbol.toPrimitive/hint-invalid.js
-    prototype/Symbol.toPrimitive/hint-number-first-invalid.js
-    prototype/Symbol.toPrimitive/hint-number-first-non-callable.js
-    prototype/Symbol.toPrimitive/hint-number-first-valid.js
-    prototype/Symbol.toPrimitive/hint-string-first-invalid.js
-    prototype/Symbol.toPrimitive/hint-string-first-non-callable.js
-    prototype/Symbol.toPrimitive/hint-string-first-valid.js
-    prototype/Symbol.toPrimitive/length.js
-    prototype/Symbol.toPrimitive/name.js
+    prototype/Symbol.toPrimitive/called-as-function.js
```
gbrail, line 700: It looks like we made a bunch of new tests pass, but we also made a test that previously passed fail, from what it looks like. Do you know why this appears to be failing now?

gbrail:
Yes, it looks good, but I have two minor points -- can you check them out? Thanks!

tests/testsrc/test262.properties
rbri: Have seen this alteady but there was no time for that. Will check

rbri: called-as-function.js seems to be a more general problem we have the same with the toJSON prototype.
This function passes in the past because a type error was expected (and there was no toPrimitiveSupport).

Will have another look at this and maybe open an issue for that, but in gerneral it is nothing introduced by this one.

gbrail:
OK -- thanks for looking at the test and removing that huge problem with the semicolon ;-)

PR #1686, opened 2024-10-07
Add a dynamic linker that understands consts

gbrail:
This is an optimization that plugs in to the dynamic linking framework to make bytecode operations that should result in constant values efficiently return a const and avoid a whole bunch of hash table lookup operations. It can make scripts much faster IF and only if they use the "const" keyword.


p-bakker:
Awesome! 

Question though: Rhino currently only supports function & scope-level consts, not block-level. While attempting to add block-level const support, I so far just focused on interpreted mode, but I assume already or otherwise due to this change, changes would be required on the bytecode generation side as well?

gbrail:
This works at a much lower level -- when it comes time to look up a property or a value on the current scope, that's when this code gets activated, and it checks to see whether that property was marked as a constant. So, whatever happens at the higher levels, this code is unaffected.

If you're having to make changes to the "interpreter" to make scope-level consts work (as opposed to the AST or the IR), then yes, we'll probably have to make similar changes to the bytecode generation. Once you have an idea of what those changes are I can look at the bytecode generation if you'd like.

PR #1687, opened 2024-10-10
Bring back the docs!

gbrail:
I retrieved all the relevant documentation I could find from the Wayback Machine, and converted it to Markdown, and added it to the "docs" subdirectory.

This includes some really useful stuff about how to embed Rhino, as well as some stuff that's just out of date.

I fixed a few obvious problems but I'd love it if some of you could check this for accuracy before we go.


p-bakker:
I don't get this, this is all already retrieved and included in our docs on rhino.github.io, no?

p-bakker:
Think we should instead toss out the /docs for in our main repo and update some of the pointers in the readme to point to the relevant pages on rhino.github.io

gbrail:
Right, I lost track of that -- even better. But I'll close this and open a
new one to fix the readme. Thanks!

Issue #1688, opened 2024-10-10
Problems with optional chaining and nullish coalesce operators


andreabergia:
While working on #1684 me and my colleagues @0xe and @nabacg have discovered a few problems with the two newly implemented operators.

For the optional chaining `?.` we do not support the syntax `a?.[ expr ]`, which #1684 would fix.
Also, while the original PR #1593 implemented partial support for optional function calling, i.e. `a.b?.()` works, it does not support properly other kind of function calls such as `a?.()` or `a[b]?.()`.

For nullish coalesce operator `??` we have instead discovered that we "leak" some temporary state. In #1624 a fix was implemented so that something like `f() ?? 42` would evaluate `f()` just once by creating a temporary name. However, that temp name leaks out to the containing scope, i.e.:

```js
a = 1;
a ?? 42;
this.$0  // evaluates to "1"
Object.keys(this)  // will contain $0
```

Furthermore, we can even overwrite variables:

```js
$0 = false;
a = true;
a ?? 42;
$0 // true
```

Given the similarities between these operators and the problems and missing features, we are trying to come up with a single approach that works with all cases. We hope to submit a PR in the next few weeks to fix all of this.

andreabergia:
Should be fixed since the related PRs have been merged.

PR #1624, opened 2024-09-12
FIX: Evaluation of ??

0xe:
Fixes #1623

PR #1694, opened 2024-10-11
Reimplement optional chain and nullish coalesce

andreabergia:
This reimplements #1591, #1593, and #1624 to fix the problems highlighted in #1688. It also replaces #1684.
Basically it implements both `??` and `?.` operators.

The structure in the IR is simplified; there is just a node property to mark that the property access is optional.

The bytecode generated (for both the compiled JVM class, and our interpreter) for `a?.b` is something of this kind:

```
compute 'a'
if isNullOrUndefined {
  put undefined on the stack
} else {
  put 'b'
  call ScriptRuntime.getElem
}
```

I have added two new conditional jump instructions in the interpreter, `IF_NULL_UNDEF` and `IF_NOT_NULL_UNDEF`, to implement the comparison and the if with a small number of instructions. There is also a new method `OptRuntime.isNullOrUndefined` for compiled mode, to avoid repeating three `if` in a row in each generated class (and, anyway, the JVM might do the inlining itself if it looks useful).

The approach is inspired by what other JS engines do; for example, `v8-debug` will print this for `a?.b`:

```
    0 S> 0x3d1200040048 @    0 : 21 00 00          LdaGlobal [0], [0]
         0x3d120004004b @    3 : c9                Star1
         0x3d120004004c @    4 : a4 08             JumpIfUndefinedOrNull [8] (0x3d1200040054 @ 12)
    1 E> 0x3d120004004e @    6 : 2f f8 01 02       GetNamedProperty r1, [1], [2]
         0x3d1200040052 @   10 : 8f 03             Jump [3] (0x3d1200040055 @ 13)
         0x3d1200040054 @   12 : 0e                LdaUndefined
         0x3d1200040055 @   13 : ca                Star0
    4 S> 0x3d1200040056 @   14 : af                Return
```

One non-obvious detail is that if you have `a?.b.c`, the access to `c` is _implied_ to be optional. This is necessary to implement the short circuiting, as required by the spec. The current implementation is not super optimized; it will basically do:

```
compute 'a'
if isNullOrUndefined {
  put undefined on the stack
} else {
  put 'b'
  call ScriptRuntime.getElem
}
if isNullOrUndefined {
  put undefined on the stack
} else {
  put 'c'
  call ScriptRuntime.getElem
}
```

Obviously the second `if` should be collapsed with the first, but I found this really complex to implement, so I did not. We get a bit worse performances than other JS engines, but hopefully this pattern is not too common to be a significant problem.

For normal property access `a.b` there is _no_ change in bytecode or runtime instructions. 

Note that this PR is only partial - function calls such as `a.b?.()`, which were already present in #1593, are implemented in https://github.com/mozilla/rhino/pull/1702


rbri:
@andreabergia :-D

     Execution failed for task ':tests:spotlessJavaCheck'.
      The following files had format violations:
      src/test/java/org/mozilla/javascript/tests/OptionalChainingOperatorTest.java

gbrail:
This looks good to me -- any more opinions? Otherwise I'd like to merge it.

rbri:
+1 for merge

and i can't wait to see the other great stuff from @andreabergia merged too...

PR #1701, opened 2024-10-16
fix endless loop in ScriptRuntimem toString() and toNumber() and add tests to ensure backward comatibility

rbri:
while integrating the latest changes into htmlunit core-js i noticed that i have introduced a no longer backward compatible behaviour. Sorry - and in fact this introduces an endless loop for some calls.

This fixes that...


gbrail:
Looks good -- thanks!

rbri:
Thanks @gbrail

Issue #1641, opened 2024-09-22
Default Parameters + Destructuring  - Pending items


0xe:
Address items pending from https://github.com/mozilla/rhino/pull/1640

- Destructuring with defaults in for loops. This should evaluate to 3, instead throws NPE:

```javascript
function a() {}; (function() { 
            for (let {x = a()} = {}; ; ) { 
                return 3; 
            }
        })()
```

- Reading [Symbol.iterator] from default array/object

```javascript
function f([x = 1, y = 2] = {x: 3, y: 4}) {
 return x + y;
 }
```

This should throw a TypeError as `[Symbol.iterator]` is not implemented for that object on the rhs, instead uses defaults to return `3`.

```javascript
function f([x = 1, y = 2] = {x: 3, y: 4, 
                              [Symbol.iterator]: function*() { 
                                yield 3; 
                                yield 4; 
                              }}) {
                                return x + y;
                              }
```

And this should work because we call into iterator to get the default values. Instead we throw a NPE currently.

p-bakker:
@0xe any chance you'll be picking up this one not that the previous part has been merged?

0xe:
> @0xe any chance you'll be picking up this one not that the previous part has been merged?

Yep. I'm planning on starting with destructuring with defaults in for loops.

0xe:
Looking to add support for destructuring with defaults in for loops, I found some problems in how default values were transformed. I've created a PR to fix those first: https://github.com/mozilla/rhino/pull/1708.

gbrail:
Looks like we have made progress since the last time we updated this issue. Can anyone update where we are with this? I'd like to decide if we're going to fix this in the next week or so or move it to a new release.

0xe:
> Looks like we have made progress since the last time we updated this issue. Can anyone update where we are with this? I'd like to decide if we're going to fix this in the next week or so or move it to a new release.

Sorry, this fell off the radar a bit. I'm still working on fixing this but won't be able to make it by next week.

PR #1708, opened 2024-10-22
FIX: Transform default values correctly in destructuring

0xe:
Default values for destructuring assignments weren't transformed correctly. Partially addresses #1641.

For example, all of these expressions with default values were not being evaluated correctly:

```javascript
function a() { return 12;};  function b([x = a()] = []) { return x }; b()
function a() { return 12;};  function b([x = a()] = [1]) { return x }; b()
function a() { return 12;};  function b([x = 1] = [a()]) { return x }; b()
var a = { p1: { p2: 121}}; function b([x = a.p1.p2] = []) { return x }; b()
```

```javascript
function a() { return 12;};  function b({x = a()} = {}) { return x }; b()
function a() { return 12;};  function b({x = a()} = {x: 1}) { return x }; b()
var a = { p1: { p2: 121}}; function b({x = a.p1.p2} = {}) { return x }; b()
function a() { return 12;};  function b({x = 1} = {x: a()}) { return x }; b()
```

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/IRFactory.java
+++ b/rhino/src/main/java/org/mozilla/javascript/IRFactory.java
@@ -613,8 +613,10 @@ public final class IRFactory {
             if (dfns != null) {
                 for (var i : dfns) {
                     Node a = i[0];
-                    AstNode b = (AstNode) i[1];
-                    a.replaceChild(b, transform(b));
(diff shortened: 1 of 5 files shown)


rbri:
the number of fixed tests is really impressive.good job.

gbrail:
This is great -- thanks!

PR #1709, opened 2024-10-24
Convert NativeArrayBuffer, NativeDataView, and TypedArrays to lambda

camnwalter:
This PR converts the ArrayBuffer and adjacent classes to using LambdaConstructor.

I haven't added any new functionality, just a basic conversion from IdScriptable to using lambdas. Hopefully will get to adding more correct functionality in the future...

This should fix these inside #963 :
- TypedArray.buffer/byteLength/byteOffset
- DataView.buffer/byteLength/byteOffset
- ArrayBuffer.byteLength
as they should all be accessor properties now.

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/typedarrays/NativeArrayBuffer.java
+++ b/rhino/src/main/java/org/mozilla/javascript/typedarrays/NativeArrayBuffer.java
@@ -41,7 +41,6 @@ public class NativeArrayBuffer extends ScriptableObject {
                         1,
                         LambdaConstructor.CONSTRUCTOR_NEW,
                         NativeArrayBuffer::js_constructor);
-
         constructor.setPrototypePropertyAttributes(DONTENUM | READONLY | PERMANENT);
 
         constructor.defineConstructorMethod(
@@ -58,7 +57,6 @@ public class NativeArrayBuffer extends ScriptableObject {
                 cx, "byteLength", NativeArrayBuffer::js_byteLength, DONTENUM | READONLY);
 
         ScriptableObject.defineProperty(scope, CLASS_NAME, constructor, DONTENUM);
-
         if (sealed) {
             constructor.sealObject();
         }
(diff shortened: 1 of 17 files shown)


rhino/src/main/java/org/mozilla/javascript/typedarrays/NativeArrayBuffer.java
```diff
@@ -106,120 +130,68 @@ public NativeArrayBuffer slice(double s, double e) {
...
+            Context cx,
+            Scriptable scope,
+            Scriptable thisObj,
+            LambdaConstructor defaultConstructor,
+            Object[] args) {
+        NativeArrayBuffer self = getSelf(thisObj);
+        double start = isArg(args, 0) ? ScriptRuntime.toNumber(args[0]) : 0;
+        double end = isArg(args, 1) ? ScriptRuntime.toNumber(args[1]) : self.getLength();
+        int endI =
+                ScriptRuntime.toInt32(
```
gbrail, line 157: I may be wrong here, but aren't most of the values here -- "self.getLength()" -- integers already? And is this method supposed to work if "start" and "end" are not integers? It looks like we are converting a bunch of integers into a floating-point value and then back to integers again.

Or are the "start" and "end" parameters here actually doubles because they could theoretically represent a value larger than 2^31? And if so can we handle that with our native byte arrays anyway?

In other words I'm asking if this whole thing shouldn't be done with integer arithmetic, and there may well be a reason!

rhino/src/main/java/org/mozilla/javascript/typedarrays/NativeFloat32Array.java
```diff
@@ -39,24 +40,48 @@ public String getClassName() {
...
+                                        lscope,
+                                        args,
+                                        NativeFloat32Array::new,
+                                        BYTES_PER_ELEMENT));
+        constructor.setPrototypePropertyAttributes(DONTENUM | READONLY | PERMANENT);
+        NativeTypedArrayView.init(cx, scope, constructor, NativeFloat32Array::realThis);
+        constructor.defineProperty(
+                cx,
+                "BYTES_PER_ELEMENT",
+                (Scriptable thisObj) -> BYTES_PER_ELEMENT,
```
gbrail, line 61: Is there a reason why this is defined as a lambda function call -- it seems like rather than making the property value a function that returns a constant, can't we just set the value directly as an Integer object? This pattern happens in all the subclasses so my same comment would apply there too.

rhino/src/main/java/org/mozilla/javascript/typedarrays/NativeTypedArrayView.java
```diff
@@ -151,6 +150,355 @@ private static int toIndex(double num) {
...
+                0,
+                (Context lcx, Scriptable lscope, Scriptable thisObj, Object[] args) ->
+                        js_reverse(lcx, lscope, thisObj, args, realThis),
+                DONTENUM,
+                DONTENUM | READONLY);
+        constructor.definePrototypeMethod(
+                scope,
+                "set",
+                0,
+                (Context lcx, Scriptable lscope, Scriptable thisObj, Object[] args) -> {
```
gbrail, line 379: Minor, but this seems like a bit too much code to be embedded inline like this, and it'd probably benefit from being a separate method.

gbrail (commented):
Just a few small things, but in general this is really good progress and I like how it fixes so many tests as well!

rhino/src/main/java/org/mozilla/javascript/typedarrays/NativeArrayBuffer.java
camnwalter: I'll have to look into it. I think I just copy pasted what was done before, but I'll see.

rhino/src/main/java/org/mozilla/javascript/typedarrays/NativeFloat32Array.java
camnwalter: Will change. Thanks

rhino/src/main/java/org/mozilla/javascript/typedarrays/NativeArrayBuffer.java
camnwalter: It does allow for non-integers but they are supposed to get converted to integers before all the max/mins in the following lines.

camnwalter:
Rebased on master

rhino/src/main/java/org/mozilla/javascript/typedarrays/NativeArrayBuffer.java
camnwalter: However, after seeing what tests failed, you were right. It looks like they can be more than 2^31, but in that case, the value would be greater than the length of the byte array, so when taking the min, it will just be the original byte array length, avoiding any issues.

gbrail:
Thanks -- this looks good!

PR #1711, opened 2024-10-25
Add INDY ops for math operations

gbrail:
Right now, this makes the bytecode a tiny bit more readable IMO, and doesn't really affect performance. 

Once we have these in, however, we can add additional type-specific linkers which will shorten the codepaths to these common operations for common types, and avoid the long strings of "instanceof" checks that make up most of the generic math operations today.


gbrail:
This seems uncontroversial so I'm going to merge it, and happy to entertain questions any time!

PR #1714, opened 2024-10-29
Allow passing `Undefined.instance` to `ScriptRuntime.evalSpecial`

andreabergia:
The method `ScriptRuntime.evalSpecial` takes a generic `Object` for `thisArg`. However, it then casts it to `Scriptable`.
We have some (rare) code paths where we end up passing `Undefined.instance` to that method, where we get a `ClassCastException`.
This simple PR allows it and adds a test.


rbri (approved):

gbrail:
Looks good to me too -- thanks!

PR #1721, opened 2024-11-07
make sure, equivalentValues(Object) is used if available

rbri:
The HTMLAllCollection implements some strange behavior - https://developer.mozilla.org/en-US/docs/Web/API/HTMLAllCollection#special_type_conversion_behavior. Of course we have to support this in HtmlUnit - see https://github.com/HtmlUnit/htmlunit/issues/896

Luckily Rhino supports this kind of customized equals behavior by overriding the ScriptableObject#equivalentValues() method.

But there are two bugs
* the equivalentValues() method is called if the lhs is null or undefined but NOT if the rhs is null or undefined (interpreter).
* the optimized version works correct if the lhs is undefined or rhs is undefined but does not call the equivalentValues() method in case lhs/rhs is null

This PR fixes both cases and also adds test cases.

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/ScriptRuntime.java
+++ b/rhino/src/main/java/org/mozilla/javascript/ScriptRuntime.java
@@ -3737,7 +3737,15 @@ public class ScriptRuntime {
             if (isSymbol(y) && isObject(x)) {
                 return eq(toPrimitive(x), y);
             }
-            if (y instanceof Scriptable) {
+            if (y == null || Undefined.isUndefined(y)) {
+                if (x instanceof ScriptableObject) {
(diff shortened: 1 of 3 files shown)


rbri:
any feedback for this? I need it for HtmlUnit and i like to know if there is a chance to have it in rhino or if i have to put it in my fork.

I did already some test and i did not see any negative performance impact.

gbrail:
This makes sense to me -- I've been travelling a bit and haven't had a chance to test things myself yet. I'm also optimistic that a few of the optimizations like the ones in the bytecode here that you had to undo might be less necessary as we can keep working on dynamic linking -- I'll have some PRs for that coming up in the future. For now I'll check this out in detail later in the week.

gbrail:
Looks good to me -- thanks!

Issue #1720, opened 2024-11-07
Proposal: Debug and experiment flags


gbrail:
We're a bit stuck on Reflect and Proxy support because it isn't feature-complete, and we want to have some sort of "experiment" capability so that we can enable it while we complete the implementation, rather than relying on a huge feature branch. However, our current configuration mechanism, which is the feature flags on the Context class, is a bit cumbersome, especially for enabling something that might be temporary in nature, like an experimental feature that will eventually "graduate."

As a secondary problem, we have a bunch of hard-coded static boolean "debug" flags in the product, which are currently only enabled by changing the code. Rhino doesn't need a complicated "debug framework" like an enterprise Java app, but it'd be nice to be able to enable and disable this debugging more easily.

So, I'm proposing that we add a few system properties to Rhino, and also an environment variable:

<table>
<tr><td>Variable</td><td>Type</td><td>Meaning</td></tr>
<tr><td>rhino.experiments</td><td>System property</td><td>Enables experimental features. Value is a comma-separated list of feature names.</td></tr>
<tr><td>rhino.debug</td><td>System property</td><td>Enables debugging. Value is a comma-separated list of things to debug.</td></tr>
<tr><td>RHINO_DEBUG</td><td>Environment variable</td><td>Same as "rhino.debug". Values will be combined and the "or" of both values will be used to enable debugging.</td></tr>
</table>

All of these properties will work by scanning the comma-separated list of either experiments to enable, or debug flags to enable, and if the value is in the list, it'll enable that experiment or debug flags. Unknown entries in the list will be ignored.

The idea here, at least for experiments, is that when an experiment runs its course and can be "graduated," we will just retire the flag.

So, for example, setting:

    rhino.experiments=proxy,foobar

Enables the experiments named "proxy" and "foobar". If there is no feature named "foobar," then that part is just ignored, and the "proxy" feature is enabled.

Finally, I think that we should also support setting debug flags via environment variable, because depending on how we're running Rhino, it may be easier to set an environment variable than a system property.

rPraml:
I would suggest, that we use different strategies, where to load properties:

Let's create a `RhinoConfiguration` class, that tries to load properties from different places:
- `rhino.config` from classpath
- `rhino.config` from current directory
- read `System.getProperties()` and take everything starting with "rhino."
- maybe read alos environment.

Why do I want this: (especially reading config from classpath)

If you have a web application, you bundle your rhino.jar typically in your war file.
[...]

PR #1722, opened 2024-11-08
Reading configuration from config files

rPraml:
This could be a first draft for #1720

### DIFF
--- a/rhino/rhino-config-for-this-module.config
+++ b/rhino/rhino-config-for-this-module.config
@@ -1,0 +1,4 @@
+# This file is loaded by the RhinoPropertiesTest and can be used to modify values in this module
+#
+# rhino.printTrees=true
+# rhino.printICode=true
--- a/rhino/rhino-config-for-this-module.config
+++ b/rhino/rhino-config-for-this-module.config
@@ -1,4 +1,0 @@
-# This file is loaded by the RhinoPropertiesTest and can be used to modify values in this module
-#
-# rhino.printTrees=true
-# rhino.printICode=true
--- a/rhino/rhino-test.config
+++ b/rhino/rhino-test.config
@@ -1,7 +1,12 @@
-# THIS FILE IS NOT USED BY DEFAULT
-# There is a RhinoPropertiesLoader in this maven module
-# If you want to reconfigure something, please use rhino-config-for-this-module.config (see RhinoPropertiesTest)
+# This file can be used to enable certain options in this module for testcases.
+#
+# for example, if there is an (experimental) feature xyz, you can set
+# rhino.xyz.enabled=true
+# and write tests for features, whiche are not enabled by default for production builds
+#
+#
+# Please keep this values for RhinoPropertiesTest
+test.file.rhino-test-config.loaded=true
+test.config.bar=value4-mod
 
-testconfig.bar=value4-mod
-TESTCONFIG_BAZ=value5
 
--- a/rhino/rhino-test.config
+++ b/rhino/rhino-test.config
@@ -1,12 +1,19 @@
-# This file can be used to enable certain options in this module for testcases.
+# This file is used to configure tests in the "rhino" module.
 #
-# for example, if there is an (experimental) feature xyz, you can set
-# rhino.xyz.enabled=true
-# and write tests for features, whiche are not enabled by default for production builds
+# Normally, rhino should run without "rhino.config" and "rhino-test.config"
+# in default configuration.
+#
+# You cen set here some options for testing purposes, for example, if you
+# add a new (exprermental) feature "xyz", that shouldn't be enabled in
+# production yet, you can add a flag in your code
+#
+#    private static final boolean XYZ = RhinoConfig.get("rhino.xyz", false)
+#
+# and then set "rhino.xyz=true" in this file to enable it for tests.
 #
 #
-# Please keep this values for RhinoPropertiesTest
+#
+# Please keep this values for RhinoPropertiesTest:
+
 test.file.rhino-test-config.loaded=true
 test.config.bar=value4-mod
-
-
--- a/rhino/rhino-test.config
+++ b/rhino/rhino-test.config
@@ -1,0 +1,7 @@
+# THIS FILE IS NOT USED BY DEFAULT
+# There is a RhinoPropertiesLoader in this maven module
+# If you want to reconfigure something, please use rhino-config-for-this-module.config (see RhinoPropertiesTest)
+
+testconfig.bar=value4-mod
+TESTCONFIG_BAZ=value5
+
--- a/rhino/src/main/java/org/mozilla/javascript/config/RhinoConfig.java
+++ b/rhino/src/main/java/org/mozilla/javascript/config/RhinoConfig.java
@@ -8,6 +8,13 @@ import java.util.Objects;
  * RhinoConfig provides typesafe and static access methods to a {@link RhinoProperties} default
  * instance.
  *
+ * <p>Note: Reading a config value only once is a good practice to avoid affecting performance. For
+ * example:
+ *
+ * <pre>
+ *     private static final boolean XYZ_ENABLED = RhinoConfig.get("rhino.xyz.enabled", false);
+ * </pre>
+ *
  * @author Roland Praml, Foconis Analytics GmbH
  */
 public class RhinoConfig {
--- a/rhino/src/main/java/org/mozilla/javascript/config/RhinoConfig.java
+++ b/rhino/src/main/java/org/mozilla/javascript/config/RhinoConfig.java
@@ -1,0 +1,101 @@
+package org.mozilla.javascript.config;
+
+import java.security.AccessController;
+import java.security.PrivilegedAction;
+import java.util.Objects;
+
+/**
+ * RhinoConfig provides typesafe and static access methods to a {@link RhinoProperties} default
+ * instance.
+ *
+ * @author Roland Praml, Foconis Analytics GmbH
+ */
+public class RhinoConfig {
+
+    private static final RhinoProperties INSTANCE =
+            // We still assume, that a security manager could be in place!
+            AccessController.doPrivileged(
+                    (PrivilegedAction<RhinoProperties>) RhinoProperties::init);
+
+    /**
+     * Returns the property as string.
+     *
+     * <p>If the value is not set, <code>defaultVaule</code> is returned.
+     */
+    private static String get(String property, String defaultValue) {
+        Object ret = INSTANCE.get(property);
+        if (ret != null) {
+            return ret.toString();
+        }
+        return defaultValue;
+    }
+
+    /** Returns the property as string with null as default. */
+    public static String get(String property) {
+        return get(property, (String) null);
+    }
+
+    /**
+     * Returns the property as enum.
+     *
+     * <p>If the property is set to any of the enum names (case-insensitive), this enum value is
+     * returned, otherwise <code>defaultValue</code> is returned.
+     *
+     * <p>Note: <code>defaultValue</code> must be specified to derive the enum class and its values.
+     */
+    @SuppressWarnings("unchecked")
+    public static <T extends Enum<T>> T get(String property, T defaultValue) {
+        Objects.requireNonNull(defaultValue, "defaultValue must not be null");
+        Object ret = INSTANCE.get(property);
+        if (ret != null) {
+            Class<T> enumType = (Class<T>) defaultValue.getDeclaringClass();
+            // We make a case insentive lookup here.
+            for (T enm : enumType.getEnumConstants()) {
+                if (enm.name().equalsIgnoreCase(ret.toString())) {
+                    return enm;
+                }
+            }
+        }
+        return defaultValue;
+    }
+
+    /**
+     * Returns the property as boolean.
+     *
+     * <p>A property is true, if it is either <code>Boolean.TRUE</code> or if and only if the string
+     * representation is equal to {@code "true"} (case-insensitive). If the property is not set,
+     * <code>defaultValue</code> is returned
+     *
+     * <p>Same behaviour as {@link Boolean#getBoolean(String)}
+     */
+    public static boolean get(String property, boolean defaultValue) {
+        Object ret = INSTANCE.get(property);
+        if (ret instanceof Boolean) {
+            return (Boolean) ret;
+        } else if (ret == null) {
+            return defaultValue;
+        }
+
+        return Boolean.parseBoolean(ret.toString());
+    }
+
+    /**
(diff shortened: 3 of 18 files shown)


gbrail:
I think that this approach makes sense in that it lets configuration come from the environment, system properties, file, or classpath. I think that's a good thing and that we should follow this pattern.

For actually checking the setting of various flags, however, I think that the config class should parse all this input and set boolean (or enum) values that can be checked directly, rather than relying on a hash lookup. I think that can give us two advantages:

1) It keeps all the logic about naming and renaming things in one place
2) It will be faster, since many of these properties, like the various debug flags, will be checked many millions of times potentially while Rhino runs, and all those hash table lookups will dominate performance pretty soon.

gbrail:
If you're looking for the kinds of debug flags I'd like to potentially replace with a real debugging mechanism, I would look at these:

https://github.com/mozilla/rhino/blob/3e899f6420dc6260380d0a688016207f64f9cfec/rhino/src/main/java/org/mozilla/javascript/Token.java#L27

https://github.com/mozilla/rhino/blob/3e899f6420dc6260380d0a688016207f64f9cfec/rhino/src/main/java/org/mozilla/classfile/ClassFileWriter.java#L4476

https://github.com/mozilla/rhino/blob/3e899f6420dc6260380d0a688016207f64f9cfec/rhino/src/main/java/org/mozilla/javascript/optimizer/DefaultLinker.java#L25

And as for feature flags, the first one IMO should be for the reflect and proxy support that's currently languishing in a PR by @rbri

rPraml:
Unfortunately, I didn't have as much time as I thought, but I wanted to update the draft before Christmas of how I want to do it now.

The idea is, to check all sources (classpath, configfile, system-properties, env - the last one wins) for the settings and parse them immediately in the correct datatype. When a variable is bound to the property `rhino.debugLinker` for example. I try to check also the upper-camel `RHINO_DEBUG_LINKER` version.

@gbrail I hope, this goes in the right direction... Unfortunately (or fortunately) I won't be able to continue until next year due to vacation. :)

rPraml:
So, I found some time to update this PR.

*What do we have now*

The `RhinoProperties` class is a container of serveral config maps, that are loaded from different location: 
- looks for `rhino.config` & `rhino-test.config` in several locations. (CHECKME: do we need rhino-test.config - similar to logback-test.xml?)
- add system-env & system-properties
- or optional: Implement `RhinoConfigLoader` service with your own loading strategy

When checking for a value, `RhinoProperties` tries all configs in loaded order in exact spelling and camel case-conversion. (e.g `rhino.printICode` and `RHINO_PRINT_ICODE` - you can use both writings in any config source. (CHECKME: This camel case conversion is required for system-env. Do we want it for other config sources, too? It theoretically slows down the parse time, as several permutations are checked, but every config is only parsed once, if assigned to a static field, so this should not have a big impact)

There is a helper class `RhinoConfig` which should be used in most cases. It adds some helper methods to get properties as String, Integer or Enum.

> For actually checking the setting of various flags, however, I think that the config class should parse all this input and set boolean (or enum) values that can be checked directly, rather than relying on a hash lookup. I think that can give us two advantages:
> 
> 1. It keeps all the logic about naming and renaming things in one place
> 
> 2. It will be faster, since many of these properties, like the various debug flags, will be checked many millions of times potentially while Rhino runs, and all those hash table lookups will dominate performance pretty soon.

I tried that, but did not like the idea to add configs from different packages in one class:

1. This is not optimal, if we want to split rhino in more modules. I can imagine to have one config class per module or package (please give feedback)
2. This should be no problem in the current PR. Each config is assigned to a static constant, so it is read only one time.


Other findings, I identified:

- I tried to attach the current config to the current context (so to have one context with `PRINT_ICODE` and an other without), but this would require to pass the context/config object down to Token etc. So I think, we have to accept, that we have only ONE config per classloader and every rhino context uses the same settings. It is important in the future, to decide, if some config values should be global or at context level.
- I read all properties (also system-properties & env). So someone that have access to this class will effectively bypass the (deprecated) security-manager

it would be good, if I get some feedback, if this goes in a way we want.

@gbrail FYI

rhino/src/main/java/org/mozilla/javascript/config/RhinoConfig.java
```diff
@@ -0,0 +1,60 @@
...
+        return get(property, (String) null);
+    }
+
+    /** Returns the property as enum. Note: default value must be specified */
+    public static <T extends Enum<T>> T get(String property, T defaultValue) {
+        Object ret = RhinoProperties.get(property);
+        if (ret != null) {
+            Class<T> enumType = (Class<T>) defaultValue.getClass();
+            // We assume, that enums all are in UPPERCASES
+            return Enum.valueOf(enumType, ret.toString().toUpperCase(Locale.ROOT));
```
gbrail, line 32: Should we catch "IllegalArgumentException" in case the input value is not valid and return the default in that case?

rhino/src/main/java/org/mozilla/javascript/config/RhinoConfig.java
```diff
@@ -0,0 +1,60 @@
...
+        return defaultValue;
+    }
+
+    /** Returns the property as boolean. */
+    public static boolean get(String property, boolean defaultValue) {
+        Object ret = RhinoProperties.get(property);
+        if (ret instanceof Boolean) {
+            return (Boolean) ret;
+        } else {
+            return "1".equals(ret) || "true".equals(ret);
```
gbrail, line 43: I wonder if you'd consider numeric values other than 1 as "true," and doing a case-insensitive comparison to "true"

rhino/src/main/java/org/mozilla/javascript/config/RhinoProperties.java
```diff
@@ -0,0 +1,163 @@
...
+    public void loadFromFile(File config) {
+        if (!config.exists()) {
+            return;
+        }
+        try (InputStream in = new FileInputStream(config)) {
+            Properties props = new Properties();
+            props.load(new InputStreamReader(in, StandardCharsets.UTF_8));
+            addConfig(props);
+        } catch (IOException e) {
+            System.err.println(
```
gbrail, line 98: I don't love that we "println" anywhere, but I don't see that we have a choice in this case. But if it happens, it'll be buried in someone's big log file and they won't know what it means -- should we consider adding "Rhino:" or something to identify this message so that people can know why it's appearing?

rhino/src/main/java/org/mozilla/javascript/config/RhinoProperties.java
```diff
@@ -0,0 +1,163 @@
...
+    public void loadFromResource(URL resource) {
+        if (resource == null) {
+            return;
+        }
+        try (InputStream in = resource.openStream()) {
+            Properties props = new Properties();
+            props.load(new InputStreamReader(in, StandardCharsets.UTF_8));
+            addConfig(props);
+        } catch (IOException e) {
+            System.err.println(
```
gbrail, line 123: Same here -- we should add "Rhino" or something to this.

gbrail (commented):
I think this is good progress, and minus the attached minor comments, I think that this is the right implementation.

I also want to know how we can test this -- there are a lot of options with properties files, system properties, environment variables, and a service loader.

I really think that we should go forward with this, but only if we can put in some way to exercise more of this codebase in our test suite.

rhino/src/test/java/org/mozilla/javascript/tests/RhinoPropertiesTest.java
```diff
@@ -0,0 +1,145 @@
...
+import org.mozilla.javascript.config.RhinoPropertiesLoader;
+
+/**
+ * Testcase for various property loading mechanism. <br>
+ * Note: Testing is a bit difficult, so to test the serviceLoader mechanism, we have to register a
+ * RhinoPropertiesLoader in this maven module. Unfortunately, this means, that all other tests in
+ * this module will also use this loader. <br>
+ * The loader does NOT load the properties from the default locations as mentioned in the
+ * documentation. So, do not wonder, if setting config values in 'rhino.config' or
+ * 'rhino-test.config' do not take effect. Instead use `rhino-config-for-this-module.config`.
```
rPraml, line 20: Testing is a bit difficult, so I tried to cover all aspects in this tests

tests/build.gradle
```diff
@@ -60,6 +60,13 @@ test {
         systemProperty 'updateTest262properties', System.getProperty('updateTest262properties')
     }
 
+    System.properties.each { k,v->
+        if (k.startsWith("rhino.")) {
+            systemProperty k, v
+        }
+    }
```
rPraml, line 67: I copy all rhino system properties to the sub process. So you can pass properties with `-Drhino.xyz` from a gradle buld

rPraml (commented):
From my perspective, this is ready for merge.

I would like to add some documentation to https://rhino.github.io
Can anyone tell me, how? (i did not find the repo/instruction how to contribute)

@gbrail FYI

rhino/rhino-config-for-this-module.config
gbrail, line 1: I think this PR is looking very good, but why do we have this file and the other .config files in the top-level directory? I suppose that we need one just to test the file reading part of the codebase, but since they won't be used by the runtime or tests otherwise, do we really need all of them? That's really my only concern with this PR at this point. Thanks!

rPraml: Hello @gbrail 

> I also want to know how we can test this -- there are a lot of options with properties files, system properties, environment variables, and a service loader.

This is the problem, why I need so much test files. Let me explain for what they are.

*rhino/rhino-config-for-this-module.config*
The file rhino-config-for-this-module.config is used internally by RhinoPropertiesTest.Loader and is the only config-file loaded in this module for tests.
This file, along with rhino/rhino-test.config, serves as a placeholder for where properties should be placed.
For example, if there is a new experimental feature 'xyz' that can be enabled with rhino.experimental.xyz=true, it can be enabled in rhino-config-for-this-module.config or rhino-test.config, so that is available for test cases.
As these files are not included in the build jars, the feature stays disabled, unless someone enables it.

*rhino/rhino-test.config*
*rhino/src/test/resources/rhino-testcase.config*
*rhino/src/test/resources/rhino.config*
These files are used for various tests

*tests/rhino-test.config*
This file is mainly for documentation/example (there are only comments) and can be used to enable features for testing.

I see the problem, that this is now a bit confusing, but the files are required for the (complex) tests.

I see the following options:
- keep everything as it is (this will confuse people)
- create one (or more) `rhino-config-tests` module, to test various config ways (with/without ServiceLoader)
- remove/rewrite the complex test cases (especially the one with serviceLoader), which means, that the one or other properties-file will be obsolete.


I will prefer option 3 ;)

rhino/rhino-test.config
```diff
@@ -0,0 +1,12 @@
+# This file can be used to enable certain options in this module for testcases.
+#
+# for example, if there is an (experimental) feature xyz, you can set
+# rhino.xyz.enabled=true
+# and write tests for features, whiche are not enabled by default for production builds
+#
+#
+# Please keep this values for RhinoPropertiesTest
+test.file.rhino-test-config.loaded=true
+test.config.bar=value4-mod
```
rPraml, line 10: This file is a placeholder/for documentation and is also used to demonstrate, that a top-level file can override values in a classpath file.
Should I delete it (and the related test)?

tests/rhino-test.config
```diff
@@ -0,0 +1,12 @@
...
+# for example, if there is an (experimental) feature xyz, you can set
+# rhino.xyz.enabled=true
+# and write tests for features, whiche are not enabled by default for production builds
+
+# Some available options with default values
+# rhino.stack.style=rhino
+# rhino.printTrees=false
+# rhino.printICode=false
+# rhino.debugLinker=false
+# rhino.use_java_policy_security=false
```
rPraml, line 12: This file is really only a placeholder/for documentation. Should I delete it?

rPraml (commented):
I've refactored the tests a bit. There are currently two top level `rhino-test.config` files, which are mainly used for documentation (and one test)

I can delete these files, if you want.
Should I add the documentation to https://github.com/rhino/rhino.github.io/blob/main/docs/_docs/configuration.md or is there a better/preferred place?

gbrail:
Sorry, lost track of where in the thread to comment. 

I think it's fine to leave "rhino.test-config" files in the "rhino" and "tests" modules. Could you please add a comment explaining why they're there, and say that it's an example of how to use these files, or that people could uncomment them in order to do further testing? Once you do that I think that we should merge this.

As for the docs, I think it'd be great if you updated the docs on "rhino.io". @p-bakker owns that repo and I'm sure he'd accept a PR.

rPraml:
Hello @gbrail I rebased this PR and made these changes: https://github.com/mozilla/rhino/pull/1722/commits/28277a4c67a0720b22dba6ec4b2f59d68dde5503

gbrail:
Do you have time to look at the two warnings that come from ErrorProne?

/home/runner/work/rhino/rhino/rhino/src/main/java/org/mozilla/javascript/config/RhinoProperties.java:173: warning: [AnnotateFormatMethod] This method uses a pair of parameters as a format string and its arguments, but the enclosing method wasn't annotated @FormatMethod. Doing so gives compile-time rather than run-time protection against malformed format strings.
    private void logDebug(String msg, Object... args) {
                 ^
    (see https://errorprone.info/bugpattern/AnnotateFormatMethod)
/home/runner/work/rhino/rhino/rhino/src/main/java/org/mozilla/javascript/config/RhinoProperties.java:180: warning: [AnnotateFormatMethod] This method uses a pair of parameters as a format string and its arguments, but the enclosing method wasn't annotated @FormatMethod. Doing so gives compile-time rather than run-time protection against malformed format strings.
    private void logError(String msg, Object... args) {
                 ^
    (see https://errorprone.info/bugpattern/AnnotateFormatMethod)

I can look at them myself later if you'd like, but I'm trying to keep the code clean of those warnings so I'd like to see if you have a chance to fix the underlying issue first. Thanks!

rPraml:
Hello @gbrail,

tried to fix this, but `com.google.errorprone.annotations.FormatMethod` is not on the classpath.
I see two options:
- add errorprone dependency (maybe in an extra PR?)
- add `@SuppressWarnings("AnnotateFormatMethod")` in the meantime.

> I can look at them myself later if you'd like, but I'm trying to keep the code clean of those warnings so I'd like to see if you have a chance to fix the underlying issue first. Thanks!

In this case I would accept the offer and leave the fix to you ;)

Roland

gbrail:
OK -- I'll dig deeper into ErrorProne later in the week and fix it, and
will merge what you already have soon. Thanks!

gbrail:
Thanks for your work on this and I'll look at the warning soon.

PR #1723, opened 2024-11-11
Allow usage of `require` in the debugger and use ES6

andreabergia:
Simple PR to allow using `require` in the debugger and to default to ES6 mode.


gbrail:
Seems OK to me, I'll get to testing things myself and merging them as the week goes on.

gbrail:
This looks good to me. Thanks!

Issue #1308, opened 2023-03-06
Support source maps


gausie:
Following up on [this old comment](https://github.com/mozilla/rhino/issues/260#issuecomment-763355565), I wanted to log an issue for supporting column numbers and thus support source maps in rhino.

p-bakker:
Could you please clarify exactly what you're requesting? 

The way your request is currently worded, it sounds like you're just requesting column number support, which, if added, would automatically cause Rhino to support sourcemaps.

If so, where do you expect column numbers to appear/be supported in Rhino?

If after adding column number support in the relevant places more is needed to (fully) support sourcemaps in Rhino, please elaborate or better yet open a separate case for that.

Any chance you're willing and able to take a stab at it and provide a PR?

p-bakker:
Could you please clarify exactly what you're requesting? 

The way your request is currently worded, it sounds like you're just requesting column number support, which, if added, would automatically cause Rhino to support sourcemaps.

If so, where do you expect column numbers to appear/be supported in Rhino?

If after adding column number support in the relevant places more is needed to (fully) support sourcemaps in Rhino, please elaborate or better yet open a separate case for that.

Any chance you're willing and able to take a stab at it and provide a PR?

p-bakker:
Assuming the idea is to 'just' translate line & column numbers on stacktraces, I could see an API that allows you to register a 'SourceMapper' and when registered, Rhino calls that to convert line and column numbers

The actual  'SourceMapper' implementation is then pluggable and could be for example (a wrapper around) https://github.com/atlassian/sourcemap

For completeness, this seems the specifications of sourcemaps: https://docs.google.com/document/u/0/d/1U1RGAehQwRypUTovF1KRlpiOFze0b-_2gc6fAH0KY0k/mobilebasic?pli=1

p-bakker:
Just FYI: Looks like there will be an effort to officially standardize sourcemaps: https://ecma-international.org/news/ecma-tc39-ecmascript-initiates-a-new-task-group-to-standardize-source-maps/

gausie:
I think getting line and column numbers attached to errors is unrelated to any new standards - that would only be if we wanted to respect the standard when reporting the error. Which, I suppose would also be nice.

Issue #973, opened 2021-06-30
Implement Debug Adapter Protocol


p-bakker:
In order to be able to use Rhino in modern develoment setups, I propose to implement the [Debug Adapter Protocol](https://microsoft.github.io/debug-adapter-protocol/), which is how [VS Code](https://code.visualstudio.com/api/extension-guides/debugger-extension) and [Wild Web Developer (Eclipse)](https://github.com/eclipse/wildwebdeveloper) offer an extendible debug targets

I wonder if anyone uses the Swing-based debugger that in part of Rhino and I know there's a implementation based on the [dbgp](https://xdebug.org/docs/dbgp) protocol in [DTLK](https://github.com/eclipse/dltk.javascript/tree/master/plugins/org.eclipse.dltk.javascript.rhino.dbgp), which is used by some, but all platforms nowadays seem to move towards DAP

p-bakker:
Within Java land, there's [lsp4e](https://projects.eclipse.org/projects/technology.lsp4e), to aid easier implementation of both DAP and Language Servers

p-bakker:
Curious to know if anyone uses any form of a debugger for debugging just JavaScript running in Rhino

c3-mgruchacz:
Hi @p-bakker, wondering if there's any update on implementing Debug Adapter Protocol for Rhino?

Thanks!

p-bakker:
While I wish I had time to work on this, I haven't. I created this case as part of establishing idea's for the Rhino roadmap.

If you're interested in contributing, I'm happy to share my ideas

tonygermano:
> I wonder if anyone uses the Swing-based debugger that in part of Rhino 
I do occasionally, but I wouldn't say I'm a heavy user. I have it configured as a build task in vscode so I can launch it from there.

c3-mgruchacz:
Yes, I'm interested, I would like to hear your ideas. I think the lsp4j project that you reference in post 2 is a good starting point, essentially implement the IDebugProtocolServer interface, mapping each method to the corresponding action in the rhino debugger class Dim.

p-bakker:
Great to hear!

Yeah, I think using lsp4j as a basis, so you can focus on just the wiring things up into Rhino would be the best approach.

As for the wiring up into Rhino, bottom-line, you'll need an implementation of org.mozilla.javascript.debug.Debugger that can be registered on a Context using .setDebugger(...)

[...]

PR #1724, opened 2024-11-11
Introduce column number in AST nodes

andreabergia:
This PR introduces column numbers in all AST nodes.
It also modifies the existing tests for the parser, and adds a lot more cases in a different class. I have gone through `Parser` measuring code coverage to ensure that the tests cover (almost) all the cases - the things I didn't really test are the non-standard things like xml, array comprehension, or let expression.

There is a breaking change: in case of a function call such as:

```js
a.
 b()
```

rhino will now assign line `1` to the expression, not line `2` (i.e. we assign the start of the expression, not the position of the parenthesis for the function call). This is for coherency with other parsers - I've checked what happen via https://astexplorer.net/ and it seems like the consensus is what I have implemented.

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/IRFactory.java
+++ b/rhino/src/main/java/org/mozilla/javascript/IRFactory.java
@@ -534,7 +534,9 @@ public final class IRFactory {
 
     private Node transformExprStmt(ExpressionStatement node) {
         Node expr = transform(node.getExpression());
-        return new Node(node.getType(), expr, node.getLineno());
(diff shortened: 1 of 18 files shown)


rbri:
:-D Execution failed for task ':tests:spotlessJavaCheck'. :-D

gbrail:
This is great progress and will help people in the future. Thanks for all the hard work here! I hope that we can next progress to getting column numbers in stack traces!

p-bakker:
@andreabergia just FMI: was this change driver by the desire to properly support sourcemaps? See #1308

p-bakker:
Also see #973

PR #1726, opened 2024-11-21
Console should always flush after writing a newline

andreabergia:
Not really sure this deserves a big description. 🙂 
We just want to ensure that any `console.log` actually ends up being printed immediately and not buffered.


andreabergia:
Ok, what you say makes sense.
Our use case was simple - we were debugging rhino (in the java debugger), and we couldn't see the stuff that was printed by our JS code using `console.log`.

I'm going to change this so that `console.log` actually flushes.

gbrail:
OK -- that makes sense then. Thanks!

Issue #1730, opened 2024-11-25
Cannot set property `flags` on regexp


andreabergia:
This is not allowed by Rhino, but it should be according to the spec:

```js
var regex = /a/;
Object.defineProperty(regex, 'flags', {
  value: 'muyi'
});
```

Instead rhino complains with `TypeError: Cannot change the value of attribute "flags" because writable is false.`

rbri:
Played a bit with this

    let regex = /a/;
    console.log(regex.flags);
    console.log(regex.ignoreCase);
    console.log(regex.test('A'));

    Object.defineProperty(regex, 'flags', {
        value: 'muyi'
    });
    console.log(regex.flags);
    console.log(regex.ignoreCase);
    console.log(regex.test('A'));

This produces

    > ""
    > false
    > false
    > "muyi"
    > false
    > false

rbri:
did some hacking on that, but it looks like there are many things to fix. Maybe we have to wait until we convert RegEx into lambda

Issue #1729, opened 2024-11-25
Cannot define an object literal with a getter for a symbol property


andreabergia:
This is not allowed by rhino, but it should be:

```js
obj = { get [Symbol.match]() {} }
```

The error printed is `TypeError: The object is not a string`, but that's confusing. The reason is that there is a code path where the symbol is casted to a string, which does not work.

p-bakker:
Think this is a dup of #1614

andreabergia:
@p-bakker you're right, it is!

Issue #959, opened 2021-06-25
Support ES2018 Regex Unicode Property Escapes


p-bakker:
Proposal: https://github.com/tc39/proposal-regexp-unicode-property-escapes
MDN: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions/Unicode_Property_Escapes

PR #1731, opened 2024-11-25
Implement `String.prototype.matchAll` and related

andreabergia:
This PR implements:
- [`Symbol.matchAll`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/matchAll)
- [`String.prototype.matchAll`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/matchAll)
- [`RegExp.prototype[Symbol.matchAll]`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/Symbol.matchAll)

All test262 cases pass; those that do not, fail for pre-existing reasons unrelated to these changes:

  - unicode flag 'u' not supported - looks similar to https://github.com/mozilla/rhino/issues/959 but I'm not 100% sure
    - RegExp/prototype/Symbol.matchAll/species-constructor.js
    - String/prototype/matchAll/flags-undefined-throws.js
  - flag property on regexp is not modifiable - tracked by https://github.com/mozilla/rhino/issues/1730
    - String/prototype/matchAll/flags-nonglobal-throws.js
    - RegExp/prototype/Symbol.matchAll/this-get-flags.js
    - RegExp/prototype/Symbol.matchAll/this-tostring-flags.js
    - RegExp/prototype/Symbol.matchAll/this-tostring-flags-throws.js
  - pre-existing problem with computed property on getter/setter and symbol keys - tracked by https://github.com/mozilla/rhino/issues/1729
    - Regexp/prototype/Symbol.matchAll/isregexp-this-throws.js
    - Regexp/prototype/Symbol.matchAll/isregexp-called-once.js
    - Regexp/prototype/Symbol.matchAll/species-constructor-get-species-throws.js
  - we autobox the `this` object, so we get a `NativeString` instead of the raw string literal and the last assertion fails
    - String/prototype/matchAll/regexp-matchAll-invocation.js
    - String/prototype/matchAll/regexp-prototype-matchAll-invocation.js
  - we autobox arguments to function calls, so `f.call(true)` receivedes a `NativeBoolean` which is an object
    - Regexp/prototype/Symbol.matchAll/this-not-object-throws.js

Please squash if accepted!

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/ScriptRuntime.java
+++ b/rhino/src/main/java/org/mozilla/javascript/ScriptRuntime.java
@@ -188,7 +188,7 @@ public class ScriptRuntime {
 
         NativeArrayIterator.init(scope, sealed);
         NativeStringIterator.init(scope, sealed);
-        getRegExpProxy(cx).register(scope, sealed);
+        registerRegExp(cx, scope, sealed);
 
         NativeJavaObject.init(scope, sealed);
         NativeJavaMap.init(scope, sealed);
@@ -296,6 +296,13 @@ public class ScriptRuntime {
         }
 
         return scope;
+    }
+
+    private static void registerRegExp(Context cx, ScriptableObject scope, boolean sealed) {
+        RegExpProxy regExpProxy = getRegExpProxy(cx);
+        if (regExpProxy != null) {
+            regExpProxy.register(scope, sealed);
+        }
     }
 
     public static ScriptableObject initStandardObjects(
--- a/rhino/src/main/java/org/mozilla/javascript/ScriptRuntime.java
+++ b/rhino/src/main/java/org/mozilla/javascript/ScriptRuntime.java
@@ -188,6 +188,7 @@ public class ScriptRuntime {
 
         NativeArrayIterator.init(scope, sealed);
         NativeStringIterator.init(scope, sealed);
+        getRegExpProxy(cx).register(scope, sealed);
 
         NativeJavaObject.init(scope, sealed);
         NativeJavaMap.init(scope, sealed);
@@ -196,8 +197,6 @@ public class ScriptRuntime {
                 cx.hasFeature(Context.FEATURE_E4X) && cx.getE4xImplementationFactory() != null;
 
         // define lazy-loaded properties using their class name
-        new LazilyLoadedCtor(
-                scope, "RegExp", "org.mozilla.javascript.regexp.NativeRegExp", sealed, true);
         new LazilyLoadedCtor(
                 scope, "Continuation", "org.mozilla.javascript.NativeContinuation", sealed, true);
 
@@ -1388,6 +1387,14 @@ public class ScriptRuntime {
         return (long) Math.min(len, NativeNumber.MAX_SAFE_INTEGER);
     }
 
+    public static long toLength(Object value) {
+        double len = toInteger(value);
+        if (len <= 0.0) {
+            return 0;
+        }
+        return (long) Math.min(len, NativeNumber.MAX_SAFE_INTEGER);
+    }
+
     /** See ECMA 9.5. */
     public static int toInt32(Object val) {
         // short circuit for common integer values
@@ -1442,6 +1449,20 @@ public class ScriptRuntime {
             return Optional.of(num);
         }
         return Optional.empty();
+    }
+
+    /** Implements the abstract operation AdvanceStringIndex. See ECMAScript spec 22.2.7.3 */
+    public static long advanceStringIndex(String string, long index, boolean unicode) {
+        if (index >= NativeNumber.MAX_SAFE_INTEGER) Kit.codeBug();
+        if (!unicode) {
+            return index + 1;
+        }
+        int length = string.length();
+        if (index + 1 > length) {
+            return index + 1;
+        }
+        int cp = string.codePointAt((int) index);
+        return index + Character.charCount(cp);
     }
 
     // XXX: this is until setDefaultNamespace will learn how to store NS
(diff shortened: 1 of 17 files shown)


rbri:
@andreabergia Great ❤️ 

the architecture tests are failing with jdk 21 - will have a look later if you like

andreabergia:
> @andreabergia Great ❤️
> 
> the architecture tests are failing with jdk 21 - will have a look later if you like

Noticed. There's two ways to make this pass:
1) move the code that registers the `NativeRegExpStringIterator` in the context into the `RegExpProxy`, which I have done in https://github.com/mozilla/rhino/pull/1731/commits/03303ebb1d36fb7ebbf68e66f45633a9fd9b5c34
2) modify the `ArchitectureTest` to allow the usage from `ScriptRuntime`.

I have implemented 1), but honestly I think 2) would be a better choice. I couldn't figure out the correct incantation for archunit though.

rbri:
> There's two ways to make this pass

Hi @andreabergia,

had a quick look and i like your commited solution more than changing the architecture test. For my understanding this RegExpProxy is a way to decouple the regexp impl from the rhino core and make it exchangeable. And therefore i think having this init method in the interface is correct.

And something like

    ...
    .doNotHaveFullyQualifiedName("org.mozilla.javascript.RegExpProxy")
    .or().doNotHaveFullyQualifiedName("org.mozilla.javascript.ScriptRuntime")

should work... (untested)

andreabergia:
> had a quick look and i like your commited solution more than changing the architecture test. For my understanding this RegExpProxy is a way to decouple the regexp impl from the rhino core and make it exchangeable.

Sounds good. Thanks for checking!

rhino/src/main/java/org/mozilla/javascript/ScriptRuntime.java
```diff
@@ -188,6 +188,7 @@ public static ScriptableObject initSafeStandardObjects(
 
         NativeArrayIterator.init(scope, sealed);
         NativeStringIterator.init(scope, sealed);
+        getRegExpProxy(cx).register(scope, sealed);
```
gbrail, line 191: I think that by doing this, we're changing the implementation so that it'll bomb if "RegExpImpl" isn't in the classpath, because "getRegExpProxy" will return null in that case. I'm not sure that we have ever taken advantage of that fact, but I think that with the current lazy constructor, a less catastrophic failure will result, and like all things Rhino, we don't really know what people are doing with the general capability to have the reg exp implementation be optional. Is there a different way to do this?

rhino/src/main/java/org/mozilla/javascript/regexp/NativeRegExpStringIterator.java
```diff
@@ -0,0 +1,111 @@
...
+        }
+        return NativeRegExp.js_exec(cx, scope, regexp, new Object[] {string});
+    }
+
+    @Override
+    protected String getTag() {
+        return ITERATOR_TAG;
+    }
+
+    private Scriptable regexp;
```
gbrail, line 105: FWIW, the Rhino creators in the Old Days seemed to like to put variables at the end of the class, but for new stuff I think it's OK (and preferable) if we act like modern-day Java developers and put them at the beginning.

rhino/src/main/java/org/mozilla/javascript/regexp/NativeRegExp.java
```diff
@@ -2761,7 +2768,7 @@ public Object execIdCall(
                 return realThis(thisObj, f).toString();
 
             case Id_exec:
-                return realThis(thisObj, f).execSub(cx, scope, args, MATCH);
```
gbrail, line 2764: I see that we're also changing stuff in IdScriptableObject that's been there a while, so I'm trying to follow this. What is changing here that's requiring us to treat the name of the function object differently?

gbrail (commented):
Thanks -- this is great progress.

I have a few specific comments, the big one being whether this is going to break the (admittedly theoretical) ability for the code to function if "RegExpImpl" isn't in the classpath.

rhino/src/main/java/org/mozilla/javascript/regexp/NativeRegExpStringIterator.java
andreabergia: Done

rhino/src/main/java/org/mozilla/javascript/ScriptRuntime.java
andreabergia: Changed that in https://github.com/mozilla/rhino/pull/1731/commits/314de9e3321b8f8c1a4c28312dc5f15c19bee845

rhino/src/main/java/org/mozilla/javascript/regexp/NativeRegExp.java
andreabergia: The reason is that, in `IdScriptableObject`, the method `ensureType` (used by `realThis`) took an `IdFunctionObject`. However, that parameter was only used to get the function name. So, I've refactored it and added an overload that only takes the function name.
This allows me to have a method `NativeRegExp::js_exec` that does not go through the `IdFunctionObject` at all, and so it can easily be called from `NativeRegExpStringIterator` to implement point 3 of the spec algorithm [RegExpExec](https://tc39.es/ecma262/multipage/text-processing.html#sec-regexpexec).

andreabergia:
Rebased onto the latest `master` branch and squashed a bit

rbri:
@gbrail - will be great if you can merge this soon - i plan a new HtmlUnit release during the next days and i like to have this in

gbrail:
OK, works for me, glad you all like it ;-)

PR #1734, opened 2024-11-27
Literal object creation rework

andreabergia:
This PR changes a bit the interpreter bytecode when creating a literal object - rhino now generates and uses a slightly more compact bytecode. Furthermore, there is a preliminary change for implementing `super` discussed below.

Before this change, this JS fragment:

```js
var o = {
  a: 1,
  [b]: 42,
}
```

would generate the following interpreter bytecode:

```
 [5] REG_IND_C0
 [6] LITERAL_KEYS [a, 136] true
 [8] REG_IND_C2
 [9] LITERAL_NEW
 [10] ONE
 [11] LITERAL_SET
 [12] REG_STR_C1 "b"
 [13] NAME
 [14] REG_IND_C1
 [15] LITERAL_KEY_SET
 [16] SHORTNUMBER 42
 [19] LITERAL_SET
 [20] OBJECTLIT
```

After this PR, it will generate the following:

```
 [5] REG_IND_C0
 [6] LITERAL_NEW_OBJECT [a, 142] true
 [8] ONE
 [9] LITERAL_SET
 [10] REG_STR_C1 "b"
 [11] NAME
 [12] LITERAL_KEY_SET
 [13] SHORTNUMBER 42
 [16] LITERAL_SET
 [17] OBJECTLIT
```

The changes are:
- `LITERAL_KEY_SET` uses the same indexing algorithm as `LITERAL_SET`, therefore does not require the `REG_IND_xxx` instruction before it
- renamed old `LITERAL_NEW` into `LITERAL_NEW_ARRAY`
- unified `LITERAL_KEYS` and the old behavior of `LITERAL_NEW` for objects into a new opcode called `LITERAL_NEW_OBJECT`
- modified `LITERAL_NEW_OBJECTS` so that it puts the object on the stack and then calls `ScriptRuntime.fillObjectLiteral`, instead of calling `ScriptRuntime.newObjectLiteral`. So, previously the stack would contain `[keys, flags, values]`; now it contains `[object, keys, flags, values]`. This is a preliminary step for implementing `super` (see https://github.com/mozilla/rhino/pull/1735).

Implemented with @0xe

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/CodeGenerator.java
+++ b/rhino/src/main/java/org/mozilla/javascript/CodeGenerator.java
@@ -1332,7 +1332,7 @@ class CodeGenerator extends Icode {
                 // Will be a node of type Token.COMPUTED_PROPERTY wrapping the actual expression
                 Node computedPropertyNode = (Node) propertyId;
                 visitExpression(computedPropertyNode.first, 0);
-                addIndexOp(Icode_LITERAL_KEY_SET, i);
+                addIcode(Icode_LITERAL_KEY_SET);
                 stackChange(-1);
             }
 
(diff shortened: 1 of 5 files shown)


gbrail:
Thanks, this one looks good!

PR #1736, opened 2024-11-27
Add support for `super` inside `eval`

andreabergia:
This PR, co-authored with @0xe , implements `super` support inside `eval`. It is stacked on top of https://github.com/mozilla/rhino/pull/1735.

The semantics come for "free", i.e. trying to resolve `super` when doing `eval` works automatically because it will do the same logic as a nested arrow function, i.e. look it up from the activation (the `NativeCall`).
So, the changes are really only related to the parser and need to allow `super` inside `eval`. We have implemented this by adding a new flag in `CompilerEnvirons` that is filled automatically during an `eval` call. We have chosen this approach because it allows `super` to be used in other places, such as a debugger's watch or immediate evaluation.


rhino/src/main/java/org/mozilla/javascript/Context.java
```diff
@@ -1367,6 +1368,16 @@ public final Script compileReader(
...
             Reader in, String sourceName, int lineno, Object securityDomain) throws IOException {
+        return compileReader(in, sourceName, lineno, securityDomain, null);
+    }
+
+    public Script compileReader(
+            Reader in,
+            String sourceName,
+            int lineno,
+            Object securityDomain,
+            Consumer<CompilerEnvirons> compilerEnvironsProcessor)
```
gbrail, line 1379: I don't think that we should be adding arguments to a public method that people definitely use outside this repo. It's probably fine to use a default argument, but I think it'd be safest to add a new method that has the new argument and keep the old one around for backward compatibility.

rhino/src/main/java/org/mozilla/javascript/Context.java
```diff
@@ -1414,7 +1426,8 @@ final Script compileString(
             ErrorReporter compilationErrorReporter,
             String sourceName,
             int lineno,
-            Object securityDomain) {
+            Object securityDomain,
+            Consumer<CompilerEnvirons> compilerEnvironsProcessor) {
```
gbrail, line 1430: Same here -- this is a public method in a public class that is definitely used outside this project and we need to keep it for compatibility.

gbrail (changes requested):
This makes sense but we have to be careful about changing the public interface to Rhino.

rhino/src/main/java/org/mozilla/javascript/Context.java
rbri: Usually we mark the old as deprecated and taking care the old is not used any longer by ourself.

andreabergia: We have done that! The old method is literally just above the line you commented on. We have also not deprecated it, because it felt it would annoy every single user to have to add a `, null` at the end or get a deprecation warning...

rhino/src/main/java/org/mozilla/javascript/Context.java
andreabergia: This is _not_ public, which is why we have not added the "backward compatibility" method.

gbrail (approved):

andreabergia:
Merged onto https://github.com/mozilla/rhino/pull/1735

PR #1737, opened 2024-11-27
Handle `super.__proto__`, which has a special implementation

andreabergia:
This PR, co-authored with @0xe, adds support for the crazy `super.__proto__`.

In Rhino there is a special implementation that uses `Ref` and `GET_REF` (and similar) for things like `a.__proto__`, `a.__parent__` and xml access. We have extended this to `super`.

A really surprising thing that we discovered looking at other engines is that the semantics of `super.__proto__` is a bit confusing - it ends up being `this.__proto__` basically, because `__proto__` is resolved on `super` but then accessed on `this`. So, the implementation uses a horrible hack: it replaces `super.__proto__` with `this.__proto__` in the IR. This seems to work, according to the tests that we have added, and while a bit dirty, it seems fine to us for such a strange edge case.


gbrail (approved):
This looks OK (but it's a weird language feature). Thanks!

PR #1738, opened 2024-11-27
Handle `super.x++` and similar

andreabergia:
This PR, co-authored with @0xe, adds support for `super.x++` and similar. It is stacked on top of https://github.com/mozilla/rhino/pull/1735.

Rhino has special opcodes (and similar in compiled classes) for `a.b++`, probably for historical reasons. Given the number of bytecodes already added by https://github.com/mozilla/rhino/pull/1735, we have decided to implement this one differently - we just replace `super.x++` with `super.x = super.x + 1` in the bytecode.


rhino/src/main/java/org/mozilla/javascript/optimizer/BodyCodegen.java
```diff
@@ -3728,6 +3734,84 @@ private void visitIncDec(Node node) {
...
+        // stack: [super, p]
+        if ((incrDecrMask & Node.POST_FLAG) != 0) {
+            // For postfix we want a copy of the old value on the stack
+            cfw.add(ByteCode.SWAP); // [p, super]
+            cfw.add(ByteCode.DUP2); // [p, super, p, super]
+            cfw.add(ByteCode.POP); // [p, super, p]
+        }
+
+        // Increment or decrement the value
+        addObjectToDouble(); // Unbox
```
gbrail, line 3777: I'm not sure that this is more efficient than leaving the object as it was, and adding a "MATH:ADD"  instruction, or a call to ScriptRuntime.subtract, instead.

gbrail (commented):
This is fine but I'm not sure that you need to unwrap and re-wrap that value, rather than using the methods we already have that can operate directly.

FWIW, I have considered adding "MATH:INCREMENT" and "MATH:DECREMENT" INDY instructions, and getting rid of all the "elemIncrDecr" and "propIncrDecr" stuff, which I think is overcomplicated and doesn't help performance. If I do that it would clean this up too.

rhino/src/main/java/org/mozilla/javascript/optimizer/BodyCodegen.java
andreabergia: We've copy&pasted the code from the pre-existing `visitIncDec`. If we want to change the approach, it would be better to do it in both places (so, IMHO in a different PR)

gbrail: I did some experimenting with increment stuff last week and it's both not simple to completely change this stuff, and not faster to do it the way that I suggested. So I think that the way you have it here is fine. Thanks!

gbrail (approved):

andreabergia:
Merged onto https://github.com/mozilla/rhino/pull/1735

PR #1739, opened 2024-11-27
Handle `delete super.x` by throwing an error

andreabergia:
This PR, co-authored with @0xe, adds support for `delete super.x`. It is stacked on top of https://github.com/mozilla/rhino/pull/1735.

The [ECMAScript spec](https://tc39.es/ecma262/#sec-delete-operator-runtime-semantics-evaluation) says that this is not allowed, but it has to be a _runtime_ error and not a syntax error. Therefore, we have generated a new bytecode instruction that will just do the throw at runtime.


gbrail (approved):
Why would someone delete a super property, indeed! JavaScript is full of these wonderful edge cases that we have to test for, so thanks. FWIW I think that JavaScript is about 95% edge cases.

andreabergia:
Merged this onto https://github.com/mozilla/rhino/pull/1735

PR #1740, opened 2024-11-27
Fixed cleanup in unit test `DynamicScopeTest`

andreabergia:
The test is creating a couple of context but not closing them properly.


gbrail:
Thanks!

Issue #645, opened 2020-01-23
NullPointerException on empty for loop


luisggpina:
Input:

```
for(({});;){  }
```

Error output:

```
Exception in thread "main" java.lang.NullPointerException
	at org.mozilla.javascript.optimizer.BodyCodegen.visitObjectLiteral(Codegen.java:3232)
	at org.mozilla.javascript.optimizer.BodyCodegen.generateExpression(Codegen.java:2351)
	at org.mozilla.javascript.optimizer.BodyCodegen.generateStatement(Codegen.java:2058)
	at org.mozilla.javascript.optimizer.BodyCodegen.generateStatement(Codegen.java:1878)
	at org.mozilla.javascript.optimizer.BodyCodegen.generateStatement(Codegen.java:1878)
	at org.mozilla.javascript.optimizer.BodyCodegen.generateBodyCode(Codegen.java:1300)
	at org.mozilla.javascript.optimizer.Codegen.generateCode(Codegen.java:319)
	at org.mozilla.javascript.optimizer.Codegen.compileToClassFile(Codegen.java:194)
	at org.mozilla.javascript.optimizer.Codegen.compile(Codegen.java:103)
	at org.mozilla.javascript.Context.compileImpl(Context.java:2530)
	at org.mozilla.javascript.Context.compileString(Context.java:1525)
	at org.mozilla.javascript.Context.compileString(Context.java:1514)
```

Affects: Version 1.7.13 and current master (3d3679db8e7a6ad8d6db600e2188c0e3dcda4a65)

rbri:
Have done a test cases for this in  #1744.

I think we can close this if the pr #1744 is merged.

rbri:
https://github.com/mozilla/rhino/pull/1744 is merged.

Thanks @luisggpina for bringing this up.

PR #1744, opened 2024-12-01
implement some test cases for issues #626 and #645

rbri:


gbrail:
These look good to me -- thanks!

PR #1746, opened 2024-12-02
SlotMap.compute can lead to deadlock in ThreadSafeSlotMapContainer

nabacg:
New SlotMap compute method introduced in [b7ece52](https://github.com/mozilla/rhino/commit/b7ece52d83a0d558b28b5962c63e8eea31f0ecd8) can lead to a Deadlock during re-definition of existing object properties when used in combination with `ThreadSafeSlotMapContainer`. 

The problem occurs, because `ThreadSafeSlotMapContainer` obtains and holds a [write lock](https://github.com/mozilla/rhino/blob/70509232ad0dc5b30db8cf31c259980e0f411d65/rhino/src/main/java/org/mozilla/javascript/ThreadSafeSlotMapContainer.java#L78) for the entire duration of method execution, including `SlotComputer` lambda logic and in certain conditions (details below) `ScriptableObject.defineOwnProperty` call to `getPropertyDescriptor` will attempt to obtain a read lock on the same slotMap instance from inside `SlotComputer` lambda body, as marked on below code snippet : 

```java
public abstract class ScriptableObject
... 
    protected void defineOwnProperty(
            Context cx, Object id, ScriptableObject desc, boolean checkValid) {

        Object key = null;
        int index = 0;
...
slotMap.compute(                            <<< WRITE LOCK AQUIRED 
        key,
        index,
        (k, ix, existing) -> {
            if (checkValid) {
                ScriptableObject current =
                        existing == null ? null : 
                               existing.getPropertyDescriptor(cx, this);  <<< READ LOCK  NEEDED
                checkPropertyChange(id, current, desc);
            }
... 
 });                                         <<< WRITE LOCK RELEASED  
...


```

Conditions necessary for this problem to occur: 
 - use of `ThreadSafeSlotMapContainer`, controlled by `Context.FEATURE_THREAD_SAFE_OBJECTS` which is not ON by default
 - property re-definition, since we only call `existing.getPropertyDescriptor` if existing != null. 
 - finally, the object in question (target of property definition) doesn't have the correct parent scope defined. This is likely to occur if it's created in Java, like in provided tests. Deadlock occurs when:
      - `existing.getPropertyDescriptor(cx, this)` is called during property redefinition, it will be used as the "global scope" when Property Descriptor object is initialised. 
      - This means that same `this` object is passed to `ScriptRuntime.getBuiltinPrototype` as scope
      - and eventually an attempt is made to [read a property](https://github.com/mozilla/rhino/blob/ed9af7ee9776accfe11a5c43737e46a61f65e407/rhino/src/main/java/org/mozilla/javascript/ScriptableObject.java#L2004) on `this` object
       - which requires read lock on the same instance of `ThreadSafeSlotMapContainer` that `slotMap.compute` is already locking.

This PR adds 2 test cases reproducing the same issue, both with `Context.FEATURE_THREAD_SAFE_OBJECTS` switched ON to force use of `ThreadSafeSlotMapContainer`:
 - new `DeadlockReproTest`, the most minimal repro we could come up with
 - new test case in PropertyTest, with identical code to existing `redefineGetterProperty`  test, showing that just changing SlotMapContainer causes the deadlock to occur.

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/AccessorSlot.java
+++ b/rhino/src/main/java/org/mozilla/javascript/AccessorSlot.java
@@ -113,10 +113,34 @@ public class AccessorSlot extends Slot {
         return getter.asGetterFunction(name, scope);
     }
 
+    @Override
+    boolean isSameGetterFunction(Object function) {
+        if (function == Scriptable.NOT_FOUND) {
+            return true;
+        } else if (getter == null) {
+            return ScriptRuntime.shallowEq(Undefined.instance, function);
+        } else {
+            return getter.isSameGetterFunction(function);
+        }
+    }
+
+    @Override
+    boolean isSameSetterFunction(Object function) {
+        if (function == Scriptable.NOT_FOUND) {
+            return true;
+        } else if (setter == null) {
+            return ScriptRuntime.shallowEq(Undefined.instance, function);
+        } else {
+            return setter.isSameSetterFunction(function);
+        }
+    }
+
     interface Getter {
         Object getValue(Scriptable start);
 
         Function asGetterFunction(final String name, final Scriptable scope);
+
+        boolean isSameGetterFunction(Object getter);
     }
 
     /** This is a Getter that delegates to a Java function via a MemberBox. */
@@ -138,6 +162,11 @@ public class AccessorSlot extends Slot {
         @Override
         public Function asGetterFunction(String name, Scriptable scope) {
             return member.asGetterFunction(name, scope);
+        }
+
+        @Override
+        public boolean isSameGetterFunction(Object function) {
+            return member.isSameGetterFunction(function);
         }
     }
 
@@ -164,12 +193,20 @@ public class AccessorSlot extends Slot {
         public Function asGetterFunction(String name, Scriptable scope) {
             return target instanceof Function ? (Function) target : null;
         }
+
+        @Override
+        public boolean isSameGetterFunction(Object function) {
+            return ScriptRuntime.shallowEq(
+                    target instanceof Function ? (Function) target : Undefined.instance, function);
+        }
     }
 
     interface Setter {
         boolean setValue(Object value, Scriptable owner, Scriptable start);
 
         Function asSetterFunction(final String name, final Scriptable scope);
+
+        boolean isSameSetterFunction(Object getter);
     }
 
     /** Invoke the setter on this slot via reflection using MemberBox. */
@@ -202,6 +239,11 @@ public class AccessorSlot extends Slot {
         public Function asSetterFunction(String name, Scriptable scope) {
             return member.asSetterFunction(name, scope);
         }
+
+        @Override
+        public boolean isSameSetterFunction(Object function) {
+            return member.isSameSetterFunction(function);
+        }
     }
 
     /**
@@ -227,6 +269,12 @@ public class AccessorSlot extends Slot {
         @Override
         public Function asSetterFunction(String name, Scriptable scope) {
             return target instanceof Function ? (Function) target : null;
+        }
+
+        @Override
+        public boolean isSameSetterFunction(Object function) {
+            return ScriptRuntime.shallowEq(
+                    target instanceof Function ? (Function) target : Undefined.instance, function);
         }
     }
 }
--- a/rhino/src/main/java/org/mozilla/javascript/AccessorSlot.java
+++ b/rhino/src/main/java/org/mozilla/javascript/AccessorSlot.java
@@ -117,22 +117,22 @@ public class AccessorSlot extends Slot {
     boolean isSameGetterFunction(Object function) {
         if (function == Scriptable.NOT_FOUND) {
             return true;
-        } else if (getter == null) {
-            return ScriptRuntime.shallowEq(Undefined.instance, function);
-        } else {
-            return getter.isSameGetterFunction(function);
         }
+        if (getter == null) {
+            return ScriptRuntime.shallowEq(Undefined.instance, function);
+        }
+        return getter.isSameGetterFunction(function);
     }
 
     @Override
     boolean isSameSetterFunction(Object function) {
         if (function == Scriptable.NOT_FOUND) {
             return true;
-        } else if (setter == null) {
-            return ScriptRuntime.shallowEq(Undefined.instance, function);
-        } else {
-            return setter.isSameSetterFunction(function);
         }
+        if (setter == null) {
+            return ScriptRuntime.shallowEq(Undefined.instance, function);
+        }
+        return setter.isSameSetterFunction(function);
     }
 
     interface Getter {
--- a/rhino/src/main/java/org/mozilla/javascript/IdScriptableObject.java
+++ b/rhino/src/main/java/org/mozilla/javascript/IdScriptableObject.java
@@ -922,15 +922,15 @@ public abstract class IdScriptableObject extends ScriptableObject implements IdF
         ScriptableObject desc = super.getOwnPropertyDescriptor(cx, id);
         if (desc == null) {
             if (id instanceof String) {
-                return getBuiltInDescriptor((String) id);
+                return getBuiltInDataDescriptor((String) id);
             }
 
             if (ScriptRuntime.isSymbol(id)) {
                 if (id instanceof SymbolKey) {
-                    return getBuiltInDescriptor((SymbolKey) id);
+                    return getBuiltInDataDescriptor((SymbolKey) id);
                 }
 
-                return getBuiltInDescriptor(((NativeSymbol) id).getKey());
+                return getBuiltInDataDescriptor(((NativeSymbol) id).getKey());
             }
         }
         return desc;
@@ -954,7 +954,7 @@ public abstract class IdScriptableObject extends ScriptableObject implements IdF
         return slot;
     }
 
-    private ScriptableObject getBuiltInDescriptor(String name) {
+    private ScriptableObject getBuiltInDataDescriptor(String name) {
         Scriptable scope = getParentScope();
         if (scope == null) {
             scope = this;
@@ -990,7 +990,7 @@ public abstract class IdScriptableObject extends ScriptableObject implements IdF
         return null;
     }
 
-    private ScriptableObject getBuiltInDescriptor(Symbol key) {
+    private ScriptableObject getBuiltInDataDescriptor(Symbol key) {
         Scriptable scope = getParentScope();
         if (scope == null) {
             scope = this;
--- a/rhino/src/main/java/org/mozilla/javascript/IdScriptableObject.java
+++ b/rhino/src/main/java/org/mozilla/javascript/IdScriptableObject.java
@@ -877,7 +877,12 @@ public abstract class IdScriptableObject extends ScriptableObject implements IdF
                             setInstanceIdValue(id, value);
                         }
                     }
-                    attr = applyDescriptorToAttributeBitset(attr, desc);
+                    attr =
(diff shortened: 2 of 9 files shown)


nabacg:
Closing to prevent CI deadlocks

p-bakker:
Cancelled

rbri:
@p-bakker sorry i guess it is too late now

gbrail:
This is a great find -- thanks, and I'm going to spend a bit of time understanding it since I introduced a deadlock!

This is the first time I've seen the new "var" keyword in a PR. (I know, it's not really a "keyword.") It was introduced in Java 10 so it's fair game for us to use AFAIK!

rbri:
> This is the first time I've seen the new "var" keyword in a PR.

Not correct, because core-js is still on java 8 i had to 'backport' already some var's after the merges. For me it will be fine to live without var - i think it makes our complicated code a bit more readable at most places.

rhino/src/main/java/org/mozilla/javascript/AccessorSlot.java
```diff
@@ -113,10 +113,34 @@ Function getGetterFunction(String name, Scriptable scope) {
         return getter.asGetterFunction(name, scope);
     }
 
+    @Override
+    boolean isSameGetterFunction(Object function) {
+        if (function == Scriptable.NOT_FOUND) {
+            return true;
+        } else if (getter == null) {
```
rbri, line 120: Maybe a matter of taste - but there is no else needed because of the return

rhino/src/main/java/org/mozilla/javascript/IdScriptableObject.java
```diff
@@ -936,48 +936,82 @@ protected ScriptableObject getOwnPropertyDescriptor(Context cx, Object id) {
...
+                    return getBuiltInSlot((SymbolKey) id);
+                }
 
+                return getBuiltInSlot(((NativeSymbol) id).getKey());
+            }
+        }
+        return slot;
+    }
+
+    private ScriptableObject getBuiltInDescriptor(String name) {
```
rbri, line 957: Maybe getBuiltInDataDescriptor is a better name

aardvark179:
> This is a great find -- thanks, and I'm going to spend a bit of time understanding it since I introduced a deadlock!
> 
> This is the first time I've seen the new "var" keyword in a PR. (I know, it's not really a "keyword.") It was introduced in Java 10 so it's fair game for us to use AFAIK!

There were already three `var`s in `ScriptableObject#ensureLambdaAccessorSlot` in lines I touched with this change, so I figured it was fine to use it, and since source version is set to 11.

I actually find `var` much easier when reading complex code as the type, or meaning, is usually clear from the variable, and it often reduces noise when refactoring, but I understand tastes differ, and I'll change it if this is a big issue.

rhino/src/main/java/org/mozilla/javascript/AccessorSlot.java
aardvark179: Changed.

rhino/src/main/java/org/mozilla/javascript/IdScriptableObject.java
aardvark179: Agreed, and changed.

nabacg:
> There were already three vars in ScriptableObject#ensureLambdaAccessorSlot in lines I touched with this change,

That's most likely me, introduced as part of #1577 , so it's fairly recent. I personally also find `var`  more readable, but happy to refactor those methods especially if it's causing problems during "backports"..

nabacg:
Doing some more testing on this, if I set `Context.FEATURE_THREAD_SAFE_OBJECTS=true` and make `ThreadSafeSlotMapContainer` by used by default, I still get deadlocks in 262 tests, even on this branch. 

Following suites are the problem:
  - `Array/prototype/forEach/15.4.4.18-7-b-6.js`
  -  `Array/prototype/forEach/15.4.4.18-7-c-i-19.js`
  -  `Array/prototype/forEach/15.4.4.18-2-12.js`
  
  They all have a variation of below `Object.defineProperty` call 
  
  ```js 
    Object.defineProperty(Object.prototype, "1", {  
      get: function() {  
        return 6.99;  
      },  
      configurable: true  
    });  
   ``` 
   
 which passes 
 - `Object.prototype` as 1 argument, target of property definition
 -  descriptor as 3 argument, a literal object, so it's prototype is `Object.prototype` ( the 1 argument)
 -  `ScriptableObject.defineOwnProperty` -> calls [applyDescriptorToAttributeBitset](https://github.com/nabacg/rhino/blob/0f4e8562ba651e10f042a8ec4761efb3e76063e5/rhino/src/main/java/org/mozilla/javascript/ScriptableObject.java#L1641) with `desc` (the 3rd argument above) from inside `.compute` lambda ( holds the write lock)
	 - `applyDescriptorToAttributeBitset` internally calls [desc.getProperty](https://github.com/nabacg/rhino/blob/0f4e8562ba651e10f042a8ec4761efb3e76063e5/rhino/src/main/java/org/mozilla/javascript/ScriptableObject.java#L1855) 
	         - desc will perform property lookup on itself, then it's prototype (1 argument above)
		    - prototype's slotMap  `.query` requires a read lock on the same slotMap instance
		    
We're working on fixing this too, I mostly wanted to share the new details.

nabacg:
Having said that, I do worry that we don't catch those `ThreadSafeSlotMapContainer` related issues earlier in the process, because it's fairly easy to introduce problems like this even with fairly innocent looking refactoring.   

Would running a CI build with  `Context.FEATURE_THREAD_SAFE_OBJECTS` switched ON in parallel to current runs something worth considering ?

nabacg:
with [d350e81](https://github.com/mozilla/rhino/pull/1746/commits/d350e81e5fb71ad83b5f19564d8aaa8e5e143b9e) (which funnily was mostly refactoring my own code from #1577 ), I can finally pass all tests with `Context.FEATURE_THREAD_SAFE_OBJECTS` switched ON by default, without deadlocking anything.

gbrail:
OK -- let me look at this again, and resolve three things:

First -- I agree -- "var" is fine! We moved to Java 11 minimum in this release so there should be no problems (and TBH I don't know how "var" shows up in the bytecode so it might matter even less).

Second -- we should update the docs for SlotMap to indicate that the lambda in "compute" must not lock the slot map.

Third -- there is another way to resolve this, because I assume that the "deadlock" you mention is a problem with the non-recursive nature of the lock. We could replace all that "stamped lock" stuff in ThreadSafeSlotMapContainer with simple "synchronized" blocks, which are recursive and which optimize out pretty well when there's no contention anyway. 

But if you're good with this change, I'll look at it and it's also fine.

rhino/src/main/java/org/mozilla/javascript/ScriptableObject.java
```diff
@@ -1729,30 +1744,47 @@ public void defineProperty(
...
                 0,
-                (id, index, existing) ->
-                        ensureLambdaAccessorSlot(
-                                cx, id, index, existing, getter, setter, attributes));
+                (id, index, existing) -> {
+                    if (existing != null) {
+                        // it's dangerous to use `this` as scope inside slotMap.compute.
+                        // It can cause deadlock when ThreadSafeSlotMapContainer is used
+
+                        return replaceExistingLambdaSlot(cx, name, existing, newSlot);
```
gbrail, line 1758: Just looking at code coverage, since you're changing some fundamental stuff here, and I see that line 1758 doesn't get covered. Do you think that you could devise a test case for that and add it to your new test?

rhino/src/main/java/org/mozilla/javascript/IdScriptableObject.java
```diff
@@ -922,62 +932,96 @@ protected ScriptableObject getOwnPropertyDescriptor(Context cx, Object id) {
...
-        int attr = EMPTY;
+    private Slot queryOrFakeSlot(Context cx, Object id) {
+        var slot = querySlot(cx, id);
+        if (slot == null) {
+            if (id instanceof String) {
+                return getBuiltInSlot((String) id);
+            }
 
+            if (ScriptRuntime.isSymbol(id)) {
+                if (id instanceof SymbolKey) {
```
gbrail, line 957: Also, looking at coverage, nothing from line 957 to 961 seems to get covered either.

rhino/src/main/java/org/mozilla/javascript/ScriptableObject.java
nabacg: Thanks for pointing this out, [474c3ba](https://github.com/mozilla/rhino/pull/1746/commits/474c3ba6ad89d2067dba48ec98d0b98b2e959a3d) should take care of it.

rhino/src/main/java/org/mozilla/javascript/IdScriptableObject.java
nabacg: This one is a little more convoluted: as of right now, I think it's impossible for this code path to be executed, thus it will be challenging to test it. This is because it's only called from  `IdScriptableObject::defineOwnProperty` ( [here](https://github.com/nabacg/rhino/blob/e20ecb059baac35e24541fb78daf5fd0aa31d312/rhino/src/main/java/org/mozilla/javascript/IdScriptableObject.java#L870) and [here](https://github.com/nabacg/rhino/blob/e20ecb059baac35e24541fb78daf5fd0aa31d312/rhino/src/main/java/org/mozilla/javascript/IdScriptableObject.java#L897)) and before the call, we've already [established](https://github.com/mozilla/rhino/blob/e20ecb059baac35e24541fb78daf5fd0aa31d312/rhino/src/main/java/org/mozilla/javascript/IdScriptableObject.java#L861) that `key` must be a CharSequence, so it can't a Symbol.  So maybe the solutions is to remove this branch?

Having said that, this doesn't feel right and perhaps it's an issue with current implementation of `IdScriptableObject::defineOwnProperty`, since it's only handling String (or CharSequence) keys and essentially delegating Symbol key'ed properties to it's base class (ScriptableObject).  It's very consistent with how `getOwnPropertyDescriptor` implementation, which does try to find built in descriptor using both Strings and Symbols ([here](https://github.com/mozilla/rhino/pull/1746/files#diff-4fda7f3359a263960b860746308e275c0ec0e9e78462481905066c97eb5a69a2L930) ). I believe this currently works mostly, because we always check base class for property descriptor first, but it's a little confusing why we need that symbol lookup.  

One could also imagine, someday in the future, someone creating a new built-in `NativeX` class extending `IdScriptableObject` that defines a Symbol keyed property, that's shouldn't be redefined  ( i.e. `configurable: false` ) and current implementation of `IdScriptableObject::defineOwnProperty` would not check it ( only Strings are handled) and allow base class implementation to override it. 

I appreciate it's a little theoretical, but wanted to share this and ask for opinions on whether we should address this? Also if addressing this deserves a separate PR or should I add it here?

gbrail: OK, thanks -- I suppose one way we could deal with this could be to assert, and in the future if someone wants to implement a subclass that needs it, we'd fix it -- although in many cases we are trying to get rid of IdScriptableObject as it doesn't yield the performance benefit it once did. If you're not comfortable with that, however, I'd also be OK leaving this as it is.

nabacg: I personally wouldn't mind leaving it as is for now, I don't see an obvious place to assert that wouldn't break existing functionality or be completely redundant.

gbrail: OK, then I'm satisfied, thanks!

PR #1747, opened 2024-12-04
Support generator methods in object literals, e.g. `o = {*g() {...}}`

andreabergia:
This makes quite a few test262 cases pass.

There's a few tests that were passing and apparently are broken now, but that is a false positive. The reason is that, _before_ this PR, Rhino would not support the syntax altogether and thus would complain with a SyntaxError, whereas now we accept the method definition, and also accept some syntax that we shouldn't have inside the method body. For example one of the test cases, `method-definition/generator-param-redecl-let.js` has this body:

```js
var obj = {
    *foo(a) {
        let a = 3;
    }
};
```

Previously rhino would have thrown an error because `*foo` was not supported. After this PR, the method declaration is allowed and the test is passing, but rhino should throw a SyntaxError on `let a = 3`. However, the same test exists for non-generator methods, named `method-definition/name-param-redecl-let.js`, which was already failing because of the same underlying reason.
I've checked that this is true for all the test262 cases where there is a regression.

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/IRFactory.java
+++ b/rhino/src/main/java/org/mozilla/javascript/IRFactory.java
@@ -33,6 +33,7 @@ import org.mozilla.javascript.ast.FunctionCall;
 import org.mozilla.javascript.ast.FunctionNode;
 import org.mozilla.javascript.ast.GeneratorExpression;
 import org.mozilla.javascript.ast.GeneratorExpressionLoop;
+import org.mozilla.javascript.ast.GeneratorMethodDefinition;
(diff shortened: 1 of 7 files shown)


rbri:
😎

gbrail:
This is also promising! I'm in "everyone have meetings this week so we can be on vacation" mode at work so it'll be a few days but I will look at and test all this!

gbrail (approved):
This looks good to me. Can you possibly rebase it? Thanks!

andreabergia:
Rebased

PR #1748, opened 2024-12-04
Ensure `next` and `orderedNext` are null for slots in `HashSlotMap`.

aardvark179:
# Slot `next` and `orderedNext` in `HashSlotMap`

When an embedded slot map is promoted to being a `HashSlotMap` the slots are placed in the new map, but there `next` and `orderedNext` fields are not nulled, and nothing in the `HashSlotMap` code touches those fields. This can lead to slots holding on to an unexpected set of resources via the `orderedNext` chain which was built by `EmbeddedSlotMap`.

This PR copies the slots rather than reusing them, and sets the `next` and `orderedNext` fields to null on the copies, and sizes the linked hash map to match the number of slots being inserted on promotion to avoid unnecessary internal reallocations during this process.

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/AccessorSlot.java
+++ b/rhino/src/main/java/org/mozilla/javascript/AccessorSlot.java
@@ -16,6 +16,17 @@ public class AccessorSlot extends Slot {
         super(oldSlot);
     }
 
+    @Override
+    AccessorSlot copySlot() {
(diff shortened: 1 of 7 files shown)


gbrail:
This is a great catch -- thanks for your thorough testing. I'll check it out and get to merging this soon.

gbrail:
Thanks for fixing this!

PR #1751, opened 2024-12-09
FEAT: Implement Symbol.hasInstance for Function.prototype

0xe:
Adds support for `Function.prototype[Symbol.hasInstance]`.

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/BaseFunction.java
+++ b/rhino/src/main/java/org/mozilla/javascript/BaseFunction.java
@@ -31,10 +31,7 @@ public class BaseFunction extends IdScriptableObject implements Function {
         if (cx.getLanguageVersion() >= Context.VERSION_ES6) {
             obj.setStandardPropertyAttributes(READONLY | DONTENUM);
         }
-        IdFunctionObject constructor = obj.exportAsJSClass(MAX_PROTOTYPE_ID, scope, sealed);
-        if (cx.getLanguageVersion() >= Context.VERSION_ES6) {
-            ScriptRuntimeES6.addSymbolHasInstance(cx, scope, constructor);
-        }
+        obj.exportAsJSClass(MAX_PROTOTYPE_ID, scope, sealed);
     }
 
     /**
@@ -271,7 +268,13 @@ public class BaseFunction extends IdScriptableObject implements Function {
     @Override
     protected void initPrototypeId(int id) {
         if (id == SymbolId_hasInstance) {
-            initPrototypeValue(id, SymbolKey.HAS_INSTANCE, makeHasInstance(), CONST | DONTENUM);
+            initPrototypeMethod(
+                    FUNCTION_TAG,
+                    id,
+                    SymbolKey.HAS_INSTANCE,
+                    String.valueOf(SymbolKey.HAS_INSTANCE),
+                    1,
+                    CONST | DONTENUM);
             return;
         }
 
@@ -321,52 +324,6 @@ public class BaseFunction extends IdScriptableObject implements Function {
             }
         }
         return false;
-    }
-
-    private Object makeHasInstance() {
-        Context cx = Context.getCurrentContext();
-        ScriptableObject obj = null;
-
-        if (cx != null) {
-            Scriptable scope = this.getParentScope();
-            obj =
-                    new LambdaFunction(
-                            scope,
-                            0,
-                            new Callable() {
-                                @Override
-                                public Object call(
-                                        Context cx,
-                                        Scriptable scope,
-                                        Scriptable thisObj,
-                                        Object[] args) {
-                                    if (thisObj != null
-                                            && args.length == 1
-                                            && args[0] instanceof Scriptable) {
-                                        Scriptable obj = (Scriptable) args[0];
-                                        Object protoProp = null;
-                                        if (thisObj instanceof BoundFunction)
-                                            protoProp =
-                                                    ((NativeFunction)
-                                                                    ((BoundFunction) thisObj)
-                                                                            .getTargetFunction())
-                                                            .getPrototypeProperty();
-                                        else
-                                            protoProp =
-                                                    ScriptableObject.getProperty(
-                                                            thisObj, "prototype");
-                                        if (protoProp instanceof IdScriptableObject) {
-                                            return ScriptRuntime.jsDelegatesTo(
-                                                    obj, (Scriptable) protoProp);
-                                        }
-                                        throw ScriptRuntime.typeErrorById(
-                                                "msg.instanceof.bad.prototype", getFunctionName());
-                                    }
-                                    return false; // NOT_FOUND, null etc.
-                                }
-                            });
-        }
-        return obj;
     }
 
     @Override
@@ -424,6 +381,23 @@ public class BaseFunction extends IdScriptableObject implements Function {
                     boundArgs = ScriptRuntime.emptyArgs;
                 }
                 return new BoundFunction(cx, scope, targetFunction, boundThis, boundArgs);
+
+            case SymbolId_hasInstance:
+                if (thisObj != null && args.length == 1 && args[0] instanceof Scriptable) {
+                    Scriptable obj = (Scriptable) args[0];
+                    Object protoProp = null;
+                    if (thisObj instanceof BoundFunction)
+                        protoProp =
+                                ((NativeFunction) ((BoundFunction) thisObj).getTargetFunction())
+                                        .getPrototypeProperty();
+                    else protoProp = ScriptableObject.getProperty(thisObj, "prototype");
+                    if (protoProp instanceof IdScriptableObject) {
+                        return ScriptRuntime.jsDelegatesTo(obj, (Scriptable) protoProp);
+                    }
+                    throw ScriptRuntime.typeErrorById(
+                            "msg.instanceof.bad.prototype", getFunctionName());
+                }
+                return false; // NOT_FOUND, null etc.
         }
         throw new IllegalArgumentException(String.valueOf(id));
     }
--- a/rhino/src/main/java/org/mozilla/javascript/BaseFunction.java
+++ b/rhino/src/main/java/org/mozilla/javascript/BaseFunction.java
@@ -31,7 +31,10 @@ public class BaseFunction extends IdScriptableObject implements Function {
         if (cx.getLanguageVersion() >= Context.VERSION_ES6) {
             obj.setStandardPropertyAttributes(READONLY | DONTENUM);
         }
-        obj.exportAsJSClass(MAX_PROTOTYPE_ID, scope, sealed);
+        IdFunctionObject constructor = obj.exportAsJSClass(MAX_PROTOTYPE_ID, scope, sealed);
+        if (cx.getLanguageVersion() >= Context.VERSION_ES6) {
+            ScriptRuntimeES6.addSymbolHasInstance(cx, scope, constructor);
+        }
     }
 
     /**
@@ -80,7 +83,7 @@ public class BaseFunction extends IdScriptableObject implements Function {
     /**
      * Gets the value returned by calling the typeof operator on this object.
      *
-     * @see org.mozilla.javascript.ScriptableObject#getTypeOf()
+     * @see ScriptableObject#getTypeOf()
      * @return "function" or "undefined" if {@link #avoidObjectDetection()} returns <code>true
      *     </code>
      */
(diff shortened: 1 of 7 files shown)


rhino/src/main/java/org/mozilla/javascript/BaseFunction.java
```diff
@@ -31,7 +31,8 @@ static void init(Context cx, Scriptable scope, boolean sealed) {
         if (cx.getLanguageVersion() >= Context.VERSION_ES6) {
             obj.setStandardPropertyAttributes(READONLY | DONTENUM);
         }
-        obj.exportAsJSClass(MAX_PROTOTYPE_ID, scope, sealed);
+        IdFunctionObject constructor = obj.exportAsJSClass(MAX_PROTOTYPE_ID, scope, sealed);
+        ScriptRuntimeES6.addSymbolHasInstance(cx, scope, constructor);
```
rbri, line 35: do we need a language version check here (cx.getLanguageVersion() >= Context.VERSION_ES6)

rhino/src/test/java/org/mozilla/javascript/FunctionPrototypeSymbolHasInstanceTest.java
```diff
@@ -0,0 +1,263 @@
...
+    public void testSymbolHasInstanceIsPresent() {
+        String script =
+                ""
+                        + "var f = {\n"
+                        + "   [Symbol.hasInstance](value) { "
+                        + "   }"
+                        + "};\n"
+                        + "var g = {};\n"
+                        + "`${f.hasOwnProperty(Symbol.hasInstance)}:${g.hasOwnProperty(Symbol.hasInstance)}`";
+        Utils.runWithAllOptimizationLevels(
```
rbri, line 19: maybe you can use assertWithAllOptimizationLevelsES6 here

rhino/src/main/java/org/mozilla/javascript/BaseFunction.java
```diff
@@ -265,6 +268,11 @@ protected void fillConstructorProperties(IdFunctionObject ctor) {
 
     @Override
     protected void initPrototypeId(int id) {
+        if (id == SymbolId_hasInstance) {
+            initPrototypeValue(id, SymbolKey.HAS_INSTANCE, makeHasInstance(), 0x0F);
```
rbri, line 272: please use the real attributes instead of 0x0F for readability

rhino/src/main/java/org/mozilla/javascript/BaseFunction.java
```diff
@@ -509,6 +564,26 @@ protected boolean hasPrototypeProperty() {
         return prototypeProperty != null || this instanceof NativeFunction;
     }
 
+    @Override
+    ScriptableObject buildDataDescriptorHelper(
+            int instanceIdInfo, Scriptable scope, Object value, int attr) {
+        if (instanceIdInfo == SymbolId_hasInstance) {
+            return buildDataDescriptor(scope, value, attr, SymbolKey.HAS_INSTANCE.toString(), 1);
+        } else {
```
rbri, line 572: else not requred if you return before

rhino/src/main/java/org/mozilla/javascript/BaseFunction.java
```diff
@@ -627,6 +702,12 @@ private Object jsConstructor(Context cx, Scriptable scope, Object[] args) {
         return cx.compileFunction(global, source, evaluator, reporter, sourceURI, 1, null);
     }
 
+    @Override
+    protected int findPrototypeId(Symbol k) {
+        if (SymbolKey.HAS_INSTANCE.equals(k)) return SymbolId_hasInstance;
+        else return 0;
```
rbri, line 708: else not required

rhino/src/main/java/org/mozilla/javascript/IdScriptableObject.java
```diff
@@ -936,6 +942,21 @@ protected ScriptableObject getOwnPropertyDescriptor(Context cx, Object id) {
         return desc;
     }
 
+    /**
+     * Overridden in the base class for different descriptors
+     *
+     * @return ScriptableObject
+     */
+    ScriptableObject buildDataDescriptorHelper(
```
rbri, line 950: please add @Override... same for the next and maybe you can also copy the comment

rhino/src/main/java/org/mozilla/javascript/ScriptableObject.java
```diff
@@ -135,21 +135,39 @@ public abstract class ScriptableObject
...
+        return buildDataDescriptor(scope, value, attributes, null, -1);
+    }
+
+    protected static ScriptableObject buildDataDescriptor(
+            Scriptable scope, Object value, int attributes, String name, int length) {
         ScriptableObject desc = new NativeObject();
         ScriptRuntime.setBuiltinProtoAndParent(desc, scope, TopLevel.Builtins.Object);
         desc.defineProperty("value", value, EMPTY);
-        desc.setCommonDescriptorProperties(attributes, true);
+        desc.setCommonDescriptorProperties(attributes, true, name, length);
```
rbri, line 146: is it a good idea to introduce this new function with two (at least for me magic parameters) including all the forwarder for the old calls if we need the additional functionality only at one place?

rhino/src/main/java/org/mozilla/javascript/BaseFunction.java
```diff
@@ -313,6 +321,53 @@ static boolean isApplyOrCall(IdFunctionObject f) {
...
+                                            protoProp =
+                                                    ScriptableObject.getProperty(
+                                                            thisObj, "prototype");
+                                        if (protoProp instanceof IdScriptableObject) {
+                                            return ScriptRuntime.jsDelegatesTo(
+                                                    obj, (Scriptable) protoProp);
+                                        }
+                                        throw ScriptRuntime.typeErrorById(
+                                                "msg.instanceof.bad.prototype", getFunctionName());
+                                    } else {
```
rbri, line 362: else not needed because of the throw before

rhino/src/main/java/org/mozilla/javascript/IdScriptableObject.java
0xe: Removed these as they don't actually solve the problem with name and length attributes.

rhino/src/main/java/org/mozilla/javascript/ScriptableObject.java
0xe: Removed these as they don't actually solve the problem with name and length attributes.

rbri:
@0xe thanks for taking care of my comments

rhino/src/main/java/org/mozilla/javascript/ScriptRuntimeES6.java
```diff
@@ -46,4 +46,15 @@ public static void addSymbolUnscopables(
         ScriptableObject.putProperty(unScopablesDescriptor, "writable", false);
         constructor.defineOwnProperty(cx, SymbolKey.UNSCOPABLES, unScopablesDescriptor, false);
     }
+
+    /** Registers the symbol <code>[Symbol.hasInstance]</code> on the given constructor function. */
+    public static void addSymbolHasInstance(
```
gbrail, line 51: I get what you're trying to do here, but I don't think that we should add a new function to a "ScriptRuntime" class that we're only going to use in one place. I'd suggest that instead of all this, you just call "defineProperty" on the "constructor" object back in BaseFunction and pass it the bitset (I'll put a comment there too).

rhino/src/main/java/org/mozilla/javascript/ScriptableObject.java
```diff
@@ -831,6 +831,12 @@ public boolean hasInstance(Scriptable instance) {
         // chasing. This will be overridden in NativeFunction and non-JS
         // objects.
 
+        Context cx = Context.getCurrentContext();
+        Object hasInstance = ScriptRuntime.getObjectElem(this, SymbolKey.HAS_INSTANCE, cx);
+        if (hasInstance instanceof Callable) {
+            return (boolean)
```
gbrail, line 837: I'm not sure that I know all the semantics of just casting to boolean here, but if I know anything about JavaScript, it's 90% edge cases.

So, since the function here could literally return anything, you should use "ScriptRuntime.toBoolean" here, which can cast any number of objects correctly to a "boolean".

rhino/src/main/java/org/mozilla/javascript/BaseFunction.java
```diff
@@ -31,7 +31,10 @@ static void init(Context cx, Scriptable scope, boolean sealed) {
         if (cx.getLanguageVersion() >= Context.VERSION_ES6) {
             obj.setStandardPropertyAttributes(READONLY | DONTENUM);
         }
-        obj.exportAsJSClass(MAX_PROTOTYPE_ID, scope, sealed);
+        IdFunctionObject constructor = obj.exportAsJSClass(MAX_PROTOTYPE_ID, scope, sealed);
+        if (cx.getLanguageVersion() >= Context.VERSION_ES6) {
+            ScriptRuntimeES6.addSymbolHasInstance(cx, scope, constructor);
```
gbrail, line 36: I was going to ask why you don't just call "defineProperty" here, but actually I'm wondering why we need this at all -- adding the new symbol case in the other places that you added it should be enough for this to work.

rhino/src/main/java/org/mozilla/javascript/BaseFunction.java
```diff
@@ -313,6 +323,52 @@ static boolean isApplyOrCall(IdFunctionObject f) {
         return false;
     }
 
+    private Object makeHasInstance() {
+        Context cx = Context.getCurrentContext();
+        ScriptableObject obj = null;
+
+        if (cx != null) {
+            Scriptable scope = this.getParentScope();
+            obj =
```
gbrail, line 332: It's great to use a Lambda function here, but since this class still inherits from "IdScriptableObject", you should be able to make it work by adding a case to the switch in the "exec" function for your new "SymbolId_hasInstance" case.

gbrail (changes requested):
Thanks for working on this!

Unless I'm missing something (and in JavaScript I'm usually missing something) there is an existing pattern to add a new symbol-keyed property to a native object -- see the implementation of "SymbolId_Iterator" in NativeString, for instance. This change seems to be doing some of that and some of something else. Can you please try to do it that way? Thanks!

0xe:
> Thanks for working on this!
> 
> Unless I'm missing something (and in JavaScript I'm usually missing something) there is an existing pattern to add a new symbol-keyed property to a native object -- see the implementation of "SymbolId_Iterator" in NativeString, for instance. This change seems to be doing some of that and some of something else. Can you please try to do it that way? Thanks!

Thank you. I will try to address these in the coming days.

rhino/src/test/java/org/mozilla/javascript/FunctionPrototypeSymbolHasInstanceTest.java
```diff
@@ -0,0 +1,172 @@
...
+    public void testThrowTypeErrorOnNonObjectIncludingSymbol() {
+        String script =
+                ""
+                        + "var f = function() {}; \n"
+                        + "f.prototype = Symbol(); \n"
+                        + "f[Symbol.hasInstance]({})";
+
+        Utils.runWithAllOptimizationLevels(
+                (cx) -> {
+                    cx.setLanguageVersion(Context.VERSION_ES6);
```
rbri, line 154: there are now some assertThrows.... methods in the Utils, maybe you can use one of these

rhino/src/main/java/org/mozilla/javascript/ScriptRuntimeES6.java
0xe: You were right, those were unnecessary. I've removed them.

rhino/src/main/java/org/mozilla/javascript/ScriptableObject.java
0xe: Fixed it.

rhino/src/main/java/org/mozilla/javascript/BaseFunction.java
0xe: You were right, those were unnecessary. I've removed them.

rhino/src/main/java/org/mozilla/javascript/BaseFunction.java
0xe: Yep, this is way simpler. Changed it to switch on execIdCall.

rhino/src/test/java/org/mozilla/javascript/FunctionPrototypeSymbolHasInstanceTest.java
0xe: Thanks! I've used `Utils.assertEcmaErrorES6()`

gbrail (approved):
Yes, this looks good to me now. Thanks!

PR #1752, opened 2024-12-09
Fixed computed property in object literal with negative integers

andreabergia:
There is a code path, in compiled mode only (not in interpreted!), where something like `o = { [-1]: xxx }` would not work correctly.

The problem is a combination of factors:
- in the interpreter, we would treat `-1` as a `double`, not an `int`;
- but there are various places in `ScriptRuntime` where integer keys are allowed _only_ for positive or zero numbers, per the spec - otherwise they need to be treated as strings.

We have modified the existing test for computed properties and fixed the code.


gbrail:
Yes, good catch, thank you!

PR #1754, opened 2024-12-11
Consolidate multiple optimization levels down to one

gbrail:
Instead of 10+ optimization, levels, reduce them to one -- "interpreted mode" and compiled mode.  This includes:

* Adding an "interpreted mode" flag to Context
* Deprecate "optimization level" flags in Context
* Add an "-int" flag to the shell to support the new level
* Keep old "-opt" shell flag, but take it out of documentation
* Reduce many tests that ran in three optimization levels to just run twice


rhino-engine/src/main/java/org/mozilla/javascript/engine/RhinoScriptEngine.java
```diff
@@ -287,12 +298,22 @@ private static int parseInteger(Object v) throws ScriptException {
...
             throw new ScriptException("Value must be a string or number");
         }
     }
 
+    private static boolean parseBoolean(Object v) throws ScriptException {
+        if (v instanceof String) {
+            return Boolean.parseBoolean((String) v);
+        } else if (v instanceof Boolean) {
+            return (Boolean) v;
+        } else {
```
rbri, line 312: else not needed because of the return

rhino-tools/src/main/java/org/mozilla/javascript/tools/shell/Main.java
```diff
@@ -551,7 +551,7 @@ static void processFileSecure(Context cx, Scriptable scope, String path, Object
         Object source = readFileOrUrl(path, !isClass);
 
         byte[] digest = getDigest(source);
-        String key = path + "_" + cx.getOptimizationLevel();
+        String key = path + "_" + (cx.isInterpretedMode() ? "int" : "comp");
```
rbri, line 554: 'int' might be confusing, is there any good reason for this shortening?

rhino-tools/src/main/resources/org/mozilla/javascript/tools/resources/Messages.properties
```diff
@@ -34,7 +34,7 @@ msg.shell.usage =\
     \    -w                 Enable warnings.\n\
     \    -version 100|110|120|130|140|150|160|170|180|200\n\
     \                       Set a specific language version.\n\
-    \    -opt [-1|0-9]      Set optimization level.\n\
+    \    -int               Run in interpreted mode.\n\
```
rbri, line 37: 'int' is confusing

rhino/src/main/java/org/mozilla/javascript/CompilerEnvirons.java
```diff
@@ -98,13 +98,23 @@ public void setXmlAvailable(boolean flag) {
         xmlAvailable = flag;
     }
 
+    @Deprecated
```
rbri, line 101: I really like the patter we are using in HtmlUnit; this is a great help if you later on have to decide to remove this stuff

     *
     * @deprecated as of version 4.5.0; use {@link #validateSelectors(SelectorList, DomNode)} instead
     */
    @Deprecated

rhino/src/test/java/org/mozilla/javascript/tests/Utils.java
```diff
@@ -106,21 +84,4 @@ public static void assertWithAllOptimizationLevelsES6(
...
-                    try {
-                        cx.evaluateString(scope, js, "test", 1, null);
-                        fail("EvaluatorException expected");
-                    } catch (EvaluatorException e) {
-                        assertEquals(expectedMessage, e.getMessage());
-                    }
-
-                    return null;
-                });
-    }
```
rbri, line 125: Why this is removed?

gbrail: This function, and "getTestOptLevels", is no longer used. There's also a duplicate-ish version of this class in the "tests" module where this might be used more.

rhino/src/main/java/org/mozilla/javascript/CompilerEnvirons.java
gbrail: Yes, I will go in and add this syntax, thanks!

rhino-tools/src/main/resources/org/mozilla/javascript/tools/resources/Messages.properties
gbrail: Yes, although don't you think that '-interpreted' is too long? And '-I' is too much like the "-I" argument in a C compiler.

What if we support both "-int" and "-interpreted" command-line flags?

rhino-tools/src/main/java/org/mozilla/javascript/tools/shell/Main.java
gbrail: Not really -- AFAIK this is just used as a cache key but there's no reason why not to make it longer. (Previously it was just an integer.)

rhino-tools/src/main/java/org/mozilla/javascript/tools/jsc/Main.java
```diff
@@ -80,8 +80,7 @@ public String[] processOptions(String args[]) {
                     continue;
                 }
                 if ((arg.equals("-opt") || arg.equals("-O")) && ++i < args.length) {
-                    int optLevel = Integer.parseInt(args[i]);
-                    compilerEnv.setOptimizationLevel(optLevel);
+                    // This no longer has an effect, but parse it for backward compatibility
```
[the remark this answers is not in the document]
gbrail: Yep!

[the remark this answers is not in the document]
gbrail: Check out new comments, thanks!

rhino-tools/src/main/java/org/mozilla/javascript/tools/shell/Main.java
gbrail: Checkout the latest patch, it supports both "-int" and "-interpreted"

rhino-tools/src/main/resources/org/mozilla/javascript/tools/resources/Messages.properties
rbri: > although don't you think that '-interpreted' is too long?

sometimes, but i learned the hard way that is easier for me to type this many chars while writing the code & doc the spending the the answering all the questions ;-D

rbri:
@gbrail i just notice that we have the Utils class more or less twice
* rhino/src/test/java/org/mozilla/javascript/tests/Utils.java
* tests/src/test/java/org/mozilla/javascript/tests/Utils.java

I think we have to merge it into one - but i'm to far away from gradle modules to understand how we can make the one from the rhino module available in the test module also.

Maybe you can have a look at this (please)

gbrail:
I'd like to address the "Utils" class in a different PR, but we can try some things. However, I think that we might need a new "test-utils" module to contain that if we want to do it the cleanest possible way.

PR #1755, opened 2024-12-13
Change default version to VERSION_ES6

gbrail:
This changes the version set when a Context object is created to VERSION_ES6 by default.

It also adds language to the JavaDoc to indicate that:

* We will not be adding additional versions after VERSION_ES6, but instead will introduce new features in new versions of Rhino, like other JavaScript engines.
* Code that uses Rhino should plan on migrating to VERSION_ES6.
* Projects that use Rhino should, if possible, stop relying on older language versions, as these typically are features that are incompatible with modern JavaScript.

(Simply changing VERSION_DEFAULT is not sufficient -- VERSION_DEFAULT unfortunately drives a number of odd behaviors that many tests and probably many programs depend on.) 

A few other notes:

First, not setting the language level and assuming that Rhino is very very out of date is a common problem with new users -- I hope that this will help in that regard.

Second, a number of tests thought that they were setting the language level but they weren't because they re-used a class that created it's own Context, so that's fixed now!


p-bakker:
The name 'VERSION_ES6' doesn't really cover what it does anymore, as it basically means: 'the latest version of EcmaScript, as far as Rhino supports it'

Maybe create a new, final language version called VERSION_ECMASCRIPT or just ECMASCRIPT or if not a new language version, then just a new alias for VERSION_ES6?

gbrail:
Yes -- adding "VERSION_ECMASCRIPT" as an alias, and using that, makes sense
to me too. I'll make it an alias, though. Thanks!

gbrail:
Check it out -- I switched to VERSION_ECMASCRIPT and kept the old name for compatibility. Thanks for looking at it!

andreabergia (approved):

rhino/src/main/java/org/mozilla/javascript/Context.java
```diff
@@ -659,7 +681,16 @@ public final int getLanguageVersion() {
...
-     * the overview documentation for version-specific behavior.
+     * the overview documentation for version-specific behavior. The default version is {@link
+     * #VERSION_ES6}, which represents the newest ECMAScript features implemented by Rhino, and this
+     * is the version that should be used unless a project needs backwards compatibility with older
+     * Rhino scripts that may depend on behaviors from the earlier history of JavaScript.
+     *
+     * <p>As of version 1.8, the Rhino community has no plans to continue to add new language
+     * versions, but instead plans to track the ECMAScript standard and add new features as the
+     * language evolves in new versions of Rhino, like other JavaScript engines. Projects that use
+     * Rhino are encouraged to migrate to the {@link #VERSION_ES6} version and stop relying on older
```
rbri, line 692: VERSION_ECMASCRIPT ?

rbri (approved):

rbri:
What about deprecating some of the older versions, including the plan to remove them in some time

gbrail:
Can we address deprecating some releases in a future issue or PR? Personally I'd like to get rid of pretty much all of that version checking stuff but I have a feeling that will prevent many projects from upgrading to a recent release of Rhino. I'd love to hear how others feel.

PR #1756, opened 2024-12-13
Add nullable JS API parameter support

drubanovich-soti:
This change allows to define JS APIs which receive nullable primitive type parameters. E.g. before the change it was not possible to pass null to JS function with string parameter, instead the function received a string "null".

As standard Java doesn't have nullable types, this feature currently can only be used if the files are written in Kotlin. Kotlin metadata library is used to detect nullable parameter types.

### DIFF
--- a/rhino/src/test/java/org/mozilla/javascript/NullableArgumentsConversionTest.java
+++ b/rhino/src/test/java/org/mozilla/javascript/NullableArgumentsConversionTest.java
@@ -1,0 +1,72 @@
+package org.mozilla.javascript;
+
+import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.mozilla.javascript.FunctionObject.*;
+
+import java.util.Arrays;
+import java.util.Collection;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+import org.mozilla.javascript.tests.Utils;
+
+@RunWith(Parameterized.class)
+public class NullableArgumentsConversionTest {
+
+    @Parameterized.Parameters
+    public static Collection<Object[]> data() {
+        return Arrays.asList(
+                new Object[][] {
+                    {"string", JAVA_STRING_TYPE, false, "string"},
+                    {"string", JAVA_STRING_TYPE, true, "string"},
+                    {null, JAVA_STRING_TYPE, false, "null"},
+                    {null, JAVA_STRING_TYPE, true, null},
+                    {2, JAVA_INT_TYPE, false, 2},
+                    {2, JAVA_INT_TYPE, true, 2},
+                    {null, JAVA_INT_TYPE, false, 0},
+                    {null, JAVA_INT_TYPE, true, null},
+                    {2.0, JAVA_DOUBLE_TYPE, false, 2.0},
+                    {2.0, JAVA_DOUBLE_TYPE, true, 2.0},
+                    {null, JAVA_DOUBLE_TYPE, false, 0.0},
+                    {null, JAVA_DOUBLE_TYPE, true, null},
+                    {true, JAVA_BOOLEAN_TYPE, false, true},
+                    {true, JAVA_BOOLEAN_TYPE, true, true},
+                    {null, JAVA_BOOLEAN_TYPE, false, false},
+                    {null, JAVA_BOOLEAN_TYPE, true, null}
+                });
+    }
+
+    private final Object arg;
+    private final int typeTag;
+    private final boolean isNullable;
+    private final Object expectedConvertedArg;
+
+    public NullableArgumentsConversionTest(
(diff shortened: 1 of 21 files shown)


drubanovich-soti:
I would like to run this idea with you before I fix the build (it passes locally) and add unit tests.
Thanks!

gbrail:
Thanks for working on this! This is an interesting idea, and it'd be great to have Rhino work better with Kotlin.

However, right now Rhino has no dependencies, and I think it's important that we keep it that way. This PR makes Rhino depend on a Kotlin package. I don't know how intrusive that is (like how many nested dependencies it brings in) but I'm still very reluctant to make a core library like this something that Rhino requires.

Is there another way to do this, like making the dependency optional and searching for it via reflection, or adding a "rhino-kotlin" module that's loaded via ServiceLoader, or incorporating the part of the Kotlin library that Rhino needs in to the project directly?

drubanovich-soti:
Sure Greg, if you like the idea, I will work on a proper implementation. 
I'll try your ServiceLoader suggestion with Kotlin-specific implementation in a new "rhino-kotlin" module.

gbrail:
One idea might be to add a new module, like "rhino-kotlin", that Kotlin
users can add... Just one of many possible ideas.

I'm hoping that we can release a "1.8.0" soon and this might not make it in
that release, but we can get to it soon after!

drubanovich-soti:
I finally got some time to work on it. I tested it with my project and it works fine. 
I still plan to add some unit tests and maybe Readme.md in kotlin-rhino, but it would be great to receive any feedback for the implementation. E.g. I wasn't sure about module/package names, maybe you can provide a better alternative.

gbrail:
I think that this approach will work -- having a separate "rhino-kotlin" module sounds good to me. Of course, I really think we need to find a way to build some test, even if it only runs as part of our CI, to actually exercise this code with Kotlin.

Also, I don't necessarily understand why there is a new module called "rhino-reflection." Would this work if we only added the new "rhino-kotlin" module?

drubanovich-soti:
I've added Kotlin tests in the third commit, note that it brings in several Kotlin dependencies though.

> I don't necessarily understand why there is a new module called "rhino-reflection." Would this work if we only added the new "rhino-kotlin" module?
I can make NullabilityDetector interface to be a part of rhino. A separate "rhino-reflection" interface module gives more flexibility to clients though. E.g. clients who don't want to use Kotlin but still want to create JS functions with nullable primitive parameters can write their own NullabilityDetector implementation which relies on their custom Nullable annotations.
Another reason is that we can now move rhino-kotlin into a different repository, making this one kotlin-independent.

gbrail:
I think that we should move the contents of "rhino-reflection" into the main "rhino" package. It's a very small amount of code and I don't think we lose anything in terms of flexibility by having it be part of the main module. Then we can still provide the behavior that the new behavior only is triggered when the "rhino-kotlin" package is present.

drubanovich-soti:
Hi Greg, I added the tests and addressed the comments.

rhino/src/main/java/org/mozilla/javascript/NullabilityDetectorProvider.java
```diff
@@ -0,0 +1,26 @@
...
+    public static NullabilityDetectorProvider getInstance() {
+        if (provider == null) {
+            provider = new NullabilityDetectorProvider();
+        }
+
+        return provider;
+    }
+
+    public NullabilityDetector getImpl() {
+        Iterator<NullabilityDetector> iterator = loader.iterator();
```
gbrail, line 23: I (very recently) added ScriptRuntime.loadOneServiceImplementation to handle this case and throw an error if there are duplicates (to avoid inconsistency if the classpath is weird). Can you look at using that? It's package-public but could easily be public.

rhino/src/test/java/org/mozilla/javascript/NullableArgumentsConversionTest.java
```diff
@@ -0,0 +1,67 @@
...
+            Object arg, int typeTag, boolean isNullable, Object expectedConvertedArg) {
+        this.arg = arg;
+        this.typeTag = typeTag;
+        this.isNullable = isNullable;
+        this.expectedConvertedArg = expectedConvertedArg;
+    }
+
+    @Test
+    public void checkArgumentConversion() {
+        Context context = new Context();
```
gbrail, line 54: That context constructor is deprecated with a good reason. Take a look at the Javadoc for it. You can either use "Context.enter()" in a "try-with-resources" block, or better yet you can use Utils.runWithAllModes, which will test your stuff in both interpreted and compiled mode.

gbrail (commented):
Two small comments on this (but I'd like to fix both before we merge).

And one last thing -- can you please update README.md to add a description of the new module and a short description of why someone might want to use it? Thanks!

drubanovich-soti:
Back for the review

gbrail:
This looks good to me, and thanks for doing all this work!

drubanovich-soti:
Thanks for the quick feedback!

PR #1758, opened 2024-12-16
Shell: improve handling of unhandled promise rejection

andreabergia:
We have changed the logging to log _all_ the exceptions, rather than just the first one. Also, we quit the process with non-zero status if we are evaluating a file (i.e. not in REPL mode) and there are any unhandled promise rejections.

Example: create a file named `promise.js`:

```js
Promise.reject('some error');
Promise.reject('another error');
```

If you launch this in both node and rhino, you will see that node exits with status 1, while rhino (before this PR) with status 0. With this change, rhino too exits with a non-zero status.

### DIFF
--- a/rhino-tools/src/main/java/org/mozilla/javascript/tools/shell/Main.java
+++ b/rhino-tools/src/main/java/org/mozilla/javascript/tools/shell/Main.java
@@ -96,7 +96,7 @@ public class Main {
             }
             if (type == PROCESS_FILES) {
                 processFiles(cx, args);
-                printPromiseWarnings(cx);
+                printPromiseWarnings(cx, true);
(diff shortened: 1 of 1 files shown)


gbrail (approved):
This seems fine with me. Thanks!

gbrail:
My only question is whether we think that someone will be broken by this change in a serious enough way to require us to add a command-line switch to not terminate on unhandled promise rejections.

andreabergia:
Honestly, rhino doesn't provide any API that returns a promise, such as `fetch` or similar. So I would guess the probability is on the low side. My preference would be to enable this behavior by default and, if someone complains, consider adding a switch to have the old behavior.

gbrail:
OK -- I'm ok with this then. Thanks!

PR #1760, opened 2024-12-18
Add a "testutils" module and move common test code there

gbrail:
This adds a "testutils" module that contains code that may be used by all of the Rhino tests. Currently it contains the "Utils" class which is used by both the "rhino" and "tests" modules, but it could contain more.

In order for this to work, code in the module is kept in the "src/main/java" directory so that other modules can depend on it for testing purposes, even though none of the code in this module should ever be used outside tests.


rbri:
@gbrail thanks

PR #1762, opened 2024-12-19
Fix: handle stack frames in the correct order

andreabergia:
Commit f3c64ee29f9094bdc87a4fe3615c21f9c65fab51 removed `ObjArray` and replaced its usage with standard JDK classes. In `Interpreter`, in particular, an `ArrayDeque` is now used to store
`cx.previousInterpreterInvocations`, which is used to generate the stack frame information. However, there is one place where `toArray` is done, and the behavior for `ObjArray` and `ArrayDeque` are different (swapped).
Unfortunately no tests actually ends up exercising this difference due to the interpreter function peeling optimization done in https://github.com/mozilla/rhino/pull/1510.

We have discovered this problem because, in ServiceNow's fork, we currently need to disable the function peeling optimization.

I cannot figure out how to write an unit test for this problem, though. Any ideas?

---

To verify that the behavior are different, I've written this code:

```java
class ObjArrayTest {
		var objArray = new ObjArray();
		objArray.push(1);
		objArray.push(2);
		objArray.push(3);
		System.out.println("objArray: " + Arrays.toString(objArray.toArray()));
	
		var deque = new ArrayDeque<Integer>();
		deque.push(1);
		deque.push(2);
		deque.push(3);
		System.out.println("ArrayDeque: " + Arrays.toString(deque.toArray()));
	}
}
```

It prints:

```
objArray: [1, 2, 3]
ArrayDeque: [3, 2, 1]
```

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/Interpreter.java
+++ b/rhino/src/main/java/org/mozilla/javascript/Interpreter.java
@@ -14,6 +14,7 @@ import java.math.BigInteger;
 import java.util.ArrayDeque;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Objects;
@@ -951,7 +952,10 @@ public final class Interpreter extends Icode implements Evaluator {
                 --previousCount;
             }
             array = new CallFrame[previousCount + 1];
-            cx.previousInterpreterInvocations.toArray(array);
+
+            ArrayList<Object> tempList = new ArrayList<>(cx.previousInterpreterInvocations);
+            Collections.reverse(tempList);
+            tempList.toArray(array);
         }
         array[array.length - 1] = (CallFrame) cx.lastInterpreterFrame;
 


gbrail:
I could see how this would be a problem! I think this is a fine solution.

As for testing, is there a way to verify the stack traces in interpreted mode? I'm not totally sure where this stuff happens but I figure that we'd get backwards stack traces in this case, would we not?

gbrail:
This is a low-risk enough change that I'm happy to merge it unless someone here has a great testing idea.

andreabergia:
> As for testing, is there a way to verify the stack traces in interpreted mode? I'm not totally sure where this stuff happens but I figure that we'd get backwards stack traces in this case, would we not?

If we have three recursive call of `interpreterLoop`, then yeah, we would see them in an incorrect order. However, given the function peeling optimization done in https://github.com/mozilla/rhino/pull/1510, in most situations we will just have _one_ call of `interpreterLoop`, with multiple frames stored - and those frames are in the correct order.

So... I am not really sure how to make this explicitly happen in a test.

PR #1763, opened 2024-12-19
fix border checks for typedarray set

rbri:
(see https://github.com/HtmlUnit/htmlunit-core-js/issues/27)


gbrail:
Looks good and thanks -- one suggestion if you have time, though -- since you're changing this, would it be difficult to put those rangeError messages in mesages.properties instead?

rbri:
@gbrail good point will do that tomorrow

rbri:
@done, have added some missing tests for that messages also

gbrail:
Looks good!

Issue #1764, opened 2024-12-20
Test262 properties outdated


andreabergia:
I've noticed that the test262.properties file is outdated today. The diffs are surprising though:

```diff
-     from/calling-from-valid-1-noStrict.js non-strict Spec pretty clearly says this should be undefined
+     from/calling-from-valid-1-noStrict.js {compiled-non-strict,interpreted-non-strict} Spec pretty clearly says this should be undefined
```

So I am not sure if just rerunning and commiting is the right fix. I guess this might be related to the optimization level of https://github.com/mozilla/rhino/pull/1754 ?

gbrail:
Yes, re-running and re-committing is the right fix, with a simple PR.
However, it might be worth looking at that PR -- there is a lot of complex
logic in there for calculating how to report on the test failures in the
four remaining test modes (non strict, strict, interpreted, and optimized)
and perhaps I missed some logic -- I think that both cases are the same,
but previously we must have consolidated them into one "non-strict" case
and now we don't.

rbri:
Can we close this?

gbrail:
I think so.

FWIW, for me, on three different OSes, even forcing Java 11 for the whole build, I can't use "updateTest262properties" without seeing a whole bunch of changes to that file that aren't relevant, so I made my last update manually. I wonder if anyone else has the same problem or if I'm doing something wrong.

rbri:
Works for me with java 11.

p-bakker:
Does this case need to remain open @andreabergia ?

andreabergia:
Nope, sorry, it's fine now.

PR #1768, opened 2024-12-21
make the report backward compatible

rbri:
have implemented a poor mans approach for being backward compatible

see https://github.com/mozilla/rhino/issues/1764

There still some differences but i have no time to analyze that at the moment.


gbrail:
I think it's a good step forward and we should do this. I will take a look at the READMEs and build scripts and get rid of references to those opt levels too -- now that we have fewer of them there's less need to restrict them IMO.

Issue #1558, opened 2024-08-07
Object.assign is broken after 1.7.15 with custom ScriptableObjects


pavelhoral:
### The issue

This is related to #780, which changed how `Object.assign` works. That change makes sense as it aligns the implementation with the ECMA specification. However it is a breaking change for all custom `ScriptableObject` based classes (not only inside RhinoJS itself, but in depending projects as well).

### Symptoms

Classes extending `ScriptableObject` quite often only override basic `get` / `put` / `has` / `getIds` methods. Trying to use `Object.assign` ends with error due to missing attribute slot:

* https://github.com/mozilla/rhino/blob/cc302b4f8fac284145c51f5ceea21ec333f7a1c8/rhino/src/main/java/org/mozilla/javascript/ScriptableObject.java#L486
* https://github.com/mozilla/rhino/blob/cc302b4f8fac284145c51f5ceea21ec333f7a1c8/rhino/src/main/java/org/mozilla/javascript/ScriptableObject.java#L2598

```
js> var customNative = new Packages.org.example.javascript.CustomNative()
js> Object.assign({}, customNative)
js: Property hello not found.
        at (unknown):18
```

### Quick solution to the issue

The first solution that comes to my mind is to make sure that missing attribute slot does not end with exception... at least not in case when called from `Object.assign` as the specification clearly states the following:

![image](https://github.com/user-attachments/assets/ec7941a8-004a-488a-97e5-4c2aebb8feb4)

So _undefined_ property attributes is a well defined state... the property should be skipped.

### Proper solution

Previous _quick solution_ still means 1.7.15+ brought breaking changes for custom `ScriptableObject` implementations. I would appreciate if there would be clear example how custom `NativeObject` / `IdScriptableObject` / `ScriptableObject` should be implemented... in the case of `IdScriptableObject` implementation I guess one needs to override `findInstanceIdInfo` method?

### Footnote

Not sure if this is a partial duplicate of #1549. That issue's description is a bit cryptic for me.

p-bakker:
Added the `docs` tag as well, as in the end, we'll need to update the docs for implementing custom host objects as well: currently all examples are about implementing the Scriptable interface, but while that works, it doesn't result in host objects that `extend` Object, which is a breast practice nowadays

pavelhoral:
I was playing with this a bit in a project that is using RhinoJS and I am not able to fix it on my end as `NativeObject`'s assign method calls package private method `putImpl` that I am not able to override:

[...]

PR #1771, opened 2024-12-23
Support more spec-compliant ScriptableObject extensions

gbrail:
This fixes a problem in which ScriptableObject subclasses couldn't support some of the newer code, like Object.assign, because it depends on internal implementation details of ScriptableObject. This changes two things:

* Adds new overridable methods to ScriptableObject called "putOwnProperty".
* Call those in internal places instead of calling "putImpl" directly.
* Catch exceptions in Object.assign if "getAttributes" throws an exception, as the default implementation does.

This makes it possible to address the problem identified in https://github.com/mozilla/rhino/issues/1558.

To be fully compatible, a ScriptableObject subclass that wants to add its own properties should override "putOwnProperty" INSTEAD of overriding "put" and delegate to the superclass for any unknown properties.

Added a new test that demonstrates this technique.

I'd be interested in feedback here -- this is not a fix to all spec-compliance problems but it's not the ugliest fix in the world either.

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/AbstractEcmaObjectOperations.java
+++ b/rhino/src/main/java/org/mozilla/javascript/AbstractEcmaObjectOperations.java
@@ -221,7 +221,7 @@ public class AbstractEcmaObjectOperations {
         if (base == null) base = o;
 
         if (base instanceof ScriptableObject) {
-            if (((ScriptableObject) base).putImpl(p, 0, o, v, isThrow)) return;
+            if (((ScriptableObject) base).putOwnProperty(p, o, v, isThrow)) return;
 
             o.put(p, o, v);
         } else {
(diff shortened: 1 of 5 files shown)


rbri:
ok for me

do you have an idea why this does not fix getOwnPropertyDescriptor/trap-is-null-target-is-proxy.js
while it fixes all the others especially getOwnPropertyDescriptor/trap-is-undefined-target-is-proxy.js ?

gbrail:
Something like this might also be a path forward for getting strict mode right -- we can keep Scriptable, but add a new interface for objects that want to extend in a more spec-compliant way, and have code in ScriptableObject and ScriptRuntime adapt if the new interface is present.

gbrail:
Also, as for the tests that were fixed -- this might be because NativeProxy doesn't implement "getAttributes". Arguably it should, or we should change the contract between the runtime and ScriptableObject extensions more.

PR #1772, opened 2024-12-23
Fix another old reference to optimization levels

gbrail:
Fix a spot I missed when updating the README.

Thanks @andreabergia!


andreabergia (approved):

gbrail:
Yes, the docs in the other repo need an update, and that will probably come some later time, like in January...

PR #1774, opened 2024-12-24
Fix: methods cannot be used as constructor

### LINKED ISSUE #1299: Support ES2015 Method Definitions - prevent methods as constructor
See:
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Method_definitions
- https://stackoverflow.com/a/50971728/7049810

While Rhino supports the method shorthadn (`var x = {foo(){}}`), it doesn't:
- ~~support `super`~~
- prevent calling the method as a constructor (new x.foo())
- ~~support generator methods~~

andreabergia:
The spec says that invoking a method as a constructor is not valid, but invoking normal functions is. This is, for example, `node`:

```js
> o = { method() {}, notAMethod: function() {} }
{ method: [Function: method], notAMethod: [Function: notAMethod] }
> new o.method()
Uncaught TypeError: o.method is not a constructor
> new o.notAMethod()
notAMethod {}
```

This PR implements the same behavior in Rhino, and fixes #1299.

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/BaseFunction.java
+++ b/rhino/src/main/java/org/mozilla/javascript/BaseFunction.java
@@ -408,6 +408,11 @@ public class BaseFunction extends IdScriptableObject implements Function {
 
     @Override
     public Scriptable construct(Context cx, Scriptable scope, Object[] args) {
+        if (cx.getLanguageVersion() >= Context.VERSION_ES6 && this.getHomeObject() != null) {
+            // Only methods have home objects associated with them
+            throw ScriptRuntime.typeErrorById("msg.not.ctor", getFunctionName());
+        }
+
         Scriptable result = createObject(cx, scope);
         if (result != null) {
             Object val = call(cx, scope, result, args);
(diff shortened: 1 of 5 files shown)


gbrail:
This makes sense to me, but I wonder how much existing code it could break and whether we should put it behind a "VERSION_ES6" or a "VERSION_ECMASCRIPT" flag just in case -- then again, since it didn't break any of MozillaSuiteTest perhaps it's not a big problem.

gbrail (commented):
I'm willing to be convinced on this, but I think that since this breaks things that may have worked before, we should check that language version is >= VERSION_ES6 before throwing this exception. I'm not sure that we will put this in 1.8.0, but can merge it soon after.

andreabergia:
> I'm willing to be convinced on this, but I think that since this breaks things that may have worked before, we should check that language version is >= VERSION_ES6 before throwing this exception.

Added a check on the language version.

rhino/src/main/java/org/mozilla/javascript/BaseFunction.java
```diff
@@ -408,6 +408,11 @@ public Object call(Context cx, Scriptable scope, Scriptable thisObj, Object[] ar
 
     @Override
     public Scriptable construct(Context cx, Scriptable scope, Object[] args) {
+        if (cx.getLanguageVersion() >= Context.VERSION_ES6 && this.getHomeObject() != null) {
```
rbri, line 411: usually we do not use this (only in setters when it is required)

rbri (approved):
LGTM

gbrail:
Looks good to me -- I agree about the excessive use of "this" but I'd rather move forward ;-) Thanks!

PR #1782, opened 2024-12-30
Reduce indirection and memory overhead for slot maps.

aardvark179:
# Reducing the property related memory overhead of JavaScript objects.

## Current slot map model

Currently JavaScript Objects in RHino store their properties in slot map, which is held by a slot map container. The map itself then either holds an array, or some other collection structure, which actually holds the slots. So the structure looks something like this:

```mermaid
graph LR
Object["ScriptableObject"]
Container["SlotMapContainer"]
Map["SlotMap"]
Array["Slot[]"]
Slot1["Slot"]
Slot2["Slot"]
Slot3["Slot"]
Object1["ScriptableObject"]
Container1["SlotMapContainer"]
Map1["SlotMap"]
Array1["Slot[]"]
Slot4["Slot"]
Slot5["Slot"]
Slot6["Slot"]
Object --> Container --> Map --> Array
Array --> Slot1
Array --> Slot2
Array --> Slot3
Object1 --> Container1 --> Map1 --> Array1
Array1 --> Slot4
Array1 --> Slot5
Array1 --> Slot6
```

And in any environment which relies on thread safe slot map handling there is an additional stamped lock object like this:

```mermaid
graph LR
Object["ScriptableObject"]
Container["SlotMapContainer"]
Lock["StampedLock"]
Map["SlotMap"]
Array["Slot[]"]
Slot1["Slot"]
Slot2["Slot"]
Slot3["Slot"]
Object --> Container --> Map --> Array
Container --> Lock
Array --> Slot1
Array --> Slot2
Array --> Slot3
Object1["ScriptableObject"]
Container1["SlotMapContainer"]
Lock1["StampedLock"]
Map1["SlotMap"]
Array1["Slot[]"]
Slot4["Slot"]
Slot5["Slot"]
Slot6["Slot"]
Object1 --> Container1 --> Map1 --> Array1
Container1 --> Lock1
Array1 --> Slot4
Array1 --> Slot5
Array1 --> Slot6
```

There is one optimisation here, in that the arrays for slots in an `EmbeddedSlotMap` are not created until slots are stored in the map, but all the other objects are created eagerly. If I look at a heap dump from one of our production systems and analyse the number of properties stored in slot maps then we see the following:

| No. of props | Count   | Cummulative count | Cummulative % |
|--------------|---------|-------------------|---------------|
| 0            | 113,538 | 113,538           | 76.6428827    |
| 1            | 28,013  | 141,551           | 95.5528254    |
[...]

PR #1783, opened 2024-12-30
Start to factor out slot map container

aardvark179:
# Start to factor out slot map container

* #1782 

This PR introduces an abstract super class of `ScriptableObject` and `SlotMapContainer` called `SlotMapOwner`. It is responsible for creating, getting, and setting the slot map owned by these objects. I originally experimented with this as an interface but was writing the almost exact same code for both classes and it was hard to limit the visibility of the API which should be entirely internal to Rhino. This also moves all slot promotion to the maps themselves. I did some minor refactoring in `EmbeddedSlotMap` to only check for promotion when would expand the backing array.

This changes the slot map structure in the single threaded case from

```mermaid
graph LR
Object["ScriptableObject"]
Container["SlotMapContainer"]
Map["SlotMap"]
Array["Slot[]"]
Slot1["Slot"]
Slot2["Slot"]
Slot3["Slot"]
Object1["ScriptableObject"]
Container1["SlotMapContainer"]
Map1["SlotMap"]
Array1["Slot[]"]
Slot4["Slot"]
Slot5["Slot"]
Slot6["Slot"]
Object --> Container --> Map --> Array
Array --> Slot1
Array --> Slot2
Array --> Slot3
Object1 --> Container1 --> Map1 --> Array1
Array1 --> Slot4
Array1 --> Slot5
Array1 --> Slot6
```
to:
```mermaid
graph LR
Object["ScriptableObject"]
Map["SlotMap"]
Array["Slot[]"]
Slot1["Slot"]
Slot2["Slot"]
Slot3["Slot"]
Object1["ScriptableObject"]
Map1["SlotMap"]
Array1["Slot[]"]
Slot4["Slot"]
Slot5["Slot"]
Slot6["Slot"]
Object --> Map --> Array
Array --> Slot1
Array --> Slot2
Array --> Slot3
Object1 --> Map1 --> Array1
Array1 --> Slot4
Array1 --> Slot5
Array1 --> Slot6
```

This saves 24 bytes per object.

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/SlotMapOwner.java
+++ b/rhino/src/main/java/org/mozilla/javascript/SlotMapOwner.java
@@ -1,0 +1,39 @@
+package org.mozilla.javascript;
+
+public abstract class SlotMapOwner {
+    private static final long serialVersionUID = 1L;
+
+    /**
+     * This holds all the slots. It may or may not be thread-safe, and may expand itself to a
+     * different data structure depending on the size of the object.
+     */
+    private SlotMap slotMap;
+
+    protected SlotMapOwner() {
+        slotMap = createSlotMap(0);
+    }
+
+    protected SlotMapOwner(int capacity) {
+        slotMap = createSlotMap(capacity);
+    }
+
+    protected SlotMapOwner(SlotMap map) {
+        slotMap = map;
+    }
+
+    protected static SlotMap createSlotMap(int initialSize) {
+        Context cx = Context.getCurrentContext();
+        if ((cx != null) && cx.hasFeature(Context.FEATURE_THREAD_SAFE_OBJECTS)) {
+            return new ThreadSafeSlotMapContainer(initialSize);
+        }
+        return new SlotMapContainer(initialSize);
+    }
+
+    final SlotMap getMap() {
+        return slotMap;
+    }
+
+    final void setMap(SlotMap newMap) {
+        slotMap = newMap;
+    }
+}
--- a/rhino/src/main/java/org/mozilla/javascript/SlotMapOwner.java
+++ b/rhino/src/main/java/org/mozilla/javascript/SlotMapOwner.java
@@ -25,8 +25,13 @@ public abstract class SlotMapOwner {
         Context cx = Context.getCurrentContext();
         if ((cx != null) && cx.hasFeature(Context.FEATURE_THREAD_SAFE_OBJECTS)) {
             return new ThreadSafeSlotMapContainer(initialSize);
+        } else if (initialSize == 0) {
+            return SlotMapContainer.EMPTY_SLOT_MAP;
+        } else if (initialSize > SlotMapContainer.LARGE_HASH_SIZE) {
+            return new HashSlotMap();
+        } else {
+            return new EmbeddedSlotMap();
         }
-        return new SlotMapContainer(initialSize);
     }
 
     final SlotMap getMap() {
--- a/benchmarks/src/jmh/java/org/mozilla/javascript/benchmarks/SlotMapBenchmark.java
+++ b/benchmarks/src/jmh/java/org/mozilla/javascript/benchmarks/SlotMapBenchmark.java
@@ -44,7 +44,7 @@ public class SlotMapBenchmark {
     public Object embeddedInsert1Key(EmbeddedState state) {
         Slot newSlot = null;
         for (int i = 0; i < 100; i++) {
-            newSlot = state.emptyMap.modify(state.randomKeys[i], 0, 0);
+            newSlot = state.emptyMap.modify(null, state.randomKeys[i], 0, 0);
         }
         if (newSlot == null) {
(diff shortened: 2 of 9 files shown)


aardvark179:
Rebased after merge of #1782.

rhino/src/main/java/org/mozilla/javascript/SlotMapOwner.java
```diff
@@ -0,0 +1,60 @@
...
+import java.lang.invoke.VarHandle;
+
+public abstract class SlotMapOwner {
+    private static final long serialVersionUID = 1L;
+
+    /**
+     * This holds all the slots. It may or may not be thread-safe, and may expand itself to a
+     * different data structure depending on the size of the object.
+     */
+    private volatile SlotMap slotMap;
```
gbrail, line 13: I'm a bit concerned about making this volatile by default given how frequently it is accessed. Reading the docs for VarHandle, though, it looks like we can use it to choose volatile access semantics when we need them (in the thread-safe case) and non-volatile semantics otherwise. Do you know if that's something that we can do? I would be more comfortable with this if I could try to reduce as much overhead as possible in the single-threaded case.

rhino/src/main/java/org/mozilla/javascript/SlotMapOwner.java
```diff
@@ -0,0 +1,60 @@
...
+        }
+    }
+
+    @SuppressWarnings("AndroidJdkLibsChecker")
+    private static final VarHandle SLOT_MAP = getSlotMapHandle();
+
+    @SuppressWarnings("AndroidJdkLibsChecker")
+    private static VarHandle getSlotMapHandle() {
+        try {
+            return MethodHandles.lookup()
```
gbrail, line 46: Is there a way for us to verify whether this feature is going to break Android compatibility? I think that our hypothesis is that it won't affect Android users unless they try to use "thread safe slot maps." However I don't know whether the incompatibility would be triggered by using the VarHandle, or merely by trying to load the class. Is there any way that we can verify this?

gbrail (commented):
Thanks for being patient -- I really like this work but I have two specific concerns that are making me reluctant to just move forward with this one. PTAL at my comments and can you let me know if you have any ideas? Thanks!

rhino/src/main/java/org/mozilla/javascript/SlotMapOwner.java
aardvark179: I was expecting a question along those lines whether I made it volatile or not. I made it volatile as that was an easier way to think about things, but I believe it can be made non-volatile since all the multithreaded writes should be protected by fences. I'd like to think about this carefully before changing it though, and probably go through the §17.4 of the JLS and the VarHandles docs to make sure we aren't leaving any holes in the reasoning.

rhino/src/main/java/org/mozilla/javascript/SlotMapOwner.java
aardvark179: On Android 33 (Tiramisu) or later (released in 2022) `VarHandle`s are supported so this should work fine. I am not sure what the behaviour is like before that (whether the class is present but returns an error, or if the app will fail at install or build time). Since we already suppressed warnings on `ThreadSafeSlotMapContainer` for android compatibility concerns would it be reasonable to move the VarHandle and uses to a separate class only used by the thread safe maps?

rhino/src/main/java/org/mozilla/javascript/SlotMapOwner.java
gbrail: Yes, I'd love to see if we can make this work with the same semantics without making that field volatile -- let me know what you learn about whether we can do that!

rhino/src/main/java/org/mozilla/javascript/SlotMapOwner.java
gbrail: I would be a lot more comfortable if the VarHandle stuff could be isolated to a class that's not loaded in the non-thread-safe case. It's too bad we don't have a clear compatibility policy for Android -- I've noticed that people have lots of versions that they use and it's hard to predict what will break.

rhino/src/main/java/org/mozilla/javascript/SlotMapOwner.java
aardvark179: Okay. I'm happy that making the field non-volatile is okay. All access in multithreaded environments is either to immutable maps which are replaced with a CAS operation, or behind a lock.

rhino/src/main/java/org/mozilla/javascript/SlotMapOwner.java
aardvark179: Separated the multithreaded case to a separate class, and moved that to the later PR for thread safe stuff. I've kept it as a class in the same file so that it can have private access to the field.

gbrail:
I think this is good progress -- thanks and looking forward to the next one!

PR #1789, opened 2025-01-03
Improve feature test support

rbri:
* enhance Utils a bit to support feature testing
* migrate a feature test case from core-js to rhino (found while updating to the latest rhino source)

### DIFF
--- a/rhino/src/test/java/org/mozilla/javascript/tests/Bug421071Test.java
+++ b/rhino/src/test/java/org/mozilla/javascript/tests/Bug421071Test.java
@@ -63,16 +63,8 @@ public class Bug421071Test {
         thread.join();
     }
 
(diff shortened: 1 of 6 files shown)


andreabergia (approved):
This looks great! Removing all those various context factory makes it quite a bit more readable.

I would suggest a slightly different API name though, i.e. rather than

```java
        factory = new Utils.FeatureContextFactory(Context.FEATURE_DYNAMIC_SCOPE);
```

have something like:

```java
        factory = Utils.contextFactoryWithFeatures(Context.FEATURE_DYNAMIC_SCOPE);
```

rbri:
@andreabergia good point, will do that update

gbrail:
Looks good to me -- thanks!

PR #1794, opened 2025-01-07
Test: add a missing `@Test` annotation in `NativeProxyTest`

andreabergia:
Just adding a missing `@Test` annotation on a method of `NativeProxyTest` that I spotted while trying to understand how everything works. 🙂 

(@rbri - great job by the way!)


rbri:
@andreabergia thanks for noting that

PR #1799, opened 2025-01-12
Use ServiceLoader to load RegExp and XML implementations

gbrail:
- Use ServiceLoader to load RegExp and XML implementations
- Don't forget new files

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/Context.java
+++ b/rhino/src/main/java/org/mozilla/javascript/Context.java
@@ -23,10 +23,8 @@ import java.util.Deque;
 import java.util.EnumSet;
 import java.util.HashMap;
 import java.util.HashSet;
-import java.util.Iterator;
 import java.util.Locale;
 import java.util.Map;
-import java.util.ServiceLoader;
 import java.util.Set;
 import java.util.TimeZone;
 import java.util.function.Consumer;
@@ -2329,9 +2327,9 @@ public class Context implements Closeable {
      */
     @Deprecated
     public XMLLib.Factory getE4xImplementationFactory() {
-        Iterator<XMLLoader> i = ServiceLoader.load(XMLLoader.class).iterator();
-        if (i.hasNext()) {
-            return i.next().getFactory();
+        XMLLoader loader = ScriptRuntime.loadOneServiceImplementation(XMLLoader.class);
+        if (loader != null) {
+            return loader.getFactory();
         }
         return null;
     }
@@ -2693,13 +2691,7 @@ public class Context implements Closeable {
 
     RegExpProxy getRegExpProxy() {
         if (regExpProxy == null) {
-            ServiceLoader<RegExpProxy> l = ServiceLoader.load(RegExpProxy.class);
-            Iterator<RegExpProxy> i = l.iterator();
-            if (i.hasNext()) {
-                regExpProxy = i.next();
-            } else {
-                return null;
-            }
+            regExpProxy = ScriptRuntime.loadOneServiceImplementation(RegExpProxy.class);
         }
         return regExpProxy;
     }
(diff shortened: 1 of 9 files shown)


gbrail:
This PR uses the ServiceLoader to load the implementations of Regexp and XML objects, as opposed to what we do now, which is to and load classes based on hard-coded class names. It also moves initialization of the specific objects required by XML to the XML implementation and gets that code out of ScriptRuntime.

I think that this is a bit cleaner than we what did before, and it theoretically lets us replace these implementations in future versions of Rhino by just changing the classpath. On the other hand, in both cases we just take the first service that we find, which is not how ServiceLoader is normally used, since if there is more than one implementation we run the risk of inconsistency.

I'm open to suggestions as to other ways to do this, or whether it's worth it at all, but I like the way that this gets the XML stuff out of the main codebase at least.

This is all leading up to another PR where I'm going to try and use a simpler "lazy loading" mechanism for most of the standard objects to try and keep the loading of the standard library objects as fast as we can.

gbrail:
I recall that we discussed some of these issues with @rPraml back in https://github.com/mozilla/rhino/pull/1699. I think that this PR is a simpler one that at least gets closer to the possibility of removing a few unnecessary XML dependencies from the main module.

rhino/src/main/java/org/mozilla/javascript/ScriptRuntime.java
```diff
@@ -194,19 +197,15 @@ public static ScriptableObject initSafeStandardObjects(
...
-            String xmlImpl = cx.getE4xImplementationFactory().getImplementationClassName();
-            new LazilyLoadedCtor(scope, "XML", xmlImpl, sealed, true);
-            new LazilyLoadedCtor(scope, "XMLList", xmlImpl, sealed, true);
-            new LazilyLoadedCtor(scope, "Namespace", xmlImpl, sealed, true);
-            new LazilyLoadedCtor(scope, "QName", xmlImpl, sealed, true);
+        if (cx.hasFeature(Context.FEATURE_E4X)) {
+            XMLLoader loader = loadOneServiceImplementation(XMLLoader.class);
+            if (loader != null) {
+                loader.load(scope, sealed);
+            }
```
rPraml, line 208: CHECKME: Should we log some warning in future, if `FEATURE_E4X` is enabled (unfortunately, it is by default), but no XMLLoader found?

rhino/src/main/java/org/mozilla/javascript/Context.java
```diff
@@ -2318,11 +2319,19 @@ public boolean hasFeature(int featureIndex) {
...
+     * <p>This is no longer used in E4X -- an implementation is only provided for backward
+     * compatibility.
+     *
      * @return An XMLLib.Factory. Should not return <code>null</code> if {@link #FEATURE_E4X} is
      *     enabled. See {@link #hasFeature}.
      */
+    @Deprecated
     public XMLLib.Factory getE4xImplementationFactory() {
-        return getFactory().getE4xImplementationFactory();
+        XMLLoader loader = ScriptRuntime.loadOneServiceImplementation(XMLLoader.class);
```
rPraml, line 2330: INFO: As far as I know, this will create a new instance of XMLLoader every time.

rPraml (approved):
This PR should not introduce any behaviour change, so it looks good to me.

rhino/src/main/java/org/mozilla/javascript/ScriptRuntime.java
gbrail: That's an interesting point. Unfortunately we don't have a way to log a warning and I'm reluctant to add a dependency on a logging package or just "System.out.println", and throwing an exception will probably break someone, so I'm not sure realistically what we can do -- any ideas?

rPraml: In my other PR I had the idea to deprecate the flag. One idea is to enable xml parsing only, when the implementation is on class path.

But for now, I'm happy if this will get merged, as it does not introduce a functional change. I agree, that printing a warning could break things for others. We have the same situation in master branch, that we can parse, but not use xml, if dom3 is not present. And there is also no warning

gbrail:
Thanks for your comments -- I'm going to move forward with this. Thanks!

PR #1800, opened 2025-01-13
Convert Map and Set to use lambdas

gbrail:
- Convert NativeMap to lambda
- Partial work on lambdas for native map and set.
- Use aliases for properties that need to be the same

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/AbstractEcmaObjectOperations.java
+++ b/rhino/src/main/java/org/mozilla/javascript/AbstractEcmaObjectOperations.java
@@ -264,8 +264,19 @@ public class AbstractEcmaObjectOperations {
             Object items,
             Object callback,
             KEY_COERCION keyCoercion) {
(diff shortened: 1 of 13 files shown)


rbri:
@gbrail had a quick look at this - looks good and another minor step forward.

Just saw that some WeakMap/Set tests are still failing - i guess our code does not accept symbols as weak map keys.

https://tc39.es/ecma262/multipage/executable-code-and-execution-contexts.html#sec-canbeheldweakly

Maybe you can fix that on the way...

gbrail:
Thanks for the feedback -- I'm a bit behind on this because I ran in to some serialization difficulties with another lambda-related PR and I want to understand all that. I wish that serialization would just go away (and TBH a few years ago I believe the Java community was advocating for that) but it didn't and I believe that continuations use it so I'm going to keep on trying to figure it out.

rbri:
Super cool, another step forward

PR #1804, opened 2025-01-26
Optimize `extractSetMethod(...)` and `findFunction(...)`

ZZZank:
- setter extracting is rewritten to do setter scanning in 1 pass instead of 2 passes
- a helper method `failFastConversionWeights(...)` that can compute conversion weights in batch
- computed weights from `failFastConversionWeights(...)` will now be cached to prevent computing weight again in `preferSignature()`

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/JavaMembers.java
+++ b/rhino/src/main/java/org/mozilla/javascript/JavaMembers.java
@@ -727,28 +727,24 @@ class JavaMembers {
         //       instance of the target arg to determine that.
         //
 
-        // Make two passes: one to find a method with direct type assignment,
-        // and one to find a widening conversion.
-        for (int pass = 1; pass <= 2; ++pass) {
-            for (MemberBox method : methods) {
-                if (!isStatic || method.isStatic()) {
-                    Class<?>[] params = method.argTypes;
-                    if (params.length == 1) {
-                        if (pass == 1) {
-                            if (params[0] == type) {
-                                return method;
-                            }
-                        } else {
-                            if (pass != 2) Kit.codeBug();
-                            if (params[0].isAssignableFrom(type)) {
-                                return method;
-                            }
-                        }
+        MemberBox acceptableMatch = null;
+        for (MemberBox method : methods) {
+            if (!isStatic || method.isStatic()) {
+                Class<?>[] params = method.argTypes;
+                if (params.length == 1) {
+                    if (params[0] == type) {
+                        // perfect match, no need to continue scanning
+                        return method;
+                    }
+                    if (acceptableMatch == null
+                        && params[0].isAssignableFrom(type)) {
+                        // do not return at this point, there can still be perfect match
+                        acceptableMatch = method;
                     }
                 }
             }
         }
-        return null;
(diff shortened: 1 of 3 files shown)


ZZZank:
I've ran a test for these changes (https://github.com/ZZZank/rhino-mozilla/actions/runs/12994132974), seems to be working quite well

gbrail:
This looks fine to me, although I don't consume this bit of Rhino functionality myself much anymore and it'd be nice to see if someone else could check it out.

Also, could you please run "./gradlew test testCodeCoverageReport" and take a look at the result (which is in tests/build/reports/jacoco/testCodeCoverageReport) for the parts that you changed?

I do see a big "we should never get here" in your code, and I won't worry as much about coverage there, but there are a few places where your new code checks for things and we never get there in the tests. It's not much but given how complex this code is I would feel better if you took at look at the coverage for the stuff you touched and see if you can cover as many of the edge cases as you can. Thanks!

ZZZank:
Code Coverage report: https://github.com/ZZZank/rhino-mozilla/actions/runs/13013170787

I took a brief look into it, and it seems almost all modified codes are touched, except for line380~382:
```
->      if (firstBestFit < 0) {
->          // Nothing was found
->          return -1;
        } else if (extraBestFitsCount == 0) {
            // single best fit
            return firstBestFit;
        }
```

and line507~509:

```
        if (member.vararg) {
            typeLen--;
->          if (typeLen > args.length) {
->              return null;
            }
        } else {
```

I knew a way of testing them both, but I'm not familiar with Rhino Test System. So I will write down my plan here first, hopefully I can figure Rhino Test out or someone will help me.

First make a dummy class:
```java
class VarArgDummy {
   public static void dummy(String s0, String s1, String s2, String... varArgs) {}
}
```

import it into Rhino as `VarArgDummy`, and call:
```js
VarArgDummy.dummy("s0", "s1, but no s2")
```
It should throw an exception, and the aforementioned two pieces of codes will be touched. It will work like this:
```java
// in findFunction(...)
        if (methodsOrCtors.length == 1) { // dummy() is the only method with such name
            if (failFastConversionWeights(args, methodsOrCtors[0]) == null) { //see below, returning null
                return -1; // code touched
            }

// In failFastConversionWeights(...)
        if (member.vararg) { // the method is vararg
            typeLen--; // typeLen was originally 4
            if (typeLen > args.length) { // typeLen is 3, args.length is 2
                return null; // code touched, back to findFunction(...)
            } 
        } else {
```

gbrail:
Thanks for looking at that -- It'd be great to have a few more tests. I'd suggest looking at some of the tests in the "tests" package, especially the ones under "src/test/java/org/mozilla/javascript/test," for example, but basically any Junit test will work (most use Junit 4 but we support Junit 5 too).

You should also look at the "Utils" class in the "testutils" package, which will manage the Context properly (which is important) and also ensure that each test runs both in interpreted and in compiled mode.

ZZZank:
code coverage report after adding tests: https://github.com/ZZZank/rhino-mozilla/actions/runs/13067986035 , the two branches mentioned above are covered

gbrail:
Thanks! This looks good now!

PR #1805, opened 2025-01-26
Make lambda-based classes serializable

gbrail:
Create new interfaces and change existing ones so that native classes based on LambdaConstructor and LambdaFunction can be serialized. Even though serialization is an old and complex part of Java that should not be used in modern apps, many existing code relies on it, as does Rhino's continuation mechanism, which means that if we want to rebuild many of the native classes using lambdas, we need serialization to work.

This adds some new interfaces to make this possible, and fortunately does not require major changes other than that. It also adds a set of tests so that we can ensure that key native classes are serializable as we continue to make changes.

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/LambdaAccessorSlot.java
+++ b/rhino/src/main/java/org/mozilla/javascript/LambdaAccessorSlot.java
@@ -1,8 +1,5 @@
 package org.mozilla.javascript;
 
-import java.util.function.BiConsumer;
-import java.util.function.Function;
-
 /**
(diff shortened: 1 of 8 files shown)


gbrail:
Not sure if any of the regulars got a chance to look at this -- I feel it's pretty uncontroversial and hope to merge it later in the week unless someone is concerned. Thanks!

rbri:
Looks good for me, +1

p-bakker:
Given this, is it safe to say we can close #704 and it's related PR?

Issue #970, opened 2021-06-29
Support ES2015 block-level function declarations in strict-mode


p-bakker:
See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/block#block_scoping_rules_with_let_const_or_function_declaration_in_strict_mode

In ES5 function declarations were not allowed in strict mode in the newly introduced block scopes
As of ES2015 function declarations inside block scopes in strict mode are allowed, but they are to be block scopes

PR #1806, opened 2025-01-30
Fixed problems with hoisting

andreabergia:
This PR fixes some problems with function hoisting in nested scopes. Consider this:

```js
function outer() {
  inner();

  function inner() {}
  do {
    inBlock();
    function inBlock() {} 
  } while (false);

  var expr = function() {}
}
```

The hoisting rules say that this is sorta equivalent to:

```js
function outer() {
  // Hoisting declaraitons
  var inner;
  var inBlock;
  var expr;
  inner = function inner() {}

  inner();

  do {
    // Hoisting
    inBlock = function inBlock() {} 

    inBlock();
  } while (false);

  expr = function() {}
}
```

Rhino correctly handles hoisting at the top level of the function (i.e. for `inner`), but it does not handle it correctly for nested blocks (i.e. for `inBlock`). So in the example above, the following exception gets thrown:

```
TypeError: inBlock is not a function, it is undefined
```

This PR fixes the problem. The trick is that, whenever the IR for a block is generated, nested functions are hoisted on top of it.

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/IRFactory.java
+++ b/rhino/src/main/java/org/mozilla/javascript/IRFactory.java
@@ -499,10 +499,26 @@ public final class IRFactory {
         }
         try {
             List<Node> kids = new ArrayList<>();
+            /*
+            Function declarations inside blocks (FUNCTION_EXPRESSION_STATEMENTS) should be
+            hoisted at the top of block so that they can be referred by statements above them
+             */
+            List<Node> functions = new ArrayList<>();
+
             for (Node kid : node) {
-                kids.add(transform((AstNode) kid));
+                if (kid instanceof FunctionNode
+                        && ((FunctionNode) kid).getFunctionType()
+                                == FunctionNode.FUNCTION_EXPRESSION_STATEMENT) {
+                    functions.add(transform((AstNode) kid));
+                } else {
+                    kids.add(transform((AstNode) kid));
+                }
             }
             node.removeChildren();
+
+            for (Node function : functions) {
+                node.addChildToBack(function);
+            }
             for (Node kid : kids) {
                 node.addChildToBack(kid);
             }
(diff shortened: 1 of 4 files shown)


andreabergia:
> Based on https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/function#block-level_function_declaration, I don't think any assumptions are to be made as to whether functions declarations within blocks are hoisted or not, at least in non-strict mode.
> 
> When in Strict mode however, the function should be hoisted, but only to the top of it's containing block, so not the outer function
> 
> Also see related PR/discussion #970

We discovered this because of a problem reported by one of our customers. What we implemented is coherent with the other engines that we have tested (v8, jsc, spidermonkey) from what we can tell.

Functions declared inside blocks are _not_ hoisted to the top of the function; the equivalent `var` declaration is, though.  But, the _initialization_ is done only in the block.

Example:

```js
if (typeof 'print' === 'undefined') {
  print = console.log.bind(console);
}

(function test() {
  print(typeof inner);
  if (true) {
    print(typeof inner);
    function inner() {}
  }
})();

```

This prints: `undefined function` in v8, spidermonkey, jsc.
In rhino before this patch it prints `undefined undefined`, but it is coherent with other engines after this PR.

gbrail:
I'm all for doing it. I wonder why there aren't any test262 tests that fail because of this behavior, though. Are there other tests we can pull from the V8 test suite (We have also used the "mjsunit" tests from the V8 repo somewhere) so that we can make sure that we're not off on our own here? Otherwise I think this is a good change!

gbrail:
This looks fine to me. Thanks for doing it!

andreabergia:
> I'm all for doing it. I wonder why there aren't any test262 tests that fail because of this behavior, though. Are there other tests we can pull from the V8 test suite (We have also used the "mjsunit" tests from the V8 repo somewhere) so that we can make sure that we're not off on our own here? Otherwise I think this is a good change!

Sorry, didn't get around to check that.
I'm a bit surprised too that there were no test262 cases, honestly.

rbri:
> Sorry, didn't get around to check that.
I'm a bit surprised too that there were no test262 cases, honestly.

Same here but at least there was a test in HtmlUnit/core-js and also a fix for that. Have removed our fix and all the tests are still passing. Thanks @andreabergia for all your work (hope for more)

https://github.com/HtmlUnit/htmlunit-rhino-fork/commit/e725bbaae7fbe650bba117ad69b643335e8fae11
https://github.com/HtmlUnit/htmlunit-core-js/commit/f6df509e6965eef9f7e124d1470b7e12bb35a95d

PR #1807, opened 2025-01-31
fix test utils

ZZZank:
- [Utils#assertException(int, Class, String, String) was doing nothing](https://github.com/mozilla/rhino/commit/1fe648b8da8775ee1e8307a864c83be391c400b1), because it has no method body:

```java
    public static <T extends Exception> void assertException(
            final int languageVersion,
            final Class<T> expectedThrowable,
            final String expectedMessage,
            String js) {}
```

- There's also a new helper function `Utils#lines(String[])` to make writing multi-line scripts easier. `Utils.lines("line1", "line2", "line3")` equals to `"line1\n" + "line2\n" + "line3"`


gbrail:
This seems helpful and fine to me. Any thoughts from others? Otherwise I'll merge it soon.

rbri:
+1

gbrail:
Thanks for doing this!

PR #1808, opened 2025-02-01
Implement number classes using lambdas

gbrail:
Next step in the push to move away from IdScriptableObject is to switch the implementations of Number, BigInteger, and Boolean to be based on lambdas.

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/ScriptRuntime.java
+++ b/rhino/src/main/java/org/mozilla/javascript/ScriptRuntime.java
@@ -1076,7 +1076,6 @@ public class ScriptRuntime {
 
     public static String numberToString(double d, int base) {
         if ((base < 2) || (base > 36)) {
-            // throw Context.reportRuntimeErrorById("msg.bad.radix", Integer.toString(base));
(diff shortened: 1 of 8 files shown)


rhino/src/main/java/org/mozilla/javascript/ScriptRuntime.java
```diff
@@ -1076,7 +1076,8 @@ public static String toString(double val) {
 
     public static String numberToString(double d, int base) {
         if ((base < 2) || (base > 36)) {
-            throw Context.reportRuntimeErrorById("msg.bad.radix", Integer.toString(base));
+            // throw Context.reportRuntimeErrorById("msg.bad.radix", Integer.toString(base));
```
rbri, line 1079: i think we can remove the throw (at lease if there is somewhere a test for that)

rbri:
yes please ....

rhino/src/main/java/org/mozilla/javascript/ScriptRuntime.java
gbrail: Looks like a miss in my editing or merging -- this should just throw a range error. Thanks!

PR #1815, opened 2025-02-02
Add missing copy to fix ArrayIndexOutOfBoundException when invoke captureContinuation

Rhython:
In the `captureContinuation` method, `stack` and `stackAttribute` are considered to be the same length. The `ensureStackLength` method does not copy `stackAttributes` with the new length, so calling `captureContinuation` after this will throw an ArrayIndexOutOfBoundException. I updated my project to use v1.8.0 and found this issue so I fixed it and create this pr.

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/Interpreter.java
+++ b/rhino/src/main/java/org/mozilla/javascript/Interpreter.java
@@ -233,6 +233,7 @@ public final class Interpreter extends Icode implements Evaluator {
             if (length > stack.length) {
                 stack = Arrays.copyOf(stack, length);
                 sDbl = Arrays.copyOf(sDbl, length);
+                stackAttributes = Arrays.copyOf(stackAttributes, length);
                 // TODO: adjust idata idata.itsMaxFrameArray & idata.itsMaxStack so they start with
                 // larger stacks next time? Not clear this is always a good idea.
             }


Rhython:
Of course, I made a simplified reproduction code, but I'm not sure how to incorporate it into the unit test of the project. 

```java
public class TestRhino2 {

    private Scriptable scope;
    private Object continuation;

    public void await() {
            throw cx.captureContinuation();
        }
    }

    public void execute(String source) {
            scope = cx.initSafeStandardObjects();
            scope.put("t", scope, this);
            cx.executeScriptWithContinuations(cx.compileString(source, "test", 1, null), scope);
        } catch (ContinuationPending e) {
            continuation = e.getContinuation();
        }
    }

    public void resume(Object obj) {
            cx.resumeContinuation(continuation, scope, obj);
        } catch (ContinuationPending e) {
            continuation = e.getContinuation();
        }
    }

        TestRhino2 test = new TestRhino2();

        String js1 = """
                let mainMenu = [];
                let bonusMenu = [];
                // add enough items to inc stack size
                for (let i = 0; i < 30; i++) {
                    mainMenu.push(['MainMenu' + i, 'MainAction' + i]);
                    bonusMenu.push(['BonusMenu' + i, 'BonusAction' + i]);
                }
                Array.prototype.push.apply(mainMenu, bonusMenu) // this will cause ensureStackLength to be called
                let select = t.await() // this will cause ArrayIndexOutOfBoundsException
                """;
        test.execute(js1);
        test.resume(0);

    }

}
```

gbrail:
Well for the most part that test code there can be turned into a Junit 5 (or 4 if you really want) test case and put into the "org.mozilla.javascript.tests" package in the "tests" directory -- you'll see lots of examples there. Those examples also use the "Utils" class, which makes it easier to ensure that you handle contexts correctly and also that you test both in interpreted and non-interpreted mode.

rbri:
merged as part of #1866, many thanks for providing this and also for the test

PR #1816, opened 2025-02-05
Make all service loading static

gbrail:
This prevents the large performance regression of loading these classes from the classpath every time they are needed, and should cover existing use cases since all these features are designed to be configured statically by deciding which Rhino modules to use.

@rbri I didn't see a big regression in standard benchmarks like the V8 benchmarks but I could imagine that in a use case with a lot of reflection that things could be much worse. This should fix it, please take a look!


rbri:
thanks, @gbrail will have a look at this during the next days...

rbri:
@gbrail have done a small test, looks like your change brings back the old performance.
+1 for merging

gbrail:
Yes, this one is a pretty simple one. Thanks for checking it out!

PR #1817, opened 2025-02-05
Load more classes lazily

gbrail:
With the switch to LambdaConstructor-based native classes, initialization time for Rhino goes up because those classes create so many properties. This PR updates our lazy loading mechanism so that there's a less-complicated way to do it without using reflection, and loads many more classes this way.


gbrail:
I am expecting that this isn't going to cause problems for most of you so I'm probably going to merge it soon.

rbri:
Fine for me...

PR #1818, opened 2025-02-05
Remove deprecated methods

rPraml:
This PR would remove some deprecated methods, that were deprecated in this PR 

https://github.com/mozilla/rhino/commit/a241104b#diff-ab1ca10bfcacde38d34848919eb3c5a3bf983f651a1a8867e9c525d7a9e4ef22 on on Feb 10, 2015

This encourages us to maintain clean and manageable code.
Since these methods have been deprecated for almost 10 years, I believe developers have had sufficient time to transition to alternative options.

I found one place, where bytecode relies on a deprecated method [here](https://github.com/mozilla/rhino/commit/f04c86c3bbad8bd6128c9c2bcfd93f0952eb5e3f)

What do you think?


rbri:
@rPraml personally i'm fine with this - i fear somebody might to mention that so far you can use newer version as replacement without recompiling the old stuff.

rbri:
at least we should get rid of calling deprecated method from our own code

gbrail:
I keep mulling this over -- how many things will we break by doing this -- and have we waited long enough for it to matter? Open to opinions and data on that!

andreabergia:
> I keep mulling this over -- how many things will we break by doing this -- and have we waited long enough for it to matter? Open to opinions and data on that!

IMHO stuff that was deprecated in 1.8.0 can safely be removed if the next release is called 1.9.0 (or 2.0.0). Deprecated things are deprecated for a reason, after all. 🙂

PR #1819, opened 2025-02-05
fix: force language in tests

### LINKED ISSUE #1803: Tests assume that error messages are always in English
Tests assumed that error messages are always in English, so when running tests in a system where system default language is not English, evaluations might fail:

```
org.junit.ComparisonFailure: `var a = 3` [MESSAGE] expected:<[missing ; after statement]> but was:<[在语句后面缺少 “;”]>
	at org.junit.Assert.assertEquals(Assert.java:117)
	at org.mozilla.javascript.tests.Bug789277Test.assertMessageEquals(Bug789277Test.java:453)
	at org.mozilla.javascript.tests.Bug789277Test.parserWithIdeErrorReporter_Reader_IdeMode(Bug789277Test.java:447)
	at java.base/java.lang.reflect.Method.invoke(Method.java:580)
```

Adding an option that forces using `en_us` lang in message formatting should be enough for fixing this i guess

andreabergia:
Fixes #1803

### DIFF
--- a/rhino/build.gradle
+++ b/rhino/build.gradle
@@ -6,6 +6,12 @@ dependencies {
     testImplementation project(':testutils')
 }
 
+test {
+    systemProperty 'user.language', 'en'
+    systemProperty 'user.country', 'US'
+    systemProperty 'user.timezone', 'America/Los_Angeles'
+}
+
 publishing {
     publications {
         rhino(MavenPublication) {


rPraml:
I would like to point out that it might be a bad idea to only run the tests in English. 

In principle, I have nothing against changing it like that, but if we change it, then developers might not find errors in local build, that would, for example, appear in production in their language setting.

For example, my machine runs on `de_DE` and time zone `Europe/Berlin` and all tests are passing.

gbrail:
I think that this is in response to an issue, but I don't recall what. I'd like to make sure we know what problem we're fixing first. In general, setting a locale in the tests so that they always run the same way is a good idea, although if we want to do that we should probably do it somewhere in the build/src directory so that it affects all the modules.

However, that would imply that we only test in English, so we should probably follow up with running the tests in CI in a few different locales.

Does anyone else have experience doing this who could maybe look in to that?

andreabergia:
> I think that this is in response to an issue, but I don't recall what. I'd like to make sure we know what problem we're fixing first. 

The issue is linked in the PR description - https://github.com/mozilla/rhino/issues/1803

> However, that would imply that we only test in English, so we should probably follow up with running the tests in CI in a few different locales.
> 
> Does anyone else have experience doing this who could maybe look in to that?

I don't think that is really scalable. There's a lot of test that check the text of an error message, which is generated via the properties files. All error messages have a description in English, but very few have a description in the other two languages for which properties file exist (French and Chinese). Duplicating those tests to check the error message in, say, French wouldn't really give us a lot more confidence in the code, IMHO.

gbrail:
I think that our best bet is to:

1) Force en/US and Pacific time in all the tests -- this should be done in "rhino.java-conventions.gradle" so that it gets picked up by all the modules
2) See if there are a few specific tests we can run with other languages to sanity-check the messages and such

I'm happy to tackle the first part but for the second I'd like someone to try who knows what kinds of errors we'd be looking for.

rPraml:
I'm fine, if we set the default language in the tests for now to en/US etc.

If I find time, I can try to create a GH-action that runs in various locales/TZ.
At least, I would try tr-TR: http://www.moserware.com/2008/02/does-your-code-pass-turkey-test.html
but I'm on holiday for the next 2 weeks (in Istanbul ;-) ...

PR #1826, opened 2025-02-12
Fix a mistake in implementing static service loading

gbrail:
A previous pull request introduced "RegExpLoader" as a new service that implements regular expressions. However, we neglec ted to update the module-info file for the "rhino" module, leading to a module configuration that cannot load.

https://github.com/mozilla/rhino/pull/1816

### DIFF
--- a/rhino/src/main/java/module-info.java
+++ b/rhino/src/main/java/module-info.java
@@ -14,7 +14,7 @@ module org.mozilla.rhino {
     exports org.mozilla.javascript.xml;
     exports org.mozilla.javascript.config;
 
-    uses org.mozilla.javascript.RegExpProxy;
+    uses org.mozilla.javascript.RegExpLoader;
     uses org.mozilla.javascript.xml.XMLLoader;
     uses org.mozilla.javascript.config.RhinoPropertiesLoader;
 


rPraml:
JFYI, there are tools like https://github.com/google/auto/tree/main/service that may generate the necessary meta-inf files, but as far as I see, it does not support modifying the module-info file.

andreabergia:
This looks correct, but why wasn't this caught by any unit test? Is it possible to add any that would catch it?

gbrail:
In this case nothing was caught until I tried to launch the shell in
IntelliJ. I feel like there must be a test framework that can let us
properly evaluate our module configurations, but I don't know what it is
yet.

rPraml:
> I feel like there must be a test framework that can let us
properly evaluate our module configurations, but I don't know what it is
yet.

Using module-info in your tests (at least with maven) has a steep learning curve...
(It took a long time for us, until everything was configured properly. We had issues with some junit-libraries etc)

The next problem is, you can decide per maven/gradle project, if you want to use "module-classpath" or "classic-classpath"

So you cannot put a test in "rhino/test" that checks if service-loading via module-info will work, when the rest of the classes will use META-INF/services

It might be possible, to modify the build script and run the build twice with some option like `-DmyTestsShouldUse=module-path`

PR #1827, opened 2025-02-12
Make `NativeGlobal` not an `IdFunctionCall`

andreabergia:
This makes `NativeGlobal` not implement `IdFunctionCall` and turns all the method into `LambdaFunction` or `LambdaConstructor`.

There's a few complicated details for handling the built-in errors, but it all seems to work and fix a few small test262 cases.


gbrail:
Thanks -- FWIW I'm working on NativeString, and at one point worked on
NativeDate, so let me know before you tackle other things.

I'm also open to retiring some of the MozillaSuiteTest if it's no longer in
spec.

rhino/src/main/java/org/mozilla/javascript/NativeGlobal.java
```diff
@@ -131,89 +179,90 @@ public static void init(Context cx, Scriptable scope, boolean sealed) {
...
-                case Id_isXMLName:
-                    {
-                        Object name = (args.length == 0) ? Undefined.instance : args[0];
-                        XMLLib xmlLib = XMLLib.extractFromScope(scope);
-                        return ScriptRuntime.wrapBoolean(xmlLib.isXMLName(cx, name));
-                    }
+    private static Boolean js_isXMLName(Context cx, Scriptable scope, Object[] args) {
+        Object name = (args.length == 0) ? Undefined.instance : args[0];
+        XMLLib xmlLib = XMLLib.extractFromScope(scope);
+        return ScriptRuntime.wrapBoolean(xmlLib.isXMLName(cx, name));
```
gbrail, line 226: This is minor, but I'm noticing that we don't really need "wrapBoolean" or even "wrapInt" any more, since auto-boxing just does that. If you agree we could take it out of a few places and be even cleaner.

gbrail:
This looks great, and your "style" of doing this is nicer than what I was doing so I may make some changes.

I have one minor nit as to whether we actually need "wrapBoolean" any more. I'm open to opinions either way, but I think that we can stop using it.

rhino/src/main/java/org/mozilla/javascript/NativeGlobal.java
```diff
@@ -131,89 +179,90 @@ public static void init(Context cx, Scriptable scope, boolean sealed) {
...
+    }
 
-                case Id_parseFloat:
-                    return js_parseFloat(args);
+    private static Boolean js_isNaN(Object[] args) {
+        // The global method isNaN, as per ECMA-262 15.1.2.6.
+        boolean result;
+        if (args.length < 1) {
```
rbri, line 232: Why not using the same coding style as below and directly return here?

    return Boolean.TRUE;

rbri:
Cool....

rhino/src/main/java/org/mozilla/javascript/NativeGlobal.java
andreabergia: Sure thing! I didn't really look at those methods honestly, just refactored them out of the `IdScriptableObject`'s `switch` as they were. 🙂

andreabergia:
Rebased on top of `master and removed all various `wrap` calls.

gbrail:
Thanks!

PR #1828, opened 2025-02-13
NativeRegExp: Fix the behaviour of sticky flag

balajirrao:
The behaviour of NativeRegExp's exec with the sticky flag on was broken in cases where the regular expression does not start with a literal.

This fix doesn't fix any previously failing test262 tests. It seems that this test [test/built-ins/RegExp/prototype/exec/y-fail-lastindex.js](https://github.com/tc39/test262/blob/main/test/built-ins/RegExp/prototype/exec/y-fail-lastindex.js) is the one expected to cover the feature. However this test has always passed since the behaviour of sticky in these "simple" cases are handled as special case.

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/regexp/NativeRegExp.java
+++ b/rhino/src/main/java/org/mozilla/javascript/regexp/NativeRegExp.java
@@ -1885,7 +1885,11 @@ public class NativeRegExp extends IdScriptableObject {
             boolean anchor = false;
             while (gData.cp <= end) {
                 int match = simpleMatch(gData, input, op, program, pc, end, true);
-                if (match >= 0) {
+                if (match < 0) {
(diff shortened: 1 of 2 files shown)


gbrail:
Looks good to me. Does this really not fix any test262 tests that were failing before? It seems like there should be some coverage for this feature in a standard test suite that we can use too.

balajirrao:
@gbrail thanks for taking a look. I realised the PR description needed some work - I've improved it now.

It seems to me that Test262 isn’t meant to be exhaustive, and that the [test](https://github.com/tc39/test262/blob/main/test/built-ins/RegExp/prototype/exec/y-fail-lastindex.js) is intended to cover the feature (though the test itself could be improved with better assertions). I’m still getting familiar with how this works, so I’d really appreciate any insights you might have.

gbrail:
This also looks good, and thanks again. And this will probably require that you rebase your other PR!

PR #1830, opened 2025-02-14
Enable assigning a Number to java.lang.Character

focJSi:
Currently, assigning a JS number to a primitive Java `char` is possible, but assigning it to a `java.lang.Character` results in an `EvaluatorException`, which is rather unintuitive.

To me it seemed like there might just be a case missing in the `coerceType` implementation, handling the `java.lang.Character` similar to the primitive one.

Let me know what you think.


gbrail:
Makes sense to me and it works. Thanks!

PR #1832, opened 2025-02-15
Modularize some tests and fix module specifications

gbrail:
Some of our module specifications haven't been right, and these weren't caught in tests because the tests weren't themselves modularized. This PR addresses some of the problems by modularizing the tests for rhino-engine, rhino-tools, and rhino-xml. Here's the status of the rest:

* A bunch of tests in the main "rhino" module live in the org.mozilla.rhino package, which is already exported by the main package. We can't modularize the tests unless we move those tests to a different package, and that means making more classes "public", or re-packaging in some other creative way
* I failed to modularize rhino-kotlin -- in module mode, the tests couldn't find their associated Kotlin classes. I don't know why not.
* I failed to modularize "tests" -- the various "getResourceAsStream" methods work differently once modules are introduced, and despite lots of experiments and research I couldn't figure it out. This might even be a bug in Gradle, but regardless it was difficult.

This, at least, will ensure that we will catch most of the problems with module specifications for the main modules.

### DIFF
--- a/buildSrc/src/main/groovy/rhino.java-conventions.gradle
+++ b/buildSrc/src/main/groovy/rhino.java-conventions.gradle
@@ -23,8 +23,10 @@ dependencies {
     testImplementation "junit:junit:4.13.2"
     testRuntimeOnly "org.junit.vintage:junit-vintage-engine"
     testRuntimeOnly "org.junit.platform:junit-platform-launcher"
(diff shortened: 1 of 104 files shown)


gbrail:
This starts to get us in a much better place in terms of modularity so i'm going to merge it -- it should not affect too many people except that it fixes some small bugs in our module specs!

PR #1833, opened 2025-02-16
Deliberately set language, locale, and TZ in all build phases

gbrail:
This is (IMO) a more generic implementation of:
 
https://github.com/mozilla/rhino/pull/1819

This should result in hard-coding the language in every build phase regardless of the system where it's running.

I'd be interested to see from those of you affected if this fixes the problem that PR was trying to address.


rbri:
if fear this will pop up from time to time, maybe it is worth to add some comments to the code...

andreabergia:
Looks like a better approach than mine, thanks for doing this!

PR #1836, opened 2025-02-18
Fix RegExp.prototype[Symbol.match] and String.prototype.match

balajirrao:
This PR includes two fixes based on the spec:

1. `RegExp.prototype[Symbol.match]`: The spec requires that this method invoke the `exec` method of the regular expression in a loop.
2. `String.prototype.match`: The spec mandates that this method call `RegExp.prototype[Symbol.match]`.

As a result, a few new test262 tests now pass. The remaining failing tests in `built-ins/RegExp/prototype/Symbol.match` and `built-ins/String/prototype/match` are unrelated to this fix.

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/NativeString.java
+++ b/rhino/src/main/java/org/mozilla/javascript/NativeString.java
@@ -624,8 +624,6 @@ final class NativeString extends IdScriptableObject {
                         return ScriptRuntime.wrapBoolean(
                                 (id == Id_equals) ? s1.equals(s2) : s1.equalsIgnoreCase(s2));
                     }
-
-                case Id_match:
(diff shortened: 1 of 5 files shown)


gbrail:
This is great -- thanks! I looked at the test coverage and it looks good, and I see how many things it fixes. I'd appreciate it if you could remove that unused method (there's a warning from Errorprone about it in the build output). Thanks!

balajirrao:
@gbrail thanks for taking a look! Will fix, rebase and push once CI is back to green on master.

gbrail:
This looks very good -- thanks!

p.s. I'm about to finish reorganizing NativeString completely into lambda functions so this comes just in time

PR #1838, opened 2025-02-19
Fixed a problem with `super` and nested lambdas

andreabergia:
We discovered this problem in our fork, where we disable the "function peeling optimization", but we have managed to reproduce it with a unit test that mixes compiled and interpreted functions.

The problem is that nested lambda does not have the `homeObject` set correctly. In the interpreter we pick it up from the containing `NativeCall` (the activation), but in some situations, like without function peeling or when going from a compiled function to an interpreted one, it might not be set.

The fix we came up with is just to propagate the home object on the nested closure function, that gets wrapped in the `ArrowFunction`. It's not particularly elegant, but it seems to fix the problem, does not cause any regression, and it's pretty cheap at runtime.


gbrail:
Looks good to me and simple enough. Thanks!

PR #1843, opened 2025-02-24
RegExp: Implement lookbehind

balajirrao:
Implement lookbehind for regular expressions. All test262 tests for lookbehind pass.

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/regexp/NativeRegExp.java
+++ b/rhino/src/main/java/org/mozilla/javascript/regexp/NativeRegExp.java
@@ -289,6 +289,231 @@ public class NativeRegExp extends IdScriptableObject {
         return rval;
     }
 
+    private static void prettyPrintRE(RECompiled regexp) {
+        for (int pc = 0; regexp.program[pc] != REOP_END; ) {
+            System.out.print(pc + ": ");
+            byte op = regexp.program[pc];
+            pc++; // Increment pc after reading op
+            switch (op) {
+                case REOP_EMPTY:
+                    System.out.println("EMPTY");
+                    break;
+                case REOP_BOL:
+                    System.out.println("BOL");
+                    break;
+                case REOP_EOL:
+                    System.out.println("EOL");
+                    break;
+                case REOP_WBDRY:
+                    System.out.println("WBDRY");
+                    break;
+                case REOP_WNONBDRY:
+                    System.out.println("WNONBDRY");
+                    break;
+                case REOP_DOT:
+                    System.out.println("DOT");
+                    break;
+                case REOP_DIGIT:
+                    System.out.println("DIGIT");
+                    break;
+                case REOP_NONDIGIT:
+                    System.out.println("NONDIGIT");
+                    break;
+                case REOP_ALNUM:
+                    System.out.println("ALNUM");
+                    break;
+                case REOP_NONALNUM:
+                    System.out.println("NONALNUM");
+                    break;
+                case REOP_SPACE:
+                    System.out.println("SPACE");
+                    break;
+                case REOP_NONSPACE:
+                    System.out.println("NONSPACE");
+                    break;
+                case REOP_BACKREF:
+                    int backrefIndex = getIndex(regexp.program, pc);
+                    System.out.println("BACKREF " + backrefIndex);
+                    pc += INDEX_LEN;
+                    break;
+                case REOP_FLAT:
+                    int flatIndex = getIndex(regexp.program, pc);
+                    int flatLength = getIndex(regexp.program, pc + INDEX_LEN);
+                    System.out.print("FLAT: ");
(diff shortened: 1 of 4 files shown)


rbri:
Fantastic !!!!

gbrail:
Thanks for taking care of this! I'm away for another week but I will look at it in detail when I get back, or otherwise there are a few others who can merge this if it is holding up something else.

balajirrao:
@gbrail thanks for the heads up! No rush - it's good enough if you take a look at it after you get back.

rhino/src/main/java/org/mozilla/javascript/regexp/NativeRegExp.java
```diff
@@ -289,6 +296,246 @@ Object execSub(Context cx, Scriptable scopeObj, Object[] args, int matchType) {
         return rval;
     }
 
+    private static void prettyPrintRE(RECompiled regexp) {
```
gbrail, line 299: It looks like we don't have tests that exercise this. Is there a way we can put something in, even as a "smoke test" just to make sure that we call it and it doesn't bomb out immediately?

rhino/src/main/java/org/mozilla/javascript/regexp/NativeRegExp.java
```diff
@@ -780,6 +1029,26 @@ private static int getDecimalValue(char c, CompilerState state, String overflowM
         return value;
     }
 
+    private static RENode reverseNodeList(RENode head) {
```
gbrail, line 1032: I'm just glad that someone got to write code to "reverse a linked list" in real-world code! ;-)

rhino/src/main/java/org/mozilla/javascript/regexp/NativeRegExp.java
```diff
@@ -1359,6 +1656,22 @@ private static boolean flatNMatcher(
...
+        // in the input, start from cp - 1 and go back length chars
+        // in the regex source, do it the other way
+        for (int i = 1; i <= length; i++) {
+            if (gData.regexp.source[matchChars + length - i] != input.charAt(gData.cp - i)) {
+                return false;
+            }
+        }
+
+        gData.cp -= length;
+        return true;
```
gbrail, line 1672: It doesn't look like we ever reach this line in tests

rhino/src/main/java/org/mozilla/javascript/regexp/NativeRegExp.java
```diff
@@ -1374,6 +1687,24 @@ private static boolean flatNIMatcher(
         return true;
     }
 
+    private static boolean flatNIMatcherBackward(
```
gbrail, line 1690: It doesn't look like this function gets called in tests

gbrail (commented):
This looks great and I can see that people are excited. The only thing that I can suggest is that you run "gradle testCodeCoverageReport" and take a look at the output, which is in "tests/reports/jacoco/...." I can see a few parts of your new code that aren't covered by tests, and since you otherwise did a very thorough job I'd love to see that we could cover those new bits with some tests as well. Thanks!

rhino/src/main/java/org/mozilla/javascript/regexp/NativeRegExp.java
balajirrao: Good point! Only way I could think of was to use reflection to call the pretty printer in a test. I've added such a test.

rhino/src/main/java/org/mozilla/javascript/regexp/NativeRegExp.java
balajirrao: This is a good find. I have memory of debugging this code with a test262 test so I was not sure why this is happening. Long story short, I now know why this isn't reached.

I've "fixed" it by moving the FLAT concatenation code from[ emitBytecode stage](https://github.com/mozilla/rhino/pull/1843/commits/4e7d0f4e6345764368234e311f02b4ac719d7d52#diff-26b8d191be9a315419e6621d92ee9a0c274c8bd7f408a2d7220bce18c80b97a7L1448) to the [parsing stage](https://github.com/mozilla/rhino/pull/1843/commits/4e7d0f4e6345764368234e311f02b4ac719d7d52#diff-26b8d191be9a315419e6621d92ee9a0c274c8bd7f408a2d7220bce18c80b97a7R792). This also makes lookbehinds a bit more optimal.

rhino/src/main/java/org/mozilla/javascript/regexp/NativeRegExp.java
balajirrao: Ah, these are missing tests indeed. Added them.

balajirrao:
@gbrail thank you for the review. I've addressed your comments and the coverage of the new code now looks good.

gbrail:
Looks good now -- thanks!

PR #1845, opened 2025-03-04
RegExp: Fix bug with quantified capture groups

balajirrao:
Fixes an issue where capture groups in quantified expressions (min ≥ 2) were not cleared between iterations. For example, in `/(?:(\2)(\d)){2}/`, during the second iteration, `\2` incorrectly retained the first iteration's value instead of being reset.

Before:
```
js>  /(?:(\2)(\d)){2}/.exec("112")
112,1,2
```

After:
```
js>  /(?:(\2)(\d)){2}/.exec("112")
11,,1
```


gbrail:
Looks like a simple enough and reasonable test. can you please rebase?

gbrail:
Thanks!

PR #1850, opened 2025-03-14
Convert `NativeArray` from `IdScriptableObject` to lambda constructor.

aardvark179:
There's 3 big parts to this, separated by commit:
1. Conversion of the length property away from the old id method. This uses a new `BuiltInSlot` type which holds on to the `NativeArray` object rather than a `LambdaSlot` because the array length property is writable, and has slightly idiosyncratic behaviour. Rather than creating multiple new closures for each array created I thought it was better to factor things so that the slot would be smaller and more lightweight. Given we know the slot's raw value should never be used we could even reuse that field to hold the native array object.
2. Main conversion of `NativeArray` over to lambda construction. This breaks a small set of legacy tests in `MozillaTestSuite` as they expect incorrect attributes of `length` properties on the array methods.
3. Some refactoring and tidy up of array methods which are specified as using [https://tc39.es/ecma262/multipage/ordinary-and-exotic-objects-behaviours.html#sec-arrayspeciescreate](ArraySpeciesCreate) and some work to handle odd corner cases with overridden index properties.

I'll try and run full benchmark comparisons to ensure nothing has regressed significantly, and I'm happy to split this PR into the three major parts if that helps with reviewing.

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/BuiltInSlot.java
+++ b/rhino/src/main/java/org/mozilla/javascript/BuiltInSlot.java
@@ -1,0 +1,108 @@
+package org.mozilla.javascript;
+
+import java.io.Serializable;
+
+public class BuiltInSlot<T extends ScriptableObject> extends Slot {
+
+    public interface Getter<U extends ScriptableObject> extends Serializable {
+        Object apply(U builtIn, Scriptable start);
+    }
+
+    public interface Setter<U extends ScriptableObject> extends Serializable {
+        boolean apply(U builtIn, Object value, Scriptable owner, Scriptable start, boolean isThrow);
+    }
+
+    public interface AttributeSetter<U extends ScriptableObject> extends Serializable {
+        void apply(U builtIn, int attributes);
+    }
+
+    public interface PropDescriptionSetter<U extends ScriptableObject> extends Serializable {
+        boolean apply(
+                U builtIn,
+                BuiltInSlot<U> current,
+                Object id,
+                ScriptableObject desc,
+                boolean checkValid,
+                Object key,
+                int index);
+    }
+
+    private final T builtIn;
+    private final Getter<T> getter;
+    private final Setter<T> setter;
+    private final AttributeSetter<T> attrUpdater;
+    private final PropDescriptionSetter<T> propDescSetter;
+
+    BuiltInSlot(
+            Object name,
+            int index,
+            int attr,
+            T builtIn,
+            Getter<T> getter,
+            Setter<T> setter,
+            AttributeSetter<T> attrUpdater) {
+        this(
+                name,
+                index,
+                attr,
+                builtIn,
+                getter,
+                setter,
+                attrUpdater,
+                BuiltInSlot::defaultPropDescSetter);
+    }
+
+    BuiltInSlot(
+            Object name,
+            int index,
+            int attr,
+            T builtIn,
+            Getter<T> getter,
+            Setter<T> setter,
+            AttributeSetter<T> attrUpdater,
+            PropDescriptionSetter<T> propDescSetter) {
+        super(name, index, attr);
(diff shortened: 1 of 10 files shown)


rhino/src/main/java/org/mozilla/javascript/BuiltInSlot.java
```diff
@@ -0,0 +1,108 @@
+package org.mozilla.javascript;
+
+import java.io.Serializable;
+
+public class BuiltInSlot<T extends ScriptableObject> extends Slot {
```
rbri, line 5: As we get more an more such special classes (which is good) i think we have to try to make at least the intention as clear as possible. Therefore l like to vote for some class comment that helps in the future to decide if we can use this approach in other situations.

rbri:
Wow - 100 test passed 🚀

rhino/src/main/java/org/mozilla/javascript/BuiltInSlot.java
aardvark179: Yeah, that’s definitely something I should clean up.

gbrail:
I think that this is mostly awesome -- I want to try and understand the new slot type, though, since we already have a few other different ways to create properties that are served by lambda functions and perhaps we need to consolidate a bit -- I think that with this new one we'll have three. Is the main advantage that this one creates fewer objects?

tests/testsrc/tests/ecma/Array/15.4.4.3-1.js
```diff
@@ -31,8 +31,6 @@ writeHeaderToLog( SECTION + " Array.prototype.join()");
 var ARR_PROTOTYPE = Array.prototype;
 
 new TestCase( SECTION, "Array.prototype.join.length",           1,      Array.prototype.join.length );
-new TestCase( SECTION, "delete Array.prototype.join.length",    false,  delete Array.prototype.join.length );
```
rbri, line 34: @gbrail has use a different approach for dealing with this in his commit
image

aardvark179: Yeah, I will update this.

aardvark179:
One error I found was that I forgot to implement `copySlot` and this wasn’t caught by a test. I’ll hav w a think about how we can try and catch that in a more reliable way.

aardvark179:
I’ve done some benchmark comparisons and they mostly look good, but there are a couple of significant regression which I will need to investigate before taking this out of draft status.

aardvark179:
I've done some digging. The benchmark that is regressing is *extremely* sensitive to inlining, and has pretty much nothing to do with arrays, it's a slightly different set of objects going through `ScriptableObject.bind` and because of that it's inlining a few more methods in that compilation, and because of *that* it's not inlining as many other property access methods. I'll have a think about this and try to work out if it is a problem that will go away as we move things to lambdas, or if we'll need to do other optimisations to resolve this.

aardvark179:
Ah, I think I’ve found the problem. Unscopables were being created more eagerly and that threw off the profile for `getBase`. I have a rough fix, but I’ll need to do a bit more testing over the weekend.

tests/testsrc/tests/ecma/Array/15.4.4.3-1.js
aardvark179: Done.

rbri:
@aardvark179 - spotless.... :-D

Run './gradlew :rhino:spotlessApply' to fix these violations.

aardvark179:
Okay, we still have a few minor regressions, which mostly seem to be due to some callosities seeing a a mixture of `ScriptableObject` and `IdScriptableObject`, these have reduced with rebasing on top of the `NativeString` change, and I expect to reduce as the remaining classes are tackled.

gbrail:
This is great progress -- thanks!

PR #1851, opened 2025-03-15
use always the same ThrowTypeError object for Arguments.caller and Arguments.callee

rbri:
While having a first look at https://github.com/HtmlUnit/htmlunit/issues/943 i was suprised to see we are always creating 4 new BaseFunction objects for every Arguments object. From the spec there is only one ThrowTypeError per realm...

This fix does not support the single ThrowTypeError per realm part of the spec, it simply uses one global object for all realms. Therefor you will not see any improvements in test262 numbers (we more or less switched one for another). But i think the reduction in object creation is worth to consider this.

What do you think?
Any good idea for having real-global objects?


p-bakker:
As far as I know, except for Symbol, for which there ought to be a global, cross-realm registry, no other objects ought to be shared cross-realm

aardvark179:
If the aim here is spec compliance then the setter and getter of bother caller and caller should all be the same `%ThrowTypeError%`, but this should be the same for each realm. I don’t think we have a great model of realms at the moment, so I’m not sure how much we should care. Maybe the thing to do here is to only create those setters and getters in the very rare case of them being reified by getting a property descriptor. The act of getting one of these properties doesn’t have to create or use a reified JS function, it just has to throw the error as if one had been called.

rbri:
My aim was more to get rid of all the ctor calls. And yes they should be all the same but then we loos the different error messages for the two cases. Therefore i stayed with two instances.

> I don’t think we have a great model of realms at the moment, so I’m not sure how much we should care.
That was my impression also

gbrail:
Just looked at this in a superficial way so far, but I'm not sure about this one. I'm in favor of reducing the number of objects overall, but I'm very wary of making objects that find their way into scripts "static final" because people do indeed use Rhino in a  "multi-realm" way even though we don't have a formal way to support the "realm" concept in JavaScript. (Which I'm quite sure is younger than Rhino.) 

Sometimes, now that we have lambdas, we can solve problems like this by creating objects in the "init" function of a native object and attacking it to various lambdas by capture. Is something like that possible? Haven't looked closely enough yet to see.

PR #1852, opened 2025-03-15
Convert NativeString to use lambdas

gbrail:
In line with other recent changes, use lambdas to support NativeString.


andreabergia (approved):

tests/testsrc/tests/ecma/String/15.5.4.8-1.js
```diff
@@ -40,8 +40,9 @@ var TITLE   = "String.prototype.split";
 writeHeaderToLog( SECTION + " "+ TITLE);
 
 new TestCase( SECTION,  "String.prototype.split.length",        2,          String.prototype.split.length );
-new TestCase( SECTION,  "delete String.prototype.split.length", false,      delete String.prototype.split.length );
-new TestCase( SECTION,  "delete String.prototype.split.length; String.prototype.split.length", 2,      eval("delete String.prototype.split.length; String.prototype.split.length") );
+// These are no longer valid ECMAScript tests
```
rbri, line 43: i guess we break backward compatibility here, do we need some more words?

aardvark179: We are breaking compatibility, but it’s a real corner case. We can add words, or we could add a special case for ES5, but my guess is that this didn’t break anything significant and that’s why the behaviour was dropped, but I don’t think the spec history goes back far enough to be sure.

gbrail: That's pretty much how I feel -- this is such a minor thing that I feel that we can accept the risk. I think that the PR for NativeArray has a similar kind of change in it.

PR #1853, opened 2025-03-18
Type Consolidator & Generic support for Map/List/Method/Constructor

ZZZank:
# Java Generic support

3 PRs in total:
- TypeInfo & make `MemberBox` use TypeInfo (#1849 )
- Type Consolidation & Generic support for Method/Constructor (This PR)

## PR#2: Type Consolidator

**This PR includes changes from #1849**

Before #1849 is reviewed, this PR will keep being a draft.

A description for this PR is not yet finished, you can see descriptions in #1849 for how type consolidation works, and newly added tests for what this PR brings


ZZZank:
I think I found an inconsistency between test and implementation. The method `WrapFactory#wrap(...)` requires a param named `staticType`, that was described as:

```
     * @param staticType type hint. If security restrictions prevent to wrap object based on its
     *     class, staticType will be used instead.
```

So, in `org.mozilla.javascript.tests.SecurityControllerTest`, when there is a security restriction present, accessing members from implementation should cause an error, but the test itself claimed:
```
        // f.create produces "SomeClass extends ArrayList<String> implements
        // SomeInterface"
        // we may access array methods, like 'size' defined by ArrayList,
        // but not methods like 'bar' defined by SomeClass, because it is in a restricted package
```

The return type of `f.create()` is `SomeInterface`, so when wrapping object, the `staticType` is `SomeInterface`. In this case, since `SomeInterface` did not define the `.size()` method, the method should not be considered as accessible. That is, line92:

https://github.com/mozilla/rhino/blob/f8ce76a38b172521a9e49050ebfb83de0bb5e073/tests/src/test/java/org/mozilla/javascript/tests/SecurityControllerTest.java#L90-L93

should cause an error, when access is restricted.

It was not throwing exception before because `NativeJavaList` was always using the list class itself as `staticType`. But now it breaks because I actually passed a `staticType` from WrapFactory to NativeJavaList

rPraml:
@gbrail I would like to 'ping' you about feedback of this PR and how to proceed:
My summary:
- I do understand (at least I hope), what this PR does
- I did a review and this makes technically sense
- I am not happy about the package structure (First: it introduces cycles, Second: naming)

For the cycles, I made an experimental branch, where I tried to break it: https://github.com/FOCONIS/rhino/tree/type-consolidator-2
I can make a PR for easier review, but credits should go to @ZZZank  ;)

ZZZank:
Closing this PR since features in this PR has been implemented by several other PRs, for example #2080 and #2023

PR #1858, opened 2025-03-24
NativeRegExp: turn 'debug' into a rhino property

balajirrao:


gbrail:
Thanks for opening this. We already have one property called "rhino.debugLinker". Would you consider calling this one "rhino.debugRegexp" so that we have consistency?

gbrail:
So, this is going to make the environment variable "DEBUG_REG_EXP", and I think that we want "DEBUG_REGEXP", which would mean changing the name of the property to "debugRegexp". If you have patience for one more change I'd appreciate it, otherwise I can do it! Thanks.

balajirrao:
@gbrail ah I see, I had noticed you wrote "Regexp" - now I know why. I'll happily make the change, no problem :) Thanks for your patience.

gbrail:
Thanks for following through on this!

PR #1860, opened 2025-03-25
Fix `this` in `apply`/`call`

andreabergia:
When `this` is missing, or null, or undefined when invoking a function via `apply` or `call`, it should be replaced with the global object only in non-strict mode, according to the spec.

Note that Rhino has a feature flag `FEATURE_OLD_UNDEF_NULL_THIS` that replaces a missing `this` with the global object always, for any function. This is enabled by default in older versions of the engine (<= 1_7).

The previous implementation was not doing the correct thing, for a couple of reasons:
- it did not account for strict and non-strict functions; it had the same behavior for all functions;
- it had a different behavior in interpreter and compiled mode due to the function peeling optimization for the interpreter of #1510, where a new, interpreter-only implementation was done.

This implementation fixes the problems above. I have had to add a new flag on functions to decide whether a function was strict or not, and to implement it in the code generation for compiled classes.
This fixes a number of test262 cases.

Spec ref: https://tc39.es/ecma262/#sec-function.prototype.apply

*Note:* this implementation removes the ad-hoc implementation for interpreter, so that compiled and interpreted mode work in the same way. This has the side effect of breaking continuations in the interpreter for `apply`/`call` though - I have had to remove some tests that were not working anymore. I am not sure how big a problem this is; it's possible to keep the duplicated code and fix it in two places if we prefer to keep that functionality.

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/InterpretedFunction.java
+++ b/rhino/src/main/java/org/mozilla/javascript/InterpretedFunction.java
@@ -169,4 +169,9 @@ final class InterpretedFunction extends NativeFunction implements Script {
     public boolean hasDefaultParameters() {
         return idata.argsHasDefaults;
     }
+
+    @Override
+    public boolean isStrict() {
+        return idata.isStrict;
+    }
 }
--- a/rhino/src/main/java/org/mozilla/javascript/InterpretedFunction.java
+++ b/rhino/src/main/java/org/mozilla/javascript/InterpretedFunction.java
@@ -82,7 +82,7 @@ final class InterpretedFunction extends NativeFunction implements Script {
     @Override
     public Object call(Context cx, Scriptable scope, Scriptable thisObj, Object[] args) {
         if (!ScriptRuntime.hasTopCall(cx)) {
-            return ScriptRuntime.doTopCall(this, cx, scope, thisObj, args, idata.isStrict);
+            return ScriptRuntime.doTopCall(this, cx, scope, thisObj, args, isStrict());
         }
         return Interpreter.interpret(this, cx, scope, thisObj, args);
     }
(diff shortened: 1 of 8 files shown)


rhino/src/main/java/org/mozilla/javascript/InterpretedFunction.java
```diff
@@ -169,4 +169,9 @@ boolean hasFunctionNamed(String name) {
     public boolean hasDefaultParameters() {
         return idata.argsHasDefaults;
     }
+
+    @Override
+    public boolean isStrict() {
```
rbri, line 174: Shold we use isStrict() instead of idata.isStrict at other places also?

andreabergia: Good call, I'll update that.

rbri:
test262.properties proves that this is another major step forward ❤️ 🚀 🧙‍♂️

rhino/src/main/java/org/mozilla/javascript/InterpretedFunction.java
andreabergia: Done, thanks for the suggestion @rbri

p-bakker:
Have a look at #1444 wrt to continuations and handing of .call/apply.

I would think breaking those isn't acceptable, as people seem to rely of that working

p-bakker:
But kudo's on fixing this!

Any change of next supporting non-wrapped primitives as `this`? 🤞

gbrail:
Thanks -- I'm also encouraged by this but concerned about continuations. People absolutely use these a lot, although I suspect it's mostly legacy code. Can we find a way to keep that working?

andreabergia:
I think I've managed to restore the continuations in the latest commit, and to clean up the code a bit - there's now only _one_ implementation in `ScriptRuntime`. 🙂

PR #1861, opened 2025-03-25
Fix: `strict` flag is not set correctly at compile time

andreabergia:
If a function does not have the directive `use strict`, but it is contained in a scope which is strict, we should propagate the strict flag. However, the parser was incorrectly resetting it.

Fixes various test262 cases.

Unfortunately, fixing this broke the `Function` constructor (which is akin to `eval`). I've done a bit of a hacky fix - it works, but I'm not particularly proud of it.


andreabergia:
Rebased

rbri:
@andreabergia 
> but I'm not particularly proud of it.

We all have this kind of stuff somewhere and praying that no one will ever find it :-D

gbrail:
Thanks -- I get what you did and I think it's OK. Thanks!

PR #1863, opened 2025-03-27
NativeRegExp: Fix bugs in character class parsing

balajirrao:
This fixes two regressions introduced by #1847.

- Empty character class parsing (Thanks to @rbri for identifying the bug and providing test cases)
- Handling character class escapes in character classes


rbri:
@balajirrao looks great, at least it fixes all may failing tests and also another regression i found today. But this time with a huge js libs - so far i was not able to extract a separate test for this.

@gbrail please.....

PR #1866, opened 2025-03-29
made a test case out of the info in https://github.com/mozilla/rhino/pull/1815

rbri:
also the fix was added and it solves the problem

Because there was no activity and i think we have to process as many PR's as possible i spend some time on PR#1815 and wrote a test case.

### DIFF
--- a/rhino/src/test/java/org/mozilla/javascript/tests/ContinuationsApiTest.java
+++ b/rhino/src/test/java/org/mozilla/javascript/tests/ContinuationsApiTest.java
@@ -35,7 +35,6 @@ import org.mozilla.javascript.serialize.ScriptableOutputStream;
  * @author Norris Boyd
  */
 public class ContinuationsApiTest {
-
     Scriptable globalScope;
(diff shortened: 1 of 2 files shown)


rhino/src/test/java/org/mozilla/javascript/tests/ContinuationsApiTest.java
```diff
@@ -381,4 +386,29 @@ public void consStringSerialization() throws IOException, ClassNotFoundException
 
         assertEquals("still the same at the other end", r1.toString(), r2.toString());
     }
+
+    @Test
+    public void abcd() {
```
andreabergia, line 391: `abcd`? 🙂

rbri: Done, many thanks for the heads up

rbri:
@gbrail please squash

andreabergia (approved):

PR #1868, opened 2025-04-02
Use JVM default file encoding as fallback

balajirrao:
When the Shell is run with a file with `.js` extension, the content-type is determined to be `text/javascript`, which leads to the 8859_1 encoding to be used. AFAICT this is obsolete now. Change it to use the JVM default encoding.

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/commonjs/module/provider/UrlModuleSourceProvider.java
+++ b/rhino/src/main/java/org/mozilla/javascript/commonjs/module/provider/UrlModuleSourceProvider.java
@@ -14,6 +14,7 @@ import java.net.URI;
 import java.net.URISyntaxException;
 import java.net.URL;
 import java.net.URLConnection;
+import java.nio.charset.Charset;
 import java.util.Iterator;
 import java.util.List;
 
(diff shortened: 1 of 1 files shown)


aardvark179:
I'm not sure about this. Prior to OpenJDK 18 the default character set is not necessarily a great choice. It might be worth picking UTF-8 explicitly for consistency. Have a look at [JEP 400](https://openjdk.org/jeps/400), especially:

> In JDK 17 and earlier, the default charset is determined when the Java runtime starts. On macOS, it is UTF-8 except in the POSIX C locale. On other operating systems, it depends upon the user's locale and the default encoding, e.g., on Windows, it is a codepage-based charset such as windows-1252 or windows-31j.

balajirrao:
> I'm not sure about this. Prior to OpenJDK 18 the default character set is not necessarily a great choice. It might be worth picking UTF-8 explicitly for consistency. Have a look at [JEP 400](https://openjdk.org/jeps/400), especially:
> 
@aardvark179 Yeah, I wasn't too sure about it either. Your suggestion sounds reasonable; prior to OpenJDK 18, it wasn't really configurable either (according to JEP 400).

andreabergia:
I agree on picking UTF-8.

gbrail:
I too recall past trauma with default character sets and think we should just pick UTF-8.

rbri:
@balajirrao is this still on your radar?

PR #1869, opened 2025-04-02
NativeRegExp: Fix handling of \x and \u

balajirrao:
Handle \x and \u properly when they appear at the very end of a regexp. They should be treated as literals in this case.


gbrail:
Thanks, this looks good to me!

PR #1870, opened 2025-04-03
Reimplement `String.prototype.search`, `replace`, `replaceAll`, `split`

andreabergia:
This PR adds implementation for:

- `RegExp.prototype[Symbol.search]`
- `RegExp.prototype[Symbol.replace]`
- `RegExp.prototype[Symbol.split]`

and rewrites the implementation of:

- `String.prototype.search`
- `String.prototype.replace`
- `String.prototype.replaceAll`
- `String.prototype.split`

following the spec, in order to rely on the `RegExp` implementation when the spec says so.

A little more than a hundred test262 cases now pass. The few apparent regressions are related to things that are known to be broken and were already failing for other tests, such as rhino doing too much auto-boxing or the `RegExp.prototype.flags` implementation not being correct according to the spec. I am happy to discuss any such case in more details.

I have a couple of style decisions that I would like some feedback on:

- There is a generic algorithm in the spec named [`GetSubstitution`](https://tc39.es/ecma262/multipage/text-processing.html#sec-getsubstitution) that is used by both `String` and `RegExp`. I have placed it in a new class named `AbstractEcmaStringOperations` but maybe it would be better in the existing `AbstractEcmaObjectOperations.java`?
- I have removed some dead code from `RegExpImpl`, though I cannot remove it all or some old test cases in `MozillaSuiteTest` would break. However, this was part of the public API of rhino, so I am uncertain if removing it is the best choice. I don't really imagine anyone would be using this... but who knows. Any opinion?

I've made an effort to keep the git history very clean, so it's ok to not squash this.

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/AbstractEcmaStringOperations.java
+++ b/rhino/src/main/java/org/mozilla/javascript/AbstractEcmaStringOperations.java
@@ -67,9 +67,7 @@ public class AbstractEcmaStringOperations {
                                 int digitCount = 1;
                                 if (templateRemainder.length() > 2) {
                                     char c2 = templateRemainder.charAt(2);
-                                    if (c2 == '0' || c2 == '1' || c2 == '2' || c2 == '3'
-                                            || c2 == '4' || c2 == '5' || c2 == '6' || c2 == '7'
-                                            || c2 == '8' || c2 == '9') {
+                                    if (isAsciiDigit(c2)) {
                                         digitCount = 2;
                                     }
                                 }
@@ -129,5 +127,24 @@ public class AbstractEcmaStringOperations {
             result.append(refReplacement);
         }
         return result.toString();
+    }
+
+    private static boolean isAsciiDigit(char c) {
+        switch (c) {
+            case '0':
+            case '1':
+            case '2':
+            case '3':
+            case '4':
+            case '5':
+            case '6':
+            case '7':
+            case '8':
+            case '9':
+                return true;
+
+            default:
+                return false;
+        }
     }
 }
--- a/rhino/src/main/java/org/mozilla/javascript/AbstractEcmaStringOperations.java
+++ b/rhino/src/main/java/org/mozilla/javascript/AbstractEcmaStringOperations.java
@@ -1,0 +1,133 @@
+package org.mozilla.javascript;
+
+/** Abstract operations for string manipulation as defined by EcmaScript */
+public class AbstractEcmaStringOperations {
+    /**
+     * GetSubstitution(matched, str, position, captures, namedCaptures, replacementTemplate)
+     *
+     * <p>https://tc39.es/ecma262/multipage/text-processing.html#sec-getsubstitution
+     */
+    public static String getSubstitution(
+            Context cx,
+            Scriptable scope,
+            String matched,
+            String str,
+            int position,
+            NativeArray capturesArray,
+            Object namedCaptures,
+            String replacementTemplate) {
+        // See ECMAScript spec 22.1.3.19.1
+        int stringLength = str.length();
+        if (position > stringLength) Kit.codeBug();
+        StringBuilder result = new StringBuilder();
+        String templateRemainder = replacementTemplate;
+        while (!templateRemainder.isEmpty()) {
+            String ref = templateRemainder.substring(0, 1);
+            String refReplacement = ref;
+
+            if (templateRemainder.charAt(0) == '$') {
+                if (templateRemainder.length() > 1) {
+                    char c = templateRemainder.charAt(1);
+                    switch (c) {
+                        case '$':
+                            ref = "$$";
+                            refReplacement = "$";
+                            break;
+
+                        case '`':
+                            ref = "$`";
+                            refReplacement = str.substring(0, position);
+                            break;
+
+                        case '&':
+                            ref = "$&";
+                            refReplacement = matched;
+                            break;
+
+                        case '\'':
+                            {
+                                ref = "$'";
+                                int matchLength = matched.length();
+                                int tailPos = position + matchLength;
+                                refReplacement = str.substring(Math.min(tailPos, stringLength));
+                                break;
+                            }
+
+                        case '0':
+                        case '1':
+                        case '2':
+                        case '3':
+                        case '4':
+                        case '5':
+                        case '6':
+                        case '7':
+                        case '8':
+                        case '9':
+                            {
+                                int digitCount = 1;
+                                if (templateRemainder.length() > 2) {
+                                    char c2 = templateRemainder.charAt(2);
+                                    if (c2 == '0' || c2 == '1' || c2 == '2' || c2 == '3'
+                                            || c2 == '4' || c2 == '5' || c2 == '6' || c2 == '7'
+                                            || c2 == '8' || c2 == '9') {
+                                        digitCount = 2;
+                                    }
+                                }
+                                String digits = templateRemainder.substring(1, 1 + digitCount);
+
+                                // No need for ScriptRuntime version; we know the string is one or
+                                // two characters and
+                                // contains only [0-9]
+                                int index = Integer.parseInt(digits);
+                                long captureLen = capturesArray.getLength();
+                                if (index > captureLen && digitCount == 2) {
+                                    digitCount = 1;
+                                    digits = digits.substring(0, 1);
+                                    index = Integer.parseInt(digits);
+                                }
+                                ref = templateRemainder.substring(0, 1 + digitCount);
+                                if (1 <= index && index <= captureLen) {
+                                    Object capture = capturesArray.get(index - 1);
+                                    if (capture
+                                            == null) { // Undefined or missing are returned as null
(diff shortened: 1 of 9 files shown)


rhino/src/main/java/org/mozilla/javascript/AbstractEcmaStringOperations.java
```diff
@@ -0,0 +1,133 @@
...
+                        case '5':
+                        case '6':
+                        case '7':
+                        case '8':
+                        case '9':
+                            {
+                                int digitCount = 1;
+                                if (templateRemainder.length() > 2) {
+                                    char c2 = templateRemainder.charAt(2);
+                                    if (c2 == '0' || c2 == '1' || c2 == '2' || c2 == '3'
```
0xe, line 70: Maybe `(c2 >= '0' || c2 <= '9')` ?

rbri:
> A little more than a hundred test262 cases now pass

❤️ 

> I have removed some dead code from RegExpImpl, though I cannot remove it all or some old test cases in MozillaSuiteTest would break. However, this was part of the public API of rhino, so I am uncertain if removing it is the best choice. I don't really imagine anyone would be using this... but who knows. Any opinion?

Not sure but if you like you can have a look at [a889fa44451ee6a0610820983cc72b53ac8bc42d](https://github.com/HtmlUnit/htmlunit/blob/83e296c80463aebc5ee521d785ba853cbe697c8a/src/main/java/org/htmlunit/javascript/regexp/HtmlUnitRegExpProxy.java). Maybe this is (was) one of the real usages of RegExpImpl. For HtmlUnit we tried to map the js RegExp stuff to real Java Regular Expressions (if possible). Therefore we did a lot of customization. I guess if we do not use the stuff you removed then there is a good chance that you can safely remove it.
And since the last release i have completely removed that stuff and we are back to use the Rhino RegExp because of the progress made in this area. Many thanks & kudos to @balajirrao and @andreabergia.

andreabergia:
> > I have removed some dead code from RegExpImpl, though I cannot remove it all or some old test cases in MozillaSuiteTest would break. However, this was part of the public API of rhino, so I am uncertain if removing it is the best choice. I don't really imagine anyone would be using this... but who knows. Any opinion?
> 
> Not sure but if you like you can have a look at [a889fa44451ee6a0610820983cc72b53ac8bc42d](https://github.com/HtmlUnit/htmlunit/blob/83e296c80463aebc5ee521d785ba853cbe697c8a/src/main/java/org/htmlunit/javascript/regexp/HtmlUnitRegExpProxy.java). Maybe this is (was) one of the real usages of RegExpImpl. For HtmlUnit we tried to map the js RegExp stuff to real Java Regular Expressions (if possible). Therefore we did a lot of customization. I guess if we do not use the stuff you removed then there is a good chance that you can safely remove it.

Hm that makes me think we should _not_ remove that code, though. It's just a few lines, won't really hurt to keep them in.

rbri:
> It's just a few lines, won't really hurt to keep them in.

Maybe hurts the inlining....

rhino/src/main/java/org/mozilla/javascript/RegExpProxy.java
```diff
@@ -13,10 +13,8 @@
  */
 public interface RegExpProxy {
     // Types of regexp actions
-    public static final int RA_MATCH = 1;
     public static final int RA_REPLACE = 2;
     public static final int RA_REPLACE_ALL = 3;
-    public static final int RA_SEARCH = 4;
```
rbri, line 19: if we really like to cleanup this i think we can drop all 4 constants and use a boolean parameter to for replace_all. This might show a lot more places where we can simplify the code.

rhino/src/main/java/org/mozilla/javascript/AbstractEcmaStringOperations.java
gbrail: Or call a separate function that uses a switch? Java is weird enough that I couldn't tell you whether 100% of the time ">=" and "<=" will be safe without consulting an expert or AI!

gbrail:
Side comment -- is "RegExpImpl" getting useful enough that it should be moved to a separate module that could theoretically be used by other projects? I used to always assume that we'd be better off with Java's native regex engine if we could figure it out and you all are showing me that the opposite may be true.

gbrail:
Also, if someone is looking for work, put "convert RegExpImpl to lambdas" as a possible project!

gbrail:
Finally, I will not lose too much sleep if we decide to change the public interface of RegExpImpl given, as @rbri says, it's super hard to extend and I doubt that anyone did. I'll let you guys decide and then take a more careful look at this before merging it. Happy to see it's happening!

rbri:
> Side comment -- is "RegExpImpl" getting useful enough that it should be moved to a separate module that could theoretically be used by other projects? I used to always assume that we'd be better off with Java's native regex engine if we could figure it out and you all are showing me that the opposite may be true.

Have had something similar in mind and already discussed with @balajirrao in a different PR. Will open an issue for that and write down my point of view.

rhino/src/main/java/org/mozilla/javascript/AbstractEcmaStringOperations.java
andreabergia: I personally don't mind the way I've written it - it's boring but very simple to understand. I can refactor it to a switch or whatever if anyone feels strongly about it, of course.

andreabergia:
I've decided to drop the "dead code removal" from this PR; we can always discuss and do it in another. It was only a few lines anyway. I've also rebased it on top of master.
For me, this is ready to be merged. Can be squashed, or not, as you prefer.

rhino/src/main/java/org/mozilla/javascript/regexp/NativeRegExp.java
```diff
@@ -3502,15 +3547,236 @@ private Object js_SymbolMatchAll(
...
+                            ScriptRuntime.getObjectProp(result, "length", cx, scope));
+            long nCaptures = Math.max(resultLength - 1, 0);
+            String matched =
+                    ScriptRuntime.toString(ScriptRuntime.getObjectIndex(result, 0, cx, scope));
+            int matchLength = matched.length();
+            double positionDbl =
+                    ScriptRuntime.toInteger(
+                            ScriptRuntime.getObjectProp(result, "index", cx, scope));
+            int position;
+            if (positionDbl < 0) {
```
gbrail, line 3617: This pattern (clamping the position to the length in some way) appears more than once in this code, and looking at the code coverage, we only test one of the three possible paths.

rhino/src/main/java/org/mozilla/javascript/NativeString.java
```diff
@@ -372,9 +374,95 @@ private static void checkValidRegex(Context cx, Object[] args, int pos, String f
...
+        }
+        if (Undefined.isUndefined(separator)) {
+            return cx.newArray(scope, new Object[] {s});
+        }
+
+        int separatorLength = r.length();
+        if (separatorLength == 0) {
+            int strLen = s.length();
+            long outLen;
+            if (lim < 0) {
```
gbrail, line 427: This is the second place where we have this kind of clamping code and we only test one of the three code paths. Would you consider either making a new function in ScriptRuntime for this generic behavior, or adding some artificial test cases so that we execute all three code paths? It's this kind of tricky stuff that, when not tested, leads to problems, because otherwise the code coverage is darn near 100% for this PR.

rhino/src/main/java/org/mozilla/javascript/AbstractEcmaStringOperations.java
```diff
@@ -0,0 +1,133 @@
...
+                                            || c2 == '4' || c2 == '5' || c2 == '6' || c2 == '7'
+                                            || c2 == '8' || c2 == '9') {
```
gbrail, line 72: Sorry, but I don't think I "feel strongly" but it does feel wrong to me, because we're introducing five comparison instructions in the average case to what I imagine is a performance-sensitive bit of code. I would prefer a function that is implemented as a switch statement.

gbrail (commented):
Thanks -- this is indeed looking very nice and I'd like to try and be very careful about such a big change. I can see a place where we could improve code coverage -- could you please take a look?

rhino/src/main/java/org/mozilla/javascript/AbstractEcmaStringOperations.java
andreabergia: I really don't think this is particularly performance sensitive - it happens only when calling `String.prototype.replace` or `replaceAll` with something that is not a regexp and the replacement is something like `$23` (two digits). But I've done the change anyway. 🙂

rhino/src/main/java/org/mozilla/javascript/NativeString.java
andreabergia: I've extracted a function `ScriptRuntime.clamp` and added some unit tests for it.
The coverage really comes from test262, luckily. 🙂

gbrail:
Thank you! I appreciate those small things.

PR #1871, opened 2025-04-03
Removed `.java-version`

andreabergia:
I have added that file by mistake some time ago. It is used by `jenv` to switch between version of the JDK. However, since Rhino currently requires jdk17 for spotless and jdk11 to update test262, we end up changing it all the time, committing it, and then having to revert the changes as we notice it. I don't think it's worth keeping it in the repo, it's more noise than anything at this point IMHO.


aardvark179 (approved):
Thank you! Adding it to `.gitignore` definitely seems like a good idea.

gbrail:
Sorry, should have caught that.

FWIW, if you set:

    RHINO_TEST_JAVA_VERSION=11

in your environment, then the build system will use the Toolchains feature of Gradle to always run the tests on Java 11, while you can have 17 (or 21) as your Java version for everything else.

PR #1872, opened 2025-04-03
Function.prototype[Symbol.hasInstance] fix

balajirrao:
I found that `String[Symbol.hasInstance]({})` threw a type error, when it should be simply return `false`. This likely was a result of switching NativeString to the lambda constructor pattern.

According to the [spec](https://tc39.es/ecma262/#sec-ordinaryhasinstance), if the `prototype` property of `this`  is not an object, it must be an error. In our implementation, the "isObject" check only included `NativeObject`s and `IdScriptableObject`s (to exclude Symbols, essentially). Fix it to simply use `ScriptRuntime.isObject()`, which should do the right thing for symbols as well.


gbrail:
Thanks -- looks good to me now!

PR #1874, opened 2025-04-03
NativeRegExp: Fix regression introduced in hex and unicode hex parsing

balajirrao:
This is a fix for a regression introduced by #1869.

@gbrail I'm sorry, I was too eager to remove code and ended up introducing a regression - at least we are adding new test cases :)

### DIFF
--- a/rhino/src/test/java/org/mozilla/javascript/tests/NativeRegExpTest.java
+++ b/rhino/src/test/java/org/mozilla/javascript/tests/NativeRegExpTest.java
@@ -940,13 +940,16 @@ public class NativeRegExpTest {
         Utils.assertWithAllModes_ES6("true-true-true-false", script);
     }
 
+    // The spec says that \\u and \\x should be followed by exactly 4 and 2 hex digits respectively
+    // for them to be valid unicode escapes. Rhino allows them to be followed by less than 4 or 2
     @Test
     public void unicodeEscapeFallback() {
-        final String script = "var regex = /\\u/;\n" + "var res = '' + regex.test('u');\n" + "res;";
-        Utils.assertWithAllModes_ES6("true", script);
-
-        final String script2 =
-                "var regex = /\\x/;\n" + "var res = '' + regex.test('x');\n" + "res;";
-        Utils.assertWithAllModes_ES6("true", script2);
+        final String script =
+                "var res = '' + /\\u/.test('u') + '-';\n"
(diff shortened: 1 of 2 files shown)


rhino/src/test/java/org/mozilla/javascript/tests/NativeRegExpTest.java
```diff
@@ -948,5 +948,14 @@ public void unicodeEscapeFallback() {
...
         Utils.assertWithAllModes_ES6("true", script2);
+
+        // rhino permits hex and unicode escapes with less than 2 and 4 digits respectively
+        final String script3 =
+                "var regex = /\\x1/;\n" + "var res = '' + regex.test('\\x01');\n" + "res;";
+        Utils.assertWithAllModes_ES6("true", script3);
+
+        final String script4 =
+                "var regex = /\\u61/;\n" + "var res = '' + regex.test('a');\n" + "res;";
+        Utils.assertWithAllModes_ES6("true", script4);
```
rbri, line 959: Maybe splitting this test into four with descriptive names is an possible improvement. 
Personally i think debugging test cases with only one test inside is more effective. 
And having more test also improves the stats...

gbrail:
Let me know if you'd like to act on that test suggestion. I'll get back to this PR soon either way.

Is there some other source of test cases we should be using for some of this new Regexp work? I used to find unit tests from the "mjsunit" part of the V8 source base which were easier to work with for getting new features implemented than going straight to test262.

rhino/src/test/java/org/mozilla/javascript/tests/NativeRegExpTest.java
balajirrao: Yeah, I agree that debugging is easier with one assertion in a test case. However it feels somewhat of an overkill in this case. I think I can combine them into one assertion, which seems to have best of both worlds :)

balajirrao:
@gbrail the PR is ready. I'lll fix the regression in character class parsing in a different PR, coming right after.

> Is there some other source of test cases we should be using for some of this new Regexp work? I used to find unit tests from the "mjsunit" part of the V8 source base which were easier to work with for getting new features implemented than going straight to test262.

mjsunit looks great, thanks - it has a whole bunch of regexp testcases too. I'm going to see if I can use any of them for the unicode support that I'm working on. I'll look for other test suites too.

gbrail:
Thanks! Looks good to me.

Issue #1876, opened 2025-04-07
Arguments should never be evaluated before function


gbrail:
When calling a function located in the global scope in compiled mode, arguments are evaluated before we've established whether the function is valid and callable. 

For example, consider the following program:

`function foo() {
         print('foo called');
}

bar(foo());`

Rhino should throw ReferenceError because bar is not defined, and not call foo.

However, in compiled mode, Rhino first calls foo, then figures out that bar is not defined.

There is no test for this in test262 or elsewhere, but it's not how other JavaScript engines work.

This is because of an optimization in the bytecode that's trying to make the bytecode smaller by delegating to a helper function. I'm working on this part of the bytecode generation to make it simpler and possibly more optimizable, and this came up, so I'll address this there.

aardvark179:
I'm surprised there aren't any tests for this, as I think it is
actually well defined, and annoyingly subtle. The ref of the function should be looked up before the call is made according to [sec-function-calls-runtime-semantics](https://tc39.es/ecma262/multipage/ecmascript-language-expressions.html#sec-function-calls-runtime-semantics-evaluation) (which is where the error should occur if the name is undefined), and /then/ if the name was a real ref we end up in [sec-evaluatecall](https://tc39.es/ecma262/multipage/ecmascript-language-expressions.html#sec-evaluatecall) which specifies that the argument list is evaluated before we check if function was an actual object, and then if it's a function, etc. That seems to be consistent with what V8 does at least.

aardvark179:
> I'm surprised there aren't any tests for this

Wow, there really aren't any in test 262 that I can find. I guess I should submit something.

aardvark179:
So I did a little work on [engine262](https://github.com/engine262/engine262) to see how far things could be done in the wrong order and still pass all of test 262, and it turns out that there *is* a test explicitly covering this, but only under the ES5 section id [11.2.3-3_3](https://262.ecma-international.org/5.1/#sec-11.2.3) which does
```javascript
    function foo(){ fooCalled = true; } 
    
    var o = { }; 
assert.throws(TypeError, function() {
        o.bar.gar( foo() );
        throw new Test262Error("o.bar does not exist!");
});
assert.sameValue(fooCalled, false, 'fooCalled');
```
 and In our test262 properties file I see…
```
language/expressions/call 60/92 (65.22%)
    11.2.3-3_1.js
    11.2.3-3_2.js
    11.2.3-3_4.js
    11.2.3-3_5.js
[...]

PR #1877, opened 2025-04-09
Add INDY operations for calling functions

gbrail:
Add invokedynamic instructions for function calling in the bytecode. Do this when simply calling a Callable on the stack, or performing a lookup and a call in one step.

Replace intermediate "call", "call0", "call1", etc code for calling methods with direct invocations of the Callable interface using MethodHandle invocations.

Replace "call1", "call2" optimization code with just building the argument array in the bytecode inline like the Java compiler does.

Only dispatch and call in one operation using "callName0" and "callProp0" when there are no arguments, to prevent resolving arguments before the status of the target is known.

This fixes https://github.com/mozilla/rhino/issues/1876

Some of these changes undo optimizations introduced to the bytecode over the years, but I'm having trouble seeing what they optimize, except making the bytecode smaller, which I believe was important in the earlier days of Java. One of those optimization introduced the aforementioned bug.

I have benchmarked it and it seems performance-neutral, or at least within the bounds of compiler variation, which seems to be the biggest differentiator in performance tests right now.

Switching to the "METHOD:CALL" family of instructions here, we are directly dispatching to the Callable interface rather than going through an intermediate "call" method in OptRuntime like we used to. This doesn't seem to affect performance. A side affect is that stack traces should be easier to read, except that they now include a bunch of MethodHandle operations instead!

### DIFF
--- a/tests/testsrc/jstests/undefined-functions.js
+++ b/tests/testsrc/jstests/undefined-functions.js
@@ -1,0 +1,53 @@
+load("testsrc/assert.js");
+
+var functionCalled = false;
+
+function testFunction() {
+    functionCalled = true;
+}
+function callTestFunction() {
+   testFunction();
+}
+function callCallTestFunction() {
+  callTestFunction();
+}
+
+// Sanity tests
+functionCalled = false;
+callTestFunction();
+assertTrue(functionCalled);
+functionCalled = false;
+callTestFunction(10);
+assertTrue(functionCalled);
+functionCalled = false;
+callCallTestFunction();
+assertTrue(functionCalled);
+functionCalled = false;
+
+// Undefined function should not parse arguments first
+assertThrows(() => {
+  notFound(testFunction());
+});
+assertFalse(functionCalled);
+
+// Shouldn't matter if there are arguments
+assertThrows(() => {
+  notFound(testFunction(1, 2, 3));
+});
+assertFalse(functionCalled);
+
+// Same thing should happen in a chain
+assertThrows(() => {
+  notFound(callTestFunction(testFunction()));
+});
+assertFalse(functionCalled);
+
+
+// Calling a non-function should not parse arguments first
+let notFunction = 'Hello, World!';
+assertThrows(() => {
+  notFunction(testFunction());
+});
+assertFalse(functionCalled);
+
+'success';
--- a/rhino/src/main/java/org/mozilla/javascript/optimizer/BodyCodegen.java
+++ b/rhino/src/main/java/org/mozilla/javascript/optimizer/BodyCodegen.java
@@ -2524,12 +2524,13 @@ class BodyCodegen {
             cfw.addALoad(contextLocal);
             cfw.addALoad(variableObjectLocal);
             cfw.addALoad(thisObjLocal);
-            if (firstArgChild == null) {
-                addDynamicInvoke("METHOD:CALL_0", Signatures.METHOD_CALL_0);
-            } else {
-                generateCallArgArray(node, firstArgChild, false);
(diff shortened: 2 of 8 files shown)


aardvark179:
I'm not against this change, but I am a little concerned by the memory footprint of these call sites. If I load lodash then this creates roughly 3000 call sites with a total retained set size of over 1MB. Unused sites seem to have a retained size of 248 bytes, with sites that are used going up to about 1K each, and the more methods I call the more 1K sites I see associated with that generated class.

We already strongly limit compilation to class files because unless we are careful to only target hot code then the meta space usage outweighs most of the performance gains, and call sites are going to push us _much_ harder in that direction. Could we work to make sure that things like `TOBOOLEAN` are made as constant call sites unless or until we do real optimisation on those sites?

gbrail:
Excellent points, and I have not really been looking at memory usage when changing things like that.

I have some ideas for how we might be able to change the INDY infrastructure in Rhino to be more memory-efficient, like you suggest. Let me spend some time on that and get back to you. In many cases I have used INDY a bit preemptively not knowing what optimizations are possible, so I bet we can streamline.

gbrail:
Based on your comments, I took out the new opcodes for "METHOD:CALL", because I can't imagine optimizing those in the future, and replaced them with just direct "invokeinterface" bytecode. I'm curious to see if this helps your use case, and I can consider other INDY byte codes that realistically we'll not find an optimization for.

Another alternative is to take a less memory-intensive approach to the INDY stuff, but that might mean replacing the "dynalink" framework with one of our own.

tests/testsrc/jstests/undefined-functions.js
```diff
@@ -0,0 +1,53 @@
...
+});
+assertFalse(functionCalled);
+
+
+// Calling a non-function should not parse arguments first
+let notFunction = 'Hello, World!';
+assertThrows(() => {
+  notFunction(testFunction());
+});
+assertFalse(functionCalled);
```
aardvark179, line 51: This assertion is wrong. The assertions for the other cases are correct because in all of those the error is a reference error caused evaluation of the refexp before the parentheses as part of [13.3.6.1](https://tc39.es/ecma262/multipage/ecmascript-language-expressions.html#sec-function-calls-runtime-semantics-evaluation)—specifically the GetValue of the reference expression in step 4. In this last case however `notFunction` is a valid thing that can be looked up, so we reach section [13.3.6.2](https://tc39.es/ecma262/multipage/ecmascript-language-expressions.html#sec-evaluatecall), and will do
1. Stuff about resolving `thisValue` which doesn't depend on the ref's value being a function.
2. Ditto
3. Let argList be ? [ArgumentListEvaluation](https://tc39.es/ecma262/multipage/ecmascript-language-expressions.html#sec-runtime-semantics-argumentlistevaluation) of arguments.
4. If func [is not an Object](https://tc39.es/ecma262/multipage/ecmascript-data-types-and-values.html#sec-object-type), throw a TypeError exception.
5. If [IsCallable](https://tc39.es/ecma262/multipage/abstract-operations.html#sec-iscallable)(func) is false, throw a TypeError exception.
6. If tailPosition is true, perform [PrepareForTailCall](https://tc39.es/ecma262/multipage/ecmascript-language-functions-and-classes.html#sec-preparefortailcall)().
7. Return ? [Call](https://tc39.es/ecma262/multipage/abstract-operations.html#sec-call)(func, thisValue, argList).

gbrail:
Closing this in favor of a different PR that I'm slowwwwly drafting to fix function call argument evaluation ordering and possibly simplify some bytecode things so that this won't be necessary anyway.

PR #1882, opened 2025-04-14
Make MozillaSuiteTest faster

andreabergia:
Applied the same strategy used for `Test262SuiteTest` to use JUnit 5's parallel execution. For me, it cuts time from 2:42 to 1:33, which is nice.

At the moment, there's a couple of tests that seem to break, probably because there's some state leaking between tests; I'll try to understand better why and to fix it.

### DIFF
--- a/tests/src/test/java/org/mozilla/javascript/tests/MozillaSuiteTest.java
+++ b/tests/src/test/java/org/mozilla/javascript/tests/MozillaSuiteTest.java
@@ -16,11 +16,13 @@ import java.util.Arrays;
 import java.util.Collection;
 import java.util.HashSet;
 import java.util.List;
-import org.junit.Assert;
-import org.junit.Test;
-import org.junit.runner.RunWith;
-import org.junit.runners.Parameterized;
-import org.junit.runners.Parameterized.Parameters;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.parallel.Execution;
+import org.junit.jupiter.api.parallel.ExecutionMode;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.MethodSource;
 import org.mozilla.javascript.Context;
 import org.mozilla.javascript.drivers.JsTestsBase;
 import org.mozilla.javascript.drivers.ShellTest;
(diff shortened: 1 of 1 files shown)


aardvark179:
Just tested locally with `--max-workers=4` and this took the test time from 4 minutes 3 seconds, to 3 minutes 25! Nice work.

gbrail:
I did go down this path once before, and the problem I ran in to is that a
few of those tests require ginormous stacks. I suspect that many of those
tests are obsolete and could be disabled, and I'd support that if someone
carefully considers what they're actually testing.

andreabergia:
I've analyzed more the problem. There's two tests that seem to be breaking.

The first one is [`regress-383269-01.js`](https://github.com/mozilla/rhino/blob/master/tests/testsrc/tests/js1_5/GC/regress-383269-01.js) which simply creates a very big object. This is flaking for me locally; sometimes it works, sometimes it does not. It doesn't really test much honestly, I am not sure it's worth keeping.

The second one, [`regress-98901.js`](https://github.com/mozilla/rhino/blob/master/tests/testsrc/tests/js1_5/Regress/regress-98901.js) is a lot more interesting: it attempts to add _many_ variables. If I run this in isolation by modifying `interpreted.tests` and `compiled.tests`, it reliably fails with a stack overflow - the exception gets thrown either in the parser, or in `CodeGenerator`. If I run _some_ tests before it, sometimes it passes and sometimes it fails. If run all the suite, it reliably works. IMHO the JVM inlining sees that the method is hot enough and does enough optimization to make the test pass reliably if it executes enough code before it; otherwise, it doesn't and we get the exception.
This is an interesting test, because the parser and the code generator could definitely be implemented in a different way so that we don't have this problem; however, doing so would be a big undertaking, and I am not so sure it's worth investing the time. To merge this PR, we need to disable this test though.

Any opinion on how to proceed? I personally think that removing the two tests and enabling the parallelization is worth i, because it cuts CI times.

aardvark179:
[regress-98901.js](https://github.com/mozilla/rhino/blob/master/tests/testsrc/tests/js1_5/Regress/regress-98901.js) is meant to be test that files like that can be successfully parsed, so it seems perverse to disable it when it exposes the problem it is intended to test for.

I would be interested to see which part of the compilation process is causing the stack overflow @andreabergia, it’s likely that a fairly small focused fix around operator handling could fix this. It might also be worth generating an even bigger test case to push things to consistently fail.

andreabergia:
Sorry for the radio silence on this. I have been working on https://github.com/mozilla/rhino/pull/1901 which reproduces the same problem of [regress-98901.js](https://github.com/mozilla/rhino/blob/master/tests/testsrc/tests/js1_5/Regress/regress-98901.js). 
My new test exposes a problem on `IRFactory` which I am not totally sure how to fix, because removing recursion from that class would be rather complicated and more akin to a rewrite that a refactor, as far as I can tell.

PR #1883, opened 2025-04-16
Add double null check pattern when initializing cache values

rPraml:
I found some places in ClassCache, where maps are initialized lazy.

These places may not be 100% thread safe.
This would probably not be a problem, because in most cases it is not accessed concurrently
But I would still like to fix this.

@gbrail I've used here the `infamous "double-checked locking" pattern` hopefully in the right way (with volatile type)

Do you see here also problems, that this could fail?
I'm also open to do an eager initialization of the three maps (constructing three ConcurrentHashMaps should not be too expensive)

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/ClassCache.java
+++ b/rhino/src/main/java/org/mozilla/javascript/ClassCache.java
@@ -139,9 +139,7 @@ public class ClassCache implements Serializable {
         if (classTable == null) {
             synchronized (this) {
                 if (classTable == null) {
-                    // Use 1 as concurrency level here and for other concurrent hash maps
-                    // as we don't expect high levels of sustained concurrent writes.
-                    classTable = new ConcurrentHashMap<>(16, 0.75f, 1);
+                    classTable = new ConcurrentHashMap<>();
                 }
             }
         }
@@ -152,7 +150,7 @@ public class ClassCache implements Serializable {
         if (classAdapterCache == null) {
             synchronized (this) {
                 if (classAdapterCache == null) {
-                    classAdapterCache = new ConcurrentHashMap<>(16, 0.75f, 1);
+                    classAdapterCache = new ConcurrentHashMap<>();
                 }
             }
         }
@@ -194,7 +192,7 @@ public class ClassCache implements Serializable {
             if (interfaceAdapterCache == null) {
                 synchronized (this) {
                     if (interfaceAdapterCache == null) {
-                        interfaceAdapterCache = new ConcurrentHashMap<>(16, 0.75f, 1);
+                        interfaceAdapterCache = new ConcurrentHashMap<>();
                     }
                 }
             }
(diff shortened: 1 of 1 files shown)


gbrail:
Well -- I've always understood double-checked locking to be an incorrect
pattern, but according to various AIs and things it's now considered OK as
long as the variable is volatile, like your have it here, so it should be
OK!

rhino/src/main/java/org/mozilla/javascript/ClassCache.java
```diff
@@ -137,16 +137,24 @@ public synchronized void setCachingEnabled(boolean enabled) {
...
     Map<CacheKey, JavaMembers> getClassCacheMap() {
         if (classTable == null) {
-            // Use 1 as concurrency level here and for other concurrent hash maps
-            // as we don't expect high levels of sustained concurrent writes.
-            classTable = new ConcurrentHashMap<>(16, 0.75f, 1);
+            synchronized (this) {
+                if (classTable == null) {
+                    // Use 1 as concurrency level here and for other concurrent hash maps
+                    // as we don't expect high levels of sustained concurrent writes.
+                    classTable = new ConcurrentHashMap<>(16, 0.75f, 1);
```
gbrail, line 144: Just as a general principle -- do we have a sense that it actually makes sense for us to hard-code our own values for cache size, ratio, etc? From time to time the java libraries change the defaults as they learn more. I would be happier if we'd just use the default constructor here and elsewhere unless we have actual data.

gbrail (commented):
Since we're being really detail oriented here -- would you consider just using the default constructor for ConcurrentHashMap in all these places, since I don't believe we have data that supports our theory of setting different values than the defaults.

rhino/src/main/java/org/mozilla/javascript/ClassCache.java
rPraml: yes, especially in this case, the implementation of ConcurrentHashMap seems to be completely rewritten from JDK7 to 8

https://github.com/openjdk/jdk7/blob/master/jdk/src/share/classes/java/util/concurrent/ConcurrentHashMap.java#L735
https://github.com/openjdk/jdk8/blob/master/jdk/src/share/classes/java/util/concurrent/ConcurrentHashMap.java#L892

There is still https://github.com/mozilla/rhino/blob/master/rhino/src/main/java/org/mozilla/javascript/commonjs/module/provider/CachingModuleScriptProviderBase.java#L34 which seems to have adopted some code from JDK6 ConcurrentHashMap and still uses the individual constructors

rPraml:
BTW1: I made some benchmarks:

call to `getClassCacheMap` volatile vs non volatile : nearly not measurable
call to `new ClassCache()` 
- with lazy init ~0.411 ns/op
- with eager init ~4,746 ns/op

BTW2: When removing the `volatile`, errorprone will detect the wrong DoubleCheckedLocking pattern:
```
rhino/src/main/java/org/mozilla/javascript/ClassCache.java:139: warning: [DoubleCheckedLocking] Double-checked locking on non-volatile fields is unsafe
        if (classTable == null) {
        ^
    (see https://errorprone.info/bugpattern/DoubleCheckedLocking)
  Did you mean 'private transient volatile  Map<CacheKey, JavaMembers> classTable = new ConcurrentHashMap<>();'?
```

gbrail:
Thanks -- this looks good to me!

PR #1886, opened 2025-04-23
fix `ClassCache#get(...)` and `ClassCache.CacheKey#hashCode()`

### LINKED ISSUE #1884: `ClassCache#get(Scriptable)` behaviour inconsistent with the behaviour described in Javadoc
According to return value section of the javadoc of `ClassCache#get(Scriptable)`, this method should always return a valid `ClassCache` instance even if none was found: 

```
     * @return previously associated ClassCache object or a new instance of ClassCache if no
     *     ClassCache object was found.
```

but the actual behavior is throwing an exception if none was found:

https://github.com/mozilla/rhino/blob/6474e6c4cb808ab664d61c13e68e5e2f9c818f6c/rhino/src/main/java/org/mozilla/javascript/ClassCache.java#L73-L77

ZZZank:
- The original hashing of `CacheKey` will completely ignore the `cls` field when `sec` is present, now it will combine hash code for both fields.
- `ClassCache.get(...)` will now try to associate a new ClassCache object when none was found, then return the new object, as described in `@return` javadoc of it.

Fixes #1884

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/ClassCache.java
+++ b/rhino/src/main/java/org/mozilla/javascript/ClassCache.java
@@ -46,7 +46,7 @@ public class ClassCache implements Serializable {
         public int hashCode() {
             int result = cls.hashCode();
             if (sec != null) {
-                result = sec.hashCode() * 31;
+                result = result * 31 + sec.hashCode();
             }
             return result;
         }
--- a/rhino/src/main/java/org/mozilla/javascript/ClassCache.java
+++ b/rhino/src/main/java/org/mozilla/javascript/ClassCache.java
@@ -44,11 +44,10 @@ public class ClassCache implements Serializable {
 
         @Override
         public int hashCode() {
-            int result = cls.hashCode();
             if (sec != null) {
-                result = result * 31 + sec.hashCode();
+                return sec.hashCode() ^ cls.hashCode();
             }
-            return result;
+            return cls.hashCode();
         }
 
         @Override
--- a/rhino/src/main/java/org/mozilla/javascript/ClassCache.java
+++ b/rhino/src/main/java/org/mozilla/javascript/ClassCache.java
@@ -62,17 +62,30 @@ public class ClassCache implements Serializable {
     /**
      * Search for ClassCache object in the given scope. The method first calls {@link
      * ScriptableObject#getTopLevelScope(Scriptable scope)} to get the top most scope and then tries
-     * to locate associated ClassCache object in the prototype chain of the top scope.
+     * to locate associated ClassCache object in the prototype chain of the top scope. If none was
+     * found, it will try to associate a new ClassCache object to the top scope.
      *
      * @param scope scope to search for ClassCache object.
      * @return previously associated ClassCache object or a new instance of ClassCache if no
      *     ClassCache object was found.
      * @see #associate(ScriptableObject topScope)
+     * @throws IllegalArgumentException if the top scope of provided scope have no associated
+     *     ClassCache, and cannot have ClassCache associated due to the top scope not being a {@link
+     *     ScriptableObject}
      */
     public static ClassCache get(Scriptable scope) {
         ClassCache cache = (ClassCache) ScriptableObject.getTopScopeValue(scope, AKEY);
         if (cache == null) {
-            throw new RuntimeException("Can't find top level scope for " + "ClassCache.get");
+            // we expect this to not happen frequently, so computing top scope twice is acceptable
+            var topScope = ScriptableObject.getTopLevelScope(scope);
+            if (!(topScope instanceof ScriptableObject)) {
+                // Note: it's originally a RuntimeException, the super class of
+                // IllegalArgumentException, so this will not break error catching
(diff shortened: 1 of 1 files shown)


rhino/src/main/java/org/mozilla/javascript/ClassCache.java
```diff
@@ -46,7 +46,7 @@ public CacheKey(Class<?> cls, Object securityContext) {
         public int hashCode() {
             int result = cls.hashCode();
             if (sec != null) {
-                result = sec.hashCode() * 31;
+                result = result * 31 + sec.hashCode();
```
gbrail, line 49: Do we have any knowledge / data about the "* 31" bit, which I assume was added long ago in history. It doesn't do anything to make this solution more correct -- if anything, it might make the hash table behave worse! Why don't we take it out and just add (or OR) the two values together?

ZZZank: 31 is a "magic number" used by many simple hash implementations, and I believe the `* 31` part is just simulating `Arrays.hashCode(...)` behaviour, which will multiply the intermediate result by 31 before adding hashcode from the next array element.

In context here, yes, `* 31` does look redundant, but I personally prefer XOR over add or logical add (aka OR), because I (somehow) believe it does better in mixing bits.

rhino/src/main/java/org/mozilla/javascript/ClassCache.java
```diff
@@ -62,17 +61,30 @@ public boolean equals(Object obj) {
...
+            // we expect this to not happen frequently, so computing top scope twice is acceptable
+            var topScope = ScriptableObject.getTopLevelScope(scope);
+            if (!(topScope instanceof ScriptableObject)) {
+                // Note: it's originally a RuntimeException, the super class of
+                // IllegalArgumentException, so this will not break error catching
+                throw new IllegalArgumentException(
+                        "top scope have no associated ClassCache and cannot have ClassCache associated due to not being a ScriptableObject");
+            }
+            cache = new ClassCache();
+            cache.associate(((ScriptableObject) topScope));
```
rPraml, line 87: @ZZZank Have you encountered a problem with the RuntimeException? Or have you simply found that the documentation doesn't match the implementation?

This is not thread safe. If two threads tries to init the cache, you might have a race condition here.
At least in our application, we share one topLevelScope for multiple threads (and this scope gets a ClassCache assigned on construction)
 
Normally, when you use initStandardObjects on the top scope, a ClassCache should be added automatically. If you use "initSafeStandardObjects" (which means I don't want to use LiveConnect) it would be totally OK, if there is no classCache and I get the exception, because it means, that some code will try to access LiveConnect, that should not (and may be tries to break out the sandbox)

In my opinion, instead of implementing a possible non-threadsafe lazy init, the documentation should therefore be adapted, even if this is strictly speaking a change to the API.

BTW: The implementation was changed here: https://github.com/mozilla/rhino/commit/e59e56822f2a210cd7b39aae266b9beb343b90ef#diff-69ef3670c4b5abfaf14ea2f9571184d5193995fbea54dab635d3daff3f06c06b but the javadoc was not adjusted

ZZZank: This problem was found when I was trying to use `ClassCache#get(...)` inside `FunctionObject#<init>(...)`, in which case MozillaSuiteTest will fail. There're 4876 failed MozillaSuiteTest in total, the first of them is:

```
runMozillaTest[0, js=testsrc/tests/bigo-test.js, opt=false]

java.lang.RuntimeException: Can't find top level scope for ClassCache.get
	at org.mozilla.javascript.ClassCache.get(ClassCache.java:84)
	at org.mozilla.javascript.FunctionObject.<init>(FunctionObject.java:84)
	at org.mozilla.javascript.ScriptableObject.defineFunctionProperties(ScriptableObject.java:2101)
	at org.mozilla.javascript.drivers.ShellTest.runNoFork(ShellTest.java:434)
	at org.mozilla.javascript.tests.MozillaSuiteTest.runMozillaTest(MozillaSuiteTest.java:197)
	at java.base/java.lang.reflect.Method.invoke(Method.java:580)
```

rhino/src/main/java/org/mozilla/javascript/ClassCache.java
aardvark179: Simple xoring for combining hashes isn’t great as two similar hashes will end up zeroing a lot  of bits, and gives the same hash for A, B as for B, A. Multiplying by a small prime tends to spread a set of bits out over the available space and keeps this good for the mod N operation you’ll commonly end up using to choose a bucket in the table.

There are better / faster hashing functions out there, but they often need a final mixing steps so aren’t so well suited to `hashCode()` methods, and I don’t know how well they would combine with prime multiply and add hash codes.

gbrail:
OK, thanks -- I get the "+31 thing" in that context. Thanks for doing this!

gbrail:
Anyway, I'm OK with the XOR for now seeing that we have a few different opinions and we can always test or adjust later if we need.

Issue #1887, opened 2025-04-26
Regression from NativeArray migration to lambda - 'length' property changed on the prototype instead of creating a new


rbri:
This was found while applying the latest changes to HtmlUnit. Most of the tests are passing but a test with an old version of the Dojo lib failed.
Spent a lot of time to nail down the problem; but i think i catched it.

Dojo introduces a new object named NodeList (strange idea to name it like the dom object but...). The interesting part here is the prototype. This is simply an empty array. The minimal code looks like this.

    var WithArrayPrototype = function(array) {
        this.length = array.length;
        return this;
    }"

    var nlp = WithArrayPrototype.prototype = [];"

    var test = new WithArrayPrototype(['abc']);"
    '' + nlp.length + ' # ' + test.length;

Running this in Rhino after the NativeArray change produces '1 # 1' instead of '0 # 1' (as the old code and all the browsers)

rbri:
In this case Rhino changes the prototype length property instead of defining a new length property for the receiver.

From a first guess it looks like the put does a search on the prototype chain and (after the NativeArray change) and finds the 'length' property of the prototype.

aardvark179:
Okay, that’s weird. I’ll look into it.

rbri:
> I'm a little surprised that there isn't something in test262 about this

Same here

aardvark179:
Sorry, been a bit busy. I have a fix, but while experimenting on something else found a way that it broke things differently. I want to make sure I get the correct behaviour for property definition in combination with built in slots, and I'm not convinced it's quite right yet.

PR #1888, opened 2025-04-26
Issue #1887

rbri:


gbrail:
I think we should keep this test -- can you rebase it and then we can merge it? It should pass now.

rbri:
@gbrail build is green now, rebase is done.

gbrail:
Thanks!

PR #1889, opened 2025-05-03
Fix the setting of `length` on array instances, and ensure that property definitions still work correctly

### LINKED ISSUE #1887: Regression from NativeArray migration to lambda - 'length' property changed on the prototype instead of creating a new

aardvark179:
This resolves #1887 and without breaking the various test262 cases around setting array length via redefining the property. descriptor.

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/BuiltInSlot.java
+++ b/rhino/src/main/java/org/mozilla/javascript/BuiltInSlot.java
@@ -112,7 +112,24 @@ public class BuiltInSlot<T extends ScriptableObject> extends Slot {
     @Override
     @SuppressWarnings("unchecked")
     public boolean setValue(Object value, Scriptable owner, Scriptable start, boolean isThrow) {
-        return setter.apply(((T) this.value), value, owner, start, isThrow);
+        if ((getAttributes() & ScriptableObject.READONLY) != 0) {
+            if (isThrow) {
+                throw ScriptRuntime.typeErrorById("msg.modify.readonly", name);
+            }
+            return true;
+        }
+        if (owner == start) {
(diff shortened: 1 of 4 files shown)


aardvark179:
@rbri, sorry this took a while, understanding the issues with the failing tests when I fixed the basic bug took longer than I had hoped.

rbri:
@aardvark179 no prob and many thanks for workin on this

rbri:
just running my test suites, will report

rbri:
all my HtmlUnit tests are working fine again with this fix. Hope to see this merged soon.
Thanks @aardvark179!

gbrail:
I see that this fixes the problem and it's good -- thanks for tracking down the problem and fixing it!

I don't feel like I every totally understand the "start" business in the Scriptable / ScriptableObject / Slot codebase, and I wonder if we should try to document it better. I also would love to see whether there's a way that we can get out of the SlotMap interfaces having to raise certain exceptions that have to do with compliance.

I'm happy to merge this -- it's just the long-term implications I'd like to consider. Thanks!

PR #1890, opened 2025-05-05
Using 'decycle' to detect package cycles

rPraml:
This is the beginning of an approach to keep the codebase cycle-free. 
So I would like to introduce a check tool, that enforces cycle-free packages for newly developed features

I will create 2-3 PRs where I would try remove existing cycles.

ChatGPT means (and I would agree this):
> Keeping packages cycle-free in Java is important for several reasons:
> - Maintainability: When packages have cyclic dependencies, it becomes difficult to understand and manage the relationships between them. This can make it harder to make changes, add new features, or fix bugs in the codebase.
> - Reusability: Cyclic dependencies can limit the reusability of code. When packages are tightly coupled, it becomes more challenging to extract and reuse individual components in other projects or modules.
> - Testability: Cyclic dependencies can also make it harder to write unit tests for the code. When packages are tightly intertwined, it can be difficult to isolate and test individual components without bringing in unnecessary dependencies.
> - Scalability: As the project grows, cyclic dependencies can lead to increased complexity and potential for bugs. By keeping packages cycle-free, it is easier to scale the project and add new functionality without introducing unexpected issues. 
>
> Overall, keeping packages cycle-free makes the codebase more maintainable, reusable, testable, and scalable, which ultimately leads to a more robust and reliable software application.

### DIFF
--- a/rhino/build.gradle
+++ b/rhino/build.gradle
@@ -42,3 +42,22 @@ publishing {
         }
     }
 }
+decycle {
+    // package o.m.j.json contains only one class. Maybe move it to o.m.j
+    ignoring from: "org.mozilla.javascript.json.JsonParser", to: "org.mozilla.javascript.**"
+    // o.m.j.commonjs is strongly coupled to o.m.j
+    ignoring from: "org.mozilla.javascript.commonjs.**", to: "org.mozilla.javascript.**"
+    // o.m.j.debug is strongly coupled to o.m.j
+    ignoring from: "org.mozilla.javascript.debug.**", to: "org.mozilla.javascript.**"
+    // o.m.j.xml is strongly coupled to o.m.j (xmlimpl already in separate module)
+    ignoring from: "org.mozilla.javascript.xml.**", to: "org.mozilla.javascript.**"
+
+    // TODO: easy to remove
+    ignoring from: "org.mozilla.classfile.**", to: "org.mozilla.javascript.Kit"
+
+    // TODO: Should be possible to remove cycle between o.m.j.ast and o.m.j
+    ignoring from: "org.mozilla.javascript.ast.**", to: "org.mozilla.javascript.**"
+
+    // TODO: o.m.j.typedarrays can be decoupled with serviceLoader pattern
+    ignoring from: "org.mozilla.javascript.typedarrays.**", to: "org.mozilla.javascript.**"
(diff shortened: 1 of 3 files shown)


rhino/build.gradle
```diff
@@ -42,3 +42,22 @@ publishing {
...
+    ignoring from: "org.mozilla.javascript.xml.**", to: "org.mozilla.javascript.**"
+
+    // TODO: easy to remove
+    ignoring from: "org.mozilla.classfile.**", to: "org.mozilla.javascript.Kit"
+
+    // TODO: Should be possible to remove cycle between o.m.j.ast and o.m.j
+    ignoring from: "org.mozilla.javascript.ast.**", to: "org.mozilla.javascript.**"
+
+    // TODO: o.m.j.typedarrays can be decoupled with serviceLoader pattern
+    ignoring from: "org.mozilla.javascript.typedarrays.**", to: "org.mozilla.javascript.**"
```
rPraml, line 62: will be fixed by #1893

gbrail:
I have the same question that @aardvark179 had on these -- is eliminating package cycles (as opposed to module cycles) really worth the work? I would rather focus on possible future modularization possibilities, like the work that @rPraml (I think it was you, sorry) was doing on moving LiveConnect to a different module. Are we missing a problem that these changes fix? or do you need help or reviewing of some of that other stuff? I'm happy to see other ideas and data on this but my first reaction is that I'm not sure that these changes are worth it.

gbrail:
I'm still not sure about eliminating all cycles on principle, but I get the idea of not introducing new ones, and I think it's OK to add this tool to the build. Unless others still feel strongly I think that we should merge this.

gbrail:
I'm still OK with merging this so that at least we can make informed future decisions about package cycles. Can you please rebase this since it looks like we introduced a new cycle? Thanks!

rPraml:
Hello @gbrail, I've updated the decycle configuration and tried to document why the plugin is there.

Please check the description to see if this fits.

I also updated the current state and tried to be very concrete here. I use as few wildcards as possible. 
It gives (at least) for me a good feeling, where we can break cycles (especially when moving liveConnect to its own package, which is still a goal for me)

gbrail:
Thanks! This works and I get why so let's merge it.

PR #1891, opened 2025-05-05
Remove package cycle org.mozilla.javascript <-> org.mozilla.classfile

rPraml:
See #1890

### DIFF
--- a/rhino/src/main/java/org/mozilla/classfile/ClassFileWriter.java
+++ b/rhino/src/main/java/org/mozilla/classfile/ClassFileWriter.java
@@ -1455,8 +1455,15 @@ public class ClassFileWriter {
             if (startIdx < itsSuperBlockStartsTop) {
                 SuperBlock sb = superBlocks[startIdx];
                 // check, if it is really the matching one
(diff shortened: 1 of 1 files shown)


rhino/src/main/java/org/mozilla/classfile/ClassFileWriter.java
```diff
@@ -1456,7 +1455,8 @@ private SuperBlock getSuperBlockFromOffset(int offset) {
             if (startIdx < itsSuperBlockStartsTop) {
                 SuperBlock sb = superBlocks[startIdx];
                 // check, if it is really the matching one
-                if (offset < sb.getStart() || offset >= sb.getEnd()) Kit.codeBug();
+                if (offset < sb.getStart() || offset >= sb.getEnd())
+                    throw new IndexOutOfBoundsException();
```
rbri, line 1459: please add a message with at least the current range and the wrong index that was used

rPraml: Thanks for the hint. I've made also a separate PR where I've fixed some other places (see #1892)

PR #1893, opened 2025-05-05
Remove package cycle between o.m.j <-> o.m.j.typedarrays

rPraml:
Will remove another package cycle. (See #1890)

### DIFF
--- a/rhino/src/main/java/module-info.java
+++ b/rhino/src/main/java/module-info.java
@@ -20,6 +20,8 @@ module org.mozilla.rhino {
 
     provides org.mozilla.javascript.RegExpLoader with
             org.mozilla.javascript.regexp.RegExpLoaderImpl;
+    provides org.mozilla.javascript.TypedArraysLoader with
+            org.mozilla.javascript.typedarrays.TypedArraysLoaderImpl;
 
     requires java.compiler;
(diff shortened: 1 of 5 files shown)


aardvark179:
I’m not sure I’m entirely convinced by these PRs. I agree that cyclic dependencies between modules can be a bad idea, and *some* parts of Rhino’s code are clearly separate (e.g. the classfile writer is something used by Rhino rather than something tightly coupled to it), but the typed array implementation will always be tightly coupled to the core of Rhino and is only really in a separate package because it helps make the code base more manageable.

rPraml:
> but the typed array implementation will always be tightly coupled to the core of Rhino 

I understand, the same is probably true for the regexp implementation, where we use the same init-pattern to avoid cycles.

> and is only really in a separate package because it helps make the code base more manageable.

And this PR could improve this, by moving the initialization code to the typedarray package as well. 
(See also my comment in #1890)

Roland

andreabergia:
I agree with @aardvark179 - I personally think it is easier to maintain the way it is right now.

rbri:
> I agree with @aardvark179 - I personally think it is easier to maintain the way it is right now.

Agree, for me the array suff is more in a subpackage and definitely not in a separate module.

Issue #1895, opened 2025-05-07
ISO Date parsing with 2 msec chars instead of 3 doesn't result in a correctly parsed date


jcompagner:
this is because rhino code always expects the MSEC part to be 3 of length:

https://github.com/mozilla/rhino/blob/master/rhino/src/main/java/org/mozilla/javascript/NativeDate.java#L987

if that is then 2 it will bomb out on this line in an ERROR state:

https://github.com/mozilla/rhino/blob/master/rhino/src/main/java/org/mozilla/javascript/NativeDate.java#L997

but browser seem to parse this: new Date('2025-05-07T09:05:20.78Z') just fine

Why are we not just using the OffsetDateTime object of java or Instant to do that parsing ? Is that to lenient? or are we then missing some stuff that js does support? (in the end ISO is ISO right?)

aardvark179:
> ECMAScript defines a string interchange format for date-times based upon a simplification of the ISO 8601 calendar date extended format. The format is as follows: YYYY-MM-DDTHH:mm:ss.sssZ

There’s good reasons to have a specific implementation, but this does look like a bug.

rbri:
i fear i was the one that did the last changes on date - will provide a fix for this

jcompagner:
i don't think this is recent or new, that code is (with i think some formatting 4 years ago on top of it) 13 years old.

thats why i asked why not use now the java parse implementation for this (those follow the same iso)

rbri:
Found some more incompatibilities... stay tuned

rbri:
@jcompagner ca we close this?

jcompagner:
code changes are a bit hard to fully check, but the test that are added seem to cover now the full range so this parsing should be a lot better now.

PR #1896, opened 2025-05-08
improved date string parser; support optional ms digits

rbri:
see issue #1895


rbri:
done with this from my point of view - ready for review and hopefully merging...

PR #1897, opened 2025-05-09
Make wrapped Java object not a JS array/map when static type is not list/map

### LINKED ISSUE #1885: Java object will sometimes be treated as an array/map when it should not be
The easiest way of reproducing this is the following code:

```java
public interface Example {
    static Example get() {
        return new Impl();
    }

    class Impl extends ArrayList<Object> implements Example {
    }
}
```

Calling `Example.get().length` will return `0`, instead of an `undefined`. Calling `Example.get()[0]` will will return value in the backing list, instead of `no public instance field or method named "0"` error.

Similarly, if the `Impl` class extends `HashMap`, calling `Example.get()["string-here"]` will return value in the backing map, instead of always returning `undefined`

ZZZank:
Fixes #1885

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/WrapFactory.java
+++ b/rhino/src/main/java/org/mozilla/javascript/WrapFactory.java
@@ -113,10 +113,12 @@ public class WrapFactory {
      */
     public Scriptable wrapAsJavaObject(
             Context cx, Scriptable scope, Object javaObject, Class<?> staticType) {
-        if (List.class.isAssignableFrom(javaObject.getClass())) {
-            return new NativeJavaList(scope, javaObject);
-        } else if (Map.class.isAssignableFrom(javaObject.getClass())) {
-            return new NativeJavaMap(scope, javaObject);
+        if (staticType != null) {
+            if (List.class.isAssignableFrom(staticType)) {
+                return new NativeJavaList(scope, javaObject);
+            } else if (Map.class.isAssignableFrom(staticType)) {
+                return new NativeJavaMap(scope, javaObject);
+            }
         }
         return new NativeJavaObject(scope, javaObject, staticType);
     }
--- a/tests/src/test/java/org/mozilla/javascript/tests/JavaObjectWrappingTest.java
+++ b/tests/src/test/java/org/mozilla/javascript/tests/JavaObjectWrappingTest.java
@@ -1,0 +1,84 @@
+package org.mozilla.javascript.tests;
+
+import org.junit.jupiter.api.Test;
+import org.mozilla.javascript.Context;
+import org.mozilla.javascript.ContextFactory;
+import org.mozilla.javascript.Undefined;
+import org.mozilla.javascript.testutils.Utils;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+
+/**
+ * @author ZZZank
+ */
+public class JavaObjectWrappingTest {
+    private static final String PATH = "Packages." + Interf.class.getName();
+
+    @Test
+    public void list() {
+        // methods exposed via static type should be accessible
+        Utils.assertWithAllModes(1, PATH + ".listBased().size()");
+        // values inside list should not be leaked, if static type is not a list
+        Utils.assertEvaluatorExceptionES6(
+            String.format("Java class \"%s\" has no public instance field or method named", ListBasedImpl.class.getName()),
+            PATH + ".listBased()[0]"
+        );
+        Utils.assertWithAllModes(Undefined.instance, PATH + ".listBased().length");
+    }
+
+    @Test
+    public void map() {
+        // methods exposed via static type should be accessible
+        Utils.assertWithAllModes(1, PATH + ".mapBased().size()");
+        // values inside map should not be leaked, if static type is not a map
+        Utils.assertWithAllModes(Undefined.instance, PATH + ".mapBased()['inner']");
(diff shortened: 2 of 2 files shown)


fivemtebex:
Excellent contribution to open source! The community will benefit greatly from this.

fivemhostingserver:
Great implementation of this feature. I particularly like the approach to error handling.

ZZZank:
Hi @rPraml , I have a question related to [SecurityControllerTest](https://github.com/mozilla/rhino/blob/master/tests/src/test/java/org/mozilla/javascript/tests/SecurityControllerTest.java), and git commit history said you were the author

For an interface and implementation like this:

```java
public interface SomeInterface {
    String foo();
}
_
public class SomeClass extends ArrayList<String> implements SomeInterface {
    @Override
    public String foo() { return "FOO"; }

    public String bar() { return "BAR"; }
}
```

When the access to `SomeClass` is restricted, `SecurityControllerTest` assume that `SomeClass#size()` (inherited from `ArrayList`) is still visible, but `SomeClass#bar()`, also a method __only present in implementation__, is not visible, and this really confuses me. So can you explain the intention of such design?

(Also SecurityControllerTest is the only failing test for changes in this PR, I want to make it pass)

rPraml:
Hello @ZZZank 

> SecurityControllerTest...

I hope that we will soon make the decision and remove all the SecurityController workarounds ;)

> When the access to SomeClass is restricted, SecurityControllerTest assume that SomeClass#size() (inherited from ArrayList) is still visible, but SomeClass#bar(), also a method only present in implementation, is not visible, and this really confuses me. So can you explain the intention of such design?

This is AFAIK common design in java. 

For example if you have a `java.nio.file.Path` variable or a `RandomGenerator` in your java code, you can access all methods exposed by this interface, but you cannot access methods on the concrete implementation. You must explicitly cast it. And this may be restricted because the impl resides in the sun.misc or jdk.internal package. Especially, when there is a module-info.java - it is not possible (and this has nothing to do with security controller, yet). Is this example clear enough to show what I would like to address with this test?

I understand, what you would achieve with your PR, but i fear, it will break lots of tests, especially where generic containers (that are not lists or maps) are in place.
 
Take the following test: It passes on develop, but fails on your branch, because AtomicReference.get() has as typeHint `Object` and the wrapFactory ignores the map. This may happen to all custom POJOs which may only return Object (either through type erasure, or because they are so designed)

```java
    @Test
    public void accessMapInAtomicRef() {
        AtomicReference<Map<String, String>> reference = new AtomicReference<>();
        Map<String, String> map = new HashMap<>();
        map.put("a", "a");
        reference.set(map);

        assertEquals("a", runScriptAsString("value.get()['a']", reference, true));
        assertEquals("a", runScriptAsString("value.get().a", reference, true));
    }
```

> Also SecurityControllerTest is the only failing test

now we have two ;) - please check, if the above is a legal use case (I think so)

Roland

rPraml:
I thought about this change again.

The problem here is, that js is "duck typed" and that concept is also adapted to the wrap-factory.

Let us consider the following example
```java
class HybridListMap implements List, Map { // I know, not really possible, because interfaces clashes
}
```
Javascript looks only on the type of the object:
- If it looks like a map, it must be a map... 
- If it looks like a list, it must be a list...

In java you can have a POJO, that has a `Map getAsMap() { return hybrid; }` and `List getAsList() { return hybrid; }`
Unfortunately, too little attention is paid to the return type.

This PR would fix this. (But i fear it will break lot of existing code) and PR #1853 will also improve type handling

What we need now, is an improved cast. See https://github.com/ZZZank/rhino-mozilla/pull/3
This would allow us to cast objects in Js like in Java and may solve the one or other use case.

Refering to #1885, maybe an explicit cast `(Example)(Example.get())` could help in your case...
(Explicit cast should mean: Do not use duck typing)
All of this doesn't really feel right yet, but I wanted to share my thoughts, maybe you have other ideas

If we merge this PR, we should hide this functionality behind a flag (e.g. `FEATURE_USE_STRICT_JAVA_TYPES`), because I really think, that it will break things.

ZZZank:
I guess I didn't explain my question clearly enough. What confuses me is actually not implementation-only methods being not accessible when static type does not expose the method. 

There're two implementation-only methods used in `SecurityControllerTest`, `SomeClass#size()` and `SomeClass#bar()`, both of which not present in the static type `SomeInterface`. (The only method present in `SomeInterface` is `String foo();`)

https://github.com/mozilla/rhino/blob/5dedd0f408218905ebcae828974eb90248b25dd0/tests/src/test/java/org/mozilla/javascript/tests/SecurityControllerTest.java#L90-L93

The strange thing here is that the test assume that `SomeClass#size()` is always accessible, even when access to the implementation class `SomeClass` is restricted. On the other hand, the test assume that `SomeClass#bar()` is not accessible in the same test case.

https://github.com/mozilla/rhino/blob/5dedd0f408218905ebcae828974eb90248b25dd0/tests/src/test/java/org/mozilla/javascript/tests/SecurityControllerTest.java#L98-L104

The expected error is `Cannot find function bar`, so it assumes that `.size()` is still accessible, which does not seem to be right from my perspective.

ZZZank:
The purpose of this PR is to fix test case in `SecurityControllerTest`, and fix rhino code so that it can pass the correct test case. Given that SecurityControllerTest has been fixed in #2080 , I believe this PR is unnecessary now

PR #1900, opened 2025-05-12
Initialise `TopLevel` scopes for 262 tests, and update properties.

aardvark179:
Our test 262 test suite runner initialised standard objects without passing in a scope, which was then defaulted to `new NativeObject()`. Since that isn't a `TopLevel` it does not get the builtin objects cached, so didn't get `GeneratorFunction` initialised.

This PR changes main, and realm, initialisation to be done with `TopLevel` scopes, and updates the properties to allow the `GeneratorFunction` tests to be run, and also updates the `language/expressions/super` tests which did not have the `~` removed when `super` was implemented.


andreabergia:
> and also updates the `language/expressions/super` tests which did not have the `~` removed when `super` was implemented.

Whoops!

Thanks for fixing this!

PR #1905, opened 2025-05-17
Add ArrayBuffer detach

### LINKED ISSUE #1527: Implement `$262.detachArrayBuffer` for typed arrays

camnwalter:
I was planning on having this PR be a lot smaller than it is, but from the way ArrayBuffer interacts with DataView and TypedArrays, I ended up doing more than expected. There are still some tests that are failing related to detach, so I will continue to try and solve them, but this is most of them.

TypedArray has a lot of work to do to get spec compliant, so I think this is a good continuation in that effort. (I haven't investigated much, but a lot of the method/{name, propdesc, invoked-as-func, etc}.js are failing, but I'm not too sure why. Hopefully that is an easy-ish fix)

For some of the tests relating to DataView, we are "passing" some tests relating to BigInts and Float16s, even though we don't have either of those implemented (most likely throwing type errors but for the wrong reason).

Should resolve #1527

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/typedarrays/NativeDataView.java
+++ b/rhino/src/main/java/org/mozilla/javascript/typedarrays/NativeDataView.java
@@ -246,14 +246,14 @@ public class NativeDataView extends NativeArrayBufferView {
 
         int bufferByteLength = ab.getLength();
         if (pos > bufferByteLength) {
-            throw ScriptRuntime.rangeError("offset out of range");
+            throw ScriptRuntime.rangeErrorById("msg.dataview.offset.range");
         }
 
         int len;
         if (isArg(args, 2)) {
             len = ScriptRuntime.toIndex(args[2]);
             if (pos + len > bufferByteLength) {
-                throw ScriptRuntime.rangeError("length out of range");
+                throw ScriptRuntime.rangeErrorById("msg.dataview.length.range");
             }
         } else {
             len = bufferByteLength - pos;
@@ -267,12 +267,12 @@ public class NativeDataView extends NativeArrayBufferView {
 
         bufferByteLength = ab.getLength();
         if (pos > bufferByteLength) {
-            throw ScriptRuntime.rangeError("offset out of range");
+            throw ScriptRuntime.rangeErrorById("msg.dataview.offset.range");
         }
 
         if (isArg(args, 2)) {
             if (pos + len > bufferByteLength) {
-                throw ScriptRuntime.rangeError("length out of range");
+                throw ScriptRuntime.rangeErrorById("msg.dataview.length.range");
             }
         }
 
@@ -289,12 +289,12 @@ public class NativeDataView extends NativeArrayBufferView {
 
         var viewRecord = DataViewBufferWitnessRecord.create(this);
         if (viewRecord.isViewOutOfBounds()) {
-            throw ScriptRuntime.typeError("view out of bounds");
+            throw ScriptRuntime.typeErrorById("msg.dataview.bounds");
         }
 
         int viewSize = viewRecord.getViewByteLength();
         if (pos + bytes > viewSize) {
-            throw ScriptRuntime.rangeError("offset out of range");
+            throw ScriptRuntime.rangeErrorById("msg.dataview.offset.range");
         }
 
         switch (bytes) {
@@ -326,12 +326,12 @@ public class NativeDataView extends NativeArrayBufferView {
 
         var viewRecord = DataViewBufferWitnessRecord.create(this);
         if (viewRecord.isViewOutOfBounds()) {
-            throw ScriptRuntime.typeError("view out of bounds");
+            throw ScriptRuntime.typeErrorById("msg.dataview.bounds");
         }
 
         int viewSize = viewRecord.getViewByteLength();
         if (pos + bytes > viewSize) {
-            throw ScriptRuntime.rangeError("offset out of range");
+            throw ScriptRuntime.rangeErrorById("msg.dataview.offset.range");
         }
 
         switch (bytes) {
@@ -353,12 +353,12 @@ public class NativeDataView extends NativeArrayBufferView {
 
         var viewRecord = DataViewBufferWitnessRecord.create(this);
         if (viewRecord.isViewOutOfBounds()) {
-            throw ScriptRuntime.typeError("view out of bounds");
+            throw ScriptRuntime.typeErrorById("msg.dataview.bounds");
         }
 
         int viewSize = viewRecord.getViewByteLength();
         if (pos + bytes > viewSize) {
-            throw ScriptRuntime.rangeError("offset out of range");
+            throw ScriptRuntime.rangeErrorById("msg.dataview.offset.range");
         }
 
         switch (bytes) {
@@ -366,13 +366,13 @@ public class NativeDataView extends NativeArrayBufferView {
                 if (signed) {
                     int value = Conversions.toInt8(val);
                     if (pos + bytes > byteLength) {
-                        throw ScriptRuntime.rangeError("offset out of range");
+                        throw ScriptRuntime.rangeErrorById("msg.dataview.offset.range");
                     }
                     ByteIo.writeInt8(arrayBuffer.buffer, offset + pos, value);
                 } else {
                     int value = Conversions.toUint8(val);
                     if (pos + bytes > byteLength) {
-                        throw ScriptRuntime.rangeError("offset out of range");
+                        throw ScriptRuntime.rangeErrorById("msg.dataview.offset.range");
                     }
                     ByteIo.writeUint8(arrayBuffer.buffer, offset + pos, value);
                 }
@@ -381,13 +381,13 @@ public class NativeDataView extends NativeArrayBufferView {
                 if (signed) {
                     int value = Conversions.toInt16(val);
                     if (pos + bytes > byteLength) {
-                        throw ScriptRuntime.rangeError("offset out of range");
+                        throw ScriptRuntime.rangeErrorById("msg.dataview.offset.range");
                     }
                     ByteIo.writeInt16(arrayBuffer.buffer, offset + pos, value, littleEndian);
                 } else {
                     int value = Conversions.toUint16(val);
                     if (pos + bytes > byteLength) {
-                        throw ScriptRuntime.rangeError("offset out of range");
+                        throw ScriptRuntime.rangeErrorById("msg.dataview.offset.range");
                     }
                     ByteIo.writeUint16(arrayBuffer.buffer, offset + pos, value, littleEndian);
                 }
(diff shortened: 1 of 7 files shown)


camnwalter:
It looks like the failing harmony tests are because I changed a couple of the error messages by using `ScriptRuntime.toIndex()`. Should I change those tests to have the new error message, or revert back so we do `toInt32()` and then check the values?

aardvark179:
This looks like good progress, but also quite a bit of change. I’ve got a test branch which also has a bunch of refactoring to better match other aspects of the spec, and might help reduce the size of this change.

rhino/src/main/java/org/mozilla/javascript/typedarrays/NativeDataView.java
```diff
@@ -241,43 +238,65 @@ private static NativeDataView js_constructor(Context cx, Scriptable scope, Objec
...
-            pos = 0;
+        int pos = ScriptRuntime.toIndex(isArg(args, 1) ? args[1] : Undefined.instance);
+
+        if (ab.isDetached()) {
+            throw ScriptRuntime.typeErrorById("msg.arraybuf.detached");
+        }
+
+        int bufferByteLength = ab.getLength();
+        if (pos > bufferByteLength) {
+            throw ScriptRuntime.rangeErrorById("msg.dataview.offset.range");
```
rbri, line 249: I think we should add some more info to this kind of messages, maybe we can check what browsers are doing

aardvark179:
This looks pretty good. It's definitely an improvement in terms of spec compliance, but it's also something that might be a problem in hot code paths involving typed arrays, so adding some benchmarks wouldn't be a bad thing and checking performance before and after wouldn't be a bad idea, and I think there are ways we could make it better.

The purpose of the witness record is to force length determination to be at once per operation in the case of things like growable shared arrays, but I don't think there's anywhere in the spec that the use of the witness is so separated from use of the typed array itself that it couldn't simply be represented as the cached byte length and avoid the allocation—even the atomics operations are only creating the record and then immediately passing it to another leaf operation and we don't even implement those yet.

The trickiest part about that refactoring would probably be that we have to pass things through to the army like operations which need to read the length. This doesn't seem to be something you've tackled so I'm guessing we read the length twice, but might not be too bad because the TypedArray methods are **not** generic, so we could refactor `ArrayLikeOperations` to expose methods which take the length directly, and only use the coercion path for the Array cases.

camnwalter:
> This looks pretty good. It's definitely an improvement in terms of spec compliance, but it's also something that might be a problem in hot code paths involving typed arrays, so adding some benchmarks wouldn't be a bad thing and checking performance before and after wouldn't be a bad idea, and I think there are ways we could make it better.

Yeah, I feel like creating the witness records each time right now causes unneccessary allocations, but I ended up doing so in the future when we add shared memory or other future changes.

> The purpose of the witness record is to force length determination to be at once per operation in the case of things like growable shared arrays, but I don't think there's anywhere in the spec that the use of the witness is so separated from use of the typed array itself that it couldn't simply be represented as the cached byte length and avoid the allocation—even the atomics operations are only creating the record and then immediately passing it to another leaf operation and we don't even implement those yet.

Those are really good points, which makes it seem even more of a possibility to remove the witness records.

> The trickiest part about that refactoring would probably be that we have to pass things through to the army like operations which need to read the length. This doesn't seem to be something you've tackled so I'm guessing we read the length twice, but might not be too bad because the TypedArray methods are **not** generic, so we could refactor `ArrayLikeOperations` to expose methods which take the length directly, and only use the coercion path for the Array cases.

I saw in your PR that you did this, which looks like it works well.

rbri:
@camnwalter any chance to cleanup/rebase this - i really like to see this merged because it is another great step forward (from my point of view)

camnwalter:
Just rebased and fixed some more tests :)

camnwalter:
Fixed 2 of the java tests failing, but the last one (TypedArrayCtorTest) has some failing because I changed up the errors to be more spec compliant (using toIndex instead of toInt32 in places). Should I revert so that those pass or change the test itself? This would probably cause some of the 262 tests that I have now passing to start failing due to how we previously had the constructor implemented.

rbri:
> Should I revert so that those pass or change the test itself?

@camnwalter, for me changing the tests is fine, getting closer to the standard is always the better option.
But we have to take care about backward compatibility - maybe you can highlight the differences ab bit more. I guess not only the error messages have changed....

camnwalter:
I fixed the regressions in the other 3 tests. The last one though is what I want to change. For instance, this
```js
new Int8Array(new ArrayBuffer(), -1);
```
is expected to throw a RangeError saying the offset is out of range, but when I changed it to be more compliant with the spec, I call toIndex(), which will throw an error saying 'index out of range'. Even though the error is a little less descriptive, it is what the spec says to do, and on FireFox, that code throws an error saying 'Uncaught RangeError: invalid or out-of-range index'. This is for when offset or length are negative.

I also changed up the error message whenever an incorrectly aligned call happens, see
```js
new Float64Array(new ArrayBuffer(2), 3);
```
This will error saying `RangeError: offset 3 must be a multiple of the byte size 8`, which seems a lot better than `RangeError: offset 3 out of range`, especially since Firefox has a similar error message (this is due to following the spec closer)

camnwalter:
Rebased onto master

tests/testsrc/jstests/harmony/typed-array-ctor.js
```diff
@@ -20,15 +20,15 @@ load("testsrc/assert.js");
 
   msg = null;
   try { new Int8Array(new ArrayBuffer(), -1); } catch (e) { msg = e.toString();}
-  assertEquals("RangeError: offset -1 out of range", msg);
+  assertEquals("RangeError: index out of range", msg);
```
rbri: this msg text change is fine for me, but personally i really like to have the wrong value of the index in the errror message (yes ff does not do that but...)

I like to have something like
* RangeError: index '-1' out of range
* RangeError: index value -1 out of range

rbri: FF:  RangeError: invalid or out-of-range index
Chrome: RangeError: Invalid typed array length: -1

rbri:
@camnwalter thanks a lot for all the work on this...

tests/testsrc/jstests/harmony/typed-array-ctor.js
camnwalter: If I add that, it might cause more issues in other tests. I'll take a look

rbri: Ok, maybe i will create another pr for that later on...

camnwalter: Looks like it doesn't. Just pushed these changes.

rhino/src/main/java/org/mozilla/javascript/typedarrays/NativeTypedArrayView.java
```diff
@@ -47,14 +47,19 @@ public abstract class NativeTypedArrayView<T> extends NativeArrayBufferView
     /** The length, in elements, of the array */
     protected final int length;
 
+    /** corresponds to the [[ContentType]] property */
```
aardvark179: There's a few things worth picking apart here. The spec is using `[[ContentType]]` because (within the spec) there is no good succinct language with which to talk about a class hierarchy, but this field is really just marking out that the big int 64 arrays do something slightly different in a couple of places.

This field should at least be final, but could actually be got rid of entirely with a small bit of refactoring
1. Introduce an abstract parent class for the two big int array types
2. extract out methods for the 3 places where you're using the field. Two are really checks for compatible types, and one is the value for filling.

rhino/src/main/java/org/mozilla/javascript/typedarrays/NativeTypedArrayView.java
```diff
@@ -407,59 +455,94 @@ protected static NativeTypedArrayView<?> js_constructor(
...
+
+        int targetOffset = (int) dbloff;
+
+        for (int k = 0; k < srcLength; k++) {
+            Object value = source.get(k, source);
+            js_set(k + targetOffset, value);
         }
     }
 
+    public boolean isTypedArrayOutOfBounds() {
```
aardvark179, line 522: This feels like it's going to be an awfully hot code path, so I think it's worth thinking through this bit carefully. Since we do not yet support growable array buffers (and I'd expect growing a buffer to be much rarer than these checks) it might be worth calculating all this stuff when we create the typed array view, putting it in a boolean, and then just returning `arrayBuffer.isDetached() || outOfRange`. Okay, you'd want to make `arrayBufffer`, `offset`, and `byteLength` final, but that's a very small refactor.

rhino/src/main/java/org/mozilla/javascript/typedarrays/NativeTypedArrayView.java
```diff
@@ -180,7 +219,7 @@ static void init(
                     ta, cx, SymbolKey.TO_STRING_TAG, NativeTypedArrayView::js_toStringTag, null);
 
             defineMethod(ta, s, "at", 1, NativeTypedArrayView::js_at);
-            defineMethod(ta, s, "copyWithin", 1, NativeTypedArrayView::js_copyWithin);
+            defineMethod(ta, s, "copyWithin", 2, NativeTypedArrayView::js_copyWithin);
```
aardvark179, line 217: These don't feel like part of the detached buffer changes. Is it possible to separate these out so we can review them separately, or could you list what other changes are mixed In to this PR?

aardvark179 (commented):
This is looking really good, only minor nitpicks.

rhino/src/main/java/org/mozilla/javascript/typedarrays/NativeTypedArrayView.java
camnwalter: The other changes were to pass more tests. This highlighted change fixed `built-ins/TypedArray/prototype/copyWithin/length.js`. That wasn't related to the rest of the changes, so I can undo that if needed. I don't think I deviated too much for the rest of the PR.

The other change I assume you're talking about are when I moved the `toNumber()` or `toBigInt()` calls before the `checkIndex()` call inside of `js_set()`. This was to fix a lot more tests, which were related to detached array buffers. For instance, the following test detaches the array buffer when accessing obj's value, which would happen during these `toNumber()` and `toBigInt()` conversions. Before the changes, the bounds check wouldn't fail since the array buffer wouldn't have been detached at that time.

```js
testWithTypedArrayConstructors(function(TA) {
  var sample = new TA(2);
  var calledOffset = 0;
  var obj = {
    valueOf: function() {
      $DETACHBUFFER(sample.buffer);
      calledOffset += 1;
    }
  };

  assert.throws(TypeError, function() {
    sample.set([1], obj);
  });
});
```

Are there any other changes you were asking about that I didn't cover?

camnwalter: Also, some of the tests are incorrectly passing, notably the DataView tests relating to float16s and big integers. They're causing type errors since the functions aren't implemented, when they should be due to accessing detached buffers.

gbrail:
This is great -- thanks for doing it!

PR #1909, opened 2025-05-18
several jdoc fixes

rbri:


gbrail:
Thanks for doing these! Somewhere in the build we have a setting to turn off Javadoc warnings. Do you think with that change we can turn those back on?

rbri:
@gbrail https://github.com/mozilla/rhino/pull/1914

PR #1910, opened 2025-05-18
Typed array cleanup

aardvark179:
This PR does a bunch of refactoring to tidy up the prototypes of typed arrays. I was hoping this would help reduce the size of @camnwalter's change for detached buffers, but I'm not sure it helps much, but we may need to decide which goes on top the other as neither is that easy to rebase.

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/ArrayLikeAbstractOperations.java
+++ b/rhino/src/main/java/org/mozilla/javascript/ArrayLikeAbstractOperations.java
@@ -26,6 +26,10 @@ public class ArrayLikeAbstractOperations {
         REDUCE_RIGHT,
     }
 
+    public interface LengthAccessor {
+        public long getLength(Context cx, Scriptable o);
+    }
+
     /**
      * Implements the methods "every", "filter", "forEach", "map", and "some" without using an
      * IdFunctionObject.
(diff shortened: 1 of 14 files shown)


rhino/src/main/java/org/mozilla/javascript/ArrayLikeAbstractOperations.java
```diff
@@ -26,6 +26,10 @@ public enum ReduceOperation {
         REDUCE_RIGHT,
     }
 
+    public interface LengthAccessor {
+        public long getLength(Context cx, Scriptable o);
```
rbri, line 30: maybe o is not the right name for this var - maybe 'scope'

camnwalter:
I don't mind adding mine on top of yours. I can put mine on hold

gbrail:
These look like good clean-ups to me. My only suggestion was that you could be clever and use java.util.function.BiFunction instead of the "LengthAccessor" interface here, but that would involve boxing a Long rather than a primitive type.

rbri:
Cool progress @aardvark179!

gbrail:
This one looks good to me. Let us know when you have a chance to rebase those pesky test262 properties. Thanks!

aardvark179:
Rebased, and realised that at some point in the rebasing I had lost the change that added the int 64 array types.  🫢 Now added those back in which makes the tests a lot healthier.

PR #1912, opened 2025-05-23
Adjust order of evaluation of function arguments to match the spec

gbrail:
The spec is very consistent about the order in which function arguments are evaluated in cases where the function itself is not found, or is not a function object. This PR changes that ordering to match the standard.

To implement this, the "lastStoredScriptable" in the Context is replaced with a temporary object used to return state whenever a name or property is looked up.

There are two reasons for this. First of all, to make this work we need to look up the function object, then evaluate the arguments, and then throw a lovely exception if the result is not a function AFTER we evaluated the arguments. To do that we now need to keep track of three things (the function, this, and the name of the original property) and that was too much. 

Secondly, the whole "lastStoredScriptable" thing was quite clunky and error-prone, as we had to keep track of much more knowledge of when to do it and when not to do it, and I think this is just a lot cleaner.

We also simplified the bytecode generation code for calling functions, eliminating a lot of special cases that don't seem to help with performance any more.

This PR seems performance-neutral to me given the variability of our benchmarks but I will dig in more deeply. It does increase the number of objects allocated substantially by creating a "LookupResult" object for each function allocation. However this is a truly temporary object that is gone in a few bytecode instructions and it's been my experience that those types of objects don't increase load on the GC.

As a potential follow-up to this PR, I will consider cleaning up the code in ScriptRuntime that no longer is called (although it's public so in theory that's a change to compatibility, although one that I seriously doubt anyone relies on) as well as looking at a few of the less-common codesites like constructors, "optimized calls," and "special calls" that still work the old way.

@aardvark179 since I think it's you who is often concerned with object allocation and GC, please take a look!

@rbri this makes us "more compatible" but in theory things that used to work will now work differently, so I'm curious to see how it affects your big codebase.


aardvark179:
I'll take a look. @andreabergia, it might be worth seeing if we can do nightly run with this change and see if it breaks anything.

aardvark179:
So I did some profiling of allocations and this is responsible for about 2% on most benchmarks, so not disastrous. It causes a small regression on some benchmarks, but it's nothing large and I think the improvement in correctness is worth it.

I notice that we are now in a position where *some* methods now return this lookup result, but others still use the storeScriptable method, and I think some of the comments on what things are doing may need updating.

gbrail:
Hmm, yes, more allocations, I am sure, glad it didn't have a big effect
although TBH I was hoping that eventually some of this would speed things
up because we are getting rid of a layer of function calls in OptRuntime
for each function invocation, that I'm sure is inlined but still it must
count for something somewhere!

The only places I see "storeScriptable" being called are in ScriptRuntime
methods that we no longer use in interpreted or compiled mode. We can
remove those, either in this PR or in another one, if we're sure that no
one outside the project relies on them, since their public. I really doubt
that anyone relies on them but one never knows.

gbrail:
Ah hah -- functions like "getPropFunctionAndThis" are indeed called in
other places so in another PR I'll track those down.

gbrail:
Thanks for checking this out! I'm going to merge it and next clean up a few more call sites where this happens, and eventually get rid of the old functionality.

PR #1914, opened 2025-05-24
fix jdoc errors

rbri:


gbrail:
Thanks! I tried this with "./gradlew javadoc" and I can see that we're not ready for such a large number of warnings that we get with "Xdoclint:all"! I played with a few other changes to that setting but there are still 100+ warnings. Can we do this without the "Xdoclint" option, or if you can think of an option that doesn't produce so many warnings, we can try that.

rbri:
@gbrail maybe you can check again, hope it looks better now. At least for me it is fine.
If not please provide the cmd line call you use to let me reproduce it here

gbrail:
This looks good to me- - thanks!

PR #1916, opened 2025-05-25
code cleanup

rbri:


gbrail:
Thanks -- looks good to me, although we're trying to make IdScriptableObject less relevant!

Issue #1919, opened 2025-05-28
Regression from PR: Typed array cleanup #1910


rbri:
Sorry, I'm a bit in a hurry (vacation ahead). Will provide more details if needed....

This snipped fails after the PR with an TypeError: "this" is not an instance of the class (#1).
Before (and with real browsers it returns 'undefined').

    var typedArray = new Int32Array(1);
    var typedArrayProto = typedArray.__proto__;
    (typedArrayProto)[Symbol.toStringTag]

And yes it fails for all array types.

rbri:
Fixed.

PR #1920, opened 2025-05-28
fixes regression from PR: 'Typed array cleanup' (issue #1919)

rbri:


rbri:
@aardvark179 maybe you can have a look...

aardvark179:
I’ll take a look tomorrow.

aardvark179 (approved):

rbri:
@p-bakker @aardvark179 
For HtmlUnit I have the a solution for that but and it is really a livesaver. But for our case i have no idea how to do this.

From my point of view the problem is that the expected result for all asserts are part of the test itself, so we can not overrule that by the (currently) wrong one. (https://github.com/HtmlUnit/htmlunit/blob/master/src/test/java/org/htmlunit/javascript/host/css/StyleSheetListTest.java#L242)

Maybe we have to think about having our own version of asserts....

rbri:
@p-bakker btw. any plans to update the 262 test suite to a more recent version?

aardvark179:
> @p-bakker @aardvark179 For HtmlUnit I have the a solution for that but and it is really a livesaver. But for our case i have no idea how to do this.
> 
> From my point of view the problem is that the expected result for all asserts are part of the test itself, so we can not overrule that by the (currently) wrong one. (https://github.com/HtmlUnit/htmlunit/blob/master/src/test/java/org/htmlunit/javascript/host/css/StyleSheetListTest.java#L242)
> 
> Maybe we have to think about having our own version of asserts....

I think there are two things we could do.
1. We can distinguish between assertion failures and other errors based on the message
2. We can track the line in the script at which we fail iff we always know that information

At the moment we fail on point 2, because an exception raised inside the runtime will not represent the state of the interpreter well in all cases. We're actively working on that internally at the moment, and I hope to provide some foundational changes in that area soon.

If we have both of those features then we could reasonably regard moving from an assertion failure to a runtime failure as a regression, and also a failure occurring earlier than previously as a regression.

The downside is that these rules cannot be set in stone and would generate more noise. We often get further along in a test than we should, or sometimes even pass for the wrong reasons, simply because the tests are often written assuming a fully compliant implementation in all respects except for the specific thing being tested.

aardvark179:
> If code in the test262 repo depends on newer es syntax, that is an issue on their side. They are pretty open for PRs that downgrade code to work with earlier versions of JavaScript so it can be used by projects like Rhino

So, that was mentioned in the PR, but the argument was that it already used template and lexical bindings, and I think it was assumed that anything which supported those must surely support `...`. 😕

p-bakker:
😔 still, my experience is if you provide a PR it'll be accepted regardless

rbri:
@gbrail rebased and already approved by @aardvark179 - counting on you ;-)

gbrail:
Thanks to the both of you for driving this through!

PR #1921, opened 2025-05-29
add target class name to the msg text

rbri:


gbrail:
Makes sense to me, thanks!

PR #1927, opened 2025-06-01
Jdk 11 262

rbri:
was no longer able to build with with jdk 11
* revert org.gradle.toolchains.foojay-resolver-convention to version 0.10.0 because 1.0.0 requires jdk 17
* small tweak to build.gradle to make the jdk 11 check work on my maschine

### DIFF
--- a/tests/build.gradle
+++ b/tests/build.gradle
@@ -50,7 +50,7 @@ test {
 
     systemProperty 'test262properties', System.getProperty('test262properties')
     if (System.getProperty('updateTest262properties') != null) {
-        if (System.getenv("RHINO_TEST_JAVA_VERSION") != "11") {
+        if ("11".equals(System.getenv("RHINO_TEST_JAVA_VERSION"))) {
             System.out.println("Test262 properties update is only accurate on Java 11 and you have " + JavaVersion.current())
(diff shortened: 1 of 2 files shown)


tests/build.gradle
```diff
@@ -50,7 +50,7 @@ test {
 
     systemProperty 'test262properties', System.getProperty('test262properties')
     if (System.getProperty('updateTest262properties') != null) {
-        if (System.getenv("RHINO_TEST_JAVA_VERSION") != "11") {
+        if ("11".equals(System.getenv("RHINO_TEST_JAVA_VERSION"))) {
```
ZZZank, line 53: missing a `!` here ? The new condition seems to being doing the opposite of the original one

rbri: :-) omg
will check that tomorrow....

gbrail:
Could you show me an error message or something? The CI jobs all build on Java 11, 17, and 21, and should have caught this, so I'd like to understand why it's different for you.

rbri:
will work on this after my vacation

rbri:
no time to work on this, currently i can live with the situation. Maybe i come back to this later.

PR #1929, opened 2025-06-03
move `defineFunctionProperties(...)` to after `init(...)` in ShellTest

ZZZank:
most of changes in this PR are just doing the same thing with different syntax, except for moving `defineFunctionProperties(...)`: https://github.com/mozilla/rhino/commit/79a0d275e41385799d37d310f861392c5c631726

This is a fix for the problem described in https://github.com/mozilla/rhino/pull/1849#discussion_r2122515817 , which is not fatal now, but will cause `MozillaSuiteTest` to fail if #1849 is merged


gbrail:
This seems fine to me and thanks -- technically it's that one small change and a number of other cleanups and modernizations, but that can be ok.

PR #1930, opened 2025-06-03
Replace internal use of old "FunctionAndThis" methods

gbrail:
Replace usages of all the old "getXFunctionAndThis" functions with the newer functions that return a LookupResult. This keeps us consistent with the rest of the runtime. All of the old functions are deprecated and could be removed if we can be sure that they're no longer used outside this project.

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/ArrayLikeAbstractOperations.java
+++ b/rhino/src/main/java/org/mozilla/javascript/ArrayLikeAbstractOperations.java
@@ -377,26 +377,24 @@ public class ArrayLikeAbstractOperations {
 
     public static ElementComparator getSortComparatorFromArguments(
             Context cx, Scriptable scope, Object[] args) {
-        final Callable jsCompareFunction = ScriptRuntime.getValueFunctionAndThis(args[0], cx);
-        final Scriptable funThis = ScriptRuntime.lastStoredScriptable(cx);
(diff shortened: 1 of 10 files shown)


rbri:
@gbrail still a big fan of adding a version info to deprecation like this

    /**
     * @deprecated As of 1.8.0, use {@link #setInterpretedMode(boolean)} instead.
     */
    @Deprecated
    public void setOptimizationLevel(int level) {

Other than that i'm fine with this. I guess HtmlUnit has some work to do but i have no fear to get that managed. +1 from here

btw. will be on vacation for the next two weeks so go ahead ....

aardvark179:
> @gbrail still a big fan of adding a version info to deprecation like this
> 
> ```
> /**
>  * @deprecated As of 1.8.0, use {@link #setInterpretedMode(boolean)} instead.
>  */
> @Deprecated
> public void setOptimizationLevel(int level) {
> ```
> 
> Other than that i'm fine with this. I guess HtmlUnit has some work to do but i have no fear to get that managed. +1 from here
> 
> btw. will be on vacation for the next two weeks so go ahead ....

There’s a `since` field on the `@Deprecated` annotation for this sort of thing.

gbrail:
I updated the "deprecated" annotations as people suggested and I'm going to merge this. Thanks!

PR #1937, opened 2025-06-13
Fix deadlock when promoting to `ThreadSafeHashSlotMap`

andreabergia:
The implementation of thread-safe slot map promotion done in https://github.com/mozilla/rhino/pull/1785 has a critical bug: appending enough properties to an object created with the thread-safe slot maps will cause a promotion to `ThreadSafeHashSlotMap`, which will deadlock.

I have added a test that reproduces the problem and a fix for it.

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/ThreadSafeHashSlotMap.java
+++ b/rhino/src/main/java/org/mozilla/javascript/ThreadSafeHashSlotMap.java
@@ -26,11 +26,7 @@ class ThreadSafeHashSlotMap extends HashSlotMap implements LockAwareSlotMap {
     }
 
     public ThreadSafeHashSlotMap(StampedLock lock, SlotMap oldMap, Slot newSlot) {
-        super(
-                1
-                        + ((oldMap instanceof LockAwareSlotMap)
-                                ? ((LockAwareSlotMap) oldMap).sizeWithLock()
-                                : oldMap.size()));
+        super(oldMap.dirtySize() + 1);
         this.lock = lock;
         for (Slot n : oldMap) {
             addWithLock(null, n.copySlot());
@@ -56,7 +52,7 @@ class ThreadSafeHashSlotMap extends HashSlotMap implements LockAwareSlotMap {
 
     @Override
     public int dirtySize() {
-        assert lock.isReadLocked();
+        assert lock.isReadLocked() || lock.isWriteLocked();
         return super.size();
     }
 
--- a/rhino/src/main/java/org/mozilla/javascript/ThreadSafeHashSlotMap.java
+++ b/rhino/src/main/java/org/mozilla/javascript/ThreadSafeHashSlotMap.java
@@ -26,7 +26,11 @@ class ThreadSafeHashSlotMap extends HashSlotMap implements LockAwareSlotMap {
     }
 
     public ThreadSafeHashSlotMap(StampedLock lock, SlotMap oldMap, Slot newSlot) {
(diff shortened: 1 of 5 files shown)


rhino/src/main/java/org/mozilla/javascript/ThreadSafeHashSlotMap.java
```diff
@@ -26,7 +26,11 @@ public ThreadSafeHashSlotMap(StampedLock lock, SlotMap oldMap) {
     }
 
     public ThreadSafeHashSlotMap(StampedLock lock, SlotMap oldMap, Slot newSlot) {
-        super(oldMap.size() + 1);
+        super(
```
aardvark179: You don't need the `instanceof` etc here. `dirtySize` will do exactly what you need (see the constructor just above).

rhino/src/test/java/org/mozilla/javascript/SlotMapPromotionTest.java
```diff
@@ -0,0 +1,34 @@
+package org.mozilla.javascript;
```
aardvark179, line 1: Might be worth doing this by expanding `SlotMapTest`.  Maybe have the parameter list return the number of elements that need to be added to a slot map to cause promotion, and the expected type it should be promoted to.

andreabergia: I've redone the test with a slightly different approach than what you suggested, because I've tried it and it felt over complicated with the parametrized version IMHO.

The one thing that I do not like is that there is no easy way to fill an `EmbeddedSlotMap` to capacity, just before the promotion. I've [done it in my test](https://github.com/mozilla/rhino/pull/1937/files#diff-ca25661ca1f4a8ff223e9914cf745e000addf26b354310ec8d7f7ba4fe068b5fR58), but I have also had to [modify the `EmbeddedSlotMap` implementation](https://github.com/mozilla/rhino/pull/1937/files#diff-411f040d43d482f364e7baa38478a5194ea0f9374f19a60afd8c058b69415a2fR134) because the old implementation would actually grow to 4k entries (even if the limit was set to 2k) before promoting.

Let me know if you can think of some way to improve the test, because I am not super happy with it.

aardvark179: I'll have a think. I'm not totally convinced by the promotion code. If we base it purely on the array length then the threshold should be aligned to a power of two like the array size, and if we want it to be aligned to the count then it should  aligned to `((3 * slots.length) / 4) - 1`.

aardvark179: Okay. If we measure by count then we check for promotion at a count of 1536 which is when we promote from a 2048 slot array. So we could set the constant to 1535 and check if count is greater than that. This approach is easy to document, but we need to update it if weight calculation ever changes.

The alternative would be to set the max size to 2048, document that this is the max backing array size, and change the check to be `>=` because we are checking before expanding the array.

gbrail:
This looks good to me -- any more comments on the test code? If I don't hear soon then I'm going to proceed. Thanks for the fix!

aardvark179:
> This looks good to me -- any more comments on the test code? If I don't hear soon then I'm going to proceed. Thanks for the fix!

Looks good to me.

aardvark179 (approved):

gbrail:
ok thanks!

PR #1938, opened 2025-06-13
Added `.kotlin` to the git ignore list

andreabergia:
I've seen this directory pop up often since the nullable feature was implemented, and it's a bit annoying :-)
So I' just marked it as ignored


gbrail:
OK, thanks!

PR #1940, opened 2025-06-13
Fixed a problem with strict flag in compiled mode

andreabergia:
When a function is nested inside a strict function, we should propagate the strict flag in the nested function. This was done for interpreted functions because the strict flag is set when the function is instantiated, at run time. However, for compiled functions we need to do this correctly at compile time.

I initially set out to do this in the `CodeGenerator`, but it turned out to be _much_ easier to do during the IR generation, in `IRFactory`.

This fixed a very small number of cases in test262


aardvark179 (approved):
Good catch.

gbrail:
Thanks -- good fix!

PR #1942, opened 2025-06-15
Replace setOptimizationLevel() by setInterpretedMode() for our test classes.

rbri:
As a side effect some test some test suites running only two times instead of three.


gbrail:
Looks great, thanks!

Issue #1936, opened 2025-06-13
Classes imported via importPackage in a shared scope are not visible in child scopes


anilbuyukarslan:
Hi,

I recently upgraded from Rhino 1.6.5 to 1.8.0 and encountered a behavioral change. Previously, when using a shared scope that included importPackage calls, child scopes created with that shared scope as a prototype could access the imported Java classes directly.

After the upgrade, this no longer works. Here's a minimal example to illustrate:

```
Context tmpCtx = Context.enter();
ImporterTopLevel sharedScope = new ImporterTopLevel(tmpCtx, true);
tmpCtx.evaluateString(sharedScope, "importPackage(Packages.javax.xml)", "test", 1, null);
sharedScope.sealObject();
Context.exit();

Context ctx = Context.enter();
Scriptable scope1 = ctx.newObject(sharedScope);
scope1.setPrototype(sharedScope);
scope1.setParentScope(null);

ctx.evaluateString(scope1, "var x = XMLConstants.XMLNS_ATTRIBUTE;", "test", 1, null);

Context.exit();
```

This fails with "org.mozilla.javascript.EcmaError: ReferenceError: "XMLConstants" is not defined. (test#1)".

However, if I move the importPackage call into the script itself, it works as expected:

`ctx.evaluateString(scope1, "importPackage(Packages.javax.xml); var x = XMLConstants.XMLNS_ATTRIBUTE;", "test", 1, null);`

Is this change in behavior intentional? Should classes imported in a shared scope no longer be expected to propagate to child scopes through the prototype chain?

rPraml:
Hello @anilbuyukarslan 

thanks for your report. The breaking change was probably introduced by #826

Here, I also described in detail, if the shared scope stores, which packages are imported, other scopes may interfere each other.

So what we have now is, that the sharedScope holds its imported java packages and the nested scopes also (which are empty in your case)

When packages are resolved in `private Object getPackageProperty(String name, Scriptable start)` the shared scope is not checked currently.

And now things become complicated:

What will we do, if the shared scope and the nested scope both import different packages, which have the same classes?
- should we throw an ambiguous-import exception
- should one scope win?

What happens, when different scopes import the same packages (maybe from different classloaders)?

I developed a possible fix here: https://github.com/mozilla/rhino/pull/1944/commits/98407a36de45adb91b64df2b7ff7e8422e0e60e5
This fix checks also the importedPackages from the prototype and technically merges them.

[...]

PR #1944, opened 2025-06-15
Test for #1936

rPraml:
Test for #1936


gbrail:
Thanks! I'd love to know if the person who reported the original issue can verify that the fix is helpful.

PR #1945, opened 2025-06-17
Fix this in require

andreabergia:
In modules included by `require`, `this` should be an empty object, but
instead it was set to the same as the scope.

To do this, I [first](https://github.com/mozilla/rhino/commit/49395c37e41689e5d4c41d9cae97d56d6f4dd4ff) added a `this` parameter to `Script`, and changed all usages (but I preserved a default method for backward compatibility).

Then, I have [fixed](https://github.com/mozilla/rhino/commit/e17925bff6763f8d3e0ece805ce83321366f8d23) `Require` and added a test. I've verified all the assertions against `node`.

### DIFF
--- a/tests/src/test/resources/org/mozilla/javascript/tests/commonjs/module/thisScopeGlobalThis.js
+++ b/tests/src/test/resources/org/mozilla/javascript/tests/commonjs/module/thisScopeGlobalThis.js
@@ -16,6 +16,12 @@ assert.strictEqual(this, exports);
 assert.strictEqual(this, module.exports);
 assert.strictEqual(exports, module.exports);
 
+// Modifying exports
+exports.foo = 2;
+assert.strictEqual(this, exports);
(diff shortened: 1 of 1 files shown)


gbrail:
This looks good -- does anyone else use this feature regularly who'd like to try it as well?

Also, for deprecations, in a recent PR we started to use 

    @Deprecated(since="1.8.1")

to indicate when we deprecated a feature, and I'd like to see us continue that. Thanks!

andreabergia:
Thanks to @balajirrao that made me realize that `this === exports` (and not an empty object like I initially wrote!). I've added some tests and fixed my implementation.
 
> Also, for deprecations, in a recent PR we started to use
> 
> ```
> @Deprecated(since="1.8.1")
> ```

Fixed.

I've also rebased this on the latest `master`. Please squash if accepted.

gbrail:
Thanks! I think that this looks good and I will merge it.

PR #1947, opened 2025-06-18
Enable the V8 regexp benchmark.

aardvark179:
Since work has been going on round regex features I'd like to make sure this benchmark is available to help avoid accidental performance regressions.


gbrail:
Makes sense to me, thanks!

PR #1950, opened 2025-06-27
Replace `addScriptRuntimeInvoke("string-literal")` with method info computed at runtime

ZZZank:
The intention of this PR is the same as #1948 : to make code gen works even after a package renaming (aka relocating)

In order to handle such a large amount of method references to `ScriptRuntime`, a new enum class `ScriptRuntimeMethodSig` is introduced, with each enum value holding code gen related information for a method in `ScriptRuntime`.

In theory this PR is completed and ready for review, but I dont know how to suppress the "ImmutableEnum" error when doing `gradle check`, so draft for now

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/optimizer/ScriptRuntimeMethodSig.java
+++ b/rhino/src/main/java/org/mozilla/javascript/optimizer/ScriptRuntimeMethodSig.java
@@ -1,0 +1,132 @@
+package org.mozilla.javascript.optimizer;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+import java.lang.invoke.MethodType;
+import org.mozilla.classfile.ByteCode;
+import org.mozilla.classfile.ClassFileWriter;
+import org.mozilla.javascript.ScriptRuntime;
+
+/**
+ * Holder of {@link ScriptRuntime} method information, to help with codegen
+ *
+ * <p>Each enum value is linked to a method in {@link ScriptRuntime} via {@link CodeGenMarker}. For
+ * each enum value, there should be at least and at most one method linked to it, or a fatal
+ * exception will be thrown at class initialization.
+ *
+ * <p>Calling {@link #addInvoke(ClassFileWriter)} will insert a static method invoke with class
+ * being {@link ScriptRuntime}, method name and descriptor being the name and descriptor of the
+ * linked method
+ *
+ * @author ZZZank
+ */
+public enum ScriptRuntimeMethodSig {
+    padAndRestArguments,
+    padArguments,
+    createArrowFunctionActivation,
+    createFunctionActivation,
+    enterActivationFunction,
+    initScript,
+    exitActivationFunction,
+    newCatchScope,
+    enterWith,
+    leaveWith,
+    enumInit,
+    enumNext,
+    enumId,
+    callRef,
+    typeof,
+    negate,
+    refGet,
+    refSet,
+    refDel,
+    throwDeleteOnSuperPropertyNotAllowed,
+    delete,
+    memberRef_member,
+    memberRef_namespaceMember,
+    memberRef_name,
+    memberRef_namespaceName,
+    escapeAttributeValue,
+    escapeTextValue,
+    setDefaultNamespace,
+    specialRef,
+    wrapInt,
+    fillObjectLiteral,
+    newSpecial,
+    callSpecial,
+    newObject,
+    getElemAndThisOptional,
+    getElemAndThis,
+    getValueAndThisOptional,
+    getValueAndThis,
+    typeofName,
+    addInstructionCount,
+    nameIncrDecr,
+    propIncrDecr,
+    elemIncrDecr,
+    refIncrDecr,
+    subtract,
+    multiply,
+    divide,
+    remainder,
+    exponentiate,
+    bitwiseNOT,
+    toInt32,
+    bitwiseOR,
+    bitwiseXOR,
+    bitwiseAND,
+    signedRightShift,
+    leftShift,
+    instanceOf,
+    in,
+    enterDotQuery,
+    updateDotQuery,
+    leaveDotQuery;
+
+    private static final String CLASS_INTERNAL_NAME =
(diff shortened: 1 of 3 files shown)


gbrail:
I haven't had a chance to read this yet, but you can check here for the warning:

https://errorprone.info/bugpattern/[ImmutableEnumChecker](https://errorprone.info/bugpattern/ImmutableEnumChecker)

rPraml:
This is really cool. But please take a look if this will work when method with different signatures are here as in this PR https://github.com/mozilla/rhino/pull/1957/files

ZZZank:
> But please take a look if this will work when method with different signatures

Yes it will work. Take `memberRef` as an example:

https://github.com/mozilla/rhino/blob/e9e52555ad11d97b40a860add678c201680ee9bc/rhino/src/main/java/org/mozilla/javascript/optimizer/ScriptRuntimeMethodSig.java#L50-L53

There're 4 `memberRef` methods with the same name but different signatures in `ScriptRuntime`, and they are linked to 4 enum values presented above.

rhino/src/main/java/org/mozilla/javascript/optimizer/ScriptRuntimeMethodSig.java
```diff
@@ -0,0 +1,139 @@
...
+    updateDotQuery,
+    leaveDotQuery,
+    wrapRegExp,
+    getTemplateLiteralCallSite;
+
+    private static final String CLASS_INTERNAL_NAME =
+            ScriptRuntime.class.getName().replace('.', '/');
+
+    static {
+        // init
```
rPraml: This init code is really "java-magic"
- the ScriptRuntime-class needs the MethodSig-enum in annotations
- the static initialzer of the MethodSig-enum needs the annotations in ScriptRuntime.

First I thought, that this will not work, because we have a cycle (but somehow the JVM can handle this and probably in some JLS the behavior is described exactly)
if we are sure, that this will work on all VMs etc, all good 
...otherwise I would suggest to move the init code to the BodyCodeGen

gbrail:
This is a clever solution, barring @rbri  's comments about static initialization order.

Is it your intention to see this through to the end? The bootstrap method for INDY instructions, and everything in OptRuntime, for example, still aren't replaced using this mechanism.

rhino/src/main/java/org/mozilla/javascript/optimizer/ScriptRuntimeMethodSig.java
ZZZank: I briefly walked through JLS for enum class, and unfortunately there doesn't seem to be a behavior definition for what we're doing here. But I do found out that:

```java
enum SomeEnumClass {
    static {
        SomeEnumClass.values();
    }
}
```
is valid, which indicates that enum members are already initialized when doing class static initialization, so what we're doing here should be fine

ZZZank:
> Is it your intention to see this through to the end?

My plan is to, at least, replace all method references in `BodyCodeGen`, so that even if I dont have much time for it, other volunteers can know how this works and continue working on it. I will certainly see this through to the end, if time permitted

ZZZank:
It took me a bit too long to pick this PR up again, but in a nutshell:
- all `ScriptRuntime` method references are replaced with `ScriptRuntimeMethodSig`
- I also fixed a deprecated `delete()` usage: [link](https://github.com/mozilla/rhino/pull/1950/changes/86a1da554b48e1ff79c302454bbb3c6348988742)

`OptRuntime` methods and `INDY` methods are not replaced yet, to make the size of this PR reasonable

rhino/src/main/java/org/mozilla/javascript/optimizer/ScriptRuntimeMethodSig.java
aardvark179: There isn't an unbreakable cycle here, because annotation constants don't need to be resolved as part of class initialisation itself.

aardvark179:
So this PR does mean you can't get the method signature wrong, and the annotation does kind of indicate the use of the runtime method, but I had to read it a few times to understand that the enum name and the method name don't really need to match, and it leaves the types in one place so if they change in an incompatible way it will remain invisible until you hit that compilation path with real code. I think we could improve this in two ways:
1. Make the API use in `BodyCodeGen` explicitly say how many variables should be consumed and produced. So something like
```java
callMethodVoid(descriptor, 4, cfw);
callMethod(otherDescriptor, 2 , cfw);
```
2. Our method descriptors should make the types explicit, and we should test that they match methods in the runtime, or we could look them up when the class is initialised.

That would give us declarations that look something like

```java
enum RuntimeMethods {
    ElemIncrDecrString("elemIncrDecr", Object.class, Object.class, Object.class, Context.class, int.class):
    
    private final String methodName;
    private final String methodSignature;
    private final boolean voidReturn;
    private final boolean paramCount;
    
    RuntimeMethods(String name, Class<?> returnType, Class<?>... paramTypes) {
        // Do lookup / validation here and set fields
    }
}
```
This would both help make working on `BodyCodeGen` easier, _and_ catch errors very quickly.

ZZZank:
I have implemented the second improvement, and helper methods for implementing the first one. Though I'm not sure whether explicitly saying param count is actually required. When the actual param count of method changes, an error will be thrown at class initialization, long before `desc.addInvoke(cfw, expectedParamCount)` call or similar

aardvark179:
> I have implemented the second improvement, and helper methods for implementing the first one. Though I'm not sure whether explicitly saying param count is actually required. When the actual param count of method changes, an error will be thrown at class initialization, long before `desc.addInvoke(cfw, expectedParamCount)` call or similar

You'll definitely get an error when a method is used,. but I can't remember how eagerly the JVM will do that part. I think I now force the issue earlier than it had been because methods get linked earlier, but I don't know if that will always be true.

ZZZank:
Perhaps I didn't make it clear enough😜: by "class initialization" I mean the initialization of `ScriptRuntimeDesc`. When there's a mismatched param count (aka mismatched param types), `NoSuchMethodException` will be thrown at class initialization, long before `desc.addInvoke(cfw)` or actual method execution

Also there''s one thing I missed after removing the marker+scan approach: the ability of navigating between method declaration and usages by Ctrl+Click on enum values. Would be nice if there's some sort of replacements for this "feature"

PR #1954, opened 2025-06-29
Preparation for interpreter refactoring.

aardvark179:
# Interpreter refactoring for error reporting and performance.

This is a the first in a set of stacked PRs covering all the changes made as part of our work on #1933. The overall goal of this has been to refactor the interpreter to allow for better error reporting, without losing any interpreter performance in the process. These changes are not presented in the order in which they were made internally and have been reorganised extensively to make them easier
to review.

## Original motivation

Internally we use a modified JVM that has been tightly coupled to the Rhino interpreter in order to enable better error reporting in some circumstances. This is done by changing the stack frames in exception stack traces to reflect the scripts being executed - similar to decoration done within the interpreter for Rhino exceptions. Since the interpreter stack structure and the information about PC are held in mutable structures this information must be captured at the point where the exception is generated, and this is too expensive to be done for all exceptions and stack traces that might be generated in the system.

Many of the prototype versions of this work showed some regression in interpreter benchmarks, and an increase in memory footprint, so in combination with this work we also worked hard to mitigate these issues. Taking those changes to their logical conclusion showed a substantial performance improvement across the benchmarks, and so we include them with this work. They also have helped provide some good pointers for further optimisation work that we may attempt.

### Exception throwing and reporting

Generating exceptions can be a pretty common occurrence, but the reporting of exception stack traces is generally much rarer, so in general we should make the process of throwing an exception as light as possible, and shift heavier work to the process of reporting. To facilitate this we want to be able to represent the stack by a linked list effectively immutable objects which can be captured by a simple operation. By their nature call frames are mutable in many respects, but we can make enough aspects of them immutable to reuse them as this structure. Let's start by thinking about what we need to decorate a Java stack trace with additional information for interpreted scripts.

#### Basic stack structure

We already have fields for parent frame and a count of how deep this
stack is, we will make these final to avoid mutations breaking our
structure.

#### Matching call frames to interpreter calls

Next, we need to be able to know which call frames map to each interpreter invocation. We have previously used a deque on the context to hold this, but we'd like to make this an immutable structure instead. We add a field referencing the call frame from the previous interpreter invocation. We will keep this separate from the parent frame field to make navigation easier when processing the call frame structure without having to traverse the entire list.

So, for each call to the interpreter we can navigate the set of frames associated with it by iterating through parent frames, with `null` indicating we've reached the end of that set, and we can iterate through the interpreter frames using the `previousInterpreterFrame` field.

#### Getting line number information

[...]

Issue #1933, opened 2025-06-10
Interpreter stack traces


aardvark179:
We run a lot of code in production in the interpreter due to resource constraints and have used a number of techniques to try and give useful error reporting and stack traces in logs and other contexts. As part of our most recent attempt to improve this we have been refactoring the interpreter to use immutable data structures to represent the stack and interpreter calls so that they can be efficiently captured when exceptions are raised and only converted to human readable form if the stack is logged or printed later.

During this refactoring process we made some performance related changes to the interpreter code to maintain performance and we are likely to land some of these as separate PRs. The combination of these changes and the refactoring should result in better interpreter performance overall.

rbri:
@aardvark179 is this done?

PR #1955, opened 2025-06-29
Callframe refactor

aardvark179:
2nd PR for #1933,  see #1954 for overall details.

We want to be able to reliably model the stack of interpreter frames in an immutable way. To make this problem more tractable we make everything intended to be final actually final by moving calculation of true maximum stack size to the creation of the frame. We record the PC of the caller in callee frames so that the stack can be reliably reconstructed, and we introduce methods to copy frames in situations where we might have previously mutated them.

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/Interpreter.java
+++ b/rhino/src/main/java/org/mozilla/javascript/Interpreter.java
@@ -2092,7 +2092,7 @@ public final class Interpreter extends Icode implements Evaluator {
                             case Icode_SETCONSTVAR1:
                                 indexReg = iCode[frame.pc++];
                             // fallthrough
-                            case Token.SETCONSTVAR:
+                            case Icode_SETCONSTVAR:
                                 stackTop =
                                         doSetConstVar(
                                                 frame,
--- a/rhino/src/main/java/org/mozilla/javascript/Interpreter.java
+++ b/rhino/src/main/java/org/mozilla/javascript/Interpreter.java
@@ -44,9 +44,12 @@ public final class Interpreter extends Icode implements Evaluator {
         // fields marked "final" in a comment are effectively final except when they're modified
         // immediately after cloning.
 
-        /*final*/ CallFrame parentFrame;
+        final CallFrame parentFrame;
         // amount of stack frames before this one on the interpretation stack
-        /*final*/ int frameIndex;
+        final short frameIndex;
+        // The frame that the iterator was executing.
+        final CallFrame previousInterpreterFrame;
+        final int parentPC;
         // If true indicates read-only frame that is a part of continuation
         boolean frozen;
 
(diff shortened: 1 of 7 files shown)


rhino/src/main/java/org/mozilla/javascript/Interpreter.java
```diff
@@ -44,9 +44,12 @@ private static class CallFrame implements Cloneable, Serializable {
...
         // immediately after cloning.
 
-        /*final*/ CallFrame parentFrame;
+        final CallFrame parentFrame;
         // amount of stack frames before this one on the interpretation stack
-        /*final*/ int frameIndex;
+        final short frameIndex;
+        // The frame that the iterator was executing.
+        final CallFrame previousInterpreterFrame;
+        final int parentPC;
```
rPraml, line 52: can we remove `parentPc`?
It should be equal with `parentFrame.pcSourceLineStart` (but this is not final)

rhino/src/main/java/org/mozilla/javascript/Interpreter.java
```diff
@@ -199,9 +254,18 @@ void initializeArgs(
 
             // Fill the frame structure
 
-            System.arraycopy(args, argShift, stack, 0, definedArgs);
+            int blen = 0;
+            if (boundArgs != null) {
+                blen = boundArgs.length;
```
rPraml:
```suggestion
                blen = Math.min(boundArgs.length, definedArgs);
```
it has an intrinsicCandidate

rhino/src/main/java/org/mozilla/javascript/Interpreter.java
aardvark179: They are the same at the moment the call is made, but `pcSourceLineStart` may be updated later once the calls returns. One of our objectives with this set of changes is that it should be possible to generate as accurate stack as possible for an exception without needing to do that work eagerly.

aardvark179: I've added some explanation to #1954 on the fields added to call Frame and why.

gbrail:
I want to help you keep moving forward, but can we address this build warning:

C:\Users\greg\src\rhino\rhino\src\main\java\org\mozilla\javascript\Interpreter.java:296: warning: [UnusedMethod] Method 'cloneFrozen' is never used.
        CallFrame cloneFrozen(CallFrame newParent, CallFrame newPreviousInterpreeterFrame) {
                  ^
    (see https://errorprone.info/bugpattern/UnusedMethod)
  Did you mean to remove this line?
1 warning

If you want to fix it in a subsequent PR, just say so.

Otherwise, this looks good to me!

aardvark179:
Sorry about that, it was an artefact of how I separated out these changes and really belonged in the next PR. I've moved it there and rebased everything on top of that.

gbrail:
This is great -- thanks!

PR #1956, opened 2025-06-29
Refactor exception handling

aardvark179:
Part of the work on #1933, see #1954 for overall details.

We move from using a deque on Context for holding the stack to it being represented purely by the linked parent frames and previous interpreter frames. We change the exception construction code to simply capture the current interpreter frame, and change the decoration code to walk this tree to produce the required stacks.

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/Context.java
+++ b/rhino/src/main/java/org/mozilla/javascript/Context.java
@@ -19,7 +19,6 @@ import java.io.Writer;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.util.ArrayDeque;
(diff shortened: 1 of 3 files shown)


gbrail:
A suggestion for these big multi-commit PRs (which I appreciate you breaking up into digestible chunks!)

Can we mark the ones "Draft" if they assume that an older PR will be merged first? I think we usually go in numeric order but in case we don't...

aardvark179:
What is really like to do is create branches in this repo, so each PR could be stacked properly on the previous one. 😀 Or creates  PRs that reflect the diffs between two branches in my fork. This is an area in which GH is still lacking.

gbrail:
Looks good to me, thanks!

PR #1957, opened 2025-06-29
Interpreter boxing / unboxing reduction.

aardvark179:
Part of the work on #1933 , see #1954 for the overall plan

While doing this work several areas in the interpreter were spotted where we box doubles in order to perform simple operations. We have endeavoured to reduce this boxing and unboxing and will use this to help guide us in further optimisation work.


gbrail:
Looks good to me, similar to the optimization that the bytecode tries to do when it knows that everything is a number. I might write the commit message when I squash this to indicate that, since the description made me think that this had to do with auto-boxing in the code. Thanks!

PR #1958, opened 2025-06-29
Extract the inner loop of the interpreter to a separate method.

aardvark179:
Part of the work on #1933, see #1954 for the overall plan.

This PR extracts the dispatch loop into a separate method as part of the preparations to extract all the instructions to be individual classes.


aardvark179:
Converting back to draft for now because I think I found a mistake in exception handling from Java.

aardvark179:
False alarm. Our fork separates out the `FEATURE_ENHANCED_JAVA_ACCESS` to allow JS code to catch Java exceptions without enabling the other features of `FEATURE_ENHANCED_JAVA_ACCESS`, and I'd forgotten that when rebasing things.

PR #1959, opened 2025-06-29
Aardvark179 interpreter dispatch refactor

aardvark179:
Part of the work on #1933, see #1954 for the overall plan.

This PR consists of about 30 commits. We add some extra handling to the main dispatch switch to allow instructions to be extracted to individual classes, a series of commits that do that for relate groups of instructions, and finally the refactoring of the dispatch loop once all instructions have been extracted.

### DIFF
--- a/benchmarks/src/jmh/java/org/mozilla/javascript/benchmarks/GeneratorBenchmark.java
+++ b/benchmarks/src/jmh/java/org/mozilla/javascript/benchmarks/GeneratorBenchmark.java
@@ -1,0 +1,64 @@
+package org.mozilla.javascript.benchmarks;
+
+import java.io.FileReader;
+import java.io.IOException;
+import java.util.concurrent.TimeUnit;
+import org.mozilla.javascript.Context;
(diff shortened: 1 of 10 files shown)


aardvark179:
When we merge this can we be careful not to do a squash? I want to ensure that @andreabergia doesn't have a bad time pulling this onto our fork. Thanks!

rbri:
Same for me, i guess it will help me with merging into core-js also

gbrail:
I think this is #2 in the 'giant interpreter rewrite' PR queue, right? I would normally squash this as I think it's long-term easier for everyone, but in this case I will see if it's mergeable without that!

I do always do a rebase though -- we have kept the commit history of master linear for a long time now and I think we should keep it up.

rbri:
Hopefully everybody will be happy with this

rbri:
@aardvark179 @andreabergia 
looks like i managed the merge for core-js, not squezzing this was a real help.

And so far the result looks also good.

PR #1960, opened 2025-06-30
Reduce allocation in arg nullability provider

ZZZank:
The `MemberBox#argNullability` is currently using a `boolean[]` to store nullability information. This means that for Java class, each `MemberBox` initialization will allocate a `new boolean[method.getParameters().length]` that only contains `false`.

This PR replaced `boolean[]` with `NullabilityAccessor` so that for pure Java class, we can prevent allocation by reusing the same `NullabilityAccessor.FALSE` object:

```java
    interface NullabilityAccessor {
        NullabilityAccessor FALSE = (ignored) -> false;

        boolean isNullable(int index);
    }
```

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/MemberBox.java
+++ b/rhino/src/main/java/org/mozilla/javascript/MemberBox.java
@@ -29,9 +29,9 @@ final class MemberBox implements Serializable {
     private static final long serialVersionUID = 6358550398665688245L;
 
     private transient Member memberObject;
-    private transient volatile List<TypeInfo> argTypeInfos;
-    private transient volatile TypeInfo returnTypeInfo;
-    private transient volatile NullabilityDetector.NullabilityAccessor argNullability;
+    private transient List<TypeInfo> argTypeInfos;
+    private transient TypeInfo returnTypeInfo;
+    private transient NullabilityDetector.NullabilityAccessor argNullability;
     transient boolean vararg;
 
     transient Function asGetterFunction;
--- a/rhino/src/main/java/org/mozilla/javascript/MemberBox.java
+++ b/rhino/src/main/java/org/mozilla/javascript/MemberBox.java
@@ -31,7 +31,7 @@ final class MemberBox implements Serializable {
     private transient Member memberObject;
     private transient volatile List<TypeInfo> argTypeInfos;
     private transient volatile TypeInfo returnTypeInfo;
-    transient boolean[] argNullability;
+    transient NullabilityDetector.NullabilityAccessor argNullability;
     transient boolean vararg;
 
     transient Function asGetterFunction;
@@ -53,7 +53,7 @@ final class MemberBox implements Serializable {
         this.memberObject = method;
         this.argNullability =
                 nullDetector == null
-                        ? new boolean[method.getParameters().length]
+                        ? NullabilityDetector.NullabilityAccessor.FALSE
                         : nullDetector.getParameterNullability(method);
         this.vararg = method.isVarArgs();
         this.argTypeInfos = factory.createList(method.getGenericParameterTypes());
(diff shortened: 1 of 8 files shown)


aardvark179:
So, this definitely decreases the allocations when Kotlin isn't involved, but increases them when Kotlin classes are being used.

How does it compare to a lazily filled cache of boolean arrays filled with `false`?

rPraml:
Can we return an `int` or `long` for cases where we have only 32 or 64 parameters (probably 99.9% of the use cases)
In the other 0.1% the method might return the result by throwing an exception or we have a fast `long` returning mehtod and a slower `boolean[]`

ZZZank:
Using `int` to replace `boolean[]` is a great idea. We can use `int` as the backend of `NullabilityAccessor`, so that for most Kotlin methods, the object allocation should be similar to original:
- Original: `boolean[]` object + `boolean` * N
- PR: `NullabilityAccessor` object + `int` * 1

rPraml:
I thought to replace the boolean[] by int, so I use different boxes. I've made a POC here: #1963
The idea is to have 3 diffent types of memberboxes: 
- one that cannot handle nullability at all
- one, that handle up to 32 parameters
- and a third "full featured" one, that hold more than 32 params.

it might be even possible to refactor also the asSetter/GetterFunction out. OTOH I don't know how "hot" these code paths are.

/edit: I like the idea with the compress method. I will not continue work on my POC - because it will only squeeze out 4-8 bytes and is not well tested... but maybe it will give you some ideas on how we can improve this even further

rhino/src/main/java/org/mozilla/javascript/NullabilityDetector.java
```diff
@@ -4,7 +4,66 @@
...
 
-    boolean[] getParameterNullability(Constructor<?> constructor);
+    NullabilityAccessor getParameterNullability(Constructor<?> constructor);
+
+    interface NullabilityAccessor {
+        NullabilityAccessor TRUE = (ignored) -> true;
+        NullabilityAccessor FALSE = (ignored) -> false;
+        NullabilityAccessor INDEX_OUT_OF_BOUNDS =
+                (i) -> {
+                    throw new IndexOutOfBoundsException(
```
gbrail, line 16: According to the code coverage, we never actually get to this point -- can you see if you can contrive some sort of a test case for that?

rhino/src/main/java/org/mozilla/javascript/MemberBox.java
```diff
@@ -111,6 +109,23 @@ List<TypeInfo> getArgTypes() {
         return argTypeInfos;
     }
 
+    public NullabilityDetector.NullabilityAccessor getArgNullability() {
+        var got = this.argNullability;
+        if (got == null) {
+            synchronized (this) {
```
gbrail: What's the idea of making this synchronized? (using double-checked locking, which is OK since argNullability is volatile)? It looks to me like we need it in the event that multiple threads are accessing the same class, but I want to make sure -- there are other places in Rhino where we sprinkle synchronized blocks around and it's not always clear to me that they actually are necessary to enable thread-safety.

gbrail (commented):
Thanks! This is clever and I can see how it can help. I have two small nits (one of them you can address by just commenting why we need to synchronize in a specific spot).

I also see the POC of another approach and some other comments. Do others have opinions on what to do here? Otherwise I think that this is something that we could merge soon.

rhino/src/main/java/org/mozilla/javascript/MemberBox.java
ZZZank: This was a "just in case" synchronization because I was told that some applications will share same `scope` between multiple threads, and will cause concurrent access to `argNullability` becuase `scope -> ClassCache -> JavaMembers -> NativeJavaMethod -> MemberBox -> MemberBox#getArgNullability()`

But now that I think about it, the return value of `getParameterNullability(...)` will always behave the same because we're giving it the same method/constructor, so synchronization is actually optional here. I will remove this synchronization and add some comments in code

rhino/src/main/java/org/mozilla/javascript/NullabilityDetector.java
ZZZank: I added some new tests in https://github.com/mozilla/rhino/pull/1960/commits/26b99be8c0de7bf478e6526e07ac4e05d3e515b1, and it should be capable of covering this piece of code (at least I placed a breakpoint here and it was hit)

gbrail:
This makes sense to me and I think that other comments have been resolved so I'm going to merge it. Thanks!

PR #1961, opened 2025-06-30
fix TypeInfoFactory used in FunctionObject

ZZZank:
The `TypeInfoFactory` used in `FunctionObject` is currently obtained via:

```java
TypeInfoFactory.get(this)
```

where the `this` is FunctionObject itself, which doesn't make any sense because it's trying to obtained a TypeInfoFactory from a not-yet-initialized object. `scope` should be used instead.

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/FunctionObject.java
+++ b/rhino/src/main/java/org/mozilla/javascript/FunctionObject.java
@@ -80,7 +80,7 @@ public class FunctionObject extends BaseFunction {
      * @see org.mozilla.javascript.Scriptable
      */
     public FunctionObject(String name, Member methodOrConstructor, Scriptable scope) {
-        var typeInfoFactory = TypeInfoFactory.get(this);
+        var typeInfoFactory = TypeInfoFactory.get(scope);
 
         if (methodOrConstructor instanceof Constructor) {
             member = new MemberBox((Constructor<?>) methodOrConstructor, typeInfoFactory);
--- a/rhino/src/main/java/org/mozilla/javascript/lc/type/TypeInfoFactory.java
+++ b/rhino/src/main/java/org/mozilla/javascript/lc/type/TypeInfoFactory.java
@@ -260,8 +260,8 @@ public interface TypeInfoFactory extends Serializable {
      * Associate this TypeInfoFactory object with the given top-level scope.
      *
      * @param topScope scope to associate this TypeInfoFactory object with.
-     * @return {@code true} if no previous TypeInfoFactory object were associated with the scope and
-     *     this TypeInfoFactory were successfully associated, false otherwise.
+     * @return {@code true} if no previous TypeInfoFactory object was associated with the scope and
+     *     this TypeInfoFactory is successfully associated, {@code false} otherwise.
      * @throws IllegalArgumentException if provided scope is not top scope
      * @see #get(Scriptable scope)
      */
@@ -291,8 +291,6 @@ public interface TypeInfoFactory extends Serializable {
             // we expect this to not happen frequently, so computing top scope twice is acceptable
             var topScope = ScriptableObject.getTopLevelScope(scope);
             if (!(topScope instanceof ScriptableObject)) {
-                // Note: it's originally a RuntimeException, the super class of
-                // IllegalArgumentException, so this will not break error catching
                 throw new IllegalArgumentException(
                         "top scope have no associated TypeInfoFactory and cannot have TypeInfoFactory associated due to not being a ScriptableObject");
             }
--- a/tests/src/test/java/org/mozilla/javascript/tests/type_info/AlwaysFailFactory.java
+++ b/tests/src/test/java/org/mozilla/javascript/tests/type_info/AlwaysFailFactory.java
@@ -1,0 +1,53 @@
+package org.mozilla.javascript.tests.type_info;
+
+import java.lang.reflect.GenericArrayType;
+import java.lang.reflect.ParameterizedType;
+import java.lang.reflect.TypeVariable;
+import java.lang.reflect.WildcardType;
+import java.util.List;
+import org.mozilla.javascript.lc.type.TypeInfo;
+import org.mozilla.javascript.lc.type.TypeInfoFactory;
+
+/**
+ * @author ZZZank
+ */
+enum AlwaysFailFactory implements TypeInfoFactory {
+    INSTANCE;
+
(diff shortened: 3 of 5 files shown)


aardvark179:
While you're in there, I think you may have a race in `TypeInfoFactory`.
```JavaScript
    static TypeInfoFactory get(Scriptable scope) {
        TypeInfoFactory got =
                (TypeInfoFactory) ScriptableObject.getTopScopeValue(scope, "TypeInfoFactory");
        if (got == null) {
            // we expect this to not happen frequently, so computing top scope twice is acceptable
            var topScope = ScriptableObject.getTopLevelScope(scope);
            if (!(topScope instanceof ScriptableObject)) {
                // Note: it's originally a RuntimeException, the super class of
                // IllegalArgumentException, so this will not break error catching
                throw new IllegalArgumentException(
                        "top scope have no associated TypeInfoFactory and cannot have TypeInfoFactory associated due to not being a ScriptableObject");
            }
            got = new ConcurrentFactory();
            got.associate(((ScriptableObject) topScope));
        }
        return got;
    }
```

You aren't checking  the result of `got.associate`, and `got` may have had another thread race to do the association.

`associate` should probably return the factory that was associated with the scope, not a boolean.

rPraml:
@ZZZank I think, we should remove the lazy init and change the code
```java
        if (got == null) {
            // we expect this to not happen frequently, so computing top scope twice is acceptable
            var topScope = ScriptableObject.getTopLevelScope(scope);
            if (!(topScope instanceof ScriptableObject)) {
                // Note: it's originally a RuntimeException, the super class of
                // IllegalArgumentException, so this will not break error catching
                throw new IllegalArgumentException(
                        "top scope have no associated TypeInfoFactory and cannot have TypeInfoFactory associated due to not being a ScriptableObject");
            }
            got = new ConcurrentFactory();
            got.associate(((ScriptableObject) topScope));
        }
```
to 
```java
        if (got == null) { // or just return null?
                throw new IllegalArgumentException(
                        "top scope have no associated TypeInfoFactory and cannot have TypeInfoFactory associated due to not being a ScriptableObject");
        }
```

Unfortunately, I don't remember why we need this lazy-init: the (de)serialization roundtrip should be fixed and I think if you get no TypeInfoFactory, this means there is a programming error. Or am I missing something?

ZZZank:
Tests failed again due to the removal of lazy init. Will do some more research tomorrow

And a note for myself: global factory for scope independent action

gbrail:
Thanks - there's a lot in this TypeInfoFactory stuff, but I see what it's doing

I'm trying to figure out the impact of this -- since the "ConcurrentFactory" is global, and doesn't do any GC via weak references, I'm trying to ensure that an environment that might use a JVM to start and stop lots and lots of tests doesn't end up with a memory leak. I don't think so because it seems like the default behavior is that each context gets its own factory associated with the top level scope, right?

ZZZank:
The default behaviour now is:
- Each top level scope will get a new `ConcurrentFactory` associated at `cx.initStandardObject(scope)`, which can then be GCed after the scope itself is GCed
- `new FunctionObject(...)` and `ScriptableObject#defineProperty(...)` will first search for factory associated with provided scope, and fallback to the global factory (with weak reference mechanism) if not found.

So either the factory itself or the cached objects in factory can be garbage collected, so no memory leak I believe

gbrail:
That seems OK to me. Looks like a few people have had a chance to look at it, so I'll get to it soon. Thanks!

gbrail:
Oooh -- first time for this -- upon rebasing, we have a failure in the "decycle" test, which we added recently.

```
Execution failed for task ':rhino:decycleMain'.
   > Violation(slicing=Package, name=cycle, dependencies=[org.mozilla.javascript ΓåÆ org.mozilla.javascript.lc.type, org.mozilla.javascript ΓåÆ org.mozilla.javascript.lc.type.impl.factory, org.mozilla.javascript.lc.type ΓåÆ org.mozilla.javascript, org.mozilla.javascript.lc.type.impl ΓåÆ org.mozilla.javascript, org.mozilla.javascript.lc.type.impl ΓåÆ org.mozilla.javascript.lc.type, org.mozilla.javascript.lc.type.impl.factory ΓåÆ org.mozilla.javascript.lc.type, org.mozilla.javascript.lc.type.impl.factory ΓåÆ org.mozilla.javascript.lc.type.impl])
```

@rPraml is our expert in this, I believe, as this either suggests that the implementation should change or we should add some exceptions in the configuration of the cycle detector.

rhino/src/main/java/org/mozilla/javascript/ScriptRuntime.java
```diff
@@ -167,6 +168,7 @@ public static ScriptableObject initSafeStandardObjects(
 
         scope.associateValue(LIBRARY_SCOPE_KEY, scope);
         new ClassCache().associate(scope);
+        new ConcurrentFactory().associate(scope);
```
rPraml, line 171: I'm fine with that. We can move that later to `LcBridge` (or whatever concept we make)
https://github.com/mozilla/rhino/pull/1983/files#diff-24249e911005a5f57fee7790be93c8a477f7584f853110544bd704cc971277e9R169

rhino/build.gradle
```diff
@@ -89,6 +89,8 @@ decycle {
     ignoring from: "org.mozilla.javascript.ScriptableObject", to: "org.mozilla.javascript.lc.type.TypeInfoFactory"
     ignoring from: "org.mozilla.javascript.FunctionObject", to: "org.mozilla.javascript.lc.type.TypeInfo*"
     ignoring from: "org.mozilla.javascript.Context", to: "org.mozilla.javascript.lc.type.TypeInfo*"
+    ignoring from: "org.mozilla.javascript.ScriptRuntime", to: "org.mozilla.javascript.lc.type.TypeInfoFactory"
```
rPraml, line 92: Do we need both? (probably... `TypeInfoFactory implements TypeInfo`)

rPraml (approved):
> @rPraml is our expert in this, I believe, as this either suggests that the implementation should change or we should add some exceptions in the configuration of the cycle detector.

Yes, this was my intention. I'm fine with these changes, as they do not conflict with the overall plan in #1983 👍

rhino/build.gradle
ZZZank: Yes we need. My guess is that the `.associate(...)` method is from `TypeInfoFactory`, and Decycle consider it as a reference to `TypeInfoFactory`

gbrail:
OK -- this looks good now. Thanks!

PR #1962, opened 2025-07-01
Fix: seal `String.prototype` correctly

andreabergia:
When `Context::initStandardObject(sealed = true)` is called, `String` is sealed, but `String.prototype` is not after the changes done in https://github.com/mozilla/rhino/pull/1852.
This patch fixes it.

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/NativeString.java
+++ b/rhino/src/main/java/org/mozilla/javascript/NativeString.java
@@ -148,6 +148,7 @@ final class NativeString extends ScriptableObject {
 
         if (sealed) {
             c.sealObject();
+            ((NativeString) c.getPrototypeProperty()).sealObject();
         }
         ScriptableObject.defineProperty(scope, CLASS_NAME, c, DONTENUM);
     }
--- a/rhino/src/test/java/org/mozilla/javascript/tests/NativeStringTest.java
+++ b/rhino/src/test/java/org/mozilla/javascript/tests/NativeStringTest.java
@@ -6,11 +6,14 @@
 package org.mozilla.javascript.tests;
 
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertThrows;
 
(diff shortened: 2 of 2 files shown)


rPraml:
Note: There is also #1085 open where I noticed that there are at least two different seal methods in rhino (maybe relevant for this PR maybe not)

rhino/src/main/java/org/mozilla/javascript/NativeString.java
```diff
@@ -148,6 +148,7 @@ static void init(Scriptable scope, boolean sealed) {
 
         if (sealed) {
             c.sealObject();
+            ((NativeString) c.getPrototypeProperty()).sealObject();
```
rPraml, line 151: I made a quick search for this pattern and maybe this is wrong for other Native-objects, too.

- NativeArray ✅
- NativeBigInt ❌
- NativeBoolean ❌
- NativeRegexp ✅
- all Native TypedArrays ❌
- and many more...
- 
Can you check if we have a similar problem in these classes?
I don't know the exact use cases (see #1085) but maybe it would be better to introduce a method `sealObjectAndProto` and make sealObject deprecated?

andreabergia: Oh yeah, it's possible this is broken for other built-ins too.
I'll check tomorrow.

aardvark179: I think I banged my head against this a couple of times when I was converting native array to lambdas.

gbrail:
I'm happy that you fixed this for now and I'd also be happy to see a more systematic fix for other things!

rhino/src/main/java/org/mozilla/javascript/NativeString.java
andreabergia: https://github.com/mozilla/rhino/pull/1964

PR #1964, opened 2025-07-02
Fix: seal all builtins prototypes correctly

andreabergia:
Extended the test and fix of https://github.com/mozilla/rhino/pull/1962 to all builtins. Now we ensure that, whenever
`ScriptRuntime::initStandardObject(sealed = true)` is called, all objects have both themselves and their prototype property sealed.

I have not reworked the API in `ScriptRuntime` suggested by @rPraml in this PR, but if someone wants to do that, we now have a test to ensure there are no regressions. 😊

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/ScriptRuntime.java
+++ b/rhino/src/main/java/org/mozilla/javascript/ScriptRuntime.java
@@ -241,7 +241,6 @@ public class ScriptRuntime {
             new LazilyLoadedCtor(scope, "Uint32Array", sealed, true, NativeUint32Array::init);
             new LazilyLoadedCtor(scope, "BigInt64Array", sealed, true, NativeBigInt64Array::init);
             new LazilyLoadedCtor(scope, "BigUint64Array", sealed, true, NativeBigUint64Array::init);
-            new LazilyLoadedCtor(scope, "Uint32Array", sealed, true, NativeUint32Array::init);
             new LazilyLoadedCtor(scope, "Float32Array", sealed, true, NativeFloat32Array::init);
             new LazilyLoadedCtor(scope, "Float64Array", sealed, true, NativeFloat64Array::init);
             new LazilyLoadedCtor(scope, "DataView", sealed, true, NativeDataView::init);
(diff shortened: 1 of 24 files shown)


rhino/src/main/java/org/mozilla/javascript/ScriptRuntime.java
```diff
@@ -241,7 +241,6 @@ public static ScriptableObject initSafeStandardObjects(
             new LazilyLoadedCtor(scope, "Uint32Array", sealed, true, NativeUint32Array::init);
             new LazilyLoadedCtor(scope, "BigInt64Array", sealed, true, NativeBigInt64Array::init);
             new LazilyLoadedCtor(scope, "BigUint64Array", sealed, true, NativeBigUint64Array::init);
-            new LazilyLoadedCtor(scope, "Uint32Array", sealed, true, NativeUint32Array::init);
```
rbri, line 244: omg

rbri:
LGTM - thanks @andreabergia

rhino/src/main/java/org/mozilla/javascript/ScriptRuntime.java
andreabergia: Yeah it was repeated :)

rPraml:
This PR looks also good to me.

I would like to take this opportunity to point out #1085 again, as I think that the right audience is currently reading this, who has expertise in this area.

Long story short:  `String.__proto__ = {}` still allows the replacement of the prototype.

andreabergia:
> Long story short: `String.__proto__ = {}` still allows the replacement of the prototype.

Interesting! There are some problems with `__proto__` that I started looking into [a few months ago](https://github.com/andreabergia/rhino/commit/d738a17cda2aeca14a04d8e8c15429aa3bb0a79d), but I didn't have time to finish that yet. This should definitely be fixed, though (but it's another PR IMHO).

gbrail:
I am out on vacation for about a week and a half -- during that time @rbri is also capable of merging PRs and always uses good judgement so please feel free!

rbri:
Will try my best, currently attending a conference,  back on Saturday...

rbri:
Thanks @andreabergia

PR #1966, opened 2025-07-04
Implement TypedArray.from

balajirrao:
Almost all related test262 cases pass except for:

- **`built-ins/TypedArray/from/iterated-array-changed-by-tonumber.js`**: When the input to `from` is an NativeArray, we could consider it an iterable - which would make the test pass. However doing this comes at a cost of copying out the elements into a list and then doing the insertion into the typed array. Therefore this PR treats NativeArrays as "array-like" and doing so makes the test fail.
- **`built-ins/TypedArrayConstructors/from/BigInt/mapfn-this-without-thisarg-non-strict.js`**
- **`built-ins/TypedArrayConstructors/from/mapfn-this-without-thisarg-non-strict.js`**: These fail for `Array.from` as well. The second and third arguments to `from` are a `mapper` function and a `thisArg` to be used as the receiver when calling the mapper. The test expects the `this` in the mapper to be `global` when. `thisArg` is `undefined`. Not sure how can I make that happen.
- **`built-ins/TypedArrayConstructors/from/nan-conversion.js`**
- **`built-ins/TypedArrayConstructors/from/new-instance-from-sparse-array.js`**: These fail due to how `NativeArray` handles sparse (with NOT_FOUND) and `undefined` when `get` is called - both are returned as `null` which `toNumber` converts to 0 (should be NaN in the float array case). Will try to fix this in a separate PR.


rhino/src/main/java/org/mozilla/javascript/typedarrays/NativeTypedArrayView.java
```diff
@@ -1236,6 +1251,98 @@ private static Object js_with(Context cx, Scriptable scope, Scriptable thisObj,
...
+            }
+            mapFn = (Function) mapArg;
+            if (args.length >= 3) {
+                thisArg = ScriptableObject.ensureScriptable(args[2]);
+            }
+        }
+
+        Object iteratorProp = ScriptableObject.getProperty(items, SymbolKey.ITERATOR);
+        if (!(iteratorProp == Scriptable.NOT_FOUND)
+                // Optimization: Although NativeArrays have iterators, we don't want to
```
aardvark179: Okay, this feels like you've done a very mechanical translation of the spec. That's fine, but let's see if we can rewrite this to be a little easier to understand and optimise later. I think you've aiming for something like:
```java
var iterator = ScriptableObject.getProperty(items, SymbolKey.ITERATOR);
boolean useIterator = iteratorIsUndefinedOrStandard(items, iterator);
List<Object> iterValues = new ArrayList<>();
int size;
if (useIterator) {
    size = fillListFromIter(items, iterator, iterValues);
} else {
    size = (int) AbstractEcmaObjectOperations.lengthOfArrayLike(cx, items);
}
// Now make the object
var result = callConstructor(...);
if (useIterator) {
    fillResultFromList(result, iterValues, mapping);
} else {
    fillResultFromItems(result, items, mapper);
}
```
Then we can think about `iteratorIsUndefinedOrStandard` on its own and how you could optimise the common case where we're starting with an array like object with a standard iterator.

rhino/src/main/java/org/mozilla/javascript/typedarrays/NativeTypedArrayView.java
```diff
@@ -1236,6 +1251,120 @@ private static Object js_with(Context cx, Scriptable scope, Scriptable thisObj,
...
+                                ? ((NativeArray) items).get(k)
+                                : ScriptRuntime.getObjectIndex(items, k, cx, scope);
+                if (mapping) {
+                    temp = mapFn.call(cx, scope, thisArg, new Object[] {temp, k});
+                }
+                typedArray.setArrayElement(k, temp);
+            }
+
+            return result;
+        } else {
```
rbri: personally i think it is mor clear for the reader without the else - you can easier see that in all other case an exception is throw.

rhino/src/main/java/org/mozilla/javascript/typedarrays/NativeTypedArrayView.java
```diff
@@ -290,6 +294,16 @@ private static void defineMethod(
         typedArray.definePrototypeMethod(scope, key, length, target, DONTENUM, DONTENUM | READONLY);
     }
 
+    private static void defineConstructorMethod(
```
rbri: do we really need to extract this single method into another method that is called only once - to understand what is happening you have to look at two places

rbri: and it looks like the function declaration is longer than the body itself

rhino/src/main/java/org/mozilla/javascript/typedarrays/NativeTypedArrayView.java
```diff
@@ -1236,6 +1251,120 @@ private static Object js_with(Context cx, Scriptable scope, Scriptable thisObj,
...
+
+        if (result instanceof NativeTypedArrayView) {
+            NativeTypedArrayView<?> typedArray = (NativeTypedArrayView<?>) result;
+            for (int k = 0; k < values.size(); k++) {
+                Object temp = values.get(k);
+                if (mapping) {
+                    temp = mapFn.call(cx, scope, thisArg, new Object[] {temp, (long) k});
+                }
+                typedArray.setArrayElement(k, temp);
+            }
```
rbri, line 1325: personally i find this construct a bit confusing - maybe 

if {
   return
}

throw

is a bit more clear

rbri: and more in sync with the code below

rhino/src/main/java/org/mozilla/javascript/typedarrays/NativeTypedArrayView.java
```diff
@@ -1236,6 +1251,120 @@ private static Object js_with(Context cx, Scriptable scope, Scriptable thisObj,
...
+        if (args.length < 1) {
+            throw ScriptRuntime.typeErrorById("msg.missing.argument");
+        }
+
+        final Scriptable items = ScriptRuntime.toObject(scope, args[0]);
+
+        Object mapArg = (args.length >= 2) ? args[1] : Undefined.instance;
+        Scriptable thisArg = Undefined.SCRIPTABLE_UNDEFINED;
+        final boolean mapping = !Undefined.isUndefined(mapArg);
+        Function mapFn = null;
```
rbri: you can declare this two mapping vars later
this will also place the mapping stuff in on block (together)

rhino/src/main/java/org/mozilla/javascript/typedarrays/NativeTypedArrayView.java
balajirrao: Fair point. I'm going to use it for `TypedArray.of` too, so that's going to be one more use

balajirrao: But I ended up not having this method - you're right, it was pointless.

rhino/src/main/java/org/mozilla/javascript/typedarrays/NativeTypedArrayView.java
```diff
@@ -1236,6 +1244,139 @@ private static Object js_with(Context cx, Scriptable scope, Scriptable thisObj,
...
+
+    private static Object fromArrayLike(
+            Context cx,
+            Scriptable scope,
+            Constructable constructable,
+            Scriptable items,
+            Function mapFn,
+            Scriptable mapFnThisArg) {
+
+        final int length = (int) AbstractEcmaObjectOperations.lengthOfArrayLike(cx, items);
```
rbri: Maybe simply casting here is not correct. Maybe you can add a test case that deals with arrays of size > max_int and check if we can support that also.

rbri (changes requested):

rhino/src/main/java/org/mozilla/javascript/typedarrays/NativeTypedArrayView.java
balajirrao: In my mind, performance was the only reason to deviate from mechanically translating the spec. I agree that your comment expresses this intent better - will make the change and see how it looks.

rhino/src/main/java/org/mozilla/javascript/typedarrays/NativeTypedArrayView.java
balajirrao: Yeah, I'm not sure about this one. The `setArrayElement` in `NativeTypedArrayView` takes an integer index. I think we can at least fail cleanly in the "too long" cases. Let me see what I can do.

rhino/src/main/java/org/mozilla/javascript/typedarrays/NativeTypedArrayView.java
balajirrao: @aardvark179 I've [rewritten](https://github.com/mozilla/rhino/pull/1966/commits/8e13f8dac7099210fc53ecfc0e4dee4968f232cc) it and managed to make the diff significantly smaller. Could you take a look ?

Also take a look at what I've done for `NativeArray` and `NativeTypedArrayView` cases - I've used the `Scriptable::get` instead of a `java.util.Array.List::get` - I found that handles cases of detached buffers and corner-cases around conversion of `null`/`undefined`/`NOT_FOUND` to a number. Although it fails a previously-passing test (`from/BigInt/new-instance-from-sparse-array.js`) due to a different kind of error being thrown.

balajirrao: > Also take a look at what I've done for `NativeArray` and `NativeTypedArrayView` cases - I've used the `Scriptable::get` instead of a `java.util.Array.List::get` - I found that handles cases of detached buffers and corner-cases around conversion of `null`/`undefined`/`NOT_FOUND` to a number. Although it fails a previously-passing test (`from/BigInt/new-instance-from-sparse-array.js`) due to a different kind of error being thrown.

@aardvark179 I thought about this again and I'm convinced that calling `get(index, start)` on NativeArrays and TypedArrays is the way. I replaced it with a call to `List::get` wrapped in a try-catch for `IndexOutOfBoundsException`. I've pushed a [commit](https://github.com/mozilla/rhino/pull/1966/commits/7a255b459843892034e4b2c99c09b49b1e154f5f) on top of my previous changes. Please take a look, thanks!

aardvark179: I’ll take a look, but it might be a day or two before I get to this on the todo list.

rhino/src/main/java/org/mozilla/javascript/typedarrays/NativeTypedArrayView.java
```diff
@@ -1236,6 +1244,114 @@ private static Object js_with(Context cx, Scriptable scope, Scriptable thisObj,
...
+        // we don't use the iterator to avoid copying the contents to determine the length.
+        // However, with this the test262 test
+        // built-ins/TypedArray/from/iterated-array-changed-by-tonumber.js
+        // doesn't pass.
+        if (!(iteratorProp == Scriptable.NOT_FOUND)
+                && !(items instanceof List) // NativeArray and NativeTypedArrayView
+                && !Undefined.isUndefined(iteratorProp)) {
+            final Object iterator = ScriptRuntime.callIterator(items, cx, scope);
+            if (!Undefined.isUndefined(iterator)) {
+                IteratorLikeIterable it = new IteratorLikeIterable(cx, scope, iterator);
```
aardvark179: I noticed we aren't closing the iterator properly here. You should do something like:
```java
IteratorLikeIterable it = new IteratorLikeIterable(cx, scope, iterator);
try {
    listFromIterator = new ArrayList<>();
    for (Object temp : it) {
         listFromIterator.add(temp);
    }
} finally {
    it.close();
}
```

rhino/src/main/java/org/mozilla/javascript/typedarrays/NativeTypedArrayView.java
```diff
@@ -1236,6 +1244,114 @@ private static Object js_with(Context cx, Scriptable scope, Scriptable thisObj,
...
+            }
+            mapFn = (Function) mapArg;
+            if (args.length >= 3) {
+                mapFnThisArg = ScriptableObject.ensureScriptable(args[2]);
+            }
+        }
+
+        List<Object> listFromIterator = null;
+        Object iteratorProp = ScriptableObject.getProperty(items, SymbolKey.ITERATOR);
+        // Optimization: When items is an instance of java.util.List and also have an iterator,
```
aardvark179, line 1272: So I was slightly confused by this comment when trying to write the follow up PR I promised you, but writing that did help me understand this. There are a couple of stages here which affect the process of reading elements, and the side effects that might have. Let's try and divide this into two parts:

1. Is the iterator a standard one, or might it have side effects?
2. Are the values returned by the iterator ones which could cause side effects when converted to the typed array type.

The test you mention uses a standard iterator (so can be optimised nicely), but returns an object which has a side effect when converted to a number (its `valueOf` property). So you could still avoid the iterator as long as you detect that situation you could deoptimise to handle the remaining values by copying them before doing further to number conversions.

rbri:
@balajirrao @aardvark179 what is the status here?
I really hope for this because it is another step forward.
And leaving things that are almost finished lying around is usually not a good idea...

aardvark179:
@balajirrao is on vacation right now. I think I’d be fine with this going in as is and doing a follow up PR for the edge case.

rhino/src/main/java/org/mozilla/javascript/typedarrays/NativeTypedArrayView.java
balajirrao: Ah, thanks for catching this!

rhino/src/main/java/org/mozilla/javascript/typedarrays/NativeTypedArrayView.java
balajirrao: Thanks for the PR @aardvark179.

rhino/src/main/java/org/mozilla/javascript/typedarrays/NativeTypedArrayView.java
balajirrao: @rbri I've added a change that bails out if the source is longer than `Integer.MAX_VALUE`.

I've addressed all the other comments as well. The PR can be squash-merged.

rhino/src/main/java/org/mozilla/javascript/typedarrays/NativeTypedArrayView.java
balajirrao: Done.

rhino/src/main/java/org/mozilla/javascript/typedarrays/NativeTypedArrayView.java
```diff
@@ -1236,6 +1244,119 @@ private static Object js_with(Context cx, Scriptable scope, Scriptable thisObj,
...
+                }
+            }
+        }
+
+        int size;
+        if (listFromIterator != null) {
+            size = listFromIterator.size();
+        } else {
+            long sizeLong = AbstractEcmaObjectOperations.lengthOfArrayLike(cx, items);
+            if (sizeLong > Integer.MAX_VALUE)
```
rbri: {} missing

rhino/src/main/java/org/mozilla/javascript/typedarrays/NativeTypedArrayView.java
```diff
@@ -1236,6 +1244,120 @@ private static Object js_with(Context cx, Scriptable scope, Scriptable thisObj,
...
+        }
+
+        Scriptable result = constructable.construct(cx, scope, new Object[] {size});
+        if (!(result instanceof NativeTypedArrayView)) {
+            throw ScriptRuntime.typeErrorById("msg.typed.array.receiver.incompatible", "from");
+        }
+
+        NativeTypedArrayView<?> typedArray = (NativeTypedArrayView<?>) result;
+        if (typedArray.length < size) {
+            throw ScriptRuntime.typeErrorById("msg.typed.array.length.too.small");
```
rbri, line 1310: Maybe this should be a rage error also - can we compare this with real browsers and prove the type also by a test

rhino/src/main/java/org/mozilla/javascript/typedarrays/NativeTypedArrayView.java
```diff
@@ -1236,6 +1244,120 @@ private static Object js_with(Context cx, Scriptable scope, Scriptable thisObj,
...
+
+        Scriptable result = constructable.construct(cx, scope, new Object[] {args.length});
+
+        if (!(result instanceof NativeTypedArrayView)) {
+            throw ScriptRuntime.typeErrorById("msg.typed.array.receiver.incompatible", "of");
+        }
+
+        NativeTypedArrayView<?> typedArray = (NativeTypedArrayView<?>) result;
+        if (typedArray.length < args.length) {
+            throw ScriptRuntime.typeErrorById("msg.typed.array.length.too.small");
```
rbri, line 1351: see previous comment

rbri:
@balajirrao sorry for this step by step thing - just found another one. Maybe you can have a look...

rbri:
@balajirrao maybe i'm blind but the changes from your last force-push are not visible in the Files Changed tab - can you please have a look

balajirrao:
@rbri you're not blind - turns out throwing range errors for `msg.typed.array.length.too.small` leads to test262 failures for example `test262/test/built-ins/TypedArrayConstructors/from/BigInt/custom-ctor-returns-smaller-instance-throws.js`. Hence I undid that change!

rbri:
Thanks for all the work!

rbri:
and the patience...

PR #1973, opened 2025-07-15
make '__parent__' property no longer available in ES6

rbri:
'__parent__' was already removed in 2010 from the browsers (https://whereswalden.com/2010/05/07/spidermonkey-change-du-jour-the-special-__parent__-property-has-been-removed/)


rbri:
any opinion here?

Currently working on moving our '__ proto __' support forward. Have a first working version ready and i really like to see this merged (together with #1974) to start from that ground.

gbrail:
I know that in the past I have been pro-backward compatibility, but I do think that something like this language feature that was removed in 2010 could be removed from the product.

(I would also support a future version of rhino that gets rid of the "language version" entirely, but that's a longer discussion...)

rbri:
@gbrail you still suprise me from time to time :-)

Made a first shot on completely removing this stuff from the code base. But this is not that simple because we have some (regression) tests dealing with that. Will try to sort my brain regarding these tests and then will provide another pr to let us compare the two solutions. But first i like to finish the work on __proto__ (a real nightmare, spending already soo much time  on this - but i need it for the next HtmlUnit)

gbrail:
I mean, I am DEFINITELY supportive of disabling this feature in ES6 mode! And yes, I'm happy to consider other things -- I know how many people have old codebases that embed Rhino and we rely on the "language version" feature so that they can use their ancient JavaScript too, but I'm increasingly seeing that there's both momentum on this project and potential for the future and getting rid of some of the baggage might be really nice!

rbri:
@gbrail now i'm really puzzled - what should i do to get the thing merged? remove completely or hide it by a flag

BTW. this was only something found on the way of https://github.com/mozilla/rhino/pull/1976
I like to get [__proto__ merged](https://github.com/mozilla/rhino/pull/1976) first, then i will update this one

gbrail:
Sorry about the confusion -- if we make __parent__ protected by the language version, then that's fine, as really old code can still work. It looks like the effect of this change might be in that old PR, however.

rbri:
@gbrail hope you like it now ;-D

or the other way - i think i have a nice fix for the super.__proto__ stuff. But having this merged first the fix will be much simpler...

gbrail:
Yes, this is just fine. Thanks!

PR #1975, opened 2025-07-17
avoid some getProperty() calls

rbri:


aardvark179:
This may sound counter intuitive, but `isAccessorDescriptor` will check for the presence of both `get` and `set`, so in the common case of a simple data descriptor you will still end up doing the same number of property accesses (`has` is effectively an access which just checks the value isn't `NOT_FOUND`), and in the less common case of accessor descriptors you will come out even, or possibly one access worse off.

That isn't to say this area isn't worth optimising more, because it has some potentially difficult interactions with thread safe slot maps which I'm trying to nail down at the moment. What I'm trying is eagerly moving the descriptor into a specialised object like this:

```java
    public static final class DescriptorInfo {
        Object enumerable;
        Object writable;
        Object configurable;
        Object getter;
        Object setter;
        Object value;
        boolean accessorDescriptor;

        DescriptorInfo(ScriptableObject desc) {
            enumerable = getProperty(desc, "enumerable");
            writable = getProperty(desc, "writable");
            configurable = getProperty(desc, "configurable");
            getter = getProperty(desc, "get");
            setter = getProperty(desc, "set");
            value = getProperty(desc, "value");
            accessorDescriptor = isAccessorDescriptor(desc);
        }
    }
```

WIP is at https://github.com/aardvark179/rhino/tree/aardvark179-compute-modification-safety

But in the longer term I think we should look at using things like property descriptors and other objects we may generate internally as a test bed for property maps.

Edit: Now I've pasted that code I realise I can do better by doing
```
            accessorDescriptor = getter != NOT_FOUND || setter != NOT_FOUND;
```

aardvark179:
Actually, follow up question. Can your change handle something like
```javascript
Object.defineProperty(a, "bob", {writable: true, set: null})
```
which should raise a type error?

rbri:
@aardvark179
>  Can your change handle something

My change was only triggered by looking at the current code - the getter/setter/value vars are only needed depending on the accessorDescriptor flag. So more or less a mechanical refactoring to avoid the call if the var value is not used later on....

gbrail:
This looks good to me -- I imagine that there are lots of opportunities like this around the codebase!

This is not something that I want to spend time with on PRs yet, but we don't really have guidelines on how often to use "final". Some folks like to use it a lot for local variables like this, others don't. Personally, I like to use it for class members, but not for parameters, for example. It might be worth writing something up.

PR #1976, opened 2025-07-18
major __proto__ support update

rbri:
This brings this in line with the spec and real browsers. Still not 100% correct but really close.


rbri:
This is the first step, have to sort out some regressions...

rbri:
@gbrail, @tuchida, @aardvark179
i think i'm finished with this. Please have a look. Hopefully we can merge this soon, because i need it for the next HtmlUnit release.

rhino/src/main/java/org/mozilla/javascript/NativeFunction.java
```diff
@@ -18,14 +18,17 @@ public abstract class NativeFunction extends BaseFunction {
 
     private static final long serialVersionUID = 8713897114082216401L;
 
+    private boolean isShorthand;
```
andreabergia, line 21: Since we store it in the function, can we avoid adding the flag also to the `InterpretedFunction::idata`?

rbri: Will check that... the whole data transfer from the AST to the compiled/interpreted worlds took a lot of coffee, maybe you are right...

rbri:
> This is great!

thanks

> I had an half-baked attempt to fix that in https://github.com/andreabergia/rhino/commit/d738a17cda2aeca14a04d8e8c15429aa3bb0a79d, but your changes are much better and more complete!

Our approach is similar, that might mean the way is the right one...

rhino/src/main/java/org/mozilla/javascript/NativeFunction.java
rbri: >  can we avoid adding the flag also to the InterpretedFunction::idata?

I think not, as far as i can see this is the only way to transport the information to the initScriptFunction() function (similar to isES6Generator.

rhino-tools/src/main/java/org/mozilla/javascript/tools/debugger/Dim.java
```diff
@@ -542,7 +543,7 @@ private Object getObjectPropertyImpl(Context cx, Object object, Object id) {
             String name = (String) id;
             if (name.equals("this")) {
                 result = scriptable;
-            } else if (name.equals("__proto__")) {
+            } else if (name.equals(NativeObject.PROTO_PROPERTY)) {
                 result = scriptable.getPrototype();
             } else if (name.equals("__parent__")) {
```
gbrail, line 548: I'm all for using proper constants -- if we're going to do this for "__proto__", shouldn't we also do it for "__parent__"?

rhino/src/main/java/org/mozilla/javascript/ScriptRuntime.java
```diff
@@ -2143,8 +2143,19 @@ public static Object refDel(Ref ref, Context cx) {
         return wrapBoolean(ref.delete(cx));
     }
 
-    static boolean isSpecialProperty(String s) {
-        return s.equals("__proto__") || s.equals("__parent__");
+    static boolean isSpecialProperty(String s, int languageVersion) {
+        if (languageVersion >= Context.VERSION_ES6) {
+            return s.equals("__parent__");
```
gbrail, line 2148: Same question here about "__parent__" vs a constant

rhino/src/main/java/org/mozilla/javascript/NativeObject.java
```diff
@@ -45,6 +47,7 @@ public Scriptable construct(Context cx, Scriptable scope, Object[] args) {
                 };
 
         var proto = new NativeObject();
+        proto.setParentScope(s);
```
gbrail, line 50: Do you know why you had to do this? It's fine if it's necessary, it's just that I am eternally confused since we use ScriptableObject and the prototype hierarchy to represent objects, and the same class and the "parent" hierarchy to represent scopes in the runtime, and I keep wondering if we should make those separate things, even in implementation, so when I see a prototype with a parent I wonder why!

gbrail (commented):
This looks mostly good and I appreciate it -- please take a look at my minor comments!

rhino-tools/src/main/java/org/mozilla/javascript/tools/debugger/Dim.java
rbri: Yes we should, but i like to do this as part of the parent story https://github.com/mozilla/rhino/pull/1973

rhino/src/main/java/org/mozilla/javascript/ScriptRuntime.java
rbri: Same answer here ;-)

rhino/src/main/java/org/mozilla/javascript/NativeObject.java
rbri: LoL, you hit the nail

it is NECESSARY - i spend 2 long debugging sessions to figure this out (this was more or less the hardest part of this change). The problem is that we are in the early phase of the overall setup of the objects/prototoypes and all that stuff. Without that, the lambda functions will not have the correct prototype (Object instead of Function) because the lookup for the buildin prototypes (for functions) will fail without having the scope (and fall back to object).

I did some thinking about adding a comment here but finally decide to not do it because i think having the details will confuse the default reader of that stuff more than it helps. But i can add some sentences if you like....

gbrail:
Thanks -- this is a nice improvement!

PR #1981, opened 2025-07-21
as a large proportion of the annexB tests are already green, we should activate them

rbri:


gbrail:
Sure! Let's do this. Thanks!

PR #1982, opened 2025-07-21
Fixes edge case in `Array::from`

andreabergia:
This fixes how `this` is handled when calling `Array.from` in some edge cases, when `null` or `undefined` were passed as the `this` argument.


rbri (approved):
LGTM

gbrail:
Yep, this looks good -- thanks!

PR #1983, opened 2025-07-21
Preparation for liveconnect separation

rPraml:
Hello,

I've once again tried to identify the classes required for LiveConnect and move them into their own package.

This PR is intended to convey the rough plan. I assume it won't be merged in this form.

However, I would like to update it from time to time to document the current progress.

What's this actually about:

In my vision, Rhino should be divided into a core part (e.g., `rhino-core`) and a part that enables interactions with other Java classes (e.g., `rhino-liveconnect`).

`rhino-core` should be as free from reflection as possible (which is already largely possible thanks to the good work done recently in the Lambda area) - all these classes are moved to the `lc.java` package now (Maybe the name will change)

The advantages I expect from this:
- No use of reflection in `rhino-core`
- This will result in better compatibility, e.g., with grallVM native-image.
- (Presumably) better performance
- Security gain - Reflection may provide a potential attack surface for escaping a sandbox (although it's no guarantee that Rhino is secure)
- Splitting it into `rhino-core` and `rhino-liveconnect` allows for a leaner version

Disadvantage
- Moving certain classes will break one or more APIs

As described above, I will try to split the PR into several smaller PRs. e.g
- remove VMBridge (it dates back to a time because of Java 1.5 - we are 11 now and there - and we have multi release jars now, if we really need this)
- Use lambda initialization for all core - classes

BTW: This looks promising now. Test262 will run without LiveConnect classes


examples/src/main/java/Shell.java
```diff
@@ -53,9 +53,38 @@ public static void main(String args[]) {
...
+                    ScriptableObject.PERMANENT | ScriptableObject.DONTENUM,
+                    ScriptableObject.DONTENUM);
+
+            shell.defineProperty(
+                    shell,
+                    "help",
+                    1,
+                    Shell::help,
+                    ScriptableObject.PERMANENT | ScriptableObject.DONTENUM,
+                    ScriptableObject.DONTENUM);
```
ZZZank, line 87: missing `quit`?

examples/src/main/java/Shell.java
```diff
@@ -53,9 +53,38 @@ public static void main(String args[]) {
 
             // Define some global functions particular to the shell. Note
             // that these functions are not part of ECMA.
-            String[] names = {"print", "quit", "version", "load", "help"};
-            shell.defineFunctionProperties(names, Shell.class, ScriptableObject.DONTENUM);
 
+            shell.defineProperty(
```
ZZZank, line 57: Is it possible to make a bulk registeration? Like so:

```java
var properties = Map.<String, SomeFunction>of(
    "print", Shell::print,
    "quit", Shell::quit,
    "version", Shell::version,
    "load", Shell::load,
    "help", Shell::help
);
for (var entry : properties.entrySet()) {
    // define property, but how to determine the "length"?
}
```

It should be easier to read and maintain

rhino-tools/src/main/java/org/mozilla/javascript/tools/shell/Global.java
```diff
@@ -108,35 +108,144 @@ public void init(Context cx) {
...
-            "serialize",
-            "spawn",
-            "sync",
-            "toint32",
-            "version",
-            "write"
-        };
-        defineFunctionProperties(names, Global.class, ScriptableObject.DONTENUM);
+
+        defineProperty(
```
ZZZank, line 112: Same as `Shell`, bulk registration would be nice

rhino/src/main/java/org/mozilla/javascript/JavaEquals.java
```diff
@@ -0,0 +1,4 @@
+package org.mozilla.javascript;
+
+/** Classes, that implement this interface are treated as equal, when <code>o1.equals(o2)</code> */
+public interface JavaEquals {}
```
ZZZank, line 4: how about adding a common `SpecialEquality` interface for `JavaEquals` and `StatelessEquals`:

```java
interface SpecialEquality {
    boolean rhino$equals(Object o);

    interface JavaEquals extends SpecialEquality {
        default boolean rhino$equals(Object o) {
            return this.equals(o);
        }
    }
}
```

Then in the future, we can handle objects with new special equality without having to modify `EqualObjectGraphs` code (by adding interface implementation)

rhino/src/main/java/org/mozilla/javascript/ScriptRuntime.java
```diff
@@ -280,22 +281,27 @@ public static ScriptableObject initStandardObjects(
 
         // These depend on the legacy initialization behavior of the lazy loading mechanism
         new LazilyLoadedCtor(
-                s, "Packages", "org.mozilla.javascript.NativeJavaTopPackage", sealed, true);
+                s, "Packages", "org.mozilla.javascript.lc.java.NativeJavaTopPackage", sealed, true);
```
ZZZank, line 284: Can we move them to `LcBridge.instance.init(...)`

gbrail:
Thanks -- I'll look at the various bits as they get implemented.

I think this is an important evolution, as it makes the core of Rhino more compact and less vulnerable to exploits.

We spent some time naming modules when we created the module structure, and I think it's important that the main module continue to be named "rhino" and not "rhino-core". The new module could be named "rhino-liveconnect," or "rhino-reflection" or "rhino-java-integration",  I'm curious to hear other suggestions!

examples/src/main/java/Shell.java
rPraml: please see https://github.com/mozilla/rhino/pull/1993/files
If static import is used, it would match in one line.
but maybe it makes sense to replace
`defineProperty(this, "defineClass", 1, Global::defineClass, DONTENUM, DONTENUM | READONLY)`
by
`defineFunctionProperty(this, "defineClass", 1, Global::defineClass)`

rhino/src/main/java/org/mozilla/javascript/ScriptRuntime.java
rPraml: yes, I'll put that on my to-do list

davidpcaldwell:
> We spent some time naming modules when we created the module structure, and I think it's important that the main module continue to be named "rhino" and not "rhino-core". The new module could be named "rhino-liveconnect," or "rhino-reflection" or "rhino-java-integration", I'm curious to hear other suggestions!

I think "LiveConnect" is by far the best term for what we're talking about here. For the other two names above, I feel like they could mean a lot of things -- "reflection" could refer to `Reflect` from ECMAScript, "Java integration" could have to do with JavaScript interoperability with Java collections etc. ... but "LiveConnect" (although an older term) tells me exactly what we're talking about.

PR #1984, opened 2025-07-21
Remove VMBridge

rPraml:
This is the first PR for #1983

VMBridge is a relic that was previously used to ensure compatibility with older JVMs (1.5, 1.8, 11,...)

Since we now have version 11 as a minimum requirement and the older implementations of the VMBridge have already been deleted, this mechanism is no longer necessary.

If we need different implementations, we can do that with multi-release-jars (supported since java 9)

So I deleted the class and inserted the methods, where they are needed

**Possible breaking change**
The current implementation tried to instantiate a `org.mozilla.javascript.VMBridge_custom`.
This will no longer work (and did also not work in a moduel-classpath at this leads to a split-package situation)


rhino/src/main/java/org/mozilla/javascript/ContextHolder.java
```diff
@@ -0,0 +1,17 @@
...
+    static void setContext(Context context) {
+        contextLocal.set(context);
+    }
+
+    static Context getContext() {
+        return contextLocal.get();
+    }
+
+    public static void clearContext() {
+        contextLocal.remove();
```
gbrail: This bit used to have an optimization -- it put an array in the context, and just get and set the array element. That way, every time we cleared the context it was a hash table lookup inside the ThreadLocal instance, whereas now it's a hash table removal, which is slower. I don't think this would show up in our benchmarks, which just create one context and clear it at the end, but in other use cases, including our tests, this would indeed slow things down.

rhino/src/main/java/org/mozilla/javascript/InterfaceAdapter.java
```diff
@@ -72,8 +76,57 @@ static Object create(Context cx, Class<?> cl, ScriptableObject object) {
...
+                                return "Proxy[" + target.toString() + "]";
+                            }
+                        }
+                        return adapter.invoke(cf, target, topScope, proxy, method, args);
+                    }
+                };
+        Object proxy;
+        try {
+            proxy = c.newInstance(handler);
+        } catch (InvocationTargetException ex) {
```
gbrail, line 120: Since we're moving this we could use a try/multi-catch here.

rhino/src/main/java/org/mozilla/javascript/jdk18/VMBridge_jdk18.java
```diff
@@ -1,131 +0,0 @@
...
-import org.mozilla.javascript.ContextFactory;
-import org.mozilla.javascript.InterfaceAdapter;
-import org.mozilla.javascript.Scriptable;
-import org.mozilla.javascript.VMBridge;
-
-public class VMBridge_jdk18 extends VMBridge {
-    private static final ThreadLocal<Object[]> contextLocal = new ThreadLocal<>();
-
-    @Override
-    protected Object getThreadContextHelper() {
```
gbrail, line 25: Here's that optimization -- I think it's important that we keep it (although we could simplify it, hide the implementation in your ContextHolder class, and use Context[] instead of Object[]) so that applications that create and destroy lots of Context objects don't see additional overhead.

gbrail (changes requested):
Thanks! I think this is a great idea. However, the old thread slot implementation had an optimization that I think we should preserve -- please take a look below -- thanks!

rPraml:
Hello @gbrail thanks for pointing this out.

I've made a benchmark and the fastest method is, to get/set the ThreadLocal direct (OpenJDK 17)

I'm wondering if I should put the ThreadLocal into the ContextHolder class or put it directly into the Context class. OTOH the hotspot compiler would optimize/inline the methods anyway, that it is not measurable.

The idea I first had was that we could use a ScopedValue via a multi-version jar for Java 21. But ScopedValues work a bit differently, so this isn't easily possible (at least not yet).

ZZZank:
An idea on the class name: how about `ContextGlobal` instead of `ContextHolder`? The `Holder` sounds like an object holding another object, while it is actually a static utility class holding an object

gbrail:
I agree ContextHolder is a bit weird but I don't have strong feelings about it, so I'll leave it up to @rPraml

rPraml:
I've decided to delete the ContextHolder (which was effectively only a wrapper for the ThreadLocal) and inlined the threadLocal calls in the context class.

Since a ScopedValue, as originally intended, would not work, I see no advantage in outsourcing an extra class

rhino/src/main/java/org/mozilla/javascript/Context.java
```diff
@@ -528,28 +535,32 @@ static final Context enter(Context cx, ContextFactory factory) {
...
     }
 
     @Override
     public void close() {
         if (enterCount < 1) Kit.codeBug();
         if (--enterCount == 0) {
-            Object helper = VMBridge.instance.getThreadContextHelper();
-            VMBridge.instance.setContext(helper, null);
-            factory.onContextReleased(this);
+            if (currentContext.get() != this) Kit.codeBug();
```
rPraml: please check, if this would make sense

gbrail: I don't think it's bad to check but it has the potential to break existing code. How about instead:

`assert currentContext.get() == this`

rPraml:
This should be fine now. Should I rebase and squash some commits?

gbrail:
As long as it stays un-conflicted with other changes (which I think it is so far) then I will squash and rebase when I merge it. Thanks!

gbrail:
LGTM and thanks!

PR #1985, opened 2025-07-21
Regexp: Use lambda instead of reflection for init

rPraml:
2nd PR for #1983 

For regexp, we use now a lambda expression for init. This allows us an easier separation of liveconnect classes later


gbrail:
Yes, makes sense to me, thanks!

PR #1987, opened 2025-07-22
runWithAllModes() should execute the test in interpreted mode first

rbri:
runWithAllModes() should execute the test in interpreted mode first because debugging the interpreter case is easier


gbrail:
Easier for you maybe ;-) 

But seriously I'm OK with this unless someone else has a serious objection.

PR #1988, opened 2025-07-22
Builtins: Use lambda instead of reflection for init 

rPraml:
3rd PR for https://github.com/mozilla/rhino/pull/1983

For Builtins, we use now a lambda expression for init. This allows us an easier separation of liveconnect classes later


gbrail:
Thanks, looks good!

PR #1990, opened 2025-07-24
improve super put processing

rbri:
Found somewhere the hint that super is more or less about starting the property search at super but calling the final setter on this. Therefore this changes the put implementation like that and 'magically' the super.__proto__ handling works now and all the other tests are still green.

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/IRFactory.java
+++ b/rhino/src/main/java/org/mozilla/javascript/IRFactory.java
@@ -2025,31 +2025,6 @@ public final class IRFactory {
             }
             parser.checkActivationName(name, Token.GETPROP);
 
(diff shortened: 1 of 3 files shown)


rbri:
@andreabergia, @0xe
you did all the work on 'super' - please have a look at this. At least all the tests are passing but maybe you know more...

andreabergia (approved):
This looks correct and much simpler than our approach! I would say you also should do the same changes to the variant of `ScriptableObject::putSuperProperty` that takes `Symbol`, for coherence, but other than that I'm happy!

rbri:
> I would say you also should do the same changes to the variant of ScriptableObject::putSuperProperty that takes Symbol

@andreabergia goot point - will do that. Thanks for noting that

rbri:
@gbrail looks like @andreabergia if fine with this - hopefully you are also

gbrail:
Thank you -- looks good!

PR #1992, opened 2025-07-24
Update test262.properties file

andreabergia:
It's slightly outdated; this PR simply updates it.


rbri:
@andreabergia but the build is green for the current one????

andreabergia:
> @andreabergia but the build is green for the current one????

One test is marked as failing, but it's actually passing. IIRC this doesn't break the build.
The other tests are marked as failing in compiled-mode only, but they actually fail in all modes. I guess that is a bug in the runner.

rbri:
I think for the moment it is fine to update this, maybe someone has time to improve the the test runner later

PR #1993, opened 2025-07-24
Global: Do not use reflection for init

rPraml:
See #1983 
use lambda expressions for init

### DIFF
--- a/rhino-tools/src/main/java/org/mozilla/javascript/tools/shell/Global.java
+++ b/rhino-tools/src/main/java/org/mozilla/javascript/tools/shell/Global.java
@@ -108,38 +108,36 @@ public class Global extends ImporterTopLevel {
         // that these functions are not part of ECMA.
         initStandardObjects(cx, sealedStdLib);
         NativeConsole.init(this, sealedStdLib, new ShellConsolePrinter());
-        String[] names = {
-            "defineClass",
-            "deserialize",
-            "doctest",
-            "gc",
-            "help",
-            "load",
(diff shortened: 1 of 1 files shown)


rPraml:
I'm still struggling with a few CodeQL errors... Does anyone have a tip?

/edit: It happens, because we do not invoke the methods with reflection - and CodeQL "reaches" the methods now. There are some possible dangerous methods like `desreialize(obj,'/etc/passwd')` (as same as dangerous as `new Packages.java.io.File('/etc/passwd').unlink()` in liveConnect)

gbrail:
Huh -- this is the first time CodeQL has "gone off" and it noticed some 20-year-old code ;-)

Seriously, I do think that we should add some path-sanitization logic in there, but in the meantime I'll see if I can figure out how to move on with this change...

rPraml:
I'm totally puzzled now
- This PR is green now. @gbrail did you retrigger or reconfigure some bulds?
- CodeQl goes completely crazy. In my test branch #1994 I tried to strip down `Global.java`. Now I have 47 violations scattered across the entire code base: https://github.com/mozilla/rhino/pull/1994/checks?check_run_id=46706261459

>  this is the first time CodeQL has "gone off"

I think I've discovered a bug in CodeQL. I assume. that some weird code in `Global.java` will block CodeQL from properly working.

rPraml:
@gbrail I've converted this to a draft. I would like to tidy up/refactor the Global class first and replace the 20 year old code by some modern versions

PR #1996, opened 2025-07-25
Refactor Global.runCommand - add timeout and use java 11 methods

rPraml:
This PR refactors the `Global.runCommand` method, as this code looks a bit ancient (<= java 1.5)

- extract the related code in `ExecUtil` 
- use try-with-resources (java 1.7 feature) to get sophisticated exception handling (when an exception in a pipe thread occurs, it may be added as supressed-exception) - **this is a behaviour change:** Exceptions in the pipe threads were formerly printed only on console, now they may be thrown to the invoker
- **new Feature:** a `timeout` parameter allows to kill the thread after the specified timeout
- Provided lots of unit tests (for linux & windows)

I would like to tidy up the Global a bit, before continuing #1983 and maybe #1998

### DIFF
--- a/rhino-tools/src/test/java/org/mozilla/javascript/tools/tests/GlobalRunCommandTest.java
+++ b/rhino-tools/src/test/java/org/mozilla/javascript/tools/tests/GlobalRunCommandTest.java
@@ -1,10 +1,9 @@
 package org.mozilla.javascript.tools.tests;
 
+import static org.junit.jupiter.api.Assertions.*;
 import static org.junit.jupiter.api.Assertions.assertArrayEquals;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertInstanceOf;
-import static org.junit.jupiter.api.Assertions.assertThrows;
-import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
@@ -43,13 +42,14 @@ public class GlobalRunCommandTest {
     private static final boolean isWindows =
             System.getProperty("os.name").toLowerCase().contains("win");
 
+    private static final int EXIT_CODE_WHEN_KILLED = isWindows ? 1 : 143;
+
     private static final String STDIN_TO_STDOUT =
             "runCommand('stdinToStdout', { input: stdIn, output: stdOut, err: stdErr, commandExecutor: commandExecutor })";
     private static final String STDIN_TO_STDERR =
             "runCommand('stdinToStderr', { input: stdIn, output: stdOut, err: stdErr, commandExecutor: commandExecutor })";
 
-    private static final boolean mockOnly =
-            RhinoConfig.get("rhino.test.runCommand.mockOnly", false);
+    private static final boolean mockOnly = RhinoConfig.get("rhino.test.runCommand.mockOnly", true);
 
     public static Stream<Arguments> executors() {
         if (mockOnly) {
@@ -320,7 +320,7 @@ public class GlobalRunCommandTest {
     @ParameterizedTest
     @MethodSource("executors")
     public void testTimeout(Global.CommandExecutor exec) {
-        String cmd = "runCommand('sleep5',{ timeout: 500, commandExecutor: commandExecutor })";
+        String cmd = "runCommand('sleep',{ timeout: 500, commandExecutor: commandExecutor })";
         Utils.runWithAllModes(
                 cx -> {
                     cx.setLanguageVersion(Context.VERSION_ES6);
@@ -328,10 +328,16 @@ public class GlobalRunCommandTest {
                     long start = System.currentTimeMillis();
                     var result = cx.evaluateString(g, cmd, "test.js", 1, null);
                     long duration = System.currentTimeMillis() - start;
+                    assertInstanceOf(Number.class, result);
+                    assertEquals(EXIT_CODE_WHEN_KILLED, ((Number) result).intValue()); // Sigterm
+                    if (exec instanceof ExecutorMock) {
+                        // NOTE: We check only the return value here, which iś 1 (windows) or 143
+                        // (linux) and do no time measurements.
+                        // (the mock expects to be killed, otherwise, it waits forever)
+                        return null;
+                    }
                     assertTrue(duration >= 500);
                     assertTrue(duration <= 1500);
-                    assertInstanceOf(Number.class, result);
-                    assertEquals(isWindows ? 1 : 143, ((Number) result).intValue()); // Sigterm
                     return null;
                 });
     }
@@ -342,7 +348,7 @@ public class GlobalRunCommandTest {
      */
     @ParameterizedTest
     @MethodSource("executors")
-    public void testThrowOnInpupt(Global.CommandExecutor exec) {
+    public void testThrowOnInput(Global.CommandExecutor exec) {
         Utils.runWithAllModes(
                 cx -> {
                     cx.setLanguageVersion(Context.VERSION_ES6);
@@ -542,7 +548,7 @@ public class GlobalRunCommandTest {
                 case "stdinToStderr":
                     cmdarray = new String[] {"/usr/bin/env", "bash", "-c", "cat >&2"};
                     break;
-                case "sleep5":
+                case "sleep":
                     cmdarray = new String[] {"/usr/bin/sleep", "5"};
                     break;
                 default:
@@ -591,7 +597,7 @@ public class GlobalRunCommandTest {
                                 "[Console]::OpenStandardInput().CopyTo([Console]::OpenStandardError())"
                             };
                     break;
-                case "sleep5":
+                case "sleep":
                     cmdarray = new String[] {"ping", "127.0.0.1", "-n", "5"};
                     break;
                 default:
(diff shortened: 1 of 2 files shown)


rPraml:
I will provide some tests next week, as I've found some bugs in the runCommand method

rhino/src/test/java/org/mozilla/javascript/tests/RhinoPropertiesTest.java
```diff
@@ -153,6 +152,7 @@ void testEnv() {
         RhinoProperties properties = new RhinoProperties();
         properties.loadDefaults();
         // TODO: can/shoud we set an environment value. so check, if we can read PATH
-        assertNotNull(properties.get("path"));
+        // for now, we check for Path (and PATH)
+        assertNotNull(properties.get("Path"));
```
rPraml, line 156: Required to pass on (my) windows. Would be great, if someone can verify this.

gbrail:
I think that these changes make sense and I am happy that you are working on it, and I'm also concerned about the testing strategy. Writing repeatable, non-flaky tests for command-line invocations, and testing timeouts (at least in Java) is going to be hard. Already, the "GlobalRunCommandTest" fails inconsistently on my machine, and I see that it depends on things like "powershell" being available, command paths being the same across Unix-y platforms, etc. I'd like to see this test avoid becoming a maintenance headache for the rest of the life of Rhino.

I can think of two things:

1) make the command tests optional, and something that we run only as a sanity check from time to time
2) Put in more mocking / faking / stubbing whatever you want to call it so that we can test the core logic around handling I/O and timeouts without relying on actually invoking external commands AND without relying on the actual clock. (Outside Java we have better support for this, not sure what the Java equivalent is.)

Writing automated tests for features like this is really hard especially with our small "budget"!

I appreciate the effort in writing tests for this and I'm curious if others have ideas about how to make them repeatable.

Thanks!

rPraml:
Using a mock is a great idea. I've added a small bit to the ExecUtil
https://github.com/mozilla/rhino/pull/1996/commits/d4b5f2ecc849c12915ab6f2d54a9656141ba9604
so that we can control process generation. 

The idea is, that I use a sort of aliases, e.g. `stdinToStdout` - the linux executor invokes `cat`, windows `powershell` and the mock returns a MockProcess, where stdIn is connected to stdOut.

> Already, the "GlobalRunCommandTest" fails inconsistently on my machine, and I see that it depends on things like "powershell" being available, command paths being the same across Unix-y platforms, etc. I'd like to see this test avoid becoming a maintenance headache for the rest of the life of Rhino.

Yes the dependency on powershell is a bit weird, but it was the easiest way to copy stdin to stdout.

Can you tell me, what was the problem? Were the tests flaky or did they miss some binaries? I've run the tests on Windows 11, Ubuntu 24.04.2 LTS (in Windows WSL) and in the github actions. I didn't run the tests on macOs.

I've added a config option `rhino.test.runCommand.mockOnly`, that can we set to true as default

rhino-tools/src/test/java/org/mozilla/javascript/tools/tests/GlobalRunCommandTest.java
```diff
@@ -0,0 +1,729 @@
...
+                    break;
+                case "dd1M":
+                    cmdarray =
+                            new String[] {
+                                "/usr/bin/dd",
+                                "bs=1024",
+                                "count=1024",
+                                "if=/dev/stdin",
+                                "of=/dev/stdout"
+                            };
```
rPraml, line 546: I need feedback to this situation:
dd transfers 1MB and terminates. So it closes the inputstream. What would we expect when we have more than 1M?
The old implementation just drops the additional data, the new implementation will fail with a "broken pipe" exception.
I'm unsure, what is the best to handle the situation.
Should we just throw unconsumed data, or notify the user, that the process has terminated before all data was read? 
(test is currently disabled)

rPraml: should be fixed now. Exceptions of the process streams are ignored (as in old impl) 
They must be handled by checking the return value

rPraml:
this is ready for a second review now. 

Currently, it will pass on windows and linux (not tested on mac)

There is ` RhinoConfig.get("rhino.test.runCommand.mockOnly", false)` 
I can change the value to `true` - then only the mock is used for testing.

Code coverage (with mock):
Code coverage

rhino-tools/src/test/java/org/mozilla/javascript/tools/tests/GlobalRunCommandTest.java
gbrail: OK -- it's six of one a half dozen of the other! But I think that being consistent probably wins out.

rhino-tools/src/test/java/org/mozilla/javascript/tools/tests/GlobalRunCommandTest.java
```diff
@@ -0,0 +1,857 @@
...
+    public void testTimeout(Global.CommandExecutor exec) {
+        String cmd = "runCommand('sleep5',{ timeout: 500, commandExecutor: commandExecutor })";
+        Utils.runWithAllModes(
+                cx -> {
+                    cx.setLanguageVersion(Context.VERSION_ES6);
+                    var g = getGlobal(cx, exec);
+                    long start = System.currentTimeMillis();
+                    var result = cx.evaluateString(g, cmd, "test.js", 1, null);
+                    long duration = System.currentTimeMillis() - start;
+                    assertTrue(duration >= 500);
```
gbrail, line 339: This is, unfortunately, the kind of test that may become "flaky" over time. Even with the new "mocks," this timeout-related logic caused a few spurious test failures on my machine (when lots of tests were running in parallel). The main goal here is to check that the timeout works AT ALL and not try to check whether it works in a specific time. Can we do that somehow?

For example, you have your timed-out class that sleeps for five seconds -- what if updates a volatile boolean after 5 seconds and we assert for that? That way we're just ensuring that the timeout fires within 5 seconds, which is quite likely, instead of checking whether it fires after 1500ms, which is a lot more likely to get us into trouble.

rhino-tools/src/test/java/org/mozilla/javascript/tools/tests/GlobalRunCommandTest.java
```diff
@@ -0,0 +1,857 @@
...
+                });
+    }
+
+    /**
+     * Test, if we get an exception, when the inputStream throws an error. It is expected, that the
+     * exception is passed to the caller
+     */
+    @ParameterizedTest
+    @MethodSource("executors")
+    public void testThrowOnInpupt(Global.CommandExecutor exec) {
```
gbrail: If you're making other changes we might as well fix the spelling of "Input" ;-)

rhino-tools/src/test/java/org/mozilla/javascript/tools/tests/GlobalRunCommandTest.java
```diff
@@ -0,0 +1,857 @@
...
+    private static final boolean isWindows =
+            System.getProperty("os.name").toLowerCase().contains("win");
+
+    private static final String STDIN_TO_STDOUT =
+            "runCommand('stdinToStdout', { input: stdIn, output: stdOut, err: stdErr, commandExecutor: commandExecutor })";
+    private static final String STDIN_TO_STDERR =
+            "runCommand('stdinToStderr', { input: stdIn, output: stdOut, err: stdErr, commandExecutor: commandExecutor })";
+
+    private static final boolean mockOnly =
+            RhinoConfig.get("rhino.test.runCommand.mockOnly", false);
```
gbrail: I think that this should default to "true" so that the tests are as compatible and predictable as possible.

gbrail (commented):
Thanks! I think that the "mockOnly" mode should be the default so that Rhino tests run in the maximum number of environments. People looking to change or verify this could always use that.

rhino-tools/src/test/java/org/mozilla/javascript/tools/tests/GlobalRunCommandTest.java
rPraml: For the mock, I only check the exit code. The mock now waits for 5 minutes, so even under high load, the 500 ms kill should occur first. (I initially considered waiting indefinitely, but I wasn’t comfortable with that since CI builds could potentially hang.)

rhino-tools/src/test/java/org/mozilla/javascript/tools/tests/GlobalRunCommandTest.java
```diff
@@ -43,13 +42,14 @@ public class GlobalRunCommandTest {
...
+    private static final int EXIT_CODE_WHEN_KILLED = isWindows ? 1 : 143;
+
     private static final String STDIN_TO_STDOUT =
             "runCommand('stdinToStdout', { input: stdIn, output: stdOut, err: stdErr, commandExecutor: commandExecutor })";
     private static final String STDIN_TO_STDERR =
             "runCommand('stdinToStderr', { input: stdIn, output: stdOut, err: stdErr, commandExecutor: commandExecutor })";
 
-    private static final boolean mockOnly =
-            RhinoConfig.get("rhino.test.runCommand.mockOnly", false);
+    private static final boolean mockOnly = RhinoConfig.get("rhino.test.runCommand.mockOnly", true);
```
rPraml, line 52: changed to `true` and add some docs in rhino-test.config

rhino-tools/src/test/java/org/mozilla/javascript/tools/tests/GlobalRunCommandTest.java
```diff
@@ -342,7 +348,7 @@ public void testTimeout(Global.CommandExecutor exec) {
      */
     @ParameterizedTest
     @MethodSource("executors")
-    public void testThrowOnInpupt(Global.CommandExecutor exec) {
+    public void testThrowOnInput(Global.CommandExecutor exec) {
```
rPraml, line 351: fixed that typo

gbrail:
Thanks -- this looks good to me now and the tests work too. Thanks!

PR #2001, opened 2025-07-29
move NativeJavaObject.init(..) and NativeJavaMap.init() to initStandardObjects()

rbri:
NativeJavaObject.init(..) and NativeJavaMap.init() have to be part of initStandardObjects instead of initSafeStandardObjects()

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/ScriptRuntime.java
+++ b/rhino/src/main/java/org/mozilla/javascript/ScriptRuntime.java
@@ -214,9 +214,6 @@ public class ScriptRuntime {
         NativeStringIterator.init(scope, sealed);
         registerRegExp(cx, scope, sealed);
 
-        NativeJavaObject.init(scope, sealed);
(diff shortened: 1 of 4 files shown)


rbri:
@rPraml can you please have a look, i guess you are the one working in that field right now

rhino/src/main/java/org/mozilla/javascript/ScriptRuntime.java
```diff
@@ -280,6 +277,9 @@ public static ScriptableObject initStandardObjects(
             Context cx, ScriptableObject scope, boolean sealed) {
         ScriptableObject s = initSafeStandardObjects(cx, scope, sealed);
 
+        NativeJavaObject.init(s, sealed);
+        NativeJavaMap.init(s, sealed);
```
rPraml, line 281: Yes, they should be definitively considered as unsafe. (The initialization will be later moved to liveconnect)

rPraml (approved):

PR #2005, opened 2025-08-04
Introduce concept of compound operations to make certain slot map operations thread safe.

aardvark179:
When converting arrays and functions from `IdScriptableObject` to being lambda based we accidentally broke thread safety. This PR fixes those issues by in introducing compound operations that can wrap up the multiple modifications that may happen in a compound operation that can be passed around internally.

With this change all tests pass with thread safe objects enabled, and we can then do a PR to enable this in CI.

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/AbstractEcmaObjectOperations.java
+++ b/rhino/src/main/java/org/mozilla/javascript/AbstractEcmaObjectOperations.java
@@ -78,7 +78,11 @@ public class AbstractEcmaObjectOperations {
 
         if (obj.isExtensible()) return false;
 
-        for (Object name : obj.getIds(true, true)) {
+        Object[] ids;
+        try (var map = obj.startCompoundOp(false)) {
+            ids = obj.getIds(map, true, true);
+        }
+        for (Object name : ids) {
             ScriptableObject desc = obj.getOwnPropertyDescriptor(cx, name);
             if (Boolean.TRUE.equals(desc.get("configurable"))) return false;
 
@@ -141,7 +145,11 @@ public class AbstractEcmaObjectOperations {
             return false;
         }
 
-        for (Object key : obj.getIds(true, true)) {
+        Object[] ids;
+        try (var map = obj.startCompoundOp(false)) {
+            ids = obj.getIds(map, true, true);
+        }
+        for (Object key : ids) {
             ScriptableObject desc = obj.getOwnPropertyDescriptor(cx, key);
 
             if (level == INTEGRITY_LEVEL.SEALED) {
--- a/rhino/src/main/java/org/mozilla/javascript/CompoundOperationMap.java
+++ b/rhino/src/main/java/org/mozilla/javascript/CompoundOperationMap.java
@@ -1,0 +1,103 @@
+package org.mozilla.javascript;
+
+import java.util.Iterator;
+
+/**
+ * This class represents a compound operation performed on a non-thread safe map. As such it does
+ * the minimum work possible and does not enforce locking, instead it simply records whether the map
+ * has been mutated as part of the operation to allow the underlying map to delegate operations as
+ * required.
+ */
+public class CompoundOperationMap implements SlotMap, AutoCloseable {
+    protected final SlotMapOwner owner;
+    protected SlotMap map;
+    boolean touched = false;
+
+    public CompoundOperationMap(SlotMapOwner owner) {
+        this.owner = owner;
+        this.map = owner.getMap();
+    }
+
+    protected void updateMap(boolean resetTouched) {
+        if (touched) {
+            map = owner.getMap();
+            touched = resetTouched ? false : touched;
+        }
+    }
+
+    public boolean isTouched() {
+        return touched;
+    }
+
+    @Override
+    public void add(SlotMapOwner owner, Slot newSlot) {
+        map.add(owner, newSlot);
+        touched = true;
+    }
+
+    @Override
+    public <S extends Slot> S compute(
+            SlotMapOwner owner, Object key, int index, SlotComputer<S> compute) {
+        updateMap(true);
+        var res = map.compute(owner, this, key, index, compute);
+        touched = true;
+        return res;
+    }
+
+    @Override
+    public <S extends Slot> S compute(
+            SlotMapOwner owner,
+            CompoundOperationMap compoundOp,
+            Object key,
+            int index,
+            SlotComputer<S> compute) {
+        assert (compoundOp == this);
+        updateMap(true);
+        var res = map.compute(owner, this, key, index, compute);
+        touched = true;
+        return res;
+    }
+
+    @Override
+    public int dirtySize() {
+        updateMap(false);
+        return map.dirtySize();
+    }
+
+    @Override
+    public boolean isEmpty() {
+        updateMap(false);
+        return map.isEmpty();
+    }
+
+    @Override
+    public Slot modify(SlotMapOwner owner, Object key, int index, int attributes) {
+        updateMap(true);
+        var res = map.modify(owner, key, index, attributes);
+        touched = true;
+        return res;
+    }
+
+    @Override
+    public Slot query(Object key, int index) {
+        updateMap(false);
+        return map.query(key, index);
+    }
+
+    @Override
+    public int size() {
+        updateMap(false);
+        return map.size();
+    }
+
+    @Override
+    public Iterator<Slot> iterator() {
+        updateMap(false);
+        return map.iterator();
+    }
+
+    @Override
+    public void close() {
+        // This version doesn't need to do anything on clean up.
+    }
+}
--- a/rhino/src/main/java/org/mozilla/javascript/NativeArray.java
+++ b/rhino/src/main/java/org/mozilla/javascript/NativeArray.java
@@ -573,28 +573,28 @@ public class NativeArray extends ScriptableObject implements List {
             NativeArray builtIn,
             BuiltInSlot<NativeArray> current,
             Object id,
-            ScriptableObject desc,
+            DescriptorInfo info,
             boolean checkValid,
             Object key,
             int index) {
         // 10.2.4.2 Step 1.
-        Object value = getProperty(desc, "value");
+        Object value = info.value;
 
         if (value == NOT_FOUND) {
             return ScriptableObject.defineOrdinaryProperty(
-                    builtIn, id, desc, checkValid, key, index);
+                    builtIn, id, info, checkValid, key, index);
         }
 
         // 10.2.4.2 Steps 2 - 6
         long newLength = checkLength(value);
 
-        Object writable = getProperty(desc, "writable");
+        Object writable = info.writable;
         // 10.2.4.2 9 is true by definition
 
         // 10.2.4.2 10-11
         if (newLength >= builtIn.length) {
             return ScriptableObject.defineOrdinaryProperty(
-                    builtIn, id, desc, checkValid, key, index);
+                    builtIn, id, info, checkValid, key, index);
         }
 
         boolean currentWritable = ((current.getAttributes() & READONLY) == 0);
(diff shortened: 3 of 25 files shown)


aardvark179:
BTW @gbrail I'm in on Pacific time for a couple of weeks, I'd love to find some time to talk with you about property map plans and things like that.

rhino/src/main/java/org/mozilla/javascript/CompoundOperationMap.java
```diff
@@ -0,0 +1,109 @@
+package org.mozilla.javascript;
+
+import java.util.Iterator;
+
+// Should be named something like compound operations map.
```
gbrail: This is a really important bit of new functionality. Can you please add actual Javadoc here so that people can understand now to apply it for future code?

rhino/src/main/java/org/mozilla/javascript/ThreadSafeCompoundOperationMap.java
```diff
@@ -0,0 +1,101 @@
+package org.mozilla.javascript;
+
+import java.util.Iterator;
+
+class ThreadSafeCompoundOperationMap extends CompoundOperationMap {
```
gbrail, line 11: A little bit of Javadoc to explain what this class is for and how it relates to other classes would be super helpful

rhino/src/main/java/org/mozilla/javascript/ThreadSafeCompoundOperationMap.java
```diff
@@ -0,0 +1,101 @@
...
+    private boolean closed = false;
```
gbrail, line 12: My first reaction is that relying on "closed" to prevent multiple closes might not do what we need unless "closed" is volatile. Have you done any additional analysis on that regarding the threading model. (And if no locking / no volatile is an option we should also document that in comments.)

gbrail (commented):
This is super cool and the design is quite clever. can you help by better documenting the new code and also the locking behavior? Thanks!

rPraml:
I have a few questions regarding multi-threading.

In our application, we use a sealed shared scope and access the shared objects read-only. I don’t think we should run into any issues with that. Do we? 

What are your use cases?

Do you also start a separate context for each thread, in which the objects are then manipulated, or do you even start multiple threads from JavaScript?

The reason I’m asking is because of the ThreadLocal that holds the current context, and I’m curious how you solved that problem.

Thanks,
Roland

aardvark179:
We also use a separate context for every thread, but there are some exceptions for things that generally read only but are expensive to initialise.

So the problem we have is that a few objects get shared between threads, so everything has to be created as a thread safe object. This was fine until `NativeArray` and `BaseFunction` moved from being I’d scriptable object. When they moved to being lambda based then setting the length of the array or the prototype of a function via a property descriptor could suddenly result in an attempt to mutate the slot map *inside* a compute operation (deleting elements in the array case, or creating properties in the function case) but in multi threaded mode the compute operation had already claimed a write lock for the map, so the inner operation could t get a new write lock.

We missed this when converting those classes because upstream doesn’t run tests in multithreaded mode, yet, and only discovered the issue when merging down and testing in our application.

rhino/src/main/java/org/mozilla/javascript/ThreadSafeCompoundOperationMap.java
gbrail: I think this is a good PR and I'd like to move forward with it but I'd really like someone to see if they can verify whether "closed" needs to be volatile in this context before we move on. Thanks!

aardvark179: I added the following comment to that class
```
/**
 * This class represents a compound operation performed on a thread safe slot map. As each compound
 * operation creates a new instance the class itself does not need to consider access by multiple
 * threads. This means that the instance fields do not need to be volatile as we are only
 * considering access from this thread.
 */
 ```

File: rhino/src/main/java/org/mozilla/javascript/NativeObject.java [hunk 4]
```diff
@@ -577,7 +583,11 @@ private static Object js_getOwnPropDescs(
         ScriptableObject obj = ensureScriptableObject(s);
 
         ScriptableObject descs = (ScriptableObject) cx.newObject(scope);
-        for (Object key : obj.getIds(true, true)) {
+        Object[] ids;
```
andreabergia, line 586: This seems like a potential race condition - we have a compound operation only around `getIds`, and then we call `getOwnPropertyDescriptor` inside the for loop.

rhino/src/main/java/org/mozilla/javascript/CompoundOperationMap.java
```diff
@@ -0,0 +1,103 @@
...
+    @Override
+    public void add(SlotMapOwner owner, Slot newSlot) {
+        map.add(owner, newSlot);
+        touched = true;
+    }
+
+    @Override
+    public <S extends Slot> S compute(
+            SlotMapOwner owner, Object key, int index, SlotComputer<S> compute) {
+        updateMap(true);
```
andreabergia, line 41: The equivalent in the base class has

```java
        assert (compoundOp == this);
```

Does it make sense to add it here as well, or are there cases where the assertion would fail?

rhino/src/main/java/org/mozilla/javascript/ThreadSafeCompoundOperationMap.java
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    @Override
+    public Slot query(Object key, int index) {
+        updateMap(false);
+        return ((LockAwareSlotMap) map).queryWithLock(key, index);
+    }
+
+    @Override
+    public int size() {
```
andreabergia, line 72: I don't see an override for `dirtySize`. I think it's correct, as we have no `LockAwareSlotMap::dirtySizeWithLock` (by definition! it's dirty!), but I'd like confirmation.

rhino/src/main/java/org/mozilla/javascript/NativeArray.java
```diff
@@ -887,6 +930,19 @@ private static void deleteElem(Scriptable target, long index) {
         }
     }
 
+    private static void deleteElem(
+            CompoundOperationMap compoundOp, NativeArray target, long index) {
+        int i = (int) index;
+        if (i == index) {
+            checkNotSealed(target, null, i);
```
andreabergia, line 938: Checking sealed here (which we don't do in the version without the `compoundOp`) looks like a separate bug fix. In that case, I think we should add it to the other version just above too.

rhino/src/main/java/org/mozilla/javascript/AbstractEcmaObjectOperations.java
```diff
@@ -78,7 +78,11 @@ static boolean testIntegrityLevel(Context cx, Object o, INTEGRITY_LEVEL level) {
 
         if (obj.isExtensible()) return false;
 
-        for (Object name : obj.getIds(true, true)) {
+        Object[] ids;
```
andreabergia, line 81: Same potential race condition

rhino/src/main/java/org/mozilla/javascript/AbstractEcmaObjectOperations.java
```diff
@@ -141,7 +145,11 @@ static boolean setIntegrityLevel(Context cx, Object o, INTEGRITY_LEVEL level) {
             return false;
         }
 
-        for (Object key : obj.getIds(true, true)) {
+        Object[] ids;
```
andreabergia, line 148: Same potential race condition

rhino/src/main/java/org/mozilla/javascript/NativeObject.java
```diff
@@ -577,7 +583,11 @@ private static Object js_getOwnPropDescs(
...
+        try (var map = obj.startCompoundOp(false)) {
```
andreabergia, line 587: Same potential race condition

rhino/src/main/java/org/mozilla/javascript/ThreadSafeCompoundOperationMap.java
andreabergia: I agree with Duncan, these objects really aren't shared across threads. Sharing an object and starting two compound operations in two threads would create two instances, one per thread. So I think not having `volatile` is correct.

rhino/src/main/java/org/mozilla/javascript/NativeArray.java
aardvark179: The non-compound op version does do `checkNotSealed`, it's just that it's done internally by the `ScriptableObject#delete` methods. I'll add a comment though because that was a trick bug to track down when I was writing this.

rhino/src/main/java/org/mozilla/javascript/CompoundOperationMap.java
aardvark179: I think you put this comment against the wrong line, but you are correct, the assertion would make sense in

rhino/src/main/java/org/mozilla/javascript/ThreadSafeCompoundOperationMap.java
aardvark179: Correct, the super class just does the `updateMap` call and then does `dirtySize`, which is all that is needed.

aardvark179:
Regarding the potential race conditions in the use of `getIds` that @andreabergia found, I think the right way to address them would be to introduce some new APIs on `ScriptableObject` to get keys and values, and to get keys and descriptors. Trying to expand the existing compound ops to cover the use of those ids started push things in a direction that didn't feel right.

I'd rather do that in a follow up PR as it feels like that could might not be a small change either (there's a few overrides of `getIds` and a reasonable number of users) and I'd rather that chunk of stuff gets reviewed separately.

andreabergia:
> Regarding the potential race conditions in the use of `getIds` that @andreabergia found, I think the right way to address them would be to introduce some new APIs on `ScriptableObject` to get keys and values, and to get keys and descriptors. Trying to expand the existing compound ops to cover the use of those ids started push things in a direction that didn't feel right.
> 
> I'd rather do that in a follow up PR as it feels like that could might not be a small change either (there's a few overrides of `getIds` and a reasonable number of users) and I'd rather that chunk of stuff gets reviewed separately.

That's fine with me. Anyway, as far I can tell those problems were pre-existing, not something that was introduced in this PR.

gbrail:
This is looking good to me -- any other changes or reviewers? Otherwise I think we're ready to go.

As for getIds -- if we're going to change that interface (which we actually can't because it's probably overridden in a million places by users) -- if we're going to replace it with something else that has no race condition, is there a more modern implementation we could use that avoids creating that big temporary array, possibly something with an Iterator or that's Iterable?

aardvark179:
I think the question will be whether people have overridden `getIds` on subclasses of `ScriptableObject` or just on `Scriptable`. If it’s the latter then we could provide a default implementation of whatever these new APIs are based on `getIds`.

Making a replacement API more efficient will be tricky because almost every use requires indexed properties before string and symbol based ones, and that’s hard with our current slot maps, but maybe we can present the correct interface now and work on the efficiency as we move towards property maps.

While you are thinking about potential interface breakage at some point we will need to change `Callable` and `Constructsble`—the former needs to change because strict mode should not do toObject on `this` and the latter because we will need to pass `new.target`. I have a test branch that does this, and it’s a big internal change, but I’m not sure how much it would really affect users. I’ll make a draft PR so people can take a look.

gbrail:
Changing Callable and Constructable MIGHT not be awful, and if we make the migration path clear then people can upgrade

Going further if we're going to change interface breakage I would like to change the type of the "scope" from Scriptable to something that recognizes that scopes aren't objects -- this will simplify code and might make it easier for future optimizations. But since that changes the signature of literally every call in Rhino it feels pretty huge to me.

For example, can anyone explain the different ways that "parent" and "prototype" interact, and has anyone used both the "parent" based inheritance and the "prototype" based inheritance on the same object? It seems like in using the same object for scopes and objects we have made things unnecessarily complicated.

gbrail:
Anyway I'd like to move on from this PR so let me run one more test and merge it. Thanks!

Issue #2006, opened 2025-08-05
Regression: ArrayIndexOutOfBoundsException in Interpreter.getArgsArray()


rbri:
HtmlUnit uses the lates Rhino code & this was not detected in any of the HtmlUnit unit tests (and there are many ;-)).

* was able to track the root cause to commit e924885d - Refactor CallFrame to keep stack as immutable linked list.
* will work on this to extract an reproducible (sample but this will require some thime)

Have opened this issue to inform the team - maybe this is of interest for some other forks also.

see [Possible JavaScript Breakage Between 4.13.0 and 4.14.0 #1001](https://github.com/HtmlUnit/htmlunit/issues/1001#issuecomment-3153819425)

```
Caused by: java.lang.ArrayIndexOutOfBoundsException: Index 2 out of bounds for length 2
	at org.htmlunit.corejs.javascript.Interpreter.getArgsArray(Interpreter.java:4080)
	at org.htmlunit.corejs.javascript.Interpreter.doCallByteCode(Interpreter.java:3042)
	at org.htmlunit.corejs.javascript.Interpreter.interpretLoop(Interpreter.java:1995)
	at org.htmlunit.corejs.javascript.Interpreter.interpret(Interpreter.java:1192)
	at org.htmlunit.corejs.javascript.InterpretedFunction.call(InterpretedFunction.java:92)
	at org.htmlunit.corejs.javascript.ContextFactory.doTopCall(ContextFactory.java:356)
	at org.htmlunit.javascript.HtmlUnitContextFactory.doTopCall(HtmlUnitContextFactory.java:298)
	at org.htmlunit.corejs.javascript.ScriptRuntime.doTopCall(ScriptRuntime.java:4794)
	at org.htmlunit.corejs.javascript.InterpretedFunction.call(InterpretedFunction.java:90)
	at org.htmlunit.javascript.host.Window.triggerOnError(Window.java:1444)
	at org.htmlunit.javascript.JavaScriptEngine.handleJavaScriptException(JavaScriptEngine.java:1046)

```

rbri:
<img width="2828" height="265" alt="Image" src="https://github.com/user-attachments/assets/e57a1dc7-46ae-49b5-80fa-f615d9e8cf11" />

looks like indexReg is not increased by blen when reaching this point in this case

@aardvark179 any idea?

aardvark179:
[...]

PR #2008, opened 2025-08-06
Handle bound args correctly when bound function is `call`.

aardvark179:
Found this issue while investigating #2006.

### DIFF
--- a/tests/src/test/java/org/mozilla/javascript/tests/ApplyTest.java
+++ b/tests/src/test/java/org/mozilla/javascript/tests/ApplyTest.java
@@ -6,14 +6,34 @@ import org.mozilla.javascript.testutils.Utils;
 public class ApplyTest {
 
     @Test
-    public void invokeBoundCallManyArgs() {
-        /* This test is a little fiddly. The call to bind causes the
-        max stack size to be high enough that it could mask the
-        bug, so we have to make the call to the bound in function
-        in another function which has a smaller stsck. */
+    public void applyThisFromBoundArgs() {
+        // The `toString` call here converts the `NativeString` that
+        // is the bound `this` to a simple `String` that we can
+        // easily assert against.
+        String code =
+                "var f = function(x) { return this.toString(); };\n"
+                        + "var a = f.apply;\n"
+                        + "var b = a.bind(f, 'Hello!');\n"
+                        + "b([1,2]);\n";
+        Utils.assertWithAllModes_ES6("Hello!", code);
+    }
+
+    @Test
+    public void applyToApplyCallsCorrectFunction() {
         String code =
                 "Function.prototype.apply.apply(function(x) {return x;}, ['b', ['Hello!', 'Goodbye!']]);";
 
         Utils.assertWithAllModes_ES6("Hello!", code);
+    }
+
+    @Test
+    public void applyToApplySetsCorrectFunctionThis() {
+        // The `toString` call here converts the `NativeString` that
+        // is the bound `this` to a simple `String` that we can
+        // easily assert against.
+        String code =
+                "Function.prototype.apply.apply(function(x) {return this.toString();}, ['b', ['Hello!', 'Goodbye!']]);";
+
+        Utils.assertWithAllModes_ES6("b", code);
     }
 }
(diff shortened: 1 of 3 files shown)


rbri:
@aardvark179 have tested this, in my case your new else case is never touched. The problem is still there.

tests/src/test/java/org/mozilla/javascript/tests/es6/BoundFunctionTest.java
```diff
@@ -18,4 +18,18 @@ public void ctorCallableThis() {
 
         Utils.assertWithAllModes_ES6("bound foo", code);
     }
+
+    @Test
+    public void invokeBoundCallManyArgs() {
+        /* This test is a little fiddly. The call to bind causes the
+        max stack size to be high enough that it could mask the
+        bug, so we have to make the call to the bound in function
+        in another function which has a smaller stsck. */
```
rbri: typo

tests/src/test/java/org/mozilla/javascript/tests/es6/BoundFunctionTest.java
```diff
@@ -18,4 +18,18 @@ public void ctorCallableThis() {
...
+        String code =
+                "function f() {return 'Hello!';};\n"
+                        + "var b = f.call.bind(f, 1, 2,3);\n"
```
rbri: format

aardvark179:
@rbri I've refactored the apply logic and it is now doing the right thing for nested applies, at least in my test. Let me know how you get on.

rbri:
@aardvark179 thanks for the fix, at least we got some progress here...
* no longer an ArrayIndexOutOfBoundsException
* function setTimeout gets called

But the thisObj provided by the call is no longer the window - it is now a function object. Therefore the current implementation of the setTimeout() function in HtmlUnit fails.
image

Will try to figure out if we have the same in real browsers.

rbri:
@aardvark179 fyi: the rest of the HtmlUnit test suite passes with you changes

tests/src/test/java/org/mozilla/javascript/tests/ApplyTest.java
```diff
@@ -0,0 +1,19 @@
+package org.mozilla.javascript.tests.es6;
```
rbri, line 1: wrong pkg name regarding to the file location

tests/src/test/java/org/mozilla/javascript/tests/ApplyTest.java
```diff
@@ -0,0 +1,19 @@
...
+
+import org.junit.Test;
+import org.mozilla.javascript.testutils.Utils;
+
+public class ApplyTest {
+
+    @Test
+    public void invokeBoundCallManyArgs() {
+        /* This test is a little fiddly. The call to bind causes the
```
rbri: seems to be a copy - still talking about the bind function

rbri:
@aardvark179 found some time for more tests

    Function.prototype.apply.apply(function(x) {return this;}, ['b', ['Hello!', 'Goodbye!']]);
    
returns 'b' in browsers but function when running with your rhino version.

Did some additional tests with the jasmine code in real browsers, and the 'this' object provided when calling the setTimeout() function (j$.getGlobal()) is correct and points to the window.

rbri:
@aardvark179 
Looks like the bug is here...
image

state.stackTop + 1, points to the wrong element on the stack, the right one seems to be here state.stackTop +blen + 1. But this is the array [this , args], we have to take the first from this.

aardvark179:
Ah, `getApplyThis` needs to pull `this` from bound arguments if they exist.

rbri:
looks better, will do more tests

rbri:
@aardvark179 all the HtmlUnit test still green - i'm a bit surprised to see no test262 changes. Any idea why there are no tests that check this?

Oh, and many thanks for workin on that

rbri:
And why is this only a problem with the interpreter?

aardvark179:
It’s only in the interpreter because that is the only place that does this particular function peeling trick. The class file compiler just lets the apply and bind implementations do their thing.

Rhino has a lot of subtle divergence with the spec on functions calling, so we fail quite a few 262 tests in this area and might not notice further breakages.

At some point t I should look at what the minimum improvement would be that we’d need to do for this to improve, but I have some wider ideas around refactoring functions so we can fix generators and stuff.

rbri:
@aardvark179 what do you think can i merge that?

aardvark179:
I think this is safe to merge now.

rbri:
Thanks @aardvark179, another step forward

PR #2009, opened 2025-08-08
Make Object.assign copy over Symbol properties

### LINKED ISSUE #1391: ES2015 Object.assign doesn't copy over (own) symbol properties
Pretty much as the title says

balajirrao:
Closes #1391

### DIFF
--- a/tests/src/test/java/org/mozilla/javascript/tests/es6/NativeObjectTest.java
+++ b/tests/src/test/java/org/mozilla/javascript/tests/es6/NativeObjectTest.java
@@ -420,4 +420,35 @@ public class NativeObjectTest {
                     return null;
                 });
     }
+
+    @Test
+    public void symbolInObjectLiteral() {
+        final String script = "var s = Symbol('foo'); var o = {[s]: 'bar'}; o[s];";
+        evaluateAndAssert(script, "bar");
+    }
+
+    @Test
+    public void addSymbolProperty() {
+        final String script = "var s = Symbol('foo'); var o = {}; o[s] = 'bar'; o[s];";
+        evaluateAndAssert(script, "bar");
+    }
+
+    @Test
+    public void objectWrapperForSymbolAsKey() {
+        final String script =
+                "var s = Symbol('foo'); var o = Object(s); var obj = {[o]: 'bar'}; obj[s]";
+        evaluateAndAssert(script, "bar");
+    }
+
+    @Test
+    public void objectAssignCopiesSymbols() {
+        final String script =
+                "var s1 = Symbol('foo');"
+                        + "var s2 = Symbol('bar');"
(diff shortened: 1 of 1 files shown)


rhino/src/main/java/org/mozilla/javascript/NativeObject.java
```diff
@@ -722,22 +722,42 @@ private static Object js_assign(
...
                         AbstractEcmaObjectOperations.put(cx, targetObj, stringId, val, true);
                     }
                 }
             }
+
+            // This is a separate loop for Symbols, as they must be
+            // copied over after string properties
+            if (sourceObj instanceof ScriptableObject) {
+                for (Object key : ids) {
+                    if (key instanceof Symbol) {
```
rbri, line 751: is this correct here or should we use ScriptRuntime.isSymbol(key)

rhino/src/main/java/org/mozilla/javascript/AbstractEcmaObjectOperations.java
```diff
@@ -249,6 +249,25 @@ static void put(Context cx, Scriptable o, int p, Object v, boolean isThrow) {
         }
     }
 
+    /**
+     * Set ( O, P, V, Throw)
+     *
+     * <p><a href="https://262.ecma-international.org/12.0/#sec-set-o-p-v-throw">7.3.4 Set (O, P, V,
```
rbri, line 255: The spec says we have to return a boolean - i know that we are not able to do the right thing in many cases - and this was a major problem during the implementation of the proxy stuff.
But maybe we can try to follow the spec here at least for new methods (even if we can't return always the right value)

What do you think?

rhino/src/main/java/org/mozilla/javascript/NativeObject.java
balajirrao: I compared this to node to conclude that it is correct. I've added a couple of test cases now.

rhino/src/main/java/org/mozilla/javascript/AbstractEcmaObjectOperations.java
balajirrao: @rbri I'm out of ideas - I think in this case there's no simple way to have it return a boolean. Do you have any ideas ?

rbri: Ok, i see now, this is more or less the symbol version of the put operations we already have. And they also do not return a boolean. Fine for me, at least it is consistent and maybe we have to refactore this one day.

rbri (approved):

rbri:
@balajirrao can you please rebase this

rbri:
Thanks a lot, another step forward

PR #2010, opened 2025-08-11
Update shadow plugin to 9.0.1

rPraml:
this fixes warnings like

> The CopyProcessingSpec.setFileMode(Integer) method has been deprecated. This is scheduled to be removed in Gradle 9.0. Please use the filePermissions(Action) method instead. Consult the upgrading guide for further information: https://docs.gradle.org/8.12.1/userguide/upgrading_version_8.html#unix_file_permissions_deprecated

The ID has changed. See https://github.com/GradleUp/shadow/blob/main/README.md

### DIFF
--- a/rhino-all/build.gradle
+++ b/rhino-all/build.gradle
@@ -1,6 +1,6 @@
 plugins {
     id 'rhino.library-conventions'
-    id 'com.github.johnrengelman.shadow' version '8.1.1'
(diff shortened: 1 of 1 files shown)


gbrail:
This is great and gets rid of that pesky warning -- thanks!

I'm merging this manually because I found one more version we should update to since you created this.

gbrail:
Merged manually.

PR #2013, opened 2025-08-12
Fix parsing problem with optional chaining when xml feature is disabled

andreabergia:
In `Parser::propertyAccess` the code that supports optional chaining is not executed if the context does not have the feature `FEATURE_E4X`. This flag is enabled by default, which is why we never noticed it. I've completely removed the old "fallback" code, because it seems obsolete. Removing it doesn't seem to cause any regression.

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/Parser.java
+++ b/rhino/src/main/java/org/mozilla/javascript/Parser.java
@@ -3087,23 +3087,6 @@ public class Parser {
             memberTypeFlags = Node.DESCENDANTS_FLAG;
         }
 
-        if (!compilerEnv.isXmlAvailable()) {
-            int maybeName = nextToken();
-            if (maybeName != Token.NAME
-                    && !(compilerEnv.isReservedKeywordAsIdentifier()
-                            && TokenStream.isKeyword(
-                                    ts.getString(),
-                                    compilerEnv.getLanguageVersion(),
-                                    inUseStrictDirective))) {
-                reportError("msg.no.name.after.dot");
-            }
-
-            Name name = createNameNode(true, Token.GETPROP);
-            PropertyGet pg = new PropertyGet(pn, name, dotPos);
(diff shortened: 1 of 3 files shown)


rbri:
@andreabergia OMG, this XML stuff is disabled in HtmlUnit but i did not do any smoke test. Many, many thanks for finding this.

rhino/src/main/java/org/mozilla/javascript/Parser.java
```diff
@@ -3087,23 +3087,6 @@ private AstNode propertyAccess(int tt, AstNode pn, boolean isOptionalChain) thro
...
-                                    inUseStrictDirective))) {
-                reportError("msg.no.name.after.dot");
-            }
-
-            Name name = createNameNode(true, Token.GETPROP);
-            PropertyGet pg = new PropertyGet(pn, name, dotPos);
-            pg.setLineColumnNumber(lineno, column);
-            return pg;
-        }
-
```
rbri, line 3106: From a quick look i think this removal enables the parser to accept tokens like Token.XMLATTR here even if the xml stuff is disabled

andreabergia: Riiiiight of course it does. 🤦‍♂️ I'll fix that.

rbri (approved):

rhino/src/main/java/org/mozilla/javascript/Parser.java
rPraml: BTW: I came across the same issue when I attempted to separate the rhino-xml module: https://github.com/mozilla/rhino/pull/1699#issuecomment-2430920178

If FEATURE_E4X is enabled but rhino-xml is not on the classpath, the parser will accept these tokens, but the compiled script will not execute correctly.

gbrail:
looks good -- thanks!

rbri:
Thanks @andreabergia fantastic catch!

PR #2017, opened 2025-08-15
Fix combination of apply and call in the interpreter.

### LINKED ISSUE #2016: Regression II: ArrayIndexOutOfBoundsException in Interpreter.getArgsArray()
There is another on (as promised in #2006) (see also https://github.com/HtmlUnit/htmlunit/issues/1001)

    Index 1 out of bounds for length 1
    java.lang.ArrayIndexOutOfBoundsException: Index 1 out of bounds for length 1
	    at org.mozilla.javascript.Interpreter.getArgsArray(Interpreter.java:4823)

Seems like the fix does not work if the activation for the function is on

To reproduce i made this new tests for org.mozilla.javascript.tests.ApplyTest

    @Test
    public void applyToCallWithActivationCallsCorrectFunction() throws Exception {
        String script =
                "function foo(x) {\n"
                        + "  function inner(s) { return s };\n"
                        + "  return inner(this);\n"
                        + "};\n"
                        + "foo.call.apply(foo, ['b', 'Hello!']);\n";

        Utils.assertWithAllModes("Hello!", script);
    }

    @Test
    public void applyToCallWithActivationSetsCorrectFunctionThis() throws Exception {
        String script =
                "function foo(x) {\n"
                        // the inner function enables activation
                        + "  function inner(s) { return s.toString() };\n"
                        + "  return inner(this);\n"
                        + "};\n"
                        + "foo.call.apply(foo, ['b', 'Hello!']);\n";

        Utils.assertWithAllModes("b", script);
    }

aardvark179:
This should fix #2016, @rbri could you see if this fixes the problems you've seen?

### DIFF
--- a/tests/src/test/java/org/mozilla/javascript/tests/ApplyTest.java
+++ b/tests/src/test/java/org/mozilla/javascript/tests/ApplyTest.java
@@ -58,8 +58,8 @@ public class ApplyTest {
     public void applyToCallWithActivationCallsCorrectFunction() throws Exception {
         String script =
                 "function foo(x) {\n"
-                        + "  function inner(s) { return x };\n"
-                        + "  return inner(this);\n"
+                        + "  function inner(s) { return s };\n"
+                        + "  return inner(x);\n"
                         + "};\n"
                         + "foo.call.apply(foo, ['b', 'Hello!']);\n";
 


tests/src/test/java/org/mozilla/javascript/tests/ApplyTest.java
```diff
@@ -53,4 +53,29 @@ public void applyToCallSetsCorrectFunctionThis() throws Exception {
 
         Utils.assertWithAllModes("b", script);
     }
+
+    @Test
+    public void applyToCallWithActivationCallsCorrectFunction() throws Exception {
+        String script =
+                "function foo(x) {\n"
+                        + "  function inner(s) { return x };\n"
+                        + "  return inner(this);\n"
```
rbri: my bad this should be 

                        + "  function inner(s) { return s };\n"
                        + "  return inner(x);\n

rbri:
Thanks @aardvark179,
from a first look it works fine, now i have to fight with some HtmlUnit adoptions that still break... but i guess you have similar effects with your fork from time to time ;-)

tests/src/test/java/org/mozilla/javascript/tests/ApplyTest.java
aardvark179: Fixed.

rbri:
@aardvark179 thanks for the fast solution

PR #2023, opened 2025-08-16
implement `TypeInfoFactory.getConsolidationMapping(...)`

ZZZank:
# Java Generic support

- Type Consolidation & Generic support for Method/Constructor (partialy in this PR)

This PR implements the `TypeInfoFactory.getConsolidationMapping(...)`, enabling us to handle classes with no type variable on itself, but have generic superclass/interface.

Take `IntStream` as an example:

```java
public interface IntStream extends BaseStream<Integer, IntStream> {
}

public interface BaseStream<T, S extends BaseStream<T, S>> {
    S onClose(Runnable closeHandler);
}
```

`TypeInfoFactory.getConsolidationMapping(...)` will generate a mapping like this: `T -> Integer, S -> IntStream`, enabling us to eliminate the type variable `S` in `S onClose(Runnable closeHandler)`.

## Type Consolidation

The idea is simple: replace the type variable with actual type, so that when the type is requested, we can get the exact type instead of what the type variabled is erased to.

For our example, it means replacing `E` with `Integer`, so that we can know the parameter type is `Integer` instead of `Object`.

My plan is to introduce a 2-stage type consolidation process to replace all type variable if possible. The logic and actual implementation will be similar to the one in this PR: https://github.com/KubeJS-Mods/Rhino/pull/55 , but details might vary.

### Stage 1 (This PR): "Flattening"

Assuming that we have these classes, as a rather extreme example:

```java
class A<Ta> { ... }
interface B<Tb> { ... }
class C<Tc> extends A<Tc> { ... }

class D<Td> extends C<Td> implements B<A<Td>> {
    Ta getA() { ... } // inherited from A
    Tb getB() { ... } // inherited from B
    Tc getC() { ... } // inherited from C
    Td getD() { ... }
    Td setD(Td td) { ... }
}
```

What the first type consolidation does is removing all type variables from superclasses/interfaces, essentially "flattened" those types to types used only by class `D` itself:

```java
class D<Td> extends C<Td> implements B<A<Td>> {
    Td getA() { ... } // inherited from A
    A<Td> getB() { ... } // inherited from B
    Td getC() { ... } // inherited from C
    Td getD() { ... }
    Td setD(Td td) { ... }
}
```

### Stage 2 (WIP): handle `Exampl<T> -> Exampl<Integer>`

This stage of type consolidation will only apply to types with generics. That is, it will be applied to `Function<String, String>`, or `List<Integer>`, but not `String` because it's not a generic class, or raw type `Map` because there's no generic info given to us to use.

Take the aforementioned `Exampl.ints` as example, when calling `Exampl.ints.add(3)`, a new mapping will be generated from the type of `Exampl.ints` (which is `List<Integer>`), forming a `E -> Integer` conversion. Apply the conversion to method parameter types, then Rhino will be able to convert object to the correct type Integer.

Noted that the `Exampl` class will not need this stage, because the class itself has no type variable, so all type variables are consolidated at stage 1:
```java
public class Exampl extends ArrayList<Integer> {
    public boolean add(Integer o) { ... } // E -> Integer
}
```


rPraml:
Great to see the progress here! I did a quick pass over the PR and didn’t spot any major issues.

Decycle: should prevent to introduce new cycles, so that splitting into liveconnect would become more difficult (I think this is not the case here - I will update my liveconnect PR in some weeks - have to refactor and merge some simple things first)

Tests: I've made test cases some times ago targeting the same issue
https://github.com/mozilla/rhino/pull/827/files#diff-b07ab806fc2c0ac96af95754139408b877dbfb6e983c9addd865d6efe4750589R1
Maybe you can add a `class X<? extends Number>` or `T extends Foo & Bar`, if this should be already supported.

ZZZank:
> Maybe you can add a `class X<? extends Number>` or `T extends Foo & Bar`, if this should be already supported.

Not supported yet. This PR will allow us to handle type variables from superclasses/interfaces, aka the `String` in:

```java
class String2ObjectMap<VALUE> extends HashMap<String, VALUE> {}
```

but the `VALUE` is left as-is, because it's declared by the class itself. This means that types like `List<? extends Number>` and `Iterable<String>` are only partially supported now.

gbrail:
This looks pretty complete to me. @rPraml if you think that this works for your use cases I'm happy to go ahead.

rPraml:
I will try to provide an update the next days...

gbrail:
This makes sense to me and works, at least with the tests, but I am wondering how we explain it to people! Maybe we just need a simple example in the PR description, along the lines of what you can do now that you couldn't do before!

rPraml:
Hello @ZZZank 

I found some time today to experiment with the new features. See this commit: https://github.com/FOCONIS/rhino/commit/31c92f478f46e40136c316fc7d2b8673b0e46c13

Here I changed the following things:
- getConsolidationMapping(Class<?> from) -> getConsolidationMapping(Type from) as I think, we have to consolidate types, that are stored in MemberBoxes etc. (e.g. `List.add(T)` has the T as argumentType)
- implement some synchronisation
- provided some tests to demonstrate use cases

I know, this is still work in progress, but at the moment, I'm having trouble understanding exactly how I'll be able to implement my use cases later on. Maybe you can tell me, how are your plans to further integrate this

(I have a rather old PR (#827) for reference, where my use cases will work)

Use case 1:
- Given is a `class IntegerArrayList extends ArrayList<Integer>`
- This java object is assigned to a javascript value `intList`
- `intList[0] = 3` must write that value as `int` and not as double

To make it work, I did store the `valueType` in NativeJavaList (see https://github.com/mozilla/rhino/pull/827/files#diff-75db5a280e5a4c0d746efc74971652d3273fadde52cfc4a9e48a2a4c6a56c6d3R53)

Now I thought, I can get the resolved type somehow with:
```java
var consolidated = TypeInfoFactory.GLOBAL.getConsolidationMapping(list.getClass());
var params = consolidated.get(TypeInfoFactory.GLOBAL.create(List.class));
// params should contain "Integer".
```
So I need to be able to ask a class that implements `List`: What generic types do you declare for `List<T>`, and get one value in return. (analog for `Map<K,V>` I will expect two values)

Use case 2:
- There is no dedicated class, but a generic field `List<Integer>` in a POJO

More use cases you can see in the tests. If all would pass, I'm happy ;)

How should we proceed here? Should I update my branch #827 and try to incorporate as much of your implementation as possible so that my tests pass?

Roland

ZZZank:
> need a simple example in the PR description, along the lines of what you can do now that you couldn't do before

Well the awkward thing is, this PR didn't add any feature to Rhino, because `TypeInfoFactory.getConsolidationMapping(...)` was never used before. But I can provide an example usage (also serves as a preview for future Java Generic support PR):

```java
class IntegerArrayList extends ArrayList<Integer> {}
var mapping = TypeInfoFactory.GLOBAL.getConsolidationMapping(IntegerArrayList.class);

var typeVar = TypeInfoFactory.GLOBAL.create(List.class.getTypeParameters()[0]); // the 'E' in 'List<E>'

var result = typeVar.consolidate(mapping); // result will always be the 'Integer' in 'extends ArrayList<Integer>'
```

In short, this method can create a mapping that maps type variables declared by superclasses/interfaces to concrete type, or type variables declared by input class.

ZZZank:
Hi @rPraml

I have checked you experiment commit, and I think many of the changes are nice-to-have but not required. Let me explain:

- .getConsolidationMapping(**Type** from)

Getting consolidation mapping directly from `List<T>` is nice and intuitive, but this might cause performance issue, because we need to do a full mapping generation for each `List<String>`, `List<Integer>`, `List<Double>`, ...

My planned solution is that we first collect a mapping from `List` raw class, and transform its type variable (`E` in `List<E>`) to actual type param (`String` in `List<String>`), so that we can reuse most of the consolidation mapping without losing accuracy. The first step (mapping from `List` class) is exactly the `.getConsolidationMapping(...)` method.

But with that said, I think `.getConsolidationMapping(Type from)` can still be sort of a helper method that combines the two steps above together.

- **Synchronisation** for `.getConsolidationMapping(...)`

Not required I believe.

The reason is that we don't care about the identity of `mapping`, we only need to make sure that `mapping` will always contains the same content when given the same class, which is what `.computeConsolidationMapping()` can ensure.

- **Test**s 

I like those tests, they are really comprehensive. But unfortunate for this PR, these features expected by the test:
- Generic support for field (all lists/maps/beans are accessed via field)
- Generic support for list / map
- `SomeObj<SomeParam>`-like object support

are not implemented at all at the moment. So let's keep it for my next PR ¯\\\_(ツ)_/¯

- Use case 1: class with generic parent

Let's continue use the example you provided: `class IntegerArrayList extends ArrayList<Integer> {}`. You can get the mapped type by doing this:
```java
var mapping = TypeInfoFactory.GLOBAL.getConsolidationMapping(IntegerArrayList.class);
var E = TypeInfoFactory.GLOBAL.create(List.class.getTypeParameters()[0]); // the 'E' in 'List<E>', lazy static field recommended

var result = E.consolidate(mapping); // result will always be the 'Integer' in 'extends ArrayList<Integer>'
```

The same goes for map:
```java
class StringIntMap extends HashMap<String, Integer> {}
var mapping = TypeInfoFactory.GLOBAL.getConsolidationMapping(StringIntMap.class);

var K = TypeInfoFactory.GLOBAL.create(Map.class.getTypeParameters()[0]);
var V = TypeInfoFactory.GLOBAL.create(Map.class.getTypeParameters()[1]);

var resultK = K.consolidate(mapping); // String
var resultV = V.consolidate(mapping); // Integer
```

- Use case 2: generic class

This is something that will be implemented in the future. You can see my progress at: https://github.com/ZZZank/rhino-mozilla/tree/type-consolidation-apply , the general idea is:

```java
var type = (TypeInfo) (List<String>); // dummy code

var stage1mapping =  TypeInfoFactory.GLOBAL.getConsolidationMapping(type.asClass()); // mapping from `List` raw class
var stage2mapping = type.extractTranformMapping(); // contains: E -> String

var inputType = (TypeInfo) (E in Collection<E>); // type to be consolidated

var result = inputType // E in Collection<E>
    .consolidate(stage1mapping) // E in List<E>
    .consolidate(stage2mapping); // String
```

gbrail:
It sounds like it makes sense to merge this, so I'm going to merge it and we can move on to the next steps that you discussed. Thanks!

PR #2026, opened 2025-08-17
Implement ES2025 Set methods

### LINKED ISSUE #1542: ES2025 New Set methods
https://github.com/tc39/proposal-set-methods

anivar:
Implements ES2025 Set methods as defined in the [TC39 proposal](https://github.com/tc39/proposal-set-methods).

## New methods
- `Set.prototype.intersection(other)` - returns elements in both sets
- `Set.prototype.union(other)` - returns elements in either set
- `Set.prototype.difference(other)` - returns elements in this but not in other
- `Set.prototype.symmetricDifference(other)` - returns elements in either but not both
- `Set.prototype.isSubsetOf(other)` - checks if this is a subset of other
- `Set.prototype.isSupersetOf(other)` - checks if this is a superset of other
- `Set.prototype.isDisjointFrom(other)` - checks if sets have no common elements

## Implementation details
- Supports Set-like objects (with `size`, `has`, and `keys` properties)
- Optimizes intersection and disjoint checks by iterating through smaller set
- Follows spec behavior for type checking and iteration
- All methods properly handle edge cases like empty sets

## Testing
Added comprehensive unit tests covering:
- Basic operations with numbers, strings, mixed types
- Empty set handling
- Set-like object compatibility
- All edge cases

Note: Some test262 tests may still fail due to their use of spread syntax and classes, which Rhino doesn't yet support.

Fixes #1542

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/NativeSet.java
+++ b/rhino/src/main/java/org/mozilla/javascript/NativeSet.java
@@ -95,6 +95,65 @@ public class NativeSet extends ScriptableObject {
                 DONTENUM,
                 DONTENUM | READONLY);
 
+        // ES2025 Set methods
+        constructor.definePrototypeMethod(
+                scope,
+                "intersection",
+                1,
+                (Context lcx, Scriptable lscope, Scriptable thisObj, Object[] args) ->
+                        realThis(thisObj, "intersection").js_intersection(lcx, lscope, args),
+                DONTENUM,
+                DONTENUM | READONLY);
+        constructor.definePrototypeMethod(
+                scope,
+                "union",
+                1,
+                (Context lcx, Scriptable lscope, Scriptable thisObj, Object[] args) ->
+                        realThis(thisObj, "union").js_union(lcx, lscope, args),
+                DONTENUM,
+                DONTENUM | READONLY);
+        constructor.definePrototypeMethod(
+                scope,
+                "difference",
+                1,
+                (Context lcx, Scriptable lscope, Scriptable thisObj, Object[] args) ->
+                        realThis(thisObj, "difference").js_difference(lcx, lscope, args),
+                DONTENUM,
+                DONTENUM | READONLY);
+        constructor.definePrototypeMethod(
+                scope,
+                "symmetricDifference",
+                1,
+                (Context lcx, Scriptable lscope, Scriptable thisObj, Object[] args) ->
+                        realThis(thisObj, "symmetricDifference")
+                                .js_symmetricDifference(lcx, lscope, args),
+                DONTENUM,
+                DONTENUM | READONLY);
+        constructor.definePrototypeMethod(
+                scope,
+                "isSubsetOf",
+                1,
+                (Context lcx, Scriptable lscope, Scriptable thisObj, Object[] args) ->
+                        realThis(thisObj, "isSubsetOf").js_isSubsetOf(lcx, lscope, args),
+                DONTENUM,
+                DONTENUM | READONLY);
+        constructor.definePrototypeMethod(
+                scope,
+                "isSupersetOf",
+                1,
+                (Context lcx, Scriptable lscope, Scriptable thisObj, Object[] args) ->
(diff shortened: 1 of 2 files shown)


anivar:
Updated the error message format again. The Test262 tests expect the simpler `msg.isnt.function` format:
- Changed from: `msg.isnt.function.in` ("Cannot call property {0} in object {1}. It is not a function, it is '{2}'.")
- Changed to: `msg.isnt.function` ("{0} is not a function, it is {1}.")

This should fix the failing tests for keys-is-callable and has-is-callable validation.

For the array-throws tests, arrays in ES6+ do have `keys()` and `has()` methods, so those should work. The tests might be expecting a different error or behavior.

rbri:
@anivar thanks a lot for this nice contribution. Can you please regenerate the test262.properties file and make it part of this PR (see https://github.com/mozilla/rhino/tree/master/tests).

At least for me it looks like your change let some more test pass....

rbri:
@anivar do you really think that using Claude Code in this way is a good idea?

image

The test262.properties is one cornerstone of our test suite - we manage that carefully over the time to be sure the implementation of new features does not introduce regressions and matches the spec. I know that regeneration is a tricky process and you have to read some documentation and usually tweak your local environment. Please do that carefully and do not forward this task to brainless machines.

And if you got unexpected results like some failing tests that are not expected, please analyze the reason and not asking claude or other AI tool for patching. If you really think there is something to improve in der RegEx field, please make a new PR.

rbri:
Have done a check for test262 of the last useful commit and it looks here like that
image

Great progress more than 60 test are passing now.

But ;-)...
Usually we try to make the implementation of new functions a standard complient as possible. It will be great if you can have a look and e.g. try to figure out if you can make tests like

    prototype/difference/combines-empty-sets.js
    prototype/difference/combines-itself.js
    prototype/difference/combines-Map.js
    prototype/difference/combines-same-sets.js
    prototype/difference/combines-sets.js
    prototype/difference/converts-negative-zero.js

also passing.

anivar:
> And if you got unexpected results like some failing tests that are not expected, please analyze the reason and not asking claude or other AI tool for patching. If you really think there is something to improve in der RegEx field, please make a new PR.

Hi @rbri,

Thanks for the review and for running the tests.

I should have been more careful with that  test262.properties regeneration. I was pushing to wrap this up over the weekend and inadvertently took shortcuts that I shouldn't have. I understand the importance of maintaining that file properly.

As for the RegExp changes, those were definitely out of scope for this PR. Got sidetracked trying to fix unrelated test failures when I should have stayed focused on the Set methods implementation.

I'll clean this up and keep the PR focused on the Set methods only. Will also ensure the test262.properties is handled properly.

Thanks for the feedback.  it helps keep things on track.

anivar:
Hi @rbri,

I've updated the difference method as you suggested. The size optimization is now in place - when `this.size > other.size`, it iterates through `other.keys()` to avoid calling `has()`.

After regenerating test262.properties, we're down from 158 to 94 failing Set tests. That's 64 tests now passing.

The 6 specific tests you mentioned are still failing though. Looking into it, they seem to fail because of the spread operator usage in the tests rather than the implementation itself. The core logic follows the spec correctly.

Let me know if you'd like me to dig deeper into why those particular tests aren't passing yet.

rbri:
> After regenerating test262.properties, we're down from 158 to 94 failing Set tests. That's 64 tests now passing.

Great!

> The 6 specific tests you mentioned are still failing though. Looking into it, they seem to fail because of the spread operator usage in the tests rather than the implementation itself. The core logic follows the spec correctly.

Ah i see, there we can only hope for the upcomming spread support...

From a quick look it looks like there is a testcase named set-like-array.js for all the different functions. I guess the implementation of the tests is similar... but for some of your new function the test seems to pass and for some not. Maybe you can have a look....

anivar:
Closing in favor of #2028 which implements a more comprehensive set of ES2025 features as suggested. The new PR includes Set methods, Promise.try, and ArrayBuffer transfer methods in a unified implementation.

PR #2027, opened 2025-08-17
cleanup wrapping for Java method args, constructor args, and method return value

ZZZank:
- extract wrapping code for method/constructor args to `MemberBox.wrapArgsInternal(...)`
- skip wrapping for method return value if return type is `void`

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/MemberBox.java
+++ b/rhino/src/main/java/org/mozilla/javascript/MemberBox.java
@@ -310,6 +310,61 @@ final class MemberBox implements Serializable {
         }
     }
 
+    Object[] wrapArgsInternal(Object[] args) {
+        var argTypes = getArgTypes();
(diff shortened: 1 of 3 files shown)


gbrail:
Sorry no one has commented on this. It looks like some refactoring / code cleanup. Is there a test case you're hoping to fix, or other motivation? I don't use the Java reflection wrapping in Rhino for much these days so I want to make sure that we have some way to verify this. Thanks!

ZZZank:
This PR was made simply bacause the arg wrapping code in `NativeJavaClass.constructInternal()` and `NativeJavaMethod.call()` are completely identical, so yes, this is a code cleanup PR.

As of testing, my first PR for Rhino introduced `NativeJavaMethodTest`, which happens to cover arg wrapping with/out vararg.

gbrail:
OK, these make sense then -- thanks!

PR #2031, opened 2025-08-20
Fix: methods should not have a prototype property

andreabergia:
The spec says that these two should be `undefined`:

```js
let o = { f(){} };
o.f.prototype;
Object.getOwnPropertyDescriptor(o.f, 'prototype')
```

However, rhino was always creating the `prototype` property even for shorthand methods.


rbri:
a small step for test262.properties but...

PR #2032, opened 2025-08-22
Symbol.for should return context-wide unique symbols

balajirrao:


gbrail:
I'd be curious to hear how various people feel about this. Since Rhino doesn't have the explicit context of a "container" like V8, we either use the Context or the top-level scope object. The problem with using the Context is that it's quite possible to create multiple contexts and use the same set of Rhino objects across all of them, which causes different problems.

So I'm curious what you're hoping to fix with this, and I wish that we had something top-level to make it all more clear!

p-bakker:
See #1079

balajirrao:
> The problem with using the Context is that it's quite possible to create multiple contexts and use the same set of Rhino objects across all of them, which causes different problems.

Ah! I hadn't thought of this. 

This fix was motivated by trying to run an Effect app. Effect defines `Symbol.for("effect/Effect")` in two files - [packages/effect/src/internal/core.ts ](https://github.com/Effect-TS/effect/blob/3b2609409ac1e8c6939d699584f00b1b99c47e2e/packages/effect/src/internal/core.ts#L83)and [packages/effect/src/internal/effectable.ts](https://github.com/Effect-TS/effect/blob/3b2609409ac1e8c6939d699584f00b1b99c47e2e/packages/effect/src/internal/effectable.ts#L14) and uses them interchangeably.

I'd love to hear ideas on how could we move forward.

p-bakker:
From my understanding of the spec, the idea is that the Symbol registry spans both Context and Top Level scopes, so should be contained in a singleton class shared by all

aardvark179:
I think that #2105 may have superseded this PR.

balajirrao:
> I think that #2105 may have superseded this PR.

Thanks @aardvark179. I'll close this one.

PR #2033, opened 2025-08-23
rename helper to assertWithAllModes_StrictAndNonStrict_ES6 to better describe what the method does

rbri:
during a review i was a bit puzzled seeing the method evaluateAndAssert() because i expected to make use of ES6 for this tests. After checking the impl i think we should rename the method. Hope you agree. 

And maybe someone has a better/shorter suggesting for a good name.


gbrail:
This is fine with me, I assume no one else will have big objections either!

PR #2036, opened 2025-08-23
Run android integration tests in emulator

rPraml:
The trigger was PR #2018 and #1152, which prompted me to revisit running Android tests in an emulator. 
It’s still not fully finished, but I wanted to share the concept anyway.

In short: What does this PR

`install-android-sdk.sh`
downloads the SDK in RHINO_ROOT/android-sdk - if you do not already have one installed

`it-android.sh`
starts an android emulator, the emulator is accessible over http://localhost:6080

`it-android`
this project is a minimal android project, with embedded rhino (based on https://github.com/czak/minimal-android-project)
- it executes all scripts in `tests` and outputs the result in a very minimalistic UI:
grafik

Open TODOs
- make a testrunner similar to test262, so that it might run in CI (DONE!)
- exclude the it-android, if no SDK is installed (DONE!)

@anivar I adapted your performance tests and I get ~297ms for 10000 iterations, with or without your fix. So there is no difference in the emulator. Do we perform the correct test? Or do I still not understand, what the problem is
(What I've understood: On native android (Dalvik/ART), the performance is worse, on a linux JVM, the performance is OK)

### DIFF
--- a/it-android/src/main/java/org/mozilla/javascript/android/TestCase.java
+++ b/it-android/src/main/java/org/mozilla/javascript/android/TestCase.java
@@ -6,7 +6,6 @@ import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.io.Reader;
 import java.io.UncheckedIOException;
-import java.net.URI;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.List;
@@ -15,9 +14,6 @@ import org.mozilla.javascript.ContextFactory;
 import org.mozilla.javascript.ScriptRuntime;
 import org.mozilla.javascript.Scriptable;
 import org.mozilla.javascript.ScriptableObject;
-import org.mozilla.javascript.commonjs.module.ModuleScript;
-import org.mozilla.javascript.commonjs.module.ModuleScriptProvider;
-import org.mozilla.javascript.commonjs.module.RequireBuilder;
 
 /**
  * Utility class, that search for testcases in "assets/tests".
@@ -36,7 +32,8 @@ public class TestCase {
             new ContextFactory() {
                 @Override
                 protected boolean hasFeature(Context cx, int featureIndex) {
-                    if (featureIndex == 20 /*Context.FEATURE_ENABLE_XML_SECURE_PARSING*/) return false;
+                    if (featureIndex == 20 /*Context.FEATURE_ENABLE_XML_SECURE_PARSING*/)
+                        return false;
                     return super.hasFeature(cx, featureIndex);
                 }
 
@@ -46,7 +43,7 @@ public class TestCase {
                     cx.setLanguageVersion(org.mozilla.javascript.Context.VERSION_ES6);
                     cx.setGeneratingDebug(false);
                     cx.setOptimizationLevel(-1);
-                    //cx.seal(null);
+                    // cx.seal(null);
                     return cx;
                 }
             };


anivar:
@rPraml I see you tested with and without the fix and got the same performance. This actually makes sense - the JIT compilation failure is environment-specific and typically occurs on:
  - Real Android devices (not emulators)
  - Older Android versions with stricter method size limits
  - ARM architectures vs x86/x64 emulators

  The issue in #1365 was reported on real devices where the interpretLoop method exceeds the JIT threshold. Your emulator environment might have different or relaxed limits, which is why you're
  seeing consistent good performance.

  This reinforces why the fix is needed - it prevents the issue on affected devices while having no impact on environments that don't hit the threshold (like your emulator).

rPraml:
Yes, until now I could **not** reproduce the issue. 
Neither in the emulator nor on my real android device. 

> the JIT compilation failure is environment-specific

It seems that this bug is very tricky. Maybe I can retry on an older emulator version (although this would be x86) 

> This reinforces why the fix is needed

And it would be good, to have a test for this to avoid a regression in future releases, as the interpreter loop is code, that will be modified by a lot of people. 

> it prevents the issue on affected devices while having no impact on environments that don't hit the threshold (like your emulator).

@anivar I know, you are busy, but maybe you can provide details on which android device (android + SDK version) you were able to reproduce the issue. Maybe this is reproduceable in the emulator with same sdk (even if it is x86 and not arm)

.github/workflows/gradle.yml
```diff
@@ -21,6 +21,12 @@ jobs:
     - name: Check out test262
       # We don't actually want all the history for this part
       run: git submodule update --init --single-branch
+      # we need this for android enmulator
+    - name: Enable KVM
+      run: |
+        echo 'KERNEL=="kvm", GROUP="kvm", MODE="0666", OPTIONS+="static_node=kvm"' | sudo tee /etc/udev/rules.d/99-kvm4all.rules
+        sudo udevadm control --reload-rules
+        sudo udevadm trigger --name-match=kvm
```
rPraml: CHECKME: Is this really needed?

.github/workflows/gradle.yml
```diff
@@ -40,6 +46,18 @@ jobs:
       with:
         name: reports-java-21
         path: '*/build/reports'
+    - name: Build and test for Android
+      uses: reactivecircus/android-emulator-runner@v2
+      with:
+        api-level: 33
+        arch: x86_64
+        script: ./gradlew it-android:connectedCheck
```
rPraml: CHECKME: Should this run as extra step or as part of java 21 buld

.gitignore
```diff
@@ -29,3 +29,7 @@ rhino.iws
 
 .java-version
 .kotlin
+
+# android specific
+android-sdk/
+local.properties
```
rPraml, line 35: required for the `install-android-sdk.sh` script. Should we provide such script or rely that everyone can install the SDK on their machines?

buildSrc/src/main/groovy/rhino.java-conventions.gradle
```diff
@@ -4,7 +4,7 @@
 
 plugins {
     id 'java-library'
-    id 'com.diffplug.spotless'
+    id 'rhino.spotless-conventions'
```
rPraml, line 7: Move that to separate PR

it-android.sh
```diff
@@ -0,0 +1,36 @@
...
+while ! (docker exec -it $CONTAINER_NAME adb shell getprop sys.boot_completed | grep -wq 1) 2>/dev/null ; do
+  echo "- Waiting for emulator to boot..."
+  sleep 10
+done
+#echo "- Emulator booted"
+#docker exec -it $CONTAINER_NAME adb uninstall com.example.rhino
+#docker exec -it $CONTAINER_NAME adb install -r -t $APP && \
+#docker exec -it $CONTAINER_NAME adb shell am start -n com.example.rhino/com.example.rhino.MainActivity
+./gradlew it-android:connectedAndroidTest
+./gradlew it-android:installDebug
```
rPraml: TODO: Would be nice, if someone else can confirm, these scripts will work

it-android/README.md
```diff
@@ -0,0 +1,12 @@
+# Android integration tests
+
+TODO: Write proper documentation
```
rPraml: Still a TODO

it-android/build.gradle
```diff
@@ -0,0 +1,68 @@
...
+
+// read the (android specific) local.properties file
+def localPropertiesFile = rootProject.file("local.properties");
+Properties localProperties = new Properties()
+if (localPropertiesFile.exists()) {
+    localProperties.load(localPropertiesFile.newDataInputStream())
+}
+
+if (!System.getenv("ANDROID_HOME") && !localProperties.get("sdk.dir")) {
+    System.out.println("No Android SDK found. Skipping build")
```
rPraml, line 37: Skip android build for users, that do not have a SDK installed

it-android/build.gradle
```diff
@@ -0,0 +1,68 @@
...
+} else {
+    apply plugin: 'com.android.application'
+
+    android {
+        namespace = "org.mozilla.javascript.android"
+
+        compileSdkVersion 33
+        defaultConfig {
+            minSdk = 33
+            targetSdk = 33
```
rPraml, line 47: Tried to use the lowest possible sdk version, so that the apk is compatible with as many as possible phones, but `./gradlew :it-android:lintDebug` means:

Error: Google Play requires that apps target API level 33 or higher. [ExpiredTargetSdkVersion]

So let's use 33 here

it-android/src/main/assets/assert.js
```diff
@@ -0,0 +1,417 @@
+// Copyright 2008 the V8 project authors. All rights reserved.
```
rPraml, line 1: This is a 1:1 copy of assert.js from MozillaTestSuite

it-android/src/androidTest/java/org/mozilla/javascript/android/test/RhinoTest.java
```diff
@@ -0,0 +1,39 @@
...
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+import org.mozilla.javascript.android.TestCase;
+
+@RunWith(Parameterized.class)
+public class RhinoTest {
```
rPraml, line 23: TODO: write some javadocs...

it-android/src/main/java/org/mozilla/javascript/android/TestCase.java
```diff
@@ -0,0 +1,90 @@
...
+
+    private final String name;
+    private final Scriptable global;
+    private final AssetManager assetManager;
+
+    private static final ContextFactory factory =
+            new ContextFactory() {
+                @Override
+                protected boolean hasFeature(Context cx, int featureIndex) {
+                    if (featureIndex == Context.FEATURE_ENABLE_XML_SECURE_PARSING) return false;
```
rPraml: Secure-parsing cannot be changed on android-devices

anivar:
@rPraml Thanks for this Android testing infrastructure!

The JIT issue from #2018 occurs when template literals generate methods exceeding ~7392KB, causing ART compiler failures on certain Android devices.

**Key factors:**
- Android 8-10 versions most affected
- Device-specific memory/CPU constraints
- Hard to reproduce in emulators

**Suggestions for your framework:**
- Test multiple API levels (26, 28, 30)
- Add bytecode size monitoring
- Include ARM architecture in CI matrix

The fix prevents the issue, but automated tests will help catch regressions.

anivar:
@rPraml Great work on the implementation details\! Here's feedback on your specific questions:

**KVM rules** (gradle.yml): Yes, needed for hardware acceleration in GitHub Actions runners. Without it, emulators run extremely slow.

**Build integration**: Run as a separate matrix job, not part of Java 21 build. This allows:
- Parallel execution
- Clear failure isolation  
- Easier to skip for non-Android changes

**SDK installation script**: Keep it. Not everyone has Android SDK locally, and it ensures consistent SDK versions across contributors.

**API Level 33**: While Play Store requires API 33, for testing we should include older versions too:
```gradle
minSdk = 26  // Android 8.0 where JIT issues start
targetSdk = 33
```

**Architecture suggestion**: Add ARM to your emulator matrix since x86_64 won't catch ARM-specific JIT issues:
```yaml
strategy:
  matrix:
    api-level: [26, 28, 30, 33]
    arch: [x86_64, arm64-v8a]
```

The local Docker approach is excellent for debugging. Overall, this is a solid foundation\!

gbrail:
This is great progress -- I know it's a pain to get this in. 

I think that we should create a separate GitHub Action for Android, rather than putting it in the main tests. That might make it easier to develop, and also it can run in parallel. We still won't merge anything until all the checks have passed.

Also, I personally don't use Docker, but Podman, and I wonder if others do too? If so I may try to adapt the script (should just be replacing "docker" with "podman").

821938089:
minsdk should be 21, 26 is too high.

821938089:
CI failure under minSdk 21 is a regression introduced by PR #1785.
For details, see https://github.com/mozilla/rhino/pull/1785#issuecomment-2683838809.

Switching the minSDK back to 26 is not the correct solution, as this will prevent Android projects with a minSDK of 21 from compiling, resulting in the same failure as the CI.

rPraml:
This task gave me a much better understanding of how Android works under the hood.

(For example, I wasn’t aware that the entire rhino.jar is converted by dx into a completely different Dalvik bytecode format and isn’t bundled into the .apk as a regular .jar file.)

**Summary**

The goal of this pull request is to provide the infrastructure for Android tests (we are not aiming to fix any bugs at this stage.)

✅ Tests run successfully for API levels 26, 28, 30, and 33 on x86_64
❌ Tests for arm64-v8a have not yet been made to run as a GitHub Action
❌ Tests for API level 21 are currently not possiblel, since PR #1785 introduced some incompatibilities
❌ I haven’t been able to reproduce bug #2018 so far

**Next Steps**

- Over the next few days, I’ll improve the documentation and related materials to establish a fully functional Android testing infrastructure as quickly as possible.
- Hopefully someone will be able to reproduce bug #2018 (possibly by running with less RAM or other system limitations, though @anivar mentioned this is likely very hard to reproduce in an emulator).
- We need to decide on the minimum supported API level (minSdkVersion).
  Thanks to @821938089 for pointing to the right direction
  @aardvark179, @rbri, @gbrail:
  I reviewed the changes introduced in #1785. It really only affects the ThreadedAccess class.
  Although `@SuppressWarnings("AndroidJdkLibsChecker")` is used, accessing `VarHandle.compareAndExchange` causes a compilation error with dx, meaning Rhino can no longer be used on older Android versions. 
  I see the following options here:  
  - We are defining API level 26 as the minSdkVersion
  - We find some workaround for the `ThreadedAccess` issue (e.g. filter this class in dexing process - I already tried a proguard-rule, but I wasn’t able to make it work) - with desugaring, we could at least achive compatibility with API 21 (See #2042)
  - Remove all non-android-libs from the core `rhino` module and provide them with a service. So we will get a `rhino-jvm` and and maybe a `rhino-dalvik` module, that provides different impls for `ThreadedAccess`
- And last but not least, some good news: Due my reasearch, I found an interesting project: https://github.com/linkedin/dexmaker It should be possible to generate dalvik bytecode.

anivar:
@rPraml Thank you for this comprehensive Android testing infrastructure! Your deep dive into the dx/Dalvik bytecode conversion and the testing setup is excellent work.

  I've been thinking about the DexMaker suggestion and wondering if there might be alternative approaches that better align with Rhino's strengths and modern Android development:

  On DexMaker:
  While DexMaker (Apache 2.0) is license-compatible with Rhino (MPL 2.0), it primarily addresses the challenge of runtime bytecode generation on Android. Note that only dexmaker-mockito-inline requires API 28+ - the core library supports older versions. Still, I wonder if the added complexity and APK size increase are justified given that most Rhino Android use cases might not need dynamic compilation?

  Alternative Approaches to Consider:
  1. Kotlin Multiplatform - Could allow sharing code between JVM and Android while keeping Rhino's pure-Java nature
  2. GraalVM Native Image - Experimental Android support could eliminate interpreter overhead entirely
  3. Focus on Rhino's unique strengths - Rather than competing on performance, optimize for use cases where Rhino excels:
    - Testing frameworks that need deep Java mocking
    - Build tools and automation scripts
    - Enterprise apps with existing Rhino dependencies
    - Educational environments where Java integration matters

  On the Module Split:
  
  Given that Dalvik was replaced by ART in Android 5.0 (2014), perhaps a simpler approach would be a single module with runtime feature detection rather than rhino-jvm/rhino-dalvik?

  On Minimum API Level:
  With Google Play requiring API 35 for new apps (as of August 31, 2025), setting minSdkVersion to 24 (Android 7.0) might be pragmatic - this would resolve the VarHandle issues while covering the vast majority of devices.

  Your testing infrastructure gives us exactly what we need to make these decisions based on real data. The ability to run tests across multiple API levels is invaluable.

  What do you think about focusing the Android strategy on Rhino's unique advantages (pure Java, no JNI, excellent Java interop) rather than trying to match the performance of V8 or QuickJS?

install-android-sdk.sh
```diff
@@ -0,0 +1,15 @@
+#/usr/bin/env bash
```
gbrail, line 1: Can you please move this file to the "it-android" directory so that the root directory doesn't get too full?

Could you also make it executable (might need to mess around with Git -- usually if you "chmod a+x" the file Git will figure it out when you first add it).

run-android-tests-locally.sh
```diff
@@ -0,0 +1,40 @@
+#/usr/bin/env bash
```
gbrail, line 1: Could you also move this file to it-android and make it executable?

gbrail (commented):
Thanks! I get where this is going and it is going to help a lot.

So I can understand -- if I run the "install-android-sdk" script, and then run the tests, will it use the SDK to run the tests on Android, or do I need to do the whole docker / podman setup for that?

I'm asking because installing the SDK works great, and the tests work fine in either way, but getting that docker / podman script to work portably across OSse and other things is going to take some work -- I had to muck around to make it work with Windows and MSYS but there are probably five other combinations that people use. Do we get some value out of this if we don't need people to run the container, but if the CI job does?

tests/test262
rPraml, line 1: This has to be reverted!

gbrail:
I'm going to close this PR as I've included all the commits in https://github.com/mozilla/rhino/pull/2146 and I would like to merge that soon. Thanks!

PR #2037, opened 2025-08-23
Improve android handling

rPraml:
This PR tries to auto-detect the API version.
- this is used now on all places, where formerly `java.vm.name == "Dalvik"` was used
- It does not load the Codegen class on android, so we always run in interpreter mode her

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/ScriptRuntime.java
+++ b/rhino/src/main/java/org/mozilla/javascript/ScriptRuntime.java
@@ -6152,8 +6152,4 @@ public class ScriptRuntime {
 
     /** This value holds the current android API version (or -1) if not running on android */
     static final int androidApi = detectAndroidApi();
-
-    static {
-        System.out.println("Android API: " + androidApi);
-    }
 }


rPraml:
build failed flaky. What's the best way to re-run a build (without force-pushing a new commit)?

anivar:
Thanks for working on this Android compatibility issue\! I've created an improved version that works with both Dalvik and ART runtimes: https://github.com/FOCONIS/rhino/pull/5

The key improvement is using `android.os.Build` class detection instead of VM name checking. Quick test confirms it works:
- Desktop: No android.os.Build → compiled mode (level 9) ✓  
- Android: Detects android.os.Build → interpreted mode (level -1)

This approach is more future-proof since ART became default in Android 5.0 (2014).

rPraml:
Thank you @anivar for the suggestion. I took it a step further and implemented reading the API version directly. I also refactored all the places that were previously checking for `java.vm.name` accordingly.

gbrail:
While we continue to work on getting the right test infrastructure for Android, this seems like a good step forward anyway.

Issue #2035, opened 2025-08-23
Regression III: arraycopy: length -1 is negative in Interpreter$CallFrame.initializeArgs and wrong result


rbri:
Found some more problems of the major interpreter change.

You can add the two tests to BoundFunctionTest

    @Test
    public void invokeBoundCallWithoutArgs() {
        String code =
                "function f(x) { return this.toString() + ' / ' + x; };\n"
                        + "var boundThis = 'abcd';\n"
                        + "var boundF = f.bind(boundThis);\n"
                        + "boundF.call('Hello!');\n";

        Utils.assertWithAllModes_ES6("abcd / undefined", code);
    }

This produces an error

    arraycopy: length -1 is negative
    java.lang.ArrayIndexOutOfBoundsException: arraycopy: length -1 is negative
	at java.base/java.lang.System.arraycopy(Native Method)
	at org.mozilla.javascript.Interpreter$CallFrame.initializeArgs(Interpreter.java:320)
	at org.mozilla.javascript.Interpreter.initFrame(Interpreter.java:4638)
	at org.mozilla.javascript.Interpreter$DoCallByteCode.execute(Interpreter.java:3427)
	at org.mozilla.javascript.Interpreter.interpretFunction(Interpreter.java:1826)
	at org.mozilla.javascript.Interpreter.interpretLoop(Interpreter.java:1602)
	at org.mozilla.javascript.Interpreter.interpret(Interpreter.java:1225)
	at org.mozilla.javascript.InterpretedFunction.call(InterpretedFunction.java:87)

And this one simply outputs a wrong result 'abcd / undefined' instead of the expected one.

    @Test
    public void invokeBoundCallOneArg() {
        String code =
                "function f(x) { return this.toString() + ' / ' + x; };\n"
                        + "var boundThis = 'abcd';\n"
                        + "var boundF = f.bind(boundThis, 'world');\n"
                        + "boundF.call('Hello!');\n";

        Utils.assertWithAllModes_ES6("abcd / world", code);
    }

I fear there is more because i got a different exception in HtmlUnit but i have to do more analysis work.

@aardvark179 hope you have some time to patch this....

rbri:
@aardvark179 hope the test from the pr https://github.com/mozilla/rhino/pull/2038 of some help

rbri:
@aardvark179 any hope for this?

rbri:
[...]

PR #2038, opened 2025-08-24
Several fixes for calling bind() in the interpreter

rbri:
As a first step, i wrote a (hopefully) complete set bind() function calling tests. This was done to highlight the regressions we still have (e.g. https://github.com/mozilla/rhino/issues/2035) from the interpreter refactoring without seeing some unit tests to fail.

There are also some complex test in the HtmlUnit test suite that are still failing.

Next step was to use the same test approach for the apply() and call() functions.

And finally i have fixed all the failing tests.


rbri:
marked this as draft because many tests are failing as of now

rbri:
The ApplyFunctionTest and CallFunctionTest are showing no new errors.
If we had them before the refactoring we where able to detect the two already fixed regressions during the refactoring.

rbri:
@aardvark179 please review
@gbrail if we are happy with this please squash

gbrail:
Yes, this looks good to me -- thanks for the detailed tests!

rbri:
wow that was fast... thanks @gbrail

PR #2039, opened 2025-08-24
Made spotless as separate convention

rPraml:
Moved the spotless configuration to a separte convention file, so that it can be used in non-java based subprojects like android (in PR #2036)


gbrail:
Makes sense to me, thanks!

PR #2041, opened 2025-08-25
Optimize undefined lookup

andreabergia:
This creates a new token to treat undefined as a special case since it comes up so frequently, as currently it is treated as a `NAME` lookup. In the vast majority of the time we end up using these optimizations, but there are a few cases where we just keep the previous behavior of treating undefined as a normal name: deleting and assigning. These cases are likely to be very infrequent so there would likely not be any noticeable performance improvements, as well as a lot of similar code to what already exists.

With the introduction of this new behavior, the main difference was to recreate looking up if undefined was redefined somewhere in this scope stack, which is the reason for the extra tests.

Most of the existing benchmarks don't even mention `undefined` explicitly; for those that do, it's an unfrequent enough operation that I could not measure any real improvement. However, this _does_ make things faster in some microbenchmarks that we've written, but not included in this PR because they aren't worth keeping around IMHO.

---

Note: This is a patch done almost entirely by @camnwalter while he was doing an internship at ServiceNow. I'm opening a PR now with his permission, instead of letting him do it on its own when he will have time for it, because (spoiler alert) I need it for [my classes work](https://github.com/andreabergia/rhino/blob/classes-03/tests/src/test/java/org/mozilla/javascript/tests/ClassesTest.java)... 🙂

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/CodeGenerator.java
+++ b/rhino/src/main/java/org/mozilla/javascript/CodeGenerator.java
@@ -725,7 +725,7 @@ class CodeGenerator extends Icode {
                     // Put undefined
                     resolveForwardGoto(putUndefinedLabel);
                     addIcode(Icode_POP);
-                    addStringOp(Token.NAME, "undefined");
+                    addToken(Token.UNDEFINED);
                     resolveForwardGoto(afterLabel);
                 } else if (node.getIntProp(Node.SUPER_PROPERTY_ACCESS, 0) == 1) {
                     addStringOp(
@@ -770,7 +770,7 @@ class CodeGenerator extends Icode {
                     // Put undefined
                     resolveForwardGoto(putUndefinedLabel);
                     addIcode(Icode_POP);
-                    addStringOp(Token.NAME, "undefined");
+                    addToken(Token.UNDEFINED);
                     resolveForwardGoto(afterLabel);
                 } else if (node.getIntProp(Node.SUPER_PROPERTY_ACCESS, 0) == 1) {
                     visitExpression(child, 0);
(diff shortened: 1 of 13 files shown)


rbri:
> I need it for my classes work...

and your spread work? Can't wait for both ;-) !!!!

gbrail:
Can you help me understand the motivation -- is this an important optimization for you all in interpreted mode?

I'm asking because if our goal was to fix this for compiled mode, it would be a matter of adding a new linker to the INDY machinery in the "optimizer" package to look for a NAME lookup for "undefined" and replace it with a constant. (And in fact, I just tried it and it just returned a constant anyway.)

So, I'm skeptical that for compiled mode code that this would speed anything up. But if it's important to fix this for interpreted mode then that's a different story, although TBH it makes me wonder whether we can figure out a way to make the same INDY optimizations we have in compiled mode work in interpreted mode as well.

andreabergia:
> Can you help me understand the motivation -- is this an important optimization for you all in interpreted mode?
> 
> I'm asking because if our goal was to fix this for compiled mode, it would be a matter of adding a new linker to the INDY machinery in the "optimizer" package to look for a NAME lookup for "undefined" and replace it with a constant. (And in fact, I just tried it and it just returned a constant anyway.)
> 
> So, I'm skeptical that for compiled mode code that this would speed anything up. But if it's important to fix this for interpreted mode then that's a different story, although TBH it makes me wonder whether we can figure out a way to make the same INDY optimizations we have in compiled mode work in interpreted mode as well.

It's a bit more important for us at ServiceNow than in general usage I think, because the lookup would happen in the top level scope, which for us is a special object with a lot of rules and access control for security. So, avoiding the lookup is worthwhile.
`BodyCodegen` was changed in this PR as well, to handle `Token.UNDEFINED` in the same was as the interpreter. At runtime, it generates a `GETSTATIC` of `Undefined.instance`.

gbrail:
OK -- I get it, personally before I had been focusing more on performance in compiled mode. I'll try it out, thanks for the work so far!

gbrail:
Thanks -- this looks good to me now!

Issue #1297, opened 2023-01-29
Support ES2015 function name behavior


p-bakker:
ES2015 standardized the behavior of the function.name property. Rhino doesn't comply. 

See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/name, https://2ality.com/2015/09/function-names-es6.html for explanations of expected behavior. See https://tc39.es/ecma262/#sec-setfunctionname and everywhere in the spec where the abstract function operation setFunctionName is used to see the how and when

Debugger implementations would benefit from this, as they would be able to display more meaningful names on each entry in the callstack

Known issues:
- var x = function(){}; x.name // doesn't yield 'x'
- let f; f = () => {}; f.name // doesn't yield 'f'
- ({x: function test(){}}).x.name // doesn't yield 'x'
- ({foo(){}}).foo.name // doesn't yield 'foo'
- var o = {get foo(){}, set foo(x){}}; var desc = Object.getOwnPropertyDescriptor(o, "foo"); desc.get.name // doesn't yield 'get foo'
- var o = {get foo(){}, set foo(x){}}; var desc = Object.getOwnPropertyDescriptor(o, "foo"); desc.set.name // doesn't yield 'set foo'

Also see #715 for a related case

Some of the scenarios that we currently cannot cover, as they rely on features Rhino doesn't currently support:
-  `let s = Symbol('someDescription'); ({[s]: () => {})[s].name` // to yield `[someDescription]` #913
- default exports, see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/name#default-exported_function_declaration #945
-  default values in initializers: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/name#initializer_and_default_value #756, #678
- method declarations: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/name#shorthand_method #1299
- Class properties and private fields #835

rbri:
will provide a patch for bound functions - because this is simple to fix :-)

PR #1949, opened 2025-06-20
Infer function names

andreabergia:
This PR solves most of the problems highlighted in https://github.com/mozilla/rhino/issues/1297 about function name inference, i.e. all of these now work:

```js
var f = function() {}; f.name  // f
var f; f = () => {}; f.name  // f
o = { x: function(){} }; o.x.name  // x
o = { x() {} }; o.x.name  // x
var o = { get x(){} }; var desc = Object.getOwnPropertyDescriptor(o, "x"); desc.get.name  // get x
var o = { set x(v){} }; var desc = Object.getOwnPropertyDescriptor(o, "x"); desc.set.name  // set x
```

Note that this PR implements name inference only _statically_, i.e. when the name can be known at parse times. The ES6 spec says that name inference should happen also for computed properties, at runtime, but this PR does not implement that:

```js
function f() { return 'f'; }
o = { [f()]() {} }
o.f.name // spec says should be 'f' but Rhino says it's empty string
```

Please squash if accepted :)

PR #2043, opened 2025-08-26
Infer function names at runtime for computed properties

andreabergia:
This continues the work of https://github.com/mozilla/rhino/pull/1949 and applies inference of function names at runtime for computed properties. There's some special handling necessary if the key is a Symbol, or if the key is `__proto__`, but I think I have added a test case for every combination.

Makes a few more test262 cases pass. Should close https://github.com/mozilla/rhino/issues/1297.

I've also changed the recently-implemented `NewLiteralStorage` to ensure that we apply only the behavior in version >= ES6.

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/BaseFunction.java
+++ b/rhino/src/main/java/org/mozilla/javascript/BaseFunction.java
@@ -226,6 +226,11 @@ public class BaseFunction extends ScriptableObject implements Function {
         return true;
     }
 
+    /** Forces setting the function's name, bypassing all "readonly" checks. */
+    void setFunctionName(String name) {
+        nameSetter(this, name, this, this, false);
+    }
+
     protected void createPrototypeProperty() {
         if (!has(PROTOTYPE_PROPERTY_NAME, this)) {
             ScriptableObject.defineBuiltInProperty(
--- a/rhino/src/main/java/org/mozilla/javascript/BaseFunction.java
+++ b/rhino/src/main/java/org/mozilla/javascript/BaseFunction.java
@@ -228,7 +228,7 @@ public class BaseFunction extends ScriptableObject implements Function {
 
     /** Forces setting the function's name, bypassing all "readonly" checks. */
     void setFunctionName(String name) {
-        nameSetter(this, name, this, this, false);
+        nameValue = name;
     }
 
     protected void createPrototypeProperty() {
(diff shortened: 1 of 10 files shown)


rhino/src/main/java/org/mozilla/javascript/BaseFunction.java
```diff
@@ -226,6 +226,11 @@ private static boolean nameSetter(
         return true;
     }
 
+    /** Forces setting the function's name, bypassing all "readonly" checks. */
+    void setFunctionName(String name) {
+        nameSetter(this, name, this, this, false);
```
aardvark179: That could just do `nameValue = name`.

aardvark179 (commented):
Looks good in general, only nit picks.

rhino/src/main/java/org/mozilla/javascript/BaseFunction.java
andreabergia: You're right. Changed it.

rhino/src/main/java/org/mozilla/javascript/NewLiteralStorage.java
```diff
@@ -92,4 +101,87 @@ public int[] getGetterSetters() {
...
+            return new NameInference(ids, -1, false);
+        } else {
+            return new NoInference(ids, -1, false);
+        }
+    }
+
+    public static NewLiteralStorage create(Context cx, int length, boolean createKeys) {
+        if (cx.getLanguageVersion() >= Context.VERSION_ES6) {
+            return new NameInference(null, length, createKeys);
+        } else {
```
rbri, line 116: at other places we decide to have the second return not in a else clause - more or less personal tast

rhino/src/main/java/org/mozilla/javascript/SymbolKey.java
```diff
@@ -37,6 +37,7 @@ public SymbolKey(String name) {
      * Returns the symbol's name. Returns empty string for anonymous symbol (i.e. something created
```
rbri: maybe we should place the jdoc at the interface and use inheritedDoc for the overridden method

rbri (approved):

rhino/src/main/java/org/mozilla/javascript/NewLiteralStorage.java
andreabergia: I personally prefer the if/else frankly, but I'm happy to change it if anyone feels strongly about it.

rhino/src/main/java/org/mozilla/javascript/SymbolKey.java
andreabergia: Good point; fixed it.

gbrail:
Thanks, this looks good!

PR #2042, opened 2025-08-26
[Experimental] Testing older android APIs

rPraml:
This is still in progress

PR #2045, opened 2025-08-27
document android libs issues in o.m.j package

rPraml:
This PR adds documentation regarding the usage of special Android libraries.

Note: The AndroidJdkLibsChecker currently checks against API level 1 only.  
Therefore, if you use certain APIs that were introduced in later Android versions, the AndroidJdkLibsChecker may flag them as incompatible – even though they might function correctly on newer Android versions.

Current status:

- SlotMap.ThreadAccess utilizes a VM-specific API (compareAndExchange), which ties us to a minimum SDK version of 26. There is currently no straightforward workaround for this.
- Other missing APIs (such as java.time.format) can be "emulated" via Android's desugaring process, enabling compatibility with a minimum SDK version of 21. (See #2042)

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/TokenStream.java
+++ b/rhino/src/main/java/org/mozilla/javascript/TokenStream.java
@@ -577,11 +577,11 @@ class TokenStream implements Parser.CurrentPositionReporter {
         return id & 0xff;
     }
 
-    @SuppressWarnings("AndroidJdkLibsChecker")
     private static boolean isValidIdentifierName(String str) {
         int i = 0;
-        for (int c : str.codePoints().toArray()) {
-            if (i++ == 0) {
+        while (i < str.length()) {
+            int c = str.codePointAt(i);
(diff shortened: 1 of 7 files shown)


rhino/src/main/java/org/mozilla/javascript/TokenStream.java
```diff
@@ -577,11 +577,11 @@ private static int stringToKeywordForES(String name, boolean isStrict) {
         return id & 0xff;
     }
 
-    @SuppressWarnings("AndroidJdkLibsChecker")
     private static boolean isValidIdentifierName(String str) {
```
rPraml, line 580: This mehtod has used https://developer.android.com/reference/java/lang/String#codePoints(), wich would be available since API 24.

I refactored this, so that we not use the String->IntStream->toArray pipeline, as I think this version might be more performant

gbrail:
Thanks, this looks good to me too!

PR #2046, opened 2025-08-27
Do not check interpreted mode in Codegen and always optimize

rPraml:
While debugging, I discovered some quirks in Codegen.

Possible it was overlooked in https://github.com/mozilla/rhino/commit/8f8c3dc6807bfe273fcacaa3b94dbc0d47b02226

@gbrail, could you check if this makes sense? Are there any code paths where Codegen is invoked in interpreted mode?

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/optimizer/Codegen.java
+++ b/rhino/src/main/java/org/mozilla/javascript/optimizer/Codegen.java
@@ -177,27 +177,26 @@ public class Codegen implements Evaluator {
     private void transform(ScriptNode tree) {
         initOptFunctions_r(tree);
 
-        boolean optimizing = !compilerEnv.isInterpretedMode();
-
+        if (!compilerEnv.isInterpretedMode()) {
+            Kit.codeBug("Codegen nust not run in interpreted Mode");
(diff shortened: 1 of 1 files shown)


rhino/src/main/java/org/mozilla/javascript/optimizer/Codegen.java
```diff
@@ -177,27 +177,26 @@ public byte[] compileToClassFile(
     private void transform(ScriptNode tree) {
         initOptFunctions_r(tree);
 
-        boolean optimizing = !compilerEnv.isInterpretedMode();
-
+        if (compilerEnv.isInterpretedMode()) {
+            Kit.codeBug("Codegen nust not run in interpreted Mode");
```
andreabergia: Typo:

```suggestion
            Kit.codeBug("Codegen must not run in interpreted mode");
```

rbri (approved):

gbrail:
Yeah, I get what's happening here -- we used to do a set of optional optimizations at higher optimization levels, and when we switched to get rid of them someone (probably me) didn't account for the new settings. Thanks!

PR #2047, opened 2025-08-27
Fix stack generation in interpreted mode

### LINKED ISSUE #2044: Error stack lost
```
try {
    null.toString();
} cath (e) {
    console(e + e.stack);
}
```
TypeError: Cannot call method “toString” of null	at <Unknown source>:4

rPraml:
This fixes #2044

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/ScriptRuntime.java
+++ b/rhino/src/main/java/org/mozilla/javascript/ScriptRuntime.java
@@ -5082,6 +5082,9 @@ public class ScriptRuntime {
             // set exception in Error objects to enable non-ECMA "stack" property
             if (errorObject instanceof NativeError) {
                 ((NativeError) errorObject).setStackProvider(re);
+                if (cx.isInterpretedMode()) {
+                    ((NativeError) errorObject).getStackDelegated();
+                }
             }
 
             if (javaException != null && isVisible(cx, javaException)) {
--- a/rhino/src/main/java/org/mozilla/javascript/ScriptRuntime.java
+++ b/rhino/src/main/java/org/mozilla/javascript/ScriptRuntime.java
@@ -5082,9 +5082,6 @@ public class ScriptRuntime {
             // set exception in Error objects to enable non-ECMA "stack" property
             if (errorObject instanceof NativeError) {
                 ((NativeError) errorObject).setStackProvider(re);
-                if (cx.isInterpretedMode()) {
-                    ((NativeError) errorObject).getStackDelegated();
-                }
             }
 
             if (javaException != null && isVisible(cx, javaException)) {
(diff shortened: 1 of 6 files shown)


testutils/src/main/java/org/mozilla/javascript/testutils/Utils.java
```diff
@@ -240,7 +240,7 @@ public static void assertWithAllModes(
                         cx.setLanguageVersion(languageVersion);
                     }
                     final Scriptable scope = cx.initStandardObjects();
-                    final Object res = cx.evaluateString(scope, script, "test.js", 0, null);
+                    final Object res = cx.evaluateString(scope, script, "test.js", 1, null);
```
rPraml, line 243: @rbri would you agree to change this value to `1`

rhino/src/test/java/org/mozilla/javascript/tests/ErrorHandlingTest.java
```diff
@@ -37,23 +38,52 @@ public static void generateJavaError() {
     @Test
     public void throwError() {
         testIt(
-                "throw new Error('foo')",
+                "\n" // 1
+                        + "throw new Error('foo')", // 2
```
rPraml, line 42: added some linebreaks here (comments are for Spotless)

rhino/src/main/java/org/mozilla/javascript/ScriptRuntime.java
```diff
@@ -5082,6 +5082,9 @@ public static Scriptable newCatchScope(
             // set exception in Error objects to enable non-ECMA "stack" property
             if (errorObject instanceof NativeError) {
                 ((NativeError) errorObject).setStackProvider(re);
+                if (cx.isInterpretedMode()) {
+                    ((NativeError) errorObject).getStackDelegated();
```
rPraml: In interpreted mode, we must "capture" the stack on creation. (although it is not needed)

The computation is done [here](https://github.com/mozilla/rhino/pull/1956/files#diff-064c6e0ae69de426d768d3c598d6975a85b9acf9defa0a605d6a21612d387e8bR1154)

@aardvark179 - maybe you can take a look at this, if this solution is OK

rPraml:
I think, this was really a regression of #1956

Formerly, the line info was stored in the exception
https://github.com/mozilla/rhino/pull/1956/files#diff-064c6e0ae69de426d768d3c598d6975a85b9acf9defa0a605d6a21612d387e8bL1009

I have re-used the `int[]` but changed it `int` which is technically an API change, if someone has inherited from RhinoException

testutils/src/main/java/org/mozilla/javascript/testutils/Utils.java
rbri: Yes fine for me

rbri:
@gbrail / @andreabergia i like to merge this - any objections?

andreabergia:
> @gbrail / @andreabergia i like to merge this - any objections?

Fine with me. I agree with Duncan that the minor change to the API is ok for me: it's not public stuff anyway.

rbri:
Thanks a lot

PR #2058, opened 2025-08-31
Implement ES2021 WeakRef and FinalizationRegistry

anivar:
Implements ES2021 WeakRef and FinalizationRegistry following the NativeWeakMap pattern.

## Implementation

Two self-contained classes, about 400 lines total:
- NativeWeakRef.java - uses Java's WeakReference directly
- NativeFinalizationRegistry.java - uses PhantomReference + ReferenceQueue for GC detection

The finalization approach is simple: poll the ReferenceQueue during register() calls to process any pending cleanups. No separate queue management or Context modifications needed.

## Android Compatibility

Uses PhantomReference instead of the Cleaner API since Cleaner isn't available on Android. This keeps the implementation portable across all Java 8+ platforms.

## ES2021 Standard Compliance

Correctly implements symbol handling per spec:
- Allows unregistered symbols as weak targets
- Rejects registered symbols created with Symbol.for since they persist globally
- Validates all target/heldValue/token constraints
- Proper Symbol.toStringTag on prototypes

## Test Results

All 76 test262 tests passing:
- 29 WeakRef tests
- 47 FinalizationRegistry tests

Also fixes 2 Object.seal tests.

Thread-safe implementation using ConcurrentHashMap for registration tracking.

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/NativeFinalizationRegistry.java
+++ b/rhino/src/main/java/org/mozilla/javascript/NativeFinalizationRegistry.java
@@ -19,8 +19,7 @@ import java.util.concurrent.ConcurrentHashMap;
  * collected. Uses {@link PhantomReference} for GC detection, which is available on all platforms
  * including Android.
  *
- * @see <a
- *     href="https://tc39.es/ecma262/#sec-finalization-registry-objects">ECMAScript
+ * @see <a href="https://tc39.es/ecma262/#sec-finalization-registry-objects">ECMAScript
  *     FinalizationRegistry Objects</a>
  */
 public class NativeFinalizationRegistry extends ScriptableObject {
@@ -33,8 +32,7 @@ public class NativeFinalizationRegistry extends ScriptableObject {
     private final Function cleanupCallback;
     private final Scriptable parentScope;
     private final Set<RegistrationReference> activeRegistrations = ConcurrentHashMap.newKeySet();
-    private final Map<TokenKey, Set<RegistrationReference>> tokenIndex =
-            new ConcurrentHashMap<>();
+    private final Map<TokenKey, Set<RegistrationReference>> tokenIndex = new ConcurrentHashMap<>();
 
     /** Initialize FinalizationRegistry constructor and prototype. */
     static Object init(Context cx, Scriptable scope, boolean sealed) {
@@ -208,18 +206,16 @@ public class NativeFinalizationRegistry extends ScriptableObject {
             cleanupCallback.call(cx, parentScope, this, new Object[] {heldValue});
         } catch (RhinoException e) {
             // Per spec, errors in cleanup callbacks don't propagate
-            Context.reportWarning(
-                    "FinalizationRegistry cleanup callback threw: " + e.getMessage());
+            Context.reportWarning("FinalizationRegistry cleanup callback threw: " + e.getMessage());
         }
     }
 
     /**
      * Check if the given object can be used as a FinalizationRegistry target.
      *
-     * <p>Per ECMAScript spec, FinalizationRegistry targets can be:
-     * - Any object (Scriptable)
-     * - Unregistered symbols (created with Symbol(), not Symbol.for())
-     * - Registered symbols (Symbol.for()) cannot be registered as they persist globally
+     * <p>Per ECMAScript spec, FinalizationRegistry targets can be: - Any object (Scriptable) -
+     * Unregistered symbols (created with Symbol(), not Symbol.for()) - Registered symbols
+     * (Symbol.for()) cannot be registered as they persist globally
      *
      * @param target the target object to validate
      * @return true if target is a valid object or unregistered symbol that can be registered
@@ -239,8 +235,8 @@ public class NativeFinalizationRegistry extends ScriptableObject {
     /**
      * Check if the given value can be used as an unregister token.
      *
-     * <p>Per ECMAScript spec, registered symbols (created with Symbol.for()) cannot be held
-     * weakly because they persist in the global registry.
+     * <p>Per ECMAScript spec, registered symbols (created with Symbol.for()) cannot be held weakly
+     * because they persist in the global registry.
      *
      * @param token the token to validate
      * @return true if token can be used as an unregister token
--- a/rhino/src/main/java/org/mozilla/javascript/NativeWeakRef.java
+++ b/rhino/src/main/java/org/mozilla/javascript/NativeWeakRef.java
@@ -51,10 +51,7 @@ public class NativeWeakRef extends ScriptableObject {
     private static Scriptable jsConstructor(Context cx, Scriptable scope, Object[] args) {
         if (args.length < 1) {
             throw ScriptRuntime.typeErrorById(
-                    "msg.method.missing.parameter",
-                    CLASS_NAME,
-                    "1",
-                    String.valueOf(args.length));
+                    "msg.method.missing.parameter", CLASS_NAME, "1", String.valueOf(args.length));
         }
 
         Object target = args[0];
@@ -87,10 +84,9 @@ public class NativeWeakRef extends ScriptableObject {
     /**
      * Check if the given object can be used as a WeakRef target.
      *
-     * <p>Per ECMAScript spec, WeakRef targets can be:
-     * - Any object (Scriptable)
-     * - Unregistered symbols (created with Symbol(), not Symbol.for())
-     * - Registered symbols (Symbol.for()) cannot be held weakly as they persist globally
+     * <p>Per ECMAScript spec, WeakRef targets can be: - Any object (Scriptable) - Unregistered
+     * symbols (created with Symbol(), not Symbol.for()) - Registered symbols (Symbol.for()) cannot
+     * be held weakly as they persist globally
      *
      * @param target the target object to validate
      * @return true if target is a valid object or unregistered symbol that can be weakly referenced
--- a/rhino/src/main/java/org/mozilla/javascript/ScriptRuntime.java
+++ b/rhino/src/main/java/org/mozilla/javascript/ScriptRuntime.java
@@ -277,7 +277,8 @@ public class ScriptRuntime {
             new LazilyLoadedCtor(scope, "WeakMap", sealed, true, NativeWeakMap::init);
             new LazilyLoadedCtor(scope, "WeakSet", sealed, true, NativeWeakSet::init);
             new LazilyLoadedCtor(scope, "WeakRef", sealed, true, NativeWeakRef::init);
-            new LazilyLoadedCtor(scope, "FinalizationRegistry", sealed, true, NativeFinalizationRegistry::init);
+            new LazilyLoadedCtor(
+                    scope, "FinalizationRegistry", sealed, true, NativeFinalizationRegistry::init);
             new LazilyLoadedCtor(scope, "BigInt", sealed, true, NativeBigInt::init);
             new LazilyLoadedCtor(scope, "Proxy", sealed, true, NativeProxy::init);
             new LazilyLoadedCtor(scope, "Reflect", sealed, true, NativeReflect::init);


rhino/src/main/java/org/mozilla/javascript/NativeFinalizationRegistry.java
```diff
@@ -0,0 +1,462 @@
...
+    private static final String REGISTER_METHOD = "register";
+    private static final String UNREGISTER_METHOD = "unregister";
+    private static final String CLEANUP_SOME_METHOD = "cleanupSome";
+
+    // Constants
+    private static final String MSG_NO_CALLBACK = "msg.finalization.registry.no.callback";
+    private static final String MSG_REGISTER_NOT_OBJECT =
+            "msg.finalization.registry.register.not.object";
+    private static final String MSG_SAME_TARGET = "msg.finalization.registry.register.same.target";
+    private static final int CONSTRUCTOR_ARITY = 1;
```
gbrail: I don't know if I normally care too much, but in general in this project we have NOT been using constants for every arity and method name parameter. I know that's not what they taught me in school, but for this project having all that information together in the one line of code where we use it makes more sense to me than all the constant back and forth.

rhino/src/main/java/org/mozilla/javascript/NativeFinalizationRegistry.java
```diff
@@ -0,0 +1,462 @@
...
+                        NativeFinalizationRegistry::constructor);
+        constructor.setPrototypePropertyAttributes(DONTENUM | READONLY | PERMANENT);
+        return constructor;
+    }
+
+    private static void configurePrototype(LambdaConstructor constructor, Scriptable scope) {
+        // Define prototype methods
+        constructor.definePrototypeMethod(
+                scope,
+                REGISTER_METHOD,
```
gbrail: See above -- I don't think that we need constants for method names and arities.

rhino/src/main/java/org/mozilla/javascript/NativeFinalizationRegistry.java
```diff
@@ -0,0 +1,462 @@
...
+        if (args.length < 2) {
+            throw ScriptRuntime.typeErrorById(MSG_REGISTER_NOT_OBJECT);
+        }
+
+        Object target = args[0];
+        Object heldValue = args[1];
+        Object unregisterToken = args.length > 2 ? args[2] : null;
+
+        if (!isValidTarget(target)) {
+            throw ScriptRuntime.typeErrorById(MSG_REGISTER_NOT_OBJECT);
```
gbrail: I don't think that it adds to clarity to have separate constants for message Ids either, and the rest of the code in the project just puts the message name in there, which saves an indirection when trying to find it in my opinion!

rhino/src/main/java/org/mozilla/javascript/NativeFinalizationRegistry.java
```diff
@@ -0,0 +1,462 @@
...
+     * @param heldValue the held value to pass to the callback
+     */
+    private void executeCleanupCallback(Object heldValue) {
+        if (cleanupCallback == null) {
+            return;
+        }
+
+        Context cx = Context.getCurrentContext();
+        if (cx == null) {
+            cx = Context.enter();
```
gbrail: We have a long thread going on about this elsewhere, but at least for now you can and should use try-with-resources here rather than exiting the context in a finally clause.

rhino/src/main/java/org/mozilla/javascript/NativeFinalizationRegistry.java
```diff
@@ -0,0 +1,462 @@
...
+        try {
+            Scriptable scope = callback.getParentScope();
+            if (scope == null) {
+                scope = cx.initStandardObjects();
+            }
+            callback.call(cx, scope, scope, new Object[] {heldValue});
+        } catch (RhinoException e) {
+            // Cleanup callback errors should not break the cleanup process
+            // In a production environment, this might be logged
+        } catch (Exception e) {
```
gbrail: Is there a reason why we'd legitimately want to catch a non-JavaScript exception here? Otherwise we should just catch RhinoException, assuming that the spec says that we still finalize even if the callback throws.

rhino/src/main/java/org/mozilla/javascript/NativeFinalizationRegistry.java
```diff
@@ -0,0 +1,462 @@
...
+    private void callCleanupCallback(Context cx, Object heldValue, Function callback) {
+        try {
+            Scriptable scope = callback.getParentScope();
+            if (scope == null) {
+                scope = cx.initStandardObjects();
+            }
+            callback.call(cx, scope, scope, new Object[] {heldValue});
+        } catch (RhinoException e) {
+            // Cleanup callback errors should not break the cleanup process
+            // In a production environment, this might be logged
```
gbrail: Not to be annoying, but this code will absolutely be used in a production environment ;-)

rhino/src/main/java/org/mozilla/javascript/NativeFinalizationRegistry.java
```diff
@@ -0,0 +1,462 @@
...
+     * Calls the specified cleanup callback with the held value.
+     *
+     * @param cx the context to use
+     * @param heldValue the held value to pass to the callback
+     * @param callback the callback function to call
+     */
+    private void callCleanupCallback(Context cx, Object heldValue, Function callback) {
+        try {
+            Scriptable scope = callback.getParentScope();
+            if (scope == null) {
```
gbrail: Is there a reason why the scope would be null here?

tests/testsrc/test262.properties
```diff
@@ -948,7 +948,21 @@ built-ins/Error 7/53 (13.21%)
     cause_abrupt.js
     proto-from-ctor-realm.js
 
-~built-ins/FinalizationRegistry
```
gbrail, line 943: This is very nice progress -- do you know why the remaining tests are failing?

rhino/src/main/java/org/mozilla/javascript/NativeFinalizationRegistry.java
```diff
@@ -0,0 +1,462 @@
...
+            processCleanupWithCallback(ref, callbackToUse, cx);
+        }
+    }
+
+    /**
+     * Processes cleanup for a specific weak reference.
+     *
+     * @param ref the weak reference to process
+     */
+    private void processCleanup(FinalizationWeakReference ref) {
```
gbrail: Looking at the code coverage, it looks like we don't have test coverage for this and the next five functions (processCleanupWithCallback, executeCleanupCallback, callCleanupCallback). Not sure how you can test those, but even some "white box tests" might help just to ensure that we have tests for it.

gbrail (commented):
Thanks -- this looks like good progress. I have a bunch of questions, please take a look -- thanks!

rhino/src/main/java/org/mozilla/javascript/NativeFinalizationRegistry.java
anivar: You're absolutely right about the project style. Looking at similar classes like NativeWeakMap, I can see that method names and arities are typically inlined directly in
  the definePrototypeMethod calls rather than using constants. I'll remove these constants:
  - REGISTER_METHOD → inline "register"
  - UNREGISTER_METHOD → inline "unregister"
  - CONSTRUCTOR_ARITY → inline 1
  - REGISTER_ARITY → inline 2
  - UNREGISTER_ARITY → inline 1

  This will make the code more consistent with the rest of the codebase and reduce the indirection as you mentioned.

rhino/src/main/java/org/mozilla/javascript/NativeFinalizationRegistry.java
anivar: I'll replace these constants with inline values in the definePrototypeMethod calls

rhino/src/main/java/org/mozilla/javascript/NativeFinalizationRegistry.java
anivar: I will follow  the project's pattern of putting the message ID directly in the error call, which as you noted, makes it easier to find the message when debugging.

rhino/src/main/java/org/mozilla/javascript/NativeFinalizationRegistry.java
anivar: Will do 
```

try (Context enteredCx = Context.enter()) {
      callCleanupCallback(enteredCx, heldValue);
  }
```

rhino/src/main/java/org/mozilla/javascript/NativeFinalizationRegistry.java
anivar: you're right that we should only catch RhinoException here

rhino/src/main/java/org/mozilla/javascript/NativeFinalizationRegistry.java
anivar: Will try  

```
 } catch (RhinoException e) {
 
      Context.reportWarning("FinalizationRegistry cleanup callback threw: " + e.getMessage());
  }
```

rhino/src/main/java/org/mozilla/javascript/NativeFinalizationRegistry.java
anivar: A JavaScript function passed to new FinalizationRegistry(callback) will always have a parent scope  which was defined somewhere.

  If somehow it's null, that's a Rhino bug we shouldn't hide. Creating a new scope with initStandardObjects() would just mask the problem and cause weird behavior.

  I'll remove the null check and trust the scope exists

tests/testsrc/test262.properties
anivar: The remaining failures fall into a few categories:

  1. cleanupSome (14 tests) - This optional method isn't implemented yet. It allows synchronous triggering of cleanup callbacks, which is complex to implement safely.
  2. Cross-realm tests - proto-from-ctor-realm.js fails because of how Rhino handles cross-realm prototype chains differently from the spec.
  3. Edge cases in register/unregister - Some tests like custom-this.js and holdings-any-value-type.js are failing due to subtle differences in type checking or parameter handling.
  4. Weak reference constraints - Tests like throws-when-target-cannot-be-held-weakly.js will be addressed when the WeakRef PR lands, as that will establish the proper constraints for what can be held weakly.

  The core functionality works - objects can be registered, cleanup callbacks fire when GC happens, and unregister works. The failures are mostly edge cases and the optional cleanupSome feature.

rhino/src/main/java/org/mozilla/javascript/NativeFinalizationRegistry.java
anivar: The cleanup methods (processCleanup, executeCleanupCallback, etc.) only run when GC actually collects registered objects.

  For test coverage, I could add white box tests that use reflection to test these methods directly:

```
  @Test
  public void testProcessCleanupDirectly() throws Exception {
      Method processCleanup = NativeFinalizationRegistry.class
          .getDeclaredMethod("processCleanup", FinalizationWeakReference.class);
      processCleanup.setAccessible(true);
      // ... test the method
  }
```

aardvark179 (changes requested):
I thought I'd written a review of this, but I must have closed the tab or something.

So, according to the spec the cleanup routines should be run when the weak ref target is no `EMPTY`, so this should probably be written using `PhantomReference`s (they are put on the queue when their referent has been collected). I am also unsure what the effect is on the JVM of create a large number of processor queues, and they can be tricky to manage correctly.

Could you try implementing this using a single queue (maybe using a `java.lang.ref.Cleaner` and putting the tasks from there onto a queue in each `Context`. We could likely combine processing of those with the micro task loop run as the context is exited.

I think that would help avoid large reference queues hanging around, and would make it easier to understand when things can be run.

anivar:
@aardvark179 Thank you for the detailed review! Yes, he did leave inline comments earlier and I responded to them and followed the suggestions. However, after refactoring the implementation, the filenames changed and I can't access those same comments anymore.

I understand your architectural suggestions:
1. Using PhantomReferences instead of WeakReferences for more deterministic cleanup behavior
2. Implementing a single shared ReferenceQueue across all FinalizationRegistry instances for efficiency
3. Integrating the cleanup process with the Context's micro-task loop for better timing control

anivar:
@aardvark179 Implementation completely rewritten per your feedback:

PhantomReferences now replace WeakReferences for correct GC semantics. Added shared ReferenceQueue with Context integration for cleanup processing in the JavaScript execution loop as requested.

Key improvements:
- Thread-safe operations with ConcurrentHashMap
- O(1) token-based unregistration 
- Memory leak prevention with proper cleanup
- Significant code simplification 
- All deprecation and ErrorProne warnings properly resolved
- 72+ tests passing including cross-realm scenarios


Ready for re-review!

aardvark179:
Sorry I haven’t got to re-reviewing this one yet. It’s been a busy few weeks, but I promise I will get to it.

It would also be great to have a chat at some point so I can understand your Rhino use case better and maybe help plan out which features are the best ones to tackle first to enable that.

anivar:
No worries @aardvark179! 

I'd be interested in that chat about Rhino use cases and feature prioritization. Feel free to reach out whenever works for your schedule.

Thanks for the architectural guidance on PhantomReference and shared queue - it made the implementation much cleaner.

anivar:
Force pushed a simplified implementation that follows the NativeWeakMap pattern. All tests passing, Android-compatible, no Context modifications needed. Ready for review.

PR #2059, opened 2025-08-31
Fix Array.from to prioritize iterable over array-like objects

### LINKED ISSUE #1518: Array.from does not follow spec to prioritize iterable over array-like
According to the spec, if the object passed to `Array.from` has an iterator method with a `Symbol.iterator` key, then that should be used before checking for a length property to see if the object is array-like. This is true even for native arrays.

```javascript
function test() {
    let counter = 0
    Array.prototype[Symbol.iterator] = function* () {
        for (let i = 0; i < this.length; i++) {counter++; yield this[i]}
    }
    Array.from(['a','b','c'])
    return counter
}
```

In Chrome
```
console.log(test())
3
```

Rhino
```
js> print(test())
0
```

anivar:
Fixes Array.from to follow ES2015 spec by checking Symbol.iterator first, even on native arrays.

Previously Array.from skipped iterator check for NativeArray instances. Now follows spec correctly.

Closes #1518

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/NativeArray.java
+++ b/rhino/src/main/java/org/mozilla/javascript/NativeArray.java
@@ -709,9 +709,7 @@ public class NativeArray extends ScriptableObject implements List {
         }
 
         Object iteratorProp = ScriptableObject.getProperty(items, SymbolKey.ITERATOR);
-        if (!(items instanceof NativeArray)
-                && (iteratorProp != Scriptable.NOT_FOUND)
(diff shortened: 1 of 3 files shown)


rbri:
Looks good so far, but spotless fails....

anivar:
Closing as duplicate of #2077. Both PRs contain identical changes to fix Array.from to prioritize iterable over array-like objects. Keeping #2077 as it has a proper branch name (fix-array-from-iterator) instead of 'master'.

PR #2060, opened 2025-09-01
Don't propagate strict flag when compiling scripts

andreabergia:
Before this change, if we compiled dynamically a script while we already had an active context in strict mode, we would have propagated the strict flag, which is incorrect. I've modified the code to not do so.

`eval` required some special handling; I have added a test to verify that we have not broken it (and I've double-checked with node). The `Function` constructor does not suffer from this problem, because it _always_ parses the code in non-strict mode (i.e. it does not propagate the flag, whereas `eval` should), per the spec.

Nothing in test262 breaks, which is encouraging.

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/Context.java
+++ b/rhino/src/main/java/org/mozilla/javascript/Context.java
@@ -2615,9 +2615,6 @@ public class Context implements Closeable {
         if (returnFunction) {
             p.calledByCompileFunction = true;
         }
-        if (isStrictMode()) {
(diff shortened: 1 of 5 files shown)


gbrail:
This seems right to me, and it shouldn't break anything that was correct in the first place ;-) Anyone else concerned about compatibility? If not then let's do it. Thanks!

andreabergia:
If it helps, we have tested this internally and it passed all our IT suite.

rbri:
fine for me, let's do it!

rbri:
Done

Issue #1968, opened 2025-07-08
Thread safety broke again, how can we enable it in testing?


aardvark179:
So, thanks you to @andreabergia for working on merging upstream changes into our fork we now know that I broke thread safe objects, twice. Once when converting `NativeArray` to lambdas and once then converting `BaseFunction`. I've got a rough fix for these problems, but I want to go through a few more iterations before it's ready, but as part of that work I'd like to enable running the 262 suite with thread safe objects enabled. This isn't as simple as just enabling it in the tests, because a failure might mean a deadlock and we don't want CI runs to get stuck forever.

I think to enable this we want to do a couple of things
1. Enable a time out for the `tests:test` task just in case *everything* deadlocks. Would 30 minutes be sufficient? CI builds take about 15 minutes at the moment, and I think most of us get faster builds locally.
2. Add `@Timeout` to `test262Case` in case a small set of tests deadlock. I think I'd need to experiment on what this value should be, locally 10 seconds is too small(!) for a couple of tests, I guess it doesn't matter if it's something big like 1 mimnute if we have a global timeout.
3. Expand our `TestMode` to add a `THREADSAFE` mode. Should this be one mode or should we test in both interpreted and compiled mode?

rbri:
>  Would 30 minutes be sufficient?

fine for me

> Expand our TestMode to add a THREADSAFE mode. Should this be one mode or should we test in both interpreted and compiled mode?

i guess we need both because the code paths are different and deadlock might occur on both ways

aardvark179:
I guess we might, yeah.

gbrail:
I think that this is a great idea. Since the standard tests are already
pretty slow, we could set it up so that it only runs in the GitHub actions
-- perhaps as a separate set of tests so that it can run in parallel with
the others.

I have also thought of running a test with "updateTest262properties" set to
true, and failing it if any changes were suggested, to catch the case where
someone fixed test262 tests but didn't update the file. But that's a
different issue!

andreabergia:
Opened https://github.com/mozilla/rhino/pull/2061 to fix this problem.

andreabergia:
Guess we can close this @gbrail ?

aardvark179:
I think so, we certainly appear to be running the tests in CI.

PR #2061, opened 2025-09-01
Run tests on the CI with thread-safe mode on

andreabergia:
In this PR, I've done two things:

- modify the default `ContextFactory` to allow configuration of the default value for `Context.FEATURE_THREAD_SAFE_OBJECTS` via the `RhinoConfig` mechanism
- added a new step in the CI to run all the tests once more with the flag set to true. I've not done it for all the JDK versions we support because it seems unnecessary. There is a 30 minutes timeout for the step. Verified via https://github.com/mozilla/rhino/pull/2062.

See the discussion in https://github.com/mozilla/rhino/issues/1968

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -62,6 +62,18 @@ jobs:
       with:
         name: reports-java-11
         path: '*/build/reports'
+    - name: Build and test Java 21 - Multithread
+      timeout-minutes: 30
+      env:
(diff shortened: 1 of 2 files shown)


gbrail:
Thanks -- this will help maintain quality, and I'm happy to talk about ways to speed up the CI too.

Also, looking at code coverage, given our variety of slot map classes, I think we should invest in some more unit tests for all that stuff and possibly even some that can deliberately cause locking conflicts.

PR #2062, opened 2025-09-01
Test of thread-safety feature where it's supposed to hang

andreabergia:
This PR is used to test only that, on a broken version, the CI will hand and eventually times out. Should not be merged.

See https://github.com/mozilla/rhino/pull/2061


andreabergia:
Closing because the test was successful.

PR #2063, opened 2025-09-01
Run CI on different java version in parallel

andreabergia:
Executes in parallel five jobs:

- the new "is tests262 up-to-date" check
- all the unit tests with:
  - java 11
  - java 17
  - java 21
  - java 21 with multithread on

Stacked on top of https://github.com/mozilla/rhino/pull/2061

Makes the CI go from 45 minutes (before #2061) or 60 minutes (after 2061) to about 17 minutes. 🕐 🚀

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -50,12 +50,12 @@ jobs:
       uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2
       if: ${{ always() }}
       with:
-        name: reports-${{ format('java-{0}', matrix.java-version) }}${{ matrix.multithread && '-multithread' || '' }}
+        name: reports-java-${{ matrix.java-version }}${{ matrix.multithread && '-multithread' || '' }}
         path: '*/build/reports'
 
   test262:
     runs-on: ubuntu-latest
-    name: Test262 Suite
+    name: Check if test262.properties is up-to-date
     steps:
     - name: Check out Rhino
       uses: actions/checkout@ff7abcd0c3c05ccf6adc123a8cd1fd4fb30fb493 # v4.2.2


.github/workflows/gradle.yml
```diff
@@ -23,45 +33,42 @@ jobs:
...
-      # By putting JVMs in reverse order, the last one is the one that
-      # will go in the path.
       with:
+        # By putting JVMs in reverse order, the last one is the one that
+        # will go in the path.
         java-version: |
-          11
-          17
+          ${{ matrix.java-version }}
           21
```
rPraml, line 41: do we need `matrix.java-version` and `21` here?

.github/workflows/gradle.yml
```diff
@@ -23,45 +33,42 @@ jobs:
...
       run: >-
-        ./gradlew check
-    - name: Upload results Java 17
+        ./gradlew check ${{ matrix.multithread && '-Drhino.useThreadSafeObjectsByDefault=true' || '' }}
+    - name: Upload results ${{ matrix.name }}
       uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2
       if: ${{ always() }}
       with:
-        name: reports-java-17
+        name: reports-${{ format('java-{0}', matrix.java-version) }}${{ matrix.multithread && '-multithread' || '' }}
```
rPraml:
```suggestion
        name: reports-java-${{ matrix.java-version }}${{ matrix.multithread && '-multithread' || '' }}
```

gbrail:
Does this mean that all the tests run in parallel on the same runner? In the past we had test timeouts when we tried that. However, we've disabled or fixed some of the most egregious tests since then so I'm willing to give it a try. Let me just check the docs and make sure I understand what we're getting ourselves into trouble here with GH.

.github/workflows/gradle.yml
```diff
@@ -23,45 +33,42 @@ jobs:
...
-        RHINO_TEST_JAVA_VERSION: "11"
-      run: >-
-        ./gradlew check
-    - name: Upload results Java 11
-      uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2
-      if: ${{ always() }}
+
+  test262:
+    runs-on: ubuntu-latest
+    name: Test262 Suite
```
rPraml: I think this step should be called
```suggestion
    name: Check if test262.properties is up-to-date
```
See: https://github.com/mozilla/rhino/pull/2052/files

.github/workflows/gradle.yml
```diff
@@ -23,45 +33,42 @@ jobs:
...
+      # We don't actually want all the history for this part
+      run: git submodule update --init --single-branch
+    - name: Set up Java
+      uses: actions/setup-java@dded0888837ed1f317902acf8a20df0ad188d165 # v5.0.0
       with:
-        name: reports-java-11
-        path: '*/build/reports'
+        java-version: |
+          11
+          21
```
rPraml, line 70: No need to install 11 + 21 
Maybe this step can be removed completely, as `ubuntu-latest` has already java installed.

I think, it does not matter under whích java version the test262.properties is verified (and maybe we can run multithread here, too)

rPraml (commented):
I like that change, as it may speed up tests (as long as they pass)
OTOH, we "waste" cpu time, if a test fails.
If it falils on java 11, it will fail with 99% on 17 + 21 also

The same for spotless (or compile error)... If there is a spotless error, we have 4 failing buld jobs.
Maybe, we can exclude spotless from the java 11/17/21 build and make an extra spotless stage?
(Can be done in a later PR also)

.github/workflows/gradle.yml
andreabergia: About this and the point below about the test262 up-to-date check: I have kept the pre-existing code from https://github.com/mozilla/rhino/pull/1627, which installed _all_ java versions.
I am unsure if it's actually necessary, or we can just install the latest and use the `-release` flag.

I'm willing to try to change this, though @gbrail do you remember anything, even though it was almost a year ago...

.github/workflows/gradle.yml
andreabergia: Good point.

.github/workflows/gradle.yml
andreabergia: Agreed, done.

andreabergia:
> Does this mean that all the tests run in parallel on the same runner? In the past we had test timeouts when we tried that. However, we've disabled or fixed some of the most egregious tests since then so I'm willing to give it a try. Let me just check the docs and make sure I understand what we're getting ourselves into trouble here with GH.

They definitely run in parallel; I assume there's multiple isolated runners. The total time spent should be about the same - the only "duplicated" steps are the git checkout and the java toolchain setup, which take only some seconds.


> I like that change, as it may speed up tests (as long as they pass) OTOH, we "waste" cpu time, if a test fails. If it falils on java 11, it will fail with 99% on 17 + 21 also
> 
> The same for spotless (or compile error)... If there is a spotless error, we have 4 failing buld jobs. Maybe, we can exclude spotless from the java 11/17/21 build and make an extra spotless stage? (Can be done in a later PR also)

That is a fair point, we definitely waste some CPU time on failures. We could consider something like doing a spotless and the test262 check _first_ and then the parallel run of all the tests, calling just `test` instead of `check`, but that would just end up taking more time when there's no failures (which I imagine is the majority of cases).

gbrail:
I've re-created this on the current master at the following PR and took some time to understand it myself -- I'm going to close this one in favor of that. Thanks!

https://github.com/mozilla/rhino/pull/2147

andreabergia:
> I've re-created this on the current master at the following PR and took some time to understand it myself -- I'm going to close this one in favor of that. Thanks!
> 
> #2147

Cool, I'm glad it was useful!

PR #2066, opened 2025-09-03
Fix a problem with toString() on arrow functions

0xe:
```javascript
eval("()=>this").toString()
```
will return `()=>thi` with the last character cut off because of a problem with calculating `tokenEnd`.

Returns `()=>this` correctly after the fix.


aardvark179:
Good find!

rhino/src/main/java/org/mozilla/javascript/TokenStream.java
```diff
@@ -780,9 +780,15 @@ final int getToken() throws IOException {
...
                 }
-                ungetChar(c);
 
-                String str = getStringFromBuffer();
+                // `ungetChar` will decrement the cursor, however the _actual_ tokenEnd should
+                // still be restored correctly after `getStringFromBuffer()` mutates it, so we save
+                // and restore it.
+                int savedTokenEnd = cursor;
+                ungetChar(c); // decrements cursor
+                String str = getStringFromBuffer(); // restores tokenEnd to cursor
```
rbri: maybe the comment is for the line below?

rhino/src/test/java/org/mozilla/javascript/tests/ArrowFunctionToStringTest.java
```diff
@@ -0,0 +1,12 @@
+package org.mozilla.javascript.tests;
+
+import org.junit.jupiter.api.Test;
+import org.mozilla.javascript.testutils.Utils;
+
+public class ArrowFunctionToStringTest {
```
rbri: We already have a ArrowFnPositionBugTest. Maybe we should rename this to ArrowFunctionTest and merge this test into it.

rhino/src/test/java/org/mozilla/javascript/tests/ArrowFunctionToStringTest.java
```diff
@@ -0,0 +1,12 @@
...
+    @Test
+    public void EvalArrowToString() {
```
rbri: lowercase at the beginning of method names in java ;-)

0xe: Addressed.

rhino/src/main/java/org/mozilla/javascript/TokenStream.java
0xe: clarified them a bit.

rhino/src/test/java/org/mozilla/javascript/tests/ArrowFunctionToStringTest.java
0xe: Moved it to that test and renamed that to `ArrowFunctionTest`

rbri:
as always - thanks

PR #2069, opened 2025-09-03
Allow using 'let' and 'yield' in non-strict mode as a name

zhurvla:


gbrail:
Can I understand the problem we're trying to solve here? It sounds to me like a backward-compatibility issue. That's a challenge that we continue to have with this project, which is because Rhino is so old, and embedded in so many places, and unlike many other engines it supports a "language level" configuration option that other engines' don't support. It's also a lot of work for the community to keep all that working, it makes the codebase more complex, and it probably affects performance.

With all that said, I COULD see the case for "let" and "yield" to be treated as identifiers and not keywords, but only if it's done in such a way that it affects only language levels below a certain point, basically as a backward-compatibility feature.

It looks like the PR in the current form affects the current language level and breaks a lot of tests, so as a starting point we definitely don't want that.

PR #2071, opened 2025-09-08
Create devcontainer.json

emnaidev-1993:


gbrail:
Can you help us understand? What's a devcontainer.json and why would we want one?

PR #2073, opened 2025-09-10
Duplicate properties in object literal should be allowed

zhurvla:
The following should work in > ES6 mode, and throw a SyntaxError: Property "foo" already defined in this object literal. in pre-es6 mode:

(function() {
	'use strict';
	o = {foo: 1, foo: 2};
	o.foo
})();


rbri:
LGTM but i like to vote for two more tests testing this once for ES6 and once for below. I think this will make our live easier in the long run

zhurvla:
@rbri Tests added

rbri:
Thanks a lot!

Issue #943, opened 2021-06-23
Support EcmaScript 2021 WeakRef & FinalizationRegistry


p-bakker:
> WeakRef, for referring to a target object without preserving it from garbage collection
> FinalizationRegistry, to manage registration and unregistration of cleanup operations performed when target objects are garbage collected

See [MDN WeakRef](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakRef)

tonygermano:
Is there any reason this couldn't just be a thin wrapper around `java.lang.ref.WeakReference`?

p-bakker:
I haven't looked into the details, but I think/hope it would be just that

PR #2074, opened 2025-09-12
Implement ES2021 WeakRef with ES2023 symbols-as-weakmap-keys support

anivar:
This PR implements the ES2021 WeakRef constructor and prototype, allowing JavaScript code to hold weak references to objects without preventing their garbage collection. This is particularly useful for implementing caches, memoization, and other memory-sensitive patterns.

## Changes

The implementation adds a new NativeWeakRef class that provides the WeakRef constructor and its deref() method. The implementation follows the ES2021 specification while also supporting unregistered symbols as weak targets, which is part of the ES2023 symbols-as-weakmap-keys specification.

Key implementation decisions:
- Uses Java's WeakReference internally for proper garbage collection semantics
- Follows the same validation patterns as NativeWeakMap for consistency
- Implements using the LambdaConstructor pattern consistent with other modern Rhino features
- Properly validates targets according to the CanBeHeldWeakly specification
- Supports ES2023 feature: unregistered symbols can be held weakly

## Testing

Unit tests: All 29 unit tests pass, covering constructor validation, deref behavior, symbol handling, and error cases.

Test262: 25 out of 29 test262 tests pass (86.21% success rate). The 4 failing tests involve advanced edge cases around cross-realm prototype handling and NewTarget customization that don't affect normal usage.

Additionally, this implementation fixes an existing test262 test (Object/seal/seal-weakref.js) that was previously failing.

## Technical Notes

The implementation correctly distinguishes between registered and unregistered symbols per ES2023 symbols-as-weakmap-keys specification. Registered symbols (created with Symbol.for()) cannot be held weakly as they lack unique identity, while unregistered symbols can be held weakly. This matches the behavior of WeakMap and WeakSet.

Partially addresses #943 (WeakRef portion - FinalizationRegistry still needs implementation)

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/NativeWeakRef.java
+++ b/rhino/src/main/java/org/mozilla/javascript/NativeWeakRef.java
@@ -1,0 +1,176 @@
+/* -*- Mode: java; tab-width: 8; indent-tabs-mode: nil; c-basic-offset: 4 -*-
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.javascript;
+
+import java.lang.ref.WeakReference;
+
+/**
+ * Implementation of the ES2021 WeakRef constructor and prototype.
+ *
+ * <p>WeakRef allows holding a weak reference to an object without preventing its garbage
+ * collection. This is useful for caches, mappings, or any scenario where you want to reference an
+ * object without keeping it alive.
+ *
+ * <p>The WeakRef object provides a single method, deref(), which returns the referenced object if
+ * it's still alive, or undefined if it has been collected.
+ *
+ * @see <a href="https://tc39.es/ecma262/#sec-weak-ref-objects">ECMAScript WeakRef Objects</a>
+ */
+public class NativeWeakRef extends ScriptableObject {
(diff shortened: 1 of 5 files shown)


gbrail (commented):
This looks helpful and straightforward -- thanks! 

Do you think that you could put a bit of detail in the comments about the four failing test cases? I feel that if we don't address them now (even if they're advanced use cases) we may never get to them, so I'd like to at least see what the possibilities are. Thanks!

anivar:
@gbrail I've analyzed the 4 failing test cases. They all relate to how the constructor handles the newTarget parameter when called through Reflect.construct.

 When calling `Reflect.construct(WeakRef, [target], newTarget)`, the spec requires using newTarget's prototype, but LambdaConstructor always uses WeakRef's prototype. This  affects cross-realm construction and custom prototypes.                                                                                                                     

The deref test fails because it should only work on objects that actually went through WeakRef construction, not just objects that inherit from WeakRef.prototype.

  These are edge cases used in proxies and polyfills. The core WeakRef functionality works - it properly holds weak references that don't prevent garbage collection."         │

Fixing this requires either extending LambdaConstructor to handle newTarget (which would benefit other built-ins too) or implementing WeakRef with a traditional constructor pattern. The current implementation delivers the core WeakRef functionality - weak references that don't prevent garbage collection - which is what most developers need.

anivar:
Closing this PR as WeakRef is now included in PR #2058 along with FinalizationRegistry. The consolidated implementation provides both ES2021 features together with ES2023 symbol support.

PR #2077, opened 2025-09-12
Fix Array.from to prioritize iterable over array-like objects

### LINKED ISSUE #1518: Array.from does not follow spec to prioritize iterable over array-like

anivar:
Fixes #1518

As @tonygermano reported, custom iterators on arrays weren't being respected. @advenk correctly traced this to the `!(items instanceof NativeArray)` check that was bypassing iterator protocol.

## The Fix
Removed the NativeArray exclusion - just deleted `!(items instanceof NativeArray) && ` from the condition. 

Now Tony's test case works correctly:
```javascript
let counter = 0;
Array.prototype[Symbol.iterator] = function* () {
    for (let i = 0; i < this.length; i++) {
        counter++; 
        yield this[i];
    }
};
Array.from(['a','b','c']);
// counter is 3 ✓
```

This aligns with V8, SpiderMonkey, and JavaScriptCore behavior. Also fixes 2 test262 failures.

Tony's performance concern about optimization is valid - we could check if the iterator is unmodified in a future PR, but spec compliance comes first.

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/NativeArray.java
+++ b/rhino/src/main/java/org/mozilla/javascript/NativeArray.java
@@ -709,9 +709,7 @@ public class NativeArray extends ScriptableObject implements List {
         }
 
         Object iteratorProp = ScriptableObject.getProperty(items, SymbolKey.ITERATOR);
-        if (!(items instanceof NativeArray)
-                && (iteratorProp != Scriptable.NOT_FOUND)
-                && !Undefined.isUndefined(iteratorProp)) {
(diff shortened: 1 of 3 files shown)


gbrail:
Looks good, thanks!

PR #2080, opened 2025-09-12
Apply type consolidation to method/list/map/array

ZZZank:
# Java Generic support

- Type Consolidation (Stage 1 in #2023 , stage 2 in this PR)
- Generic support for Method/Constructor/List/Map/Array (This PR)

## Features/Fixes in this PR

- Type consolidation for general Java object action: method & beaning
- Type consolidation for some special Java object action: `map[...]`, `list[...]`, `array[...]`, iterator
- Implement `TypeInfo#consolidate(...)` for complex types, for example `T[]` and `List<E>`
- `WrapFactory` now uses `TypeInfo` instead of `Class<?>` to hold type information. Methods with old signature is kept for basic back-ward compat.
- Better `.hashcode()` and `.equals(...)` for TypeInfo implementations. This fixes an issue where same type representation from different factory is considered as different.
- Fix `NativeJavaMap.put(int, ...)` not usable when there's no Context. This is considered as a fix because having no Context is allowed in other `NativeJavaMap` methods.
- Fix `TypeInfoFactory.consolidateAll(...)` for arg types with only one element.
- Fix `SecurityControllerTest` using the wrong method for accessibility test

## Brief walk-through of type consolidation 

Type consolidation is applied by reading and manipulating the `staticType` passed to JS-Java bridge object (`NativeJavaXXX`). When new bridge object is created, an instance of `TypeInfo` will be required. For JavaScript, new TypeInfo instance is auto generated from existed TypeInfo.

### Type consolidation for Method

Consider this example:

```javascript
/** @type {List<String>} */
let list;

list.add(0.1);
```

Here `list` is an instance of `NativeJavaObject` with its `staticType` field representing `List<String>`. When calling `list.add(0.1);`, this Java method should be located and called: `List#add(E)`, where the `E` is `E in List<E>`.

Before this PR, `E` will be resolved to its bound `Object.class`, causing a `Double` to be added into a list for `String`.

In this PR, a `E -> String` mapping will extracted from `staticType` (see [here](https://github.com/ZZZank/rhino-mozilla/blob/b9618da8b65be6c999675274ba63bdd437de6dea/rhino/src/main/java/org/mozilla/javascript/NativeJavaMethod.java#L145)) and passed to arg wrapping code (see [here](https://github.com/ZZZank/rhino-mozilla/blob/b9618da8b65be6c999675274ba63bdd437de6dea/rhino/src/main/java/org/mozilla/javascript/MemberBox.java#L336)). Then `E` will be dynamically mapped to `String` based on the mapping, so `0.1` can be properly wrapped to `"0.1"`.

This is the "Stage 2 type consolidation" mentioned in previous PRs.

### Handling inherited methods

Let's use the same example as last section, but the Java located method is `Collection#add(E)` instead of `List#add(E)`.

In this case `E` is `Collection<E>`, and a additional step is introduced to handle this. We will use the method implemented in #2023  to generate a mapping from `List.class`. This mapping will include this entry: `E from Collection<E> -> E from List<E>`. By applying this mapping before the "Stage 2 type consolidation", we can make sure that type variables will always be those declared by `List`. See [here](https://github.com/ZZZank/rhino-mozilla/blob/b9618da8b65be6c999675274ba63bdd437de6dea/rhino/src/main/java/org/mozilla/javascript/MemberBox.java#L69)

This is the "Stage 1 type consolidation" mentioned in previous PRs, because it happens earlier.

### Type consolidation for List/Map/Array

Supporting special actions for List/Map/Array is easier than supporting method.

Take list as an example. At initialization, `E from List<E>` will be consolidated based on the mapping extracted from `staticType`, and then stored as `elementType`. The `elementType` will be used when doing list special sctions. See [here](https://github.com/ZZZank/rhino-mozilla/blob/b9618da8b65be6c999675274ba63bdd437de6dea/rhino/src/main/java/org/mozilla/javascript/NativeJavaList.java#L63)

Example:
```javascript
/** @type {ArrayList<String>} */
let list;

list[0] = 0.1;
```

Then `E from List<E>` will be consolidated following this route: `E from List<E>` -> `E from ArrayList<E>` -> `String`. So when doing `list[0] = 0.1;`, the `0.1` can be automatically wrapped to `"0.1"`


ZZZank:
This PR is now a draft because of a problem I found during testing:

There's a `dynamicType` and `staticType` setting when initializing `JavaMembers`, and `dynamicType` will be preferred over `staticType`. This will cause a situation that cannot be handled by type consolidation.

Example: `staticType` is `List<String>`, `dynamicType` is the class of actual object, aka `ArrayList`. Then we will have these type variables from `ArrayList`:

```
T in Iterable<T> // covered by stage 1 type consolidation
E in Collection<E> // covered by stage 1 type consolidation
E in List<E> // covered by stage 2 type consolidation
E in AbstractList<E>
E in ArrayList<E>
```

When `list.add(0.1)`, the located method is `AbstractList#add(E)`, where the E is `E in AbstractList<E>`. This type variable is NOT covered by type consolidation.

ZZZank:
Finally ready for review. I think I should record some findings & problems I found during development of this PR, they are certainly out of scope of this PR.

1. `NativeJavaMap` can accept any type of key, but in JS, the `key` in `map[key]` will only be `String` or `int`, causing a type mismatch.

The result of the mismatch is that despite we are capable of supporting `Map<K, V>`, we can actually only support `Map<K super String, V>` and `Map<K super Integer, V>`. Should we explicitly apply this limit or ignore it even when key type is not supported?

Example: 
- Given a `Map<String, String> map`, should we explicitly deny `map[1]`? 
- Given a `Map<Class<?>, String> map` or `Map<Number, String> map`, should we skip redirecting `map['key']` to `map.get('key')`? This redirection is an optional feature described in `Context.FEATURE_ENABLE_JAVA_MAP_ACCESS`.

2. The `dynamicType` and `staticType` system `JavaMembers` is causing many problems.

In additions to the problem described in https://github.com/mozilla/rhino/pull/2080#issuecomment-3285901869 , There's a even more severe one: If the `staticType` is an interface, `dynamicType` is a class with restricted access, then members that supposed to be accessible could be missing. I found an example when debugging `SecurityControllerTest`.

When `staticType` is `SomeInterface`, `dynamicType` is `SomeImpl`:

```java
// accessible
public interface SomeInterface {
    String foo();
}

// not accessible
class SomeImpl extends ArrayList<String> implements SomeInterface {
    public String foo() { return "foo"; }
}
```

`JavaMembers` will first try to access `SomeImpl` and then try its superclass `ArrayList` after failure. `ArrayList` is accessible, so member accessing will end here, reflected members are all members from `ArrayList`. 

Here is the problem, due to `JavaMembers` only accessed `ArrayList`, `SomeInterface.foo()` is not accessed at all. It will cause such situation that, you get an instance of an interface, but somehow, you are unable to accessible any members in this interface, but able to access members from another random class.

My planned solution is to remove `dynamicType` and only use `staticType` as type hint.

gbrail:
Are any intense users of LiveConnect interested in trying this out? Otherwise it looks like another step in making this stuff more robust and I don't see anything to be worried about. I'd like to see if there are some other regular users who feel differently!

aardvark179:
I think the majority of our Java usage is through explicit scriptable implementations, but I’m sure there is stuff using live connect in there somewhere.

Since I don’t think we have good tests of the behavior we depend on I don’t think we can object to any todo g up. We are working to getting our fork caught up with mainline, and that will help us give more timely feedback on features like this, but we aren’t quite there yet.

Issue #1696, opened 2024-10-14
ES2025 Sync Iterator helpers


p-bakker:
Proposal: https://github.com/tc39/proposal-iterator-helpers

PR #2082, opened 2025-09-12
Implement ES2025 Iterator.from() static method

anivar:
## Summary

This PR adds the Iterator.from() static method, continuing the ES2025 iterator helpers implementation started in #2078.

## What's Added

Iterator.from() converts any iterable or iterator-like object into a proper iterator that inherits from Iterator.prototype. This includes:

- String primitives get special handling since they're iterable
- Arrays, Sets, Maps work through their Symbol.iterator
- Objects with a next() method are treated as iterator-like
- Objects with null/undefined Symbol.iterator are handled as iterator-like (per spec)
- Already-proper iterators are returned unchanged

## Implementation Approach

The IteratorWrapper class follows the same pattern as NativeArrayIterator - it extends ES6Iterator and uses the standard delegation pattern through isDone() and nextValue(). This keeps things consistent with how Rhino handles iterators elsewhere in the codebase.

The wrapper also handles return() and throw() methods when they exist on the underlying iterator, ensuring proper cleanup and error propagation.

## Known Limitation

The global Iterator constructor isn't exposed yet because it conflicts with the legacy Iterator() function that's used for Java interop. We'll need to figure out a migration path for this before the full iterator helpers can be used from JavaScript.

## Testing

Basic tests verify that Iterator.prototype exists with the right Symbol properties. The infrastructure is ready for more comprehensive testing once we resolve the global naming conflict.

Part of #1696  
Depends on #2078

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/IteratorWrapper.java
+++ b/rhino/src/main/java/org/mozilla/javascript/IteratorWrapper.java
@@ -9,93 +9,36 @@ package org.mozilla.javascript;
 /**
  * Wrapper for iterators returned by Iterator.from(). This class wraps any iterator and ensures it
  * inherits from Iterator.prototype.
+ *
+ * <p>Following Rhino patterns, this extends ES6Iterator and delegates all operations to the wrapped
+ * iterator object.
  */
-public class IteratorWrapper extends ScriptableObject {
+public final class IteratorWrapper extends ES6Iterator {
     private static final long serialVersionUID = 1L;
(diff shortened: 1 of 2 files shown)


anivar:
Closing in favor of the more comprehensive implementation in #2078 which includes:
- Feature flag system for backward compatibility
- Iterator.from() implementation (this PR's goal)
- Iterator.prototype.toArray() 
- Proper test infrastructure updates
- 87.96% test262 pass rate

The implementation in #2078 supersedes this PR.

PR #2087, opened 2025-09-13
Implement Math.sumPrecise for ECMAScript proposal-math-sum

anivar:
Implements ES2026 Math.sumPrecise for precise floating-point summation. Uses Shewchuk's Grow-Expansion algorithm to handle the precision loss that happens with regular addition when summing arrays of numbers.

## Implementation

This is a Phase 1 implementation supporting array-like objects only. Iterator support will come in Phase 2 after PR #2078 provides a Context-safe iterator pattern.

The implementation uses Shewchuk's algorithm with a partials array to track error components during summation. This maintains precision even when summing values with vastly different magnitudes.

## How it works

Handles the tricky cases in floating-point math:
- Values with very different magnitudes: `[1e20, 0.1, -1e20]` returns `0.1` (regular sum gives `0`)
- Special IEEE 754 values: infinities, NaN, signed zeros all per spec
- Large arrays: enforces 2^53 element limit per spec
- Type safety: validates all elements are Numbers

## Current status

Phase 1 complete:
- Array-like objects: `Math.sumPrecise([1.1, 2.2, 3.3])`
- Length validation with `ScriptRuntime.toLength()`
- Full ES2026 spec compliance for implemented path
- All test262 tests passing for array-like

Phase 2 deferred:
- Iterator protocol support waiting on PR #2078
- Will add `Math.sumPrecise(iterable)` once Context-safe iterator available

The core Shewchuk algorithm implementation is complete and ready - just needs the iterator wrapper once PR #2078 merges.

### DIFF
--- a/tests/testsrc/test262.properties
+++ b/tests/testsrc/test262.properties
@@ -3212,11 +3212,10 @@ language/computed-property-names 31/48 (64.58%)
     to-name-side-effects/class.js
     to-name-side-effects/numbers-class.js
 
-language/destructuring 9/19 (47.37%)
+language/destructuring 8/19 (42.11%)
     binding/syntax/array-rest-elements.js
     binding/syntax/destructuring-array-parameters-function-arguments-length.js
     binding/syntax/destructuring-object-parameters-function-arguments-length.js
-    binding/syntax/property-list-with-property-list.js
     binding/syntax/recursive-array-and-object-patterns.js
     binding/initialization-requires-object-coercible-null.js
     binding/initialization-requires-object-coercible-undefined.js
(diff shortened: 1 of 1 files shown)


gbrail:
So far, we haven't implemented features before the spec has been published. Looking for some other opinions on this...

p-bakker:
Math sumPrecise is a finished proposal, scheduled for ES2026, so engine makers are encouraged to implement it now: https://github.com/tc39/proposals/blob/main/finished-proposals.md

anivar:
Thank you for the review @aardvark179! I've refactored the implementation to eliminate the code duplication as requested.

## Changes made:

1. **Unified iterator approach**: Both array-like objects and iterables now use the same algorithm through a shared  helper method, eliminating the duplicate loops.

2. **Single algorithm path**: The refactoring ensures all input types (arrays, array-likes, and iterables) flow through a single iterator-based path, removing the duplication you identified.

3. **Enhanced algorithm**: While addressing the duplication, I also improved:
   - Overflow handling with biased partial tracking (handles extreme values better)
   - Special value handling for -0, infinities, and NaN
   - Edge case handling for MAX_DOUBLE boundaries
   - Extracted TwoSum algorithm into a cleaner inner class

The refactored code is cleaner, more maintainable, and all test262 tests continue to pass. The core Shewchuk algorithm is now in one place as requested.

anivar:
Thank you for the review @aardvark179! I have refactored the implementation to eliminate the code duplication as requested.

## Changes made:

1. **Unified iterator approach**: Both array-like objects and iterables now use the same algorithm through a shared sumPreciseFromIterable helper method, eliminating the duplicate loops.

2. **Single algorithm path**: The refactoring ensures all input types (arrays, array-likes, and iterables) flow through a single iterator-based path, removing the duplication you identified.

3. **Enhanced algorithm**: While addressing the duplication, I also improved:
   - Overflow handling with biased partial tracking (handles extreme values better)
   - Special value handling for -0, infinities, and NaN
   - Edge case handling for MAX_DOUBLE boundaries
   - Extracted TwoSum algorithm into a cleaner inner class

The refactored code is cleaner, more maintainable, and all test262 tests continue to pass. The core Shewchuk algorithm is now in one place as requested.

PR #2089, opened 2025-09-14
Add automated ECMAScript feature documentation system

### LINKED ISSUE #2088: Feature documentation is incomplete - only shows subset of capabilities, not all test262 results
## Current Documentation Problem

Rhino's feature documentation is incomplete and outdated. It only covers a small subset of ECMAScript features, while test262.properties shows we test and support hundreds of features across multiple categories.

## The Real Scope

Our test262.properties file (7,500+ lines) tests:
- **Built-in Objects**: Array, String, Object, Map, Set, Promise, etc. with all their methods
- **Language Features**: async/await, arrow functions, destructuring, generators, etc.
- **Annex B Features**: Legacy browser compatibility features
- **Internationalization**: Intl.DateTimeFormat, Intl.NumberFormat, etc.
- **RegExp Features**: All regex capabilities and flags

Currently, none of this comprehensive testing is reflected in our documentation.

## Examples from test262.properties

```
built-ins/Array 2816/3077 (91.52%)
built-ins/String 1154/1212 (95.21%)
built-ins/Object 3289/3410 (96.45%)
built-ins/Promise 256/639 (40.06%)
language/expressions/arrow-function 192/343 (55.98%)
annexB/built-ins 58/241 (24.07%)
intl402/DateTimeFormat 234/456 (51.32%)
```

## Impact

Developers cannot see:
- Which specific Array methods work (map, filter, reduce, etc.)
- Which String methods are supported
- Detailed language feature support
- Internationalization capabilities
- Legacy compatibility status

## Proposed Solution

Create an automated system that:
1. Parses ALL entries from test262.properties
2. Automatically categorizes features by type
3. Generates comprehensive documentation showing every tested feature
4. Updates automatically when test results change

This would provide complete transparency about Rhino's capabilities, not just selected ES features.

anivar:
Fixes #2088

## Problem
Rhino's feature documentation only shows ES2015-ES2017, but test262 results show we support many ES2018+ features with high pass rates.

## Solution
Automated system that generates accurate documentation from test262.properties results on every merge to master.

## Key Components

**External Configuration**: Feature mappings in `tools/es-features.txt` (pipe-delimited format)
**Analysis Engine**: `tools/feature_tracker.py` parses test262.properties for pass/fail counts  
**Multiple Outputs**: FEATURES.md, rhino-features.json, rhino-features.html
**Automation**: Triggers on every merge to master branch

## How It Works

1. Every merge to master triggers the workflow
2. Parses test262.properties to calculate feature pass rates
3. Generates documentation if anything changed
4. Creates PR with updated documentation
5. Handles both feature additions and reverts automatically

## Results

Documents 91 ECMAScript features across ES2015-ES2025:
- 21 features with full support (95-100% pass rate)
- 53 features with partial support (25-94% pass rate)
- 17 features with limited support (0-24% pass rate)

Previously undocumented working features like String.replaceAll, Object.hasOwn, Math.f16round, and globalThis are now properly documented with their actual pass rates.

## Benefits

- Always reflects current main branch state
- Automatic updates on every merge
- Handles feature additions and reverts transparently
- Zero maintenance burden after setup
- Easy feature additions via external configuration file

## Testing

System validated against current test262.properties with 29,000+ test results and correctly generates all documentation formats.


aardvark179:
So, I’m definitely not opposed to the idea of automating feature documentation more, but this is a huge amount to go through, and the mixture of tools used and style makes me think a lot has been generated and may need some thorough review.

To help me and others start to look at this could you maybe break it down a little for us?

1. For a large change like this an architectural overview is really helpful.
2. Which parts of this PR are generated output. Are they something that belongs in this repository or should they become build artefacts. Or, should they be something generated at the same time as the properties file is updated and be part of that process?
3. We already have Java code that parses our test 262 properties file. Would it be feasible to expand this to handle the feature documentation generation rather than add a second parser and extra tools to our build requirements?

PR #2090, opened 2025-09-15
Improve debug output on class file writing.

aardvark179:
While refactoring parts of the system which touch on class compilation I've found it difficult in some cases to debug failures. To improve that this PR:
1. Puts all debug output to stderr.
2. Prints start and end of methods.
3. Optionally allows the caller responsible for each instruction to be printed.

That last point has been *extremely* useful in speeding up debugging.

@0xe, let me know if this is helpful for the stuff you've been trying.


aardvark179:
@0xe, let me know if this is helpful for the stuff you've been trying.

gbrail:
OK -- no problem adding new debugging, but I'm going to suggest two things.

First, we have a cool new "RhinoConfig" method that we use to set all the other debug flags using system properties, environment variables, and config files. It looks like we never used it in that class. Could you check it out?

Second, ErrorProne is giving us a suggestion and I think we should take it. Thanks!

C:\Users\greg\src\rhino\rhino\src\main\java\org\mozilla\classfile\ClassFileWriter.java:231: warning: [ShortCircuitBoolean] Prefer the short-circuiting boolean operators && and || to & and |.
        if (DEBUGCODE | DEBUGMETHODS) {
                      ^
    (see https://errorprone.info/bugpattern/ShortCircuitBoolean)
  Did you mean 'if (DEBUGCODE || DEBUGMETHODS) {'?
C:\Users\greg\src\rhino\rhino\src\main\java\org\mozilla\classfile\ClassFileWriter.java:255: warning: [ShortCircuitBoolean] Prefer the short-circuiting boolean operators && and || to & and |.
        if (DEBUGCODE | DEBUGMETHODS) {
                      ^
    (see https://errorprone.info/bugpattern/ShortCircuitBoolean)
  Did you mean 'if (DEBUGCODE || DEBUGMETHODS) {'?

aardvark179:
I would have expected the use of RhinoConfog in the class file writer to trigger an error in decycle, but it doesn’t actually care because it doesn’t regard it as a cycle since it doesn’t care about package hierarchies.

It still feels wrong though, so I’m tempted to break the cycle and make Codegen read those options.

rhino/src/main/java/org/mozilla/classfile/ClassFileWriter.java
```diff
@@ -4471,11 +4516,25 @@ public String toString() {
...
+        System.err.printf("%s:%d ", f.getDeclaringClass().getSimpleName(), f.getLineNumber());
+    }
+
     private static final int FileHeaderConstant = 0xCAFEBABE;
     // Set DEBUG flags to true to get better checking and progress info.
     private static final boolean DEBUGSTACK = false;
     private static final boolean DEBUGLABELS = false;
     private static final boolean DEBUGCODE = false;
+    private static final boolean DEBUGCODEORIGINS = false;
+    private static final boolean DEBUGMETHODS = false;
```
rPraml: FYI, with #1722, I've added a config feature.
So you might add a config switch like this:
```suggestion
    private static final boolean DEBUGMETHODS = RhinoConfig.get("rhino.debugMethods", false)
```

Documetation https://github.com/rhino/rhino.github.io/pull/4/files

gbrail:
Looks good, thanks!

PR #2097, opened 2025-09-22
Fix: `eval` of a function should be undefined

andreabergia:
Calling:

```js
eval('function f(){})'
```

should be `undefined`. Before this change, it was the function, but the spec says that parenthesis are required around the function to return it, i.e.

```js
eval('(function f(){})')
```

should return a function.

Note: this is a backward-incompatible change, but a correct one in the sense that it fixes a bug. If needed, we could add a context flag to preserve the old behavior, but I haven't done that yet. I did have to update some old tests executed by `MozillaSuiteTest` though.

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/CodeGenerator.java
+++ b/rhino/src/main/java/org/mozilla/javascript/CodeGenerator.java
@@ -296,17 +296,6 @@ class CodeGenerator extends Icode {
                             throw Kit.codeBug();
                         }
                     }
-                    // For function statements or function expression statements
-                    // in scripts, we need to ensure that the result of the script
-                    // is the function if it is the last statement in the script.
-                    // For example, eval("function () {}") should return a
-                    // function, not undefined.
-                    if (!itsInFunctionFlag) {
-                        addIndexOp(Icode_CLOSURE_EXPR, fnIndex);
-                        stackChange(1);
-                        addIcode(Icode_POP_RESULT);
-                        stackChange(-1);
-                    }
                 }
                 break;
 
--- a/rhino/src/test/java/org/mozilla/javascript/EvalTest.java
+++ b/rhino/src/test/java/org/mozilla/javascript/EvalTest.java
@@ -1,0 +1,20 @@
+package org.mozilla.javascript;
+
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import org.junit.jupiter.api.Test;
+import org.mozilla.javascript.testutils.Utils;
+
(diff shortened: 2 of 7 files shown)


gbrail:
I'm in favor of fixing this, but yes, there is a lot of old code out there. What would you think of making this change dependent on language version, so that it would only kick in if VERSION >= VERSION_ES6? That would also mean that you don't need to change all those MozillaSuiteTests.

aardvark179:
I think there’s two aspects to this. One is the overall behavioural one, the other is that this is the only use of the `POP_RESULT` bytecode, and I think a difference in compiled and interpreted behaviour. If we do keep the behaviour then I think we should fix the code generator to stuff the result into a local variable and do a normal return at the end, and then maybe move that to the IR level.

andreabergia:
> I think there’s two aspects to this. One is the overall behavioural one, the other is that this is the only use of the `POP_RESULT` bytecode, and I think a difference in compiled and interpreted behaviour. If we do keep the behaviour then I think we should fix the code generator to stuff the result into a local variable and do a normal return at the end, and then maybe move that to the IR level.

You're incorrect, I think.
`POP_RESULT` is used _also_ for every statement expression, because they get wrapped in a `Token.EXPR_VOID` IR node. For example, `f(); g(); h();` will generate:

```
 [0] LINE : 1
 [3] REG_STR_C0 "f"
 [4] NAME_AND_THIS
 [5] REG_IND_C0
 [6] CALL 0
 [7] POP_RESULT
 [8] REG_STR_C1 "g"
 [9] NAME_AND_THIS
 [10] REG_IND_C0
 [11] CALL 0
 [12] POP_RESULT
 [13] REG_STR_C2 "h"
 [14] NAME_AND_THIS
 [15] REG_IND_C0
 [16] CALL 0
 [17] POP_RESULT
 [18] RETURN_RESULT
```

where you can see the three `POP_RESULT`, one after each call.

Also, my change affects _only_ in the interpreter and not the compiler because, for code that goes through `eval`, we force the usage of the interpreter.

aardvark179:
Okay, happy to tackle `POP_RESULT` separately from this.

andreabergia:
> I'm in favor of fixing this, but yes, there is a lot of old code out there. What would you think of making this change dependent on language version, so that it would only kick in if VERSION >= VERSION_ES6? That would also mean that you don't need to change all those MozillaSuiteTests.

Sounds reasonable to me. I've updated the PR.

gbrail:
Thanks!

Issue #2104, opened 2025-09-29
Incorrect scope handling in `LambdaFunction` and `LambdaConstructor`.


aardvark179:
Found this problem while cleaning up the property definition code.

The scope argument in `Callable#call` represents the **caller's scope**, not the callee's. So the interpreter and class compiler find the function's declaration scope and use that for lookup operations instead. Built in functions however do not do this. The actual Java code of the function does not have any information about its declaration scope, so the wrapping `LambdaFunction` needs to do the work instead, but does not currently do so. 
1. We pass some cross realm tests we should not. For example:
```javascript
/*---
esid: sec-symbol.keyfor
description: Global symbol registry is shared by all realms
info: |
    The GlobalSymbolRegistry is a List that is globally available. It is shared
    by all realms. Prior to the evaluation of any ECMAScript code it is
    initialized as a new empty List.
features: [cross-realm, Symbol]
---*/

var OSymbol = $262.createRealm().global.Symbol;
var parent = Symbol.for('parent');
var child = OSymbol.for('child');

assert.notSameValue(Symbol.keyFor, OSymbol.keyFor);
assert.sameValue(OSymbol.keyFor(parent), 'parent');
assert.sameValue(Symbol.keyFor(child), 'child');
```
is passing not because we have a cross realm symbol registry, but because `Symbol.for` is using the caller's scope, not the function declaration's scope, and so looking up in the same realm both times. Similarly if we move the creation of property descriptor JS objects to `Object.getOwnPropertyDescriptor` then we start failing some test 262 tests because we are, again, executing in the caller's realm.

2. We have some explicit tests that go against the spec, and I'm not sure if these are intentional, or aren't testing quite the right thing:
2.1. `DynamicScopeTest.standardMethodObjectCreate` asserts various things about the prototypes of scopes, and the parent scopes of objects. Neither of those makes sense if objects cease to have parent scopes, and scopes don't have prototypes, so may not be a real concern.
2.2. `Bug783797Test.thisProto` depends on either `String.prototype` being the same in all realms, or the callee scope not being set correctly. I think we're okay changing this test.
2.3. `NativeSetTest` was passing because it only checked that `this` printed as `[object Object]`. It should have failed as `this` should have been `globalThis`, and we print that as `[object Global]`, but it passed on other engines because they don't make that distinction. I've changed the test to check for `globalThis` explicitly, and it now passes on Rhino and other engines.
2.4. `PromiseBindTest` failed for the same reason.
[...]

PR #2105, opened 2025-09-30
Clean up symbol implementation in preparation for scope work.

aardvark179:
This change makes the global symbol registry actually global, improves the distinction between symbol values and objects, and updates the test 262 properties.

This is in preparation for fixing #2104.

### DIFF
--- a/tests/testsrc/test262.properties
+++ b/tests/testsrc/test262.properties
@@ -451,7 +451,7 @@ harness 23/116 (19.83%)
 
 built-ins/AggregateError 25/25 (100.0%)
 
-built-ins/Array 261/3077 (8.48%)
+built-ins/Array 259/3077 (8.42%)
     fromAsync 95/95 (100.0%)
     from/proto-from-ctor-realm.js
     length/define-own-prop-length-coercion-order-set.js
@@ -461,7 +461,6 @@ built-ins/Array 261/3077 (8.48%)
     prototype/concat/Array.prototype.concat_non-array.js
     prototype/concat/create-proto-from-ctor-realm-array.js
     prototype/concat/create-proxy.js
-    prototype/concat/is-concat-spreadable-is-array-proxy-revoked.js
     prototype/copyWithin/coerced-values-start-change-start.js
     prototype/copyWithin/coerced-values-start-change-target.js
     prototype/copyWithin/resizable-buffer.js {unsupported: [resizable-arraybuffer]}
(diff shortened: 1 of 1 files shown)


gbrail:
Do have a sense of why various order-of-evaluation tests now pass, but only in interpreted mode? Is that something that we can follow up on later?

aardvark179:
That is something I thought it was worth following up on later, we clearly haven’t made the situation worse in general, but there may still be something in OptRuntime that is doing an invalid short circuit.

gbrail:
So before we do this, making the global symbol registry "not really global" was something we did on purpose. We don't have the formal concept of a "realm" or "environment" in Rhino, so that's why it's attached to the top-level scope. If we do it this way, then it's possible that a symbol object constructed in one top-level scope ends up getting used by a script running in another one, and in the (perhaps unlikely) even that that symbol is associated with other state, I don't know what it does to GC and other things. If we think that we can take that risk then it's OK, but I do feel like we should be deliberate about it.

aardvark179:
Okay, checked the order of evaluation tests, and they were failing for different reasons in the compiler and interpreter, we've fixed the problem that made them fail in interprreted mode, but not the other problem wihich causes the failure in copied mode.

aardvark179:
> So before we do this, making the global symbol registry "not really global" was something we did on purpose. We don't have the formal concept of a "realm" or "environment" in Rhino, so that's why it's attached to the top-level scope. If we do it this way, then it's possible that a symbol object constructed in one top-level scope ends up getting used by a script running in another one, and in the (perhaps unlikely) even that that symbol is associated with other state, I don't know what it does to GC and other things. If we think that we can take that risk then it's OK, but I do feel like we should be deliberate about it.

I absolutely understand that concern, and it's why this had to be a proper clean up that made `SymbolKey`s (which are immutable and have no reference to a context or scope) the things we cache, and also why I made this a weak hash map to allow GC to happen.

As for realms… They actually fall out very neatly from the scope refactoring, so "we don't have realms" may not be an issue for long. 😀

And we already sort of have realms, because we can initialise standard objects on more than one unrelated scope, and pass objects around between them, and we have tests that this works. It's just that the concept hasn't been formalised, and we have to trace up through the scopes to find those standard objects.

aardvark179:
Ah, the byte code compiler has always miscompiled `a() | b()` as
1. Call a
2. Convert result to numeric.
3. Call b
4. Convert result to numeric.
5. Do or operation.

it should actually be 1, 3, 2, 4, 5.

That should definitely be fixed separately.

aardvark179:
I have a fix for all the order of evaluation tests in compiled mode, it’s a couple of SEAP byte codes for
Most things , and a slightly more tricky thing for unsigned right shift because it puts longs on the stack.

it has to go after this change, because symbols still need to be fixed for the tests to pass.

gbrail:
OK, I'm seeing the logic of changing how the global registry works and it is looking good, let me test locally but otherwise thanks and this looks good so far.

gbrail (approved):
This looks good to me now -- thanks! I'll leave this here for a short time in case anyone else wants to review, but not for long.

It makes me wonder whether we want to get a "ConcurrentWeakHashMap" class from some other open source project and adapt it to Rhino, since this isn't the first time something like that would be helpful.

aardvark179:
The hardest part would be handling the race between removing values from the map, alongside concurrent access. You can avoid that when equality of values is enough, but with symbols identity is the defining factor and that makes thing s much harder to make truly concurrent.

gbrail:
OK, this is looking good then. Thanks!

PR #2106, opened 2025-09-30
Fix: builtin methods should have no `prototype` property

andreabergia:
Small change to ensures that built-in methods such as `Math.sin` do not have a `prototype` property.

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/ScriptableObject.java
+++ b/rhino/src/main/java/org/mozilla/javascript/ScriptableObject.java
@@ -1412,13 +1412,10 @@ public abstract class ScriptableObject extends SlotMapOwner
         defineProperty(name, f, attributes);
     }
 
-	public void defineProperty(
-			Scriptable scope,
-			String name,
-			int length,
-			SerializableCallable target) {
-		defineProperty(scope, name, length, target, DONTENUM, DONTENUM | READONLY);
-	}
+    public void defineProperty(
+            Scriptable scope, String name, int length, SerializableCallable target) {
+        defineProperty(scope, name, length, target, DONTENUM, DONTENUM | READONLY);
+    }
 
     public void defineBuiltinProperty(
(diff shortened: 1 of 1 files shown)


aardvark179:
How many places is a `LambdaFunction` created that should have a prototype?  `LambdaConstructor` springs to mind, but where else. I wonder if it would be better to make no prototype the default and force the explicit passing of a prototype property in the remaining cases.

andreabergia:
> How many places is a `LambdaFunction` created that should have a prototype? `LambdaConstructor` springs to mind, but where else. I wonder if it would be better to make no prototype the default and force the explicit passing of a prototype property in the remaining cases.

I've done some more refactoring and changed the default. Seems a bit cleaner now. 
I've also rebased this against master.

aardvark179:
Sorry to suggest _another_ round here, but I think almost every use of the methods for defining built ins have the exact same `DONTENUM, DONTENUM | READONLY` arguments attributes. Could we maybe make those the default and remove all of that visual noise while we're at it? We made things like `NativeArray` more readable by defining utility methods which just set those parameters.

Also fine for that to be a followup PR.

aardvark179:
@andreabergia I made a branch on top of yours at https://github.com/aardvark179/rhino/tree/aardvark179-builtin-proto-property .

The reason I'm interested in us picking away at this API is that the more regular we can make it, and the more we can take things like scope and context out of it, the easier it will be to hoist this stuff out of global scope initialisation and push it to class initialisation. Then when we manage to get property maps and shapes implemented we'll have a really easy job to use them when initialising built in objects.

andreabergia:
> @andreabergia I made a branch on top of yours at https://github.com/aardvark179/rhino/tree/aardvark179-builtin-proto-property .

Looks good, I've merged all your stuff.

gbrail:
This looks good and it's a big one. Thanks!

PR #2108, opened 2025-10-03
Add smoke tests for debugger and class compiler.

aardvark179:
These are fairly simple tests to ensure that those parts of the system are at least working for simple cases.


rbri:
really like this small steps forward

gbrail:
Thanks!

PR #2111, opened 2025-10-04
Fix `rhino-all` shadow jar task to avoid stack overflow.

aardvark179:
@gbrail this should solve the issue with shadower 9.2.2, at least according to https://github.com/GradleUp/shadow/issues/1781. It appears to work locally for me.


gbrail:
This works for me, thanks!

PR #2113, opened 2025-10-06
Improve support for yield*

0xe:


rhino/src/main/java/org/mozilla/javascript/ES6Generator.java
```diff
@@ -199,7 +199,8 @@ private Scriptable resumeDelegeeReturn(Context cx, Scriptable scope, Object valu
         try {
             // Call "return" but don't throw if it can't be found
             Object retResult = callReturnOptionally(cx, scope, value);
-            // Per spec, treat undefined return value same as null
+            // Treat undefined return value same as null
+            // See https://tc39.es/ecma262/#sec-iteratorclose (7.4.11, 4b)
```
aardvark179: That is a somewhat confusing spec reference to give as the text of 4b is:
> b. If return is undefined, return ? completion.

The conversion of `null` to `undefined` is occurring only because 7.4.11.3 specifies that the abstract operation `GetMethod` is used, and *that* specifies that `undefined` is returned is the method is `null` or `undefined`. I think it's a good idea to join the dots here.

rhino/src/main/java/org/mozilla/javascript/Interpreter.java
```diff
@@ -1249,6 +1249,10 @@ public static Object resumeGenerator(
                 } catch (RuntimeException e) {
                     // Only propagate exceptions other than closingException
                     if (e != value) throw e;
+                    // Re-throw GeneratorClosedException so ES6Generator can catch and complete
+                    if (e instanceof NativeGenerator.GeneratorClosedException) {
```
rbri: maybe a separate catch block

tests/src/test/java/org/mozilla/javascript/tests/harmony/GeneratorsYieldStarReturnTest.java
```diff
@@ -0,0 +1,14 @@
...
+package org.mozilla.javascript.tests.harmony;
+
+import org.mozilla.javascript.Context;
+import org.mozilla.javascript.drivers.LanguageVersion;
+import org.mozilla.javascript.drivers.RhinoTest;
+import org.mozilla.javascript.drivers.ScriptTestsBase;
+
+@RhinoTest("testsrc/jstests/harmony/generators-yield-star-return.js")
+@LanguageVersion(Context.VERSION_ES6)
+public class GeneratorsYieldStarReturnTest extends ScriptTestsBase {}
```
rbri: personally i like the plain java test cases more - it is much easier to debug a single case and it is more obvious to point to the failing function. On the other hand the spotless re-formatting of the code is sometimes a pain.

rbri (approved):
Again great progress!

tests/src/test/java/org/mozilla/javascript/tests/harmony/GeneratorsYieldStarReturnTest.java
0xe: Makes sense, let me try to redo them as plain junit tests.

rhino/src/main/java/org/mozilla/javascript/ES6Generator.java
p-bakker: When referencing the EcmaScript spec, reference the specific version of the spec you're working off of, because chapter and section numbers within the spec can change from version to version

rhino/src/main/java/org/mozilla/javascript/ES6Generator.java
```diff
@@ -380,7 +385,9 @@ private Object callReturnOptionally(Context cx, Scriptable scope, Object value)
         // Delegate to "return" method. If it's not defined we ignore it
         Object retFnObj =
                 ScriptRuntime.getObjectPropNoWarn(delegee, ES6Iterator.RETURN_METHOD, cx, scope);
-        if (!Undefined.instance.equals(retFnObj)) {
+        // Treat a return method that's null or undefined as if it doesn't exist
+        // See https://tc39.es/ecma262/#sec-getmethod (2026, 7.3.10, 2)
+        if (retFnObj != null && !Undefined.instance.equals(retFnObj)) {
```
gbrail: Super minor, I know, but I think that we have an "Undefined.isUndefined" for this! It's important because we unfortunately use SCRIPTABLE_UNDEFINED in a few places.

rbri: :-D saw the same, but because the old version of the code does not use the Undefined.isUndefined i thought this was correct

gbrail:
Thanks!

"First world problem" here, but things are moving fast enough in this codebase that this PR and many others needs a rebase. Please check it out if you can, thanks!

0xe:
> Thanks!
> 
> "First world problem" here, but things are moving fast enough in this codebase that this PR and many others needs a rebase. Please check it out if you can, thanks!

Thanks, rebased.

gbrail:
Thanks for all the work on this -- looks good!

PR #2114, opened 2025-10-07
Fix the order of evaluation of various operations in compiled mode.

aardvark179:
Now all that symbols behave better we can resolve all the order of evaluation issues and get some more tests passing.


rbri (approved):

gbrail:
This looks great -- thanks!

PR #2115, opened 2025-10-09
Fix length with default arguments

0xe:
Spec says length should only include arguments up to the first default argument. This ensures we adhere to that.

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/ast/FunctionNode.java
+++ b/rhino/src/main/java/org/mozilla/javascript/ast/FunctionNode.java
@@ -526,10 +526,11 @@ public class FunctionNode extends ScriptNode {
                 // defaultParams stores pairs: [paramName (String), defaultValue (AstNode), ...]
                 // count up to the first parameter that has a default value
                 java.util.List<AstNode> params = fnNode.getParams();
+                String paramWithDefault = null;
                 if (params != null && !params.isEmpty()) {
                     for (int i = 0; i < defaultParams.size(); i += 2) {
                         if (defaultParams.get(i) instanceof String) {
-                            String paramWithDefault = (String) defaultParams.get(i);
+                            paramWithDefault = (String) defaultParams.get(i);
 
                             for (int paramIndex = 0; paramIndex < params.size(); paramIndex++) {
                                 AstNode param = params.get(paramIndex);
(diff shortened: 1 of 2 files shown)


rhino/src/main/java/org/mozilla/javascript/ast/FunctionNode.java
```diff
@@ -503,4 +503,56 @@ public void visit(NodeVisitor v) {
...
+                                AstNode param = params.get(paramIndex);
+                                if (param instanceof Name) {
+                                    String paramName = ((Name) param).getIdentifier();
+                                    if (paramName.equals(paramWithDefault)) {
+                                        arity = paramIndex;
+                                        break;
+                                    }
+                                }
+                            }
+                            if (arity != paramCount) break;
```
andreabergia, line 545: Shouldn't need this `if` - the only time you assign `arity` is at like 538, which is already followed by a break

rhino/src/main/java/org/mozilla/javascript/ast/FunctionNode.java
```diff
@@ -503,4 +503,56 @@ public void visit(NodeVisitor v) {
...
+            java.util.List<Object> defaultParams = fnNode.getDefaultParams();
+
+            if (defaultParams != null && !defaultParams.isEmpty()) {
+                // defaultParams stores pairs: [paramName (String), defaultValue (AstNode), ...]
+                // count up to the first parameter that has a default value
+                java.util.List<AstNode> params = fnNode.getParams();
+                if (params != null && !params.isEmpty()) {
+                    for (int i = 0; i < defaultParams.size(); i += 2) {
+                        if (defaultParams.get(i) instanceof String) {
+                            String paramWithDefault = (String) defaultParams.get(i);
```
andreabergia: Nit: I would just declare this before the `if` above

rbri (approved):

rhino/src/main/java/org/mozilla/javascript/ast/FunctionNode.java
0xe: We need this to break out when we encounter the first default parameter, no need to process the rest towards the count.

gbrail:
Thanks!

PR #2116, opened 2025-10-09
Separate idea of parent and declaration scopes, and fix scopes for built in lambda based functions.

### LINKED ISSUE #2104: Incorrect scope handling in `LambdaFunction` and `LambdaConstructor`.

aardvark179:
This fixes #2104 and will enable a follow PR to resolve issues with property descriptor creation.

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/Function.java
+++ b/rhino/src/main/java/org/mozilla/javascript/Function.java
@@ -46,6 +46,13 @@ public interface Function extends Scriptable, Callable, Constructable {
     @Override
     Scriptable construct(Context cx, Scriptable scope, Object[] args);
 
+    /**
+     * Return the scope in which this function was declared or closed over. This is the
+     * "[[Environment]]" defined in https://tc39.es/ecma262/#sec-ecmascript-function-objects.
+     * Although this is currently precisely equal to the parent scope (or "__parent__") of an object
+     * it is useful to distinguish the two concepts as parent scopes are no longer supported by the
+     * spec in general and present a significant barrier to future optimisations.
+     */
     default Scriptable getDeclarationScope() {
         return this.getParentScope();
     }


aardvark179:
Making this as a draft because I think there are things that I should tidy up in the byte code compiler, but haven't checked those yet.

aardvark179:
I've rebased this on top of master. The improved `PromiseBindTest` actually showed that I had changed the precise place where the promise queue was processed, so I have also fixed that.

I think there may be further lurking horrors around promises, but I don't think they are something we should fix until we've sorted out `this` handling everywhere and else and stomped down all the other problems round realms.

gbrail:
Thanks -- I'm all for cleaning up the weirdness and complexity about scopes. This is a pretty important and fundamental change, though -- can we put some documentation in the Function interface, and maybe even elsewhere, to explain to future users of Rhino what the difference is and what each type of scope means? Thanks!

aardvark179:
I've added a little Javadoc to `Function` describing what this method represents, and why we want to distinguish this from the general case of parent objects.

gbrail (commented):
OK -- I get what we're doing here and I am supportive of moving away from the confusing world of parent scopes for objects. I also am worried about the many projects that may have depended on some of this functionality over the years, since we had to change a few really old tests to make this work.

I still think that this is OK, but I am going to wait a day or so to see if anyone else has some strong opinions. Thanks!

gbrail:
Testing this locally just now, I get some test failures:

> Task :rhino:test

InterpreterBytecodeDumpTest > functionNamesArePrinted() FAILED
    org.opentest4j.AssertionFailedError at InterpreterBytecodeDumpTest.java:32

InterpreterBytecodeDumpTest > basicSmokeTest() FAILED
    org.opentest4j.AssertionFailedError at InterpreterBytecodeDumpTest.java:19

InterpreterBytecodeDumpTest > bigIntsArePrinted() FAILED
    org.opentest4j.AssertionFailedError at InterpreterBytecodeDumpTest.java:72

InterpreterBytecodeDumpTest > methodsArePrinted() FAILED
    org.opentest4j.AssertionFailedError at InterpreterBytecodeDumpTest.java:46

2772 tests completed, 4 failed, 3 skipped

This might be a newline-related problem, because I typically test on Windows.

Can you please take a look? Thanks!

gbrail:
Never mind, the current test failure in InterpreterBytecodeDumpTest is in master, because we don't have CI tests on Windows, so I'll fix that first...

gbrail:
Thanks!

PR #2118, opened 2025-10-10
remove a no longer used call and make an if statement looking not that depressive

rbri:
looks like an leftover from https://github.com/mozilla/rhino/pull/2091


gbrail:
Thanks! Looks good...

PR #2120, opened 2025-10-12
Fix Decycle plugin compatibility with Gradle 8.12.1

anivar:
Fixes CI Stream closed error by updating Decycle plugin to 0.11.0


gbrail:
Do you know where this alternate plugin comes from and whether it's really newer? Searching plugins.gradle.com for decycle, I get "de.obqo.decycle" version 1.2.5, which is also newer than what we're running now.  Are you sure this is the right plugin?

https://plugins.gradle.org/plugin/de.obqo.decycle

anivar:
Closing as this downgrades the plugin to an older version (0.11.0 < 1.2.3). If a decycle update is needed, should use the official plugin at version 1.2.5 as suggested by @gbrail.

PR #2121, opened 2025-10-12
fix for the parser to allow let in switch statement if the switch is the only statement in a for loop

rbri:
found by an HtmlUnit user https://github.com/HtmlUnit/htmlunit/issues/1040

    for (let i = 0; i < 1; i++)
        switch (i) {
            case 0:
                let test = 7;
                break;
    }

Currently the code above is not accepted by rhino but it is valid.

The fix is a bit of a hack but solves the problem without introducing much more pain.

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/Parser.java
+++ b/rhino/src/main/java/org/mozilla/javascript/Parser.java
@@ -1485,8 +1485,16 @@ public class Parser {
 
         AstNode discriminant = expr(false);
         pn.setExpression(discriminant);
-        enterSwitch(pn);
 
+        // a bit of a hack, but there is no separate scope for the switch statement
+        // therefore the check done in defineSymbol(int, java.lang.String, boolean)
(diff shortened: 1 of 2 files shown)


rbri:
@andreabergia interested in your oppinion on this...

andreabergia:
> @andreabergia interested in your oppinion on this...

Yeah I did look at this and I'm not a big fan; I wanna see if we can make the `Switch` ast node a proper `Scope`. I am trying a different approach, but got nerd sniped by https://github.com/mozilla/rhino/pull/2128 and https://github.com/mozilla/rhino/pull/2129 
I'll continue working on this next week.

rbri:
@andreabergia had also tried some other way and always stuck half of the way while refactoring this. Hope we can make this pice of code better step by step. Have a nice weekend...

rbri:
thanks @andreabergia for taking the time to make the real fix for that

PR #2122, opened 2025-10-13
replace <code> by {@code } as suggested by IntelliJ

rbri:


gbrail:
I'm trying to keep the build clean of errorprone warnings. Can you take a look at the one that came up (about a parameter name)? Thanks!

gbrail:
Thanks!

PR #2123, opened 2025-10-13
remove casts (as suggested by IntelliJ)

rbri:


gbrail:
OK, looks good, thanks!

PR #2125, opened 2025-10-16
Make `Symbol()` comply with the spec.

aardvark179:
Nothing in this

> 20.4.1.1 Symbol ( [ description ] )
> This function performs the following steps when called:
> 1. If NewTarget is not undefined, throw a TypeError exception.
> 2. If description is undefined, let descString be undefined.
> 3. Else, let descString be ? [ToString](https://tc39.es/ecma262/#sec-tostring)(description).
> 4. Return a new Symbol whose [[Description]] is descString.

mentions a second argument, and we have no tests for a second argument, so I have removed the handling of the second argument so we now do what the spec says. We set `descString` to `null` instead of `undefined` because that is what `SymbolKey` uses internally to represent no string.


rbri:
@aardvark179 any chance to have at least one test for this?

aardvark179:
Sorry. I had edited `SymbolTest` to add a case where the second argument should be ignored, but then failed to actually commit it.

gbrail:
OK, this one looks good too -- thanks!

PR #2128, opened 2025-10-17
Fixed exception in the interpreter when using many literal bigint

andreabergia:
The interpreter has various special icode for big int literals, but the ones to access the 5th onward literal were broken. I've fixed it and added a test.


gbrail:
Great -- thanks!

PR #2129, opened 2025-10-17
Fix interpreter bytecode printing

andreabergia:
When we enable the flag `Token.printICode`, the code generator prints all the bytecode it generates. This was broken after the refactoring done in https://github.com/mozilla/rhino/pull/2091.

In this PR I have fixed the bugs added after that refactoring and I've added some smoke tests for the feature, as there were none.

This is stacked on top of https://github.com/mozilla/rhino/pull/2128

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/Interpreter.java
+++ b/rhino/src/main/java/org/mozilla/javascript/Interpreter.java
@@ -898,12 +898,17 @@ public final class Interpreter extends Icode implements Evaluator {
                     out.println(tname + " " + indexReg);
                     ++pc;
                     break;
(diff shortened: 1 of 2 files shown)


aardvark179:
Now that instructions are classes it might be worth making the printing an abstract method that takes the byte as an argument (because more than one instruction can share a class), that way we will be less likely to miss this sort of thing in future.

andreabergia:
@aardvark179 agreed, we can improve on that. We also have some other changes in our fork that I've used to build a small web UI that we should upstream. But one thing at a time 😉

gbrail:
Thanks! Looks good!

PR #2132, opened 2025-10-18
Replace the NumberToString implementation

gbrail:
Number-to-string conversions in Rhino have historically been made using the DtoA class, which is based on 20-year-old reference C code. A while back we added support for a faster algorithm (the Grisu algorithm).

This change switches to the Schubfach algorithm, which is a relatively new algorithm used in many standard libraries including Java. We use code by Rafaello Guilietti that has also been adapted for Jackson and for the JDK.

This new algorithm is much simpler and a bit faster. It also works for all inputs, including some that Rhino has previously not been able to render correctly.

This is the first of a few steps -- the next step will address Number.toExponential, toFixed, and toPrecision, and then we will see what is left.


rbri:
@gbrail fine for me to switch to this code (same for your related PR)

PR #2133, opened 2025-10-19
Introduce `NativeJavaField` for Java Generic support for field

ZZZank:
# Java Generic support

- Type Consolidation (Stage 1 in #2023 , stage 2 in #2080 )
- Generic support for Method/Constructor/List/Map/Array (#2080 )
- Generic support for Field (This PR)

## Changes in this PR

- Added `NativeJavaField` to contain type information and apply type consolidation when required.
- Trying to set value of a final field will now always be skipped, instead of silently swallowing `IllegalAccessException` after failure.
- Some small cleanup for reflecting fields in `JavaMembers`
- Some previously disabled tests are enabled


gbrail:
Thanks -- good progress on these features!

PR #2134, opened 2025-10-20
first step of migrating the Arguments impl to lambda and fix some issue

rbri:
* a bit of cleanup for the Arguments implementation itself (preparation for the migration to lambda that will be another pr)
* fix a bunch of issues with ThrowTypeError
* add some tests for ThrowTypeError

The 'new' test262 problem ThrowTypeError/distinct-cross-realm.js is something that worked so far by chance. With the fixes it now fails because we have no real realm and the ThrowTypeError is reused.

Fun fact, the test itself is also wrong - i will open an issue aganst the test262 project later.

Started this effort because the HtmlUnit brach still has some patches for arguments handling, and i like to get rid of them. Looks like the Rhino impl is already good and only minor things required to make some progress here. Already working on the next PR, but I like to have this first merged to make the steps and the impact clear.

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/Arguments.java
+++ b/rhino/src/main/java/org/mozilla/javascript/Arguments.java
@@ -17,7 +17,27 @@ package org.mozilla.javascript;
 final class Arguments extends IdScriptableObject {
     private static final long serialVersionUID = 4275508002492040609L;
 
-    private static final String FTAG = "Arguments";
+    private static final String CLASS_NAME = "Arguments";
+
+    // Fields to hold caller, callee and length properties,
+    // where NOT_FOUND value tags deleted properties.
+    // In addition if callerObj == NULL_VALUE, it tags null for scripts, as
+    // initial callerObj == null means access to caller arguments available
+    // only in JS <= 1.3 scripts
+    private Object callerObj;
+    private Object calleeObj;
+    private Object lengthObj;
+
+    private int callerAttr = DONTENUM;
+    private int calleeAttr = DONTENUM;
+    private int lengthAttr = DONTENUM;
+
+    private NativeCall activation;
+
+    // Initially args holds activation.getOriginalArgs(), but any modification
+    // of its elements triggers creation of a copy. If its element holds NOT_FOUND,
+    // it indicates deleted index, in which case super class is queried.
+    private Object[] args;
 
     public Arguments(NativeCall activation) {
         this.activation = activation;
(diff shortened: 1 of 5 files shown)


aardvark179:
> Fun fact, the test itself is also wrong - i will open an issue aganst the test262 project later.

Could you expand on that? Looking at the test I agree it is only asserting the negative, but that suggests there is a missing test case, rather than a bug in that test.

I think you could start storing `ThrowTypeError` objects on `TopLevel` and they should be findable via the activation scope, I think that scope hierarchy is now set correctly.

rbri:
@aardvark179 i think the second function has to be strict to produce a ThrowTypeError.

aardvark179:
Ah, yes. I guess it should something like
```javascript
var other = $262.createRealm().global;
var localArgs = function() {
  "use strict";
  return arguments;
}();
var otherArgs = (new other.Function('"use strict"; return arguments;'))();
var otherArgs2 = (new other.Function('"use strict"; return arguments;'))();
var localThrowTypeError = Object.getOwnPropertyDescriptor(localArgs, "callee").get;
var otherThrowTypeError = Object.getOwnPropertyDescriptor(otherArgs, "callee").get;
var otherThrowTypeError2 = Object.getOwnPropertyDescriptor(otherArgs, "callee").get;

var localError;
try {
    localError = localThrowTypeError();
} catch (e) {
    localError = e;
}

var otherError;
try {
    otherError = otherThrowTypeError();
} catch (e) {
    otherError = e;
}

assert.sameValue(true, Error.isError(localError));
assert.sameValue(true, Error.isError(otherError));

assert.notSameValue(localThrowTypeError, otherThrowTypeError);
assert.sameValue(otherThrowTypeError, otherThrowTypeError2);
```

Let me check that in engine 252… Okay, that was more painful than I expected.

rbri:
Hi @aardvark179,
It appears you have really dug yourself deep into this rabbit hole. Perhaps you can open up the issue/PR base based on your detailed work.
And maybe it makes sense to add a simple name check after the creation of all that type errors to make sure we really have a ThrowTypeError at hand.

aardvark179:
Unfortunately `%ThrowTypeError% is defined as follows:
> This function is the %ThrowTypeError% intrinsic object.
> It is an anonymous built-in [function object](https://tc39.es/ecma262/#function-object) that is defined once for each [realm](https://tc39.es/ecma262/#realm).
> It performs the following steps when called:
> 1. Throw a TypeError exception.
>
> The value of the [[Extensible]] internal slot of this function is false.
> The "length" property of this function has the attributes { [[Writable]]: false, [[Enumerable]]: false, [[Configurable]]: false }.
> The "name" property of this function has the attributes { [[Writable]]: false, [[Enumerable]]: false, [[Configurable]]: false }.

So, the function is anonymous (meaning we can't check its name), and it is only defined as throwing a TypeError (so there is nothing much we can check about the precise nature of the error thrown.

I'll do a PR for test262.

rbri:
> So, the function is anonymous (meaning we can't check its name)

yes you are right, i was confused because there is this name property check (name.js) doing thins

    var ThrowTypeError = Object.getOwnPropertyDescriptor(function() {
      "use strict";
      return arguments;
    }(), "callee").get;

    verifyProperty(ThrowTypeError, "name", {
      value: "", writable: false, enumerable: false, configurable: false
    });

gbrail:
Thanks -- I think this is a good step forward and I'd like to move ahead with it.

rbri:
@gbrail thanks for merging, next update in this field is on the way

PR #2135, opened 2025-10-20
Make `switch` a proper `Scope`

andreabergia:
Alternative approach to https://github.com/mozilla/rhino/pull/2121 for fixing some problems we have with `let` inside of `switch`.

I still have to investigate the test262 regression, but I wanted @rbri to have a look at this first.

### DIFF
--- a/tests/src/test/java/org/mozilla/javascript/tests/es5/LetTest.java
+++ b/tests/src/test/java/org/mozilla/javascript/tests/es5/LetTest.java
@@ -1,3 +1,6 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 package org.mozilla.javascript.tests.es5;
 
 import org.junit.jupiter.api.Test;


rhino/src/main/java/org/mozilla/javascript/Parser.java
```diff
@@ -1481,81 +1481,92 @@ private SwitchStatement switchStatement() throws IOException {
...
+        pushScope(pn);
         try {
-            if (mustMatchToken(Token.RP, "msg.no.paren.after.switch", true))
-                pn.setRp(ts.tokenBeg - pos);
+            if (mustMatchToken(Token.LP, "msg.no.paren.switch", true)) pn.setLp(ts.tokenBeg - pos);
+
+            AstNode discriminant = expr(false);
+            pn.setExpression(discriminant);
+
+            // a bit of a hack, but there is no separate scope for the switch statement
```
rbri: the comment can be removed because you approach is no longer a hack

andreabergia: Done.

rbri:
@andreabergia i guess you are much more familar with the code generation stuff - for me your approach looks like the right way but i have no real idea what the implications of switching from jump to scope are. Scope extends Jump, i think this might add exactly what we miss for switch statements.

and maybe a test for code like this should also be added

    var sum = 0;
      let test = 0;
      switch (test) {
        case 0:
          let test = 7;
          sum += test;
          break;
        }
    console.log(sum)

tests/testsrc/tests/js1_7/block/regress-411279.js
```diff
@@ -50,15 +50,18 @@ function test()
 
   compareSource(expect, actual, summary);
 
-  expect = '1 undefined';
-  actual = f(1);
+  // This should throw an error because of the TDZ, but it throws a ReferenceError for the moment
```
rbri: TDZ?

andreabergia:
> @andreabergia i guess you are much more familar with the code generation stuff - for me your approach looks like the right way but i have no real idea what the implications of switching from jump to scope are. Scope extends Jump, i think this might add exactly what we miss for switch statements.

Yeah it looks correct and coherent with things like loops.

> and maybe a test for code like this should also be added
> 
> ```
> var sum = 0;
>   let test = 0;
>   switch (test) {
>     case 0:
>       let test = 7;
>       sum += test;
>       break;
>     }
> console.log(sum)
> ```

Added, it passes.

tests/testsrc/tests/js1_7/block/regress-411279.js
andreabergia: Temporal Dead Zone. I've improved the comment.

andreabergia:
I changed slightly the code - now there are no regression on `MozillaSuiteTest`, we have fixed a couple of test262 cases, and we pass all the tests written by @rbri. It's not perfect, but we have _a lot_ of things broken around let/const/lexical scopes. This is a small step in the right direction, IMHO.

rbri:
> small step in the right direction, IMHO

@andreabergia not sooo small i think, thanks for taking care of this

gbrail:
I agree that this is good progress and thanks!

PR #2142, opened 2025-10-26
further improve 'arguments' support

rbri:
Use the existing ScriptRuntime.typeErrorThrower(cx) instead of having a separate ThrowTypeError in Arguments.

This also fixes the setup of the ThrowTypeError returned by ScriptRuntime.typeErrorThrower(cx).

### DIFF
--- a/tests/src/test/java/org/mozilla/javascript/tests/es6/ArgumentsTest.java
+++ b/tests/src/test/java/org/mozilla/javascript/tests/es6/ArgumentsTest.java
@@ -346,6 +346,26 @@ public class ArgumentsTest {
     }
 
     @Test
+    public void descWhenAccessedThroughFunction() {
+        String code =
+                "let res = '';\n"
+                        + "function test() {\n"
+                        + "  let desc = Object.getOwnPropertyDescriptor(test.arguments, '0');\n"
+                        + "  res += typeof desc.get + '/' + desc.get;\n"
+                        + "  res += ' ' + typeof desc.set + '/' + desc.set;\n"
+                        + "  res += ' W-' + desc.writable;\n"
+                        + "  res += ' C-' + desc.configurable;\n"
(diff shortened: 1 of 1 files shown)


aardvark179:
This looks good, but it feels like it's only partway there, but not quite. If you're altering `ScriptRuntime.typeErrorThrower()` _anyway_, then let's do it properly:
1. `%TypeErrorThrower%` is per realm, so add it to the list of things on `TopLevel` (effectively our realm object).
2. Change `ScriptRuntime.typeError` to accept a scope (the no args version can use `cx.topCallScope`), and get it from the `TopLevel` object.

I am also not quite clear on why we need another subclass of `BaseError` rather than just a `LambdaFunction`, but maybe I'm missing something.

Error handling feels like an area we should come up with a plan to tidy up. At the moment we throw internal errors and record their type based on a name, but all those constructors are per realm, so we should probably come up with a plan to fix that. Since the `ScriptRuntime` functions are used everywhere we'd better also do this in a way that allows us to introduces any new scope argument, or whatever, gradually, otherwise we'll have a really PR.

rbri:
> This looks good, but it feels like it's only partway there, but not quite.

@aardvark179 hehe it is a draft... ;-)

will try to finish my changes first, and then i will have a look at your suggestions. As in many other areas we have to go a long way to reach the spec

gbrail:
This looks good to me -- thanks. I'm idly curious whether the "read only arguments" behavior could have been implemented by making the object's properties non-configurable and making the object non-extensible. Are there other JS objects that are supposed to silently ignore attempts to set properties like this one, or is arguments special?

rbri:
> I'm idly curious whether the "read only arguments" behavior could have been implemented by making the object's properties non-configurable

at least the property descriptor shows that they are configurable (see the last added test)

gbrail:
Thanks for the progress on this!

PR #2143, opened 2025-10-26
Refactor internal property descriptors to not be `ScriptableObjects`.

aardvark179:
More preparation for separating the concepts of scopes and scriptable objects.

Property descriptors as described in https://tc39.es/ecma262/#sec-property-descriptor-specification-type are not JavaScript objects, but merely internal Record objects with a particular set of fields. Meanwhile [FromPropertyDescriptor](https://tc39.es/ecma262/#sec-frompropertydescriptor) creates a corresponding JavaScript object and sets a prototype based on the current realm. 

While prototyping separating scopes and realms, and reducing where we relied on parent scopes it became clear that our conflation of these two objects made it impossible to get some aspects of the spec correct without altering many of our APIs to pass a scope through. We had already introduced a `DescriptorInfo` object to avoid potential deadlocks with slot map operations, so this PR expands its use to pretty much everywhere where the specs specifies a property descriptor.

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/ScriptableObject.java
+++ b/rhino/src/main/java/org/mozilla/javascript/ScriptableObject.java
@@ -1632,10 +1632,10 @@ public abstract class ScriptableObject extends SlotMapOwner
         try (var map = props.startCompoundOp(false)) {
             ids = props.getIds(map, false, true);
         }
-        ScriptableObject[] descs = new ScriptableObject[ids.length];
+        DescriptorInfo[] descs = new DescriptorInfo[ids.length];
         for (int i = 0, len = ids.length; i < len; ++i) {
             Object descObj = ScriptRuntime.getObjectElem(props, ids[i], cx);
-            ScriptableObject desc = ensureScriptableObject(descObj);
+            var desc = new DescriptorInfo(ensureScriptableObject(descObj));
             checkPropertyDefinition(desc);
             descs[i] = desc;
         }
@@ -1727,7 +1727,7 @@ public abstract class ScriptableObject extends SlotMapOwner
         public Object value = NOT_FOUND;
         boolean accessorDescriptor;
 
-        DescriptorInfo(ScriptableObject desc) {
+        public DescriptorInfo(ScriptableObject desc) {
             enumerable = getProperty(desc, "enumerable");
             writable = getProperty(desc, "writable");
             configurable = getProperty(desc, "configurable");
@@ -1737,7 +1737,18 @@ public abstract class ScriptableObject extends SlotMapOwner
             accessorDescriptor = getter != NOT_FOUND || setter != NOT_FOUND;
         }
 
-        DescriptorInfo(
+        public DescriptorInfo(
+                boolean enumerable, boolean writable, boolean configurable, Object value) {
+            this.enumerable = enumerable;
+            this.writable = writable;
+            this.configurable = configurable;
+            this.getter = NOT_FOUND;
+            this.setter = NOT_FOUND;
+            this.value = value;
+            accessorDescriptor = false;
+        }
+
+        public DescriptorInfo(
                 Object enumerable,
                 Object writable,
                 Object configurable,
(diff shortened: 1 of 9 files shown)


rhino/src/main/java/org/mozilla/javascript/ScriptableObject.java
```diff
@@ -1725,12 +1719,12 @@ protected boolean defineOwnProperty(
     }
 
     public static final class DescriptorInfo {
-        Object enumerable;
```
gbrail, line 1728: Wouldn't this code be cleaner overall if DescriptorInfo had accessors that validated the whole NOT_FOUND bit (isEnumerable, isConfigurable, etc)?

rhino/src/main/java/org/mozilla/javascript/ScriptableObject.java
```diff
@@ -1742,6 +1736,43 @@ public static final class DescriptorInfo {
...
         }
+
+        DescriptorInfo(
+                Object enumerable,
+                Object writable,
+                Object configurable,
+                Object getter,
+                Object setter,
+                Object value) {
+            this.enumerable = enumerable;
```
gbrail, line 1758: I know it makes the code bigger but would you consider a "builder" pattern for this? I worry about being able to read the code when every DescriptorInfo constructor has six parameters of the same type in a row.

rhino/src/main/java/org/mozilla/javascript/ScriptableObject.java
```diff
@@ -2003,6 +2034,20 @@ protected void checkPropertyDefinition(ScriptableObject desc) {
         }
     }
 
+    protected void checkPropertyDefinition(DescriptorInfo desc) {
```
gbrail: Should this be a member of the DescriptorInfo class?

rhino/src/main/java/org/mozilla/javascript/ScriptableObject.java
```diff
@@ -2142,6 +2187,10 @@ protected static boolean isAccessorDescriptor(ScriptableObject desc) {
         return hasProperty(desc, "get") || hasProperty(desc, "set");
     }
 
+    protected static boolean isAccessorDescriptor(DescriptorInfo desc) {
```
gbrail, line 2257: Should this be a member of the DescriptorInfo class?

rhino/src/main/java/org/mozilla/javascript/ScriptableObject.java
```diff
@@ -2152,6 +2201,10 @@ protected static boolean isGenericDescriptor(ScriptableObject desc) {
         return !isDataDescriptor(desc) && !isAccessorDescriptor(desc);
     }
 
+    protected static boolean isGenericDescriptor(DescriptorInfo desc) {
```
gbrail, line 2271: Member of the class -- it's a static function that takes one argument...

rhino/src/main/java/org/mozilla/javascript/typedarrays/NativeTypedArrayView.java
```diff
@@ -153,19 +153,19 @@ protected boolean defineOwnProperty(
                     return false;
                 }
 
-                if (Boolean.FALSE.equals(getProperty(desc, "configurable"))) {
+                if (Boolean.FALSE.equals(desc.configurable)) {
                     return false;
                 }
-                if (Boolean.FALSE.equals(getProperty(desc, "enumerable"))) {
+                if (Boolean.FALSE.equals(desc.enumerable)) {
```
gbrail: Would this code (and many places like it) be easier to read and follow if it could call "!descriptor.isEnumerable()" or if NOT_FOUND is a special third case, "descriptor.isNotEnumerable()"

aardvark179:
> Thanks! I can see how this simplifies things. But I can't help but think that some handy member functions on DescriptorInfo would reduce the need for all that "Boolean.TRUE.equals" code all over.

I agree. I held off doing that initially because an extremely mechanical change was slightly easier to produce, but it is definitely more readable with a better API. After a little experimenting I'm erring towards something like:
```java
boolean isEnumerable() // check for a simple value of true
boolean isEnumerable(boolean value) // check for an explicitly set value that matches
boolean hasEnumerable() // check that a value has been set
```

The reason for that is that there are a lot of cases where we care that an explicit value of `FALSE` has been set, and writing
```java
desc.hasEnumerable() && !desc.isEnumerable()
```
is asking for bugs in the operator precedence.

gbrail:
That makes sense -- I had been thinking that we'd need "isEnumerable" and "isNotEnumerable" but I think that your way is more Java-like!

gbrail:
Thanks -- looks good!

rbri:
@aardvark179 migrated into core-js/htmlunit, works without problems (there was a small need to fix one overwrite at my site). great - thanks

Issue #1058, opened 2021-10-14
Support Java Interop API used by Graal (and Nashorn)


p-bakker:
Both Nashorn and GraalVM support options to make migrating from Rhino easier. 

I'd say we should go the other way around: add at least the their Java Interop api (optionally, so behind a flag), so code authored for Graal/Nashorn can be easily ran on Rhino as well :-)

Nice to have though

long76:
yeah, i can't override Nashorn in my Java 8 project because `ReferenceError: "Java" is not defined. (eval#1) in eval at line number 1 at column number 0` (`var MyClass = Java.type("com.example.MyClass");`)

p-bakker:
@long76 Feel like providing a PR for this?

tonygermano:
@szegedi provided a useful link for the Java Global Object API in https://github.com/mozilla/rhino/discussions/972#discussioncomment-1481759

Does anyone know if it Is possible to pass an alternative java ClassLoader to `Java.type(obj)`? I never used Nashorn.

p-bakker:
You mean if that's possible in Nashorn/Graal? Maybe @szegedi knows, as per there mentioned comment he authored the API 🙂

p-bakker:
While adding Javaa.type(...) to aid migration from Nashorn/GraalJS to Rhino, I now think it's implementation could maybe be a small, thin wrapper around the ESM import statement

p-bakker:
@tonygermano could you maybe elaborate on your 'alternative java ClassLoader' comment/requirement? 

Does the current importPackage/Class API in Rhino offer this option?

tonygermano:
I believe it's implemented here
https://github.com/mozilla/rhino/blob/04d54698afd5495d1312daa869a6192898f99ed7/rhino/src/main/java/org/mozilla/javascript/NativeJavaTopPackage.java#L46

```javascript
$ rlwrap java -jar rhino-1.7.13.jar 
Rhino 1.7.13 2020 09 02
js> var url = java.nio.file.Path.of('commons-lang3-3.9.jar').toUri().toURL()
js> var cl = new java.net.URLClassLoader([url])
js> var myPackages = new Packages(cl)
js> Packages.org.apache.commons.lang3.StringUtils
[JavaPackage org.apache.commons.lang3.StringUtils]
js> myPackages.org.apache.commons.lang3.StringUtils
[JavaClass org.apache.commons.lang3.StringUtils]
```

PR #2155, opened 2025-11-07
Convert `TopLevel` from `IdScriptableObject` to being lambda based.

aardvark179:
This is another PR in the long road to separating scopes and scriptable objects, but it's a draft for now because it cast light on a really tricky corner of Rhino's API around `ImporterTopLevel`, and I don't have a good answer for how to square this circle.

`ImporterTopLevel` stores additional information against scopes when Java packages are imported, and then uses that to resolve class names at a later time. If two packages are imported with identically named classes then when those classes are looked up this will raise an error. As a top level scope, or an explicitly created scope this is fine if a little tricky, but it also allows for extremely sneaky additional behaviour as tested for in `SealedSharedScopeTest` which sets up scopes as follows:

```java
        try (Context tmpCtx = Context.enter()) {
            sharedScope = new ImporterTopLevel(tmpCtx, true);
            sharedScope.sealObject();
        }

        ctx = Context.enter();
        ctx.setLanguageVersion(Context.VERSION_DEFAULT);
        scope1 = ctx.newObject(sharedScope);
        scope1.setPrototype(sharedScope);
        scope1.setParentScope(null);
        scope2 = ctx.newObject(sharedScope);
        scope2.setPrototype(sharedScope);
        scope2.setParentScope(null);
```

and these are then tested like this:

```java
        evaluateString(scope1, "importPackage(javax.naming);");
        evaluateString(scope2, "importPackage(javax.xml.soap);");
        o = evaluateString(scope1, "Name");
```

For this to work `importPackage()` has to execute with the caller's scope, not it's own declaration scope, in order to store data against `scope1` or `scope2` rather than the sealed `sharedScope`, _and_ it requires `scope1` and `scope2` to have the `sharedScope` set as their prototype because only prototype chain look ups pass in the target as the starting scope.

This test / API does not survive the separation of scopes and objects (scopes no longer have prototypes), and also either requires special handling of scopes, or special handling of `this`. It can _barely_ even survive the change to lambdas.

Looking through the users of `ImporterTopLevel` on GH I think almost all of them are just using it as a top level scope and are not relying on the creating child scopes, at least not explicitly,  but I thought this warranted discussion before breaking an API, especially when we have tests for it, and we even have had bug reports against (#1463).

I wonder if we should deprecate `ImporterTopLevel` except as a top level scope, (maybe even as that) and strongly push people to using `Pacakges` for this sort of thing like
```javascript
var {Integer, String:JString} = Packages.java.lang;
```

@gbrail , @rPraml , @rbri, and @p-bakker I'm guessing you'll all have opinions on this. I've checked our internal GHE and we don't use `importClass` and or `importPackage` anywhere outside a few tests as far as I can tell.

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/ImporterTopLevel.java
+++ b/rhino/src/main/java/org/mozilla/javascript/ImporterTopLevel.java
@@ -47,8 +47,6 @@ import java.util.ArrayList;
 public class ImporterTopLevel extends TopLevel {
     private static final long serialVersionUID = -9095380847465315412L;
 
-    private static final Object IMPORTER_TAG = "Importer";
-
     public ImporterTopLevel() {}
 
     public ImporterTopLevel(Context cx) {
@@ -65,8 +63,28 @@ public class ImporterTopLevel extends TopLevel {
     }
 
     public static void init(Context cx, Scriptable scope, boolean sealed) {
-        ImporterTopLevel obj = new ImporterTopLevel();
-        obj.exportAsJSClass(MAX_PROTOTYPE_ID, scope, sealed);
+        init(cx, scope, sealed, false);
+    }
+
+    public static void init(Context cx, Scriptable scope, boolean sealed, boolean isTopScope) {
+        LambdaConstructor ctor =
+                new LambdaConstructor(scope, "ImporterTopLevel", 0, ImporterTopLevel::js_construct);
+
+        ctor.definePrototypeMethod(scope, "importClass", 1, ImporterTopLevel::js_importClass);
+        ctor.definePrototypeMethod(scope, "importPackage", 1, ImporterTopLevel::js_importPackage);
+
+        if (sealed) {
+            ctor.sealObject();
+        }
+
+        var proto = (Scriptable) ctor.getPrototypeProperty();
+
+        if (isTopScope) {
+            scope.put("importClass", scope, proto.get("importClass", proto));
+            scope.put("importPackage", scope, proto.get("importPackage", proto));
+        }
+
+        ScriptableObject.defineProperty(scope, "JavaImporter", ctor, DONTENUM);
     }
 
     public void initStandardObjects(Context cx, boolean sealed) {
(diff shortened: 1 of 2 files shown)


aardvark179:
I've done some checks on what can be done with `Packages` and we might need to fix some bugs there. I should be able to translate
```javascript
var swingNames = JavaImporter();

swingNames.importPackage(Packages.javax.swing);
swingNames.importPackage(Packages.java.awt);
swingNames.importPackage(Packages.java.awt.event);
```
into something like
```javascript
var swingNames = Packages.javax.swing;
var awtNames = Packages.java.awt;
var eventNames = Packages.java.awt.event;

Object.setPrototypeOf(awtNames, eventNames);
Object.setPrototypeOf(swingNames, awtNames);
```
This mostly works, but there are some rough edges like we report the wrong names in when printing the retrieved classes.

Edit: Oh wait, the prototype thing doesn't work as expected at all.

p-bakker:
Just throwing some extra info into the mix: while refactoring should keep things backwards compatible if possible, the importClass/ImportPackages/Packages api has it's downsides. Graal and Nashorn took another approach, see #1058

aardvark179:
How about I try and keep this working for this refactor, but we deprecate the methods with a note that in 2.0 they will throw an error when called. If we then do a release before the main scope separation work lands then that would give people time to move to the package based approach.

We could also work to introduce a Nashorn/GraalJS compatible way of doing it.

rbri:
HtmlUnit / core-js does not use ImporterTopLevel at all - fine with everything

aardvark179:
Okay, this should now all work. I haven't added deprecated annotations because, although the way to use a shared sealed parent scope and child "scopes" will need to change, I think the spirit of it can survive. At the moment, as I said above, setup is done like this:
```java
        try (Context tmpCtx = Context.enter()) {
            sharedScope = new ImporterTopLevel(tmpCtx, true);
            sharedScope.sealObject();
        }

        ctx = Context.enter();
        ctx.setLanguageVersion(Context.VERSION_DEFAULT);
        scope1 = ctx.newObject(sharedScope);
        scope1.setPrototype(sharedScope);
        scope1.setParentScope(null);
        scope2 = ctx.newObject(sharedScope);
        scope2.setPrototype(sharedScope);
        scope2.setParentScope(null);
```
This won't survive the separation of scopes and scriptable, but if we can separate out the top level scope (a realm) and the global object then I think this can become something like…
```java
        try (Context tmpCtx = Context.enter()) {
            sharedScope = new ImporterTopLevel(tmpCtx, true);
            sharedScope.sealObject();
        }

        ctx = Context.enter();
        ctx.setLanguageVersion(Context.VERSION_DEFAULT);
        scope1 = new TopLevel();
        scope1.getGlobalThis().setPrototype(sharedScope.getGlobalThis());
        scope2 = new TopLevel();
        scope2.getGlobalThis().setPrototype(sharedScope.getGlobalThis());
```

gbrail:
TBH I am not sure why people use TopLevel directly, although importClass and importPackage are in a lot of places. I do know that throughout history people have done just about everything however.

This particular fix is now pretty targeted though, and works fine for me, so I'm going to merge it. Thanks!

PR #2158, opened 2025-11-13
add tests for hasOwnProperty with Symbol and wrappend Symbol parameter

rbri:


gbrail:
Looks good to me too, thanks!

PR #2161, opened 2025-11-17
Separate top level scope and `globalThis`.

aardvark179:
Another draft PR on the road to separating scopes and scriptable objects (#2163). I've put the rationale for doing this in a separate issue(#2164) 

The global scope in EcmaScript is a a little subtle in that it consists of a declaration environment record and an object environment record for the global object (`globalThis`). The global object contains variable declarations, function declarations, generator function declarations, async function declarations, and async generator function declarations. Everything else (`const` declarations, `let` declarations, `class` declarations, etc.) goes on declaration environment and is *not* added to `globalThis`. This change could make that full separation, but currently there is one test in `MozillaSuiteTest` which depends on a `const` declaration being added to the global `this`. I wonder if we should gate that on a language version.

This is working well apart from a couple of edge cases that I think it's worth talking through:

1. `ImporterTopLevel` is mostly working, but I don't think I have a great model of how it should work. Currently The behaviour for package lookup is living on the scope half, but I'm wondering if `ImporterTopLevel` should really put all the behaviour on a custom `GlobalThis`.
2. One doc test is using a java method to get the XML implementation associated with the current global scope. Currently that is not part of `globalThis`, but maybe like java imports it should live on that half.
3. Module scopes. These have some existing awkward special handling on `FunctionObject` (which is currently the only way the existing `customGlobal` test works). The existing `ModuleScope` model also doesn't match `require` in node or common-js.

Let me expand on that last point slightly. If I have a file `blah.js` containing
```javascript
function thing() { return this; }
exports.thing = thing;
exports.thing2 = thing();
```
and from node REPL I do
```javascript
e = require("./blah.js");
thing = e.thing
thing2 = e.thing2
function thing3() { return this; }
thing4 = thing3();
```

Then we find `thing2 === globalThis` and `thing2 === thing4`. `require` simply introduces a new scope (just like a function does) rather than what we do with a module scope (a new top level scope that happens to have another top level scope as the prototype). What we have doesn't match that, but it also doesn't match the behaviour of `import` which should have an entirely separate realm (top level scope).

I think I can see two possible ways to resolve this:
1. If you want custom function provided by a java class on your scope then that java object has to be your `globalThis`, and the module scope becomes a normal child scope. This would be closest to node as far as I can tell. We might well want to keep the `ModuleScope` as a special class to mark that we are in a module for relative requires etc.
2. We keep module scope as a top level scope, but with a `globalThis` of whatever custom type provided. To avoid mutating the main `globalThis` when executing code in a module we'd need to ensure all modules execute on child scopes of this main module scope (but those need not themselves be `ModuleScope`s, and we'd need to find all `instanceof ModuleScope` checks which currently look at the `thisObj` and change them to appropriate top scope checks.

I don't know if @youngj is still around, still has their use case, and has an opinion on this. I can't tell how well covered this area really is because our Jacoco reports in `rhino` only cover the tests in `rhino`, while the coverage report in `tests` doesn't seem to be configured to show the core runtime classes.

### DIFF
--- a/rhino-engine/src/main/java/org/mozilla/javascript/engine/Builtins.java
+++ b/rhino-engine/src/main/java/org/mozilla/javascript/engine/Builtins.java
@@ -47,7 +47,7 @@ public class Builtins {
 
     private static Object print(Context cx, Scriptable scope, Scriptable thisObj, Object[] args) {
         try {
-            Builtins self = getSelf(thisObj);
+            Builtins self = getSelf(scope);
             for (Object arg : args) {
                 self.stdout.write(ScriptRuntime.toString(arg));
             }
--- a/rhino/src/main/java/org/mozilla/javascript/ArrayLikeAbstractOperations.java
+++ b/rhino/src/main/java/org/mozilla/javascript/ArrayLikeAbstractOperations.java
@@ -144,6 +144,9 @@ public class ArrayLikeAbstractOperations {
         } else {
             thisArg = ScriptRuntime.toObject(cx, scope, args[1]);
         }
+        if (thisArg instanceof TopLevel) {
+            thisArg = ((TopLevel) thisArg).getGlobalThis();
+        }
 
         Scriptable array = null;
         if (operation == IterativeOperation.FILTER || operation == IterativeOperation.MAP) {
@@ -340,6 +343,8 @@ public class ArrayLikeAbstractOperations {
         }
         Function f = (Function) callbackArg;
         Scriptable parent = ScriptableObject.getTopLevelScope(f);
+        Scriptable globalThis =
+                parent instanceof TopLevel ? ((TopLevel) parent).getGlobalThis() : parent;
         // hack to serve both reduce and reduceRight with the same loop
         boolean movingLeft = operation == ReduceOperation.REDUCE;
         Object value = args.length > 1 ? args[1] : NOT_FOUND;
@@ -354,7 +359,7 @@ public class ArrayLikeAbstractOperations {
                 value = elem;
             } else {
                 Object[] innerArgs = {value, elem, index, o};
-                value = f.call(cx, parent, parent, innerArgs);
+                value = f.call(cx, parent, globalThis, innerArgs);
             }
         }
         if (value == NOT_FOUND) {
--- a/rhino/src/main/java/org/mozilla/javascript/ArrayLikeAbstractOperations.java
+++ b/rhino/src/main/java/org/mozilla/javascript/ArrayLikeAbstractOperations.java
@@ -137,15 +137,12 @@ public class ArrayLikeAbstractOperations {
         Object callbackArg = args.length > 0 ? args[0] : Undefined.instance;
 
         Function f = getCallbackArg(cx, callbackArg);
-        Scriptable parent = ScriptableObject.getTopLevelScope(f);
+        TopLevel parent = ScriptableObject.getTopLevelScope(f);
         Scriptable thisArg;
         if (args.length < 2 || args[1] == null || args[1] == Undefined.instance) {
-            thisArg = parent;
+            thisArg = parent.getGlobalThis();
         } else {
             thisArg = ScriptRuntime.toObject(cx, scope, args[1]);
-        }
-        if (thisArg instanceof TopLevel) {
-            thisArg = ((TopLevel) thisArg).getGlobalThis();
         }
 
         Scriptable array = null;
@@ -342,9 +339,8 @@ public class ArrayLikeAbstractOperations {
             throw ScriptRuntime.notFunctionError(callbackArg);
         }
         Function f = (Function) callbackArg;
-        Scriptable parent = ScriptableObject.getTopLevelScope(f);
-        Scriptable globalThis =
-                parent instanceof TopLevel ? ((TopLevel) parent).getGlobalThis() : parent;
+        TopLevel parent = ScriptableObject.getTopLevelScope(f);
+        Scriptable globalThis = parent.getGlobalThis();
         // hack to serve both reduce and reduceRight with the same loop
         boolean movingLeft = operation == ReduceOperation.REDUCE;
         Object value = args.length > 1 ? args[1] : NOT_FOUND;
--- a/rhino/src/main/java/org/mozilla/javascript/FunctionObject.java
+++ b/rhino/src/main/java/org/mozilla/javascript/FunctionObject.java
@@ -14,7 +14,6 @@ import java.lang.reflect.Constructor;
 import java.lang.reflect.Member;
 import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
-import org.mozilla.javascript.commonjs.module.ModuleScope;
 
 public class FunctionObject extends BaseFunction {
     private static final long serialVersionUID = -5332312783643935019L;
@@ -387,22 +386,8 @@ public class FunctionObject extends BaseFunction {
                     thisObj = ((Delegator) thisObj).getDelegee();
                 }
                 if (!clazz.isInstance(thisObj)) {
-                    boolean compatible = false;
-                    if (thisObj == scope || thisObj instanceof ModuleScope) {
-                        Scriptable parentScope = getDeclarationScope();
-                        if (scope != parentScope) {
-                            // Call with dynamic scope for standalone function,
-                            // use parentScope as thisObj
-                            compatible = clazz.isInstance(parentScope);
-                            if (compatible) {
-                                thisObj = parentScope;
-                            }
-                        }
-                    }
-                    if (!compatible) {
-                        // Couldn't find an object to call this on.
-                        throw ScriptRuntime.typeErrorById("msg.incompat.call", functionName);
-                    }
+                    // Couldn't find an object to call this on.
+                    throw ScriptRuntime.typeErrorById("msg.incompat.call", functionName);
                 }
             }
 
--- a/rhino/src/main/java/org/mozilla/javascript/FunctionObject.java
+++ b/rhino/src/main/java/org/mozilla/javascript/FunctionObject.java
@@ -79,10 +79,10 @@ public class FunctionObject extends BaseFunction {
      */
     public FunctionObject(String name, Member methodOrConstructor, Scriptable scope) {
         if (methodOrConstructor instanceof Constructor) {
-            member = new MemberBox((Constructor<?>) methodOrConstructor);
+            member = new MemberBox(getDeclarationScope(), (Constructor<?>) methodOrConstructor);
             isStatic = true; // well, doesn't take a 'this'
         } else {
-            member = new MemberBox((Method) methodOrConstructor);
+            member = new MemberBox(getDeclarationScope(), (Method) methodOrConstructor);
             isStatic = member.isStatic();
         }
         String methodName = member.getName();
--- a/rhino/src/main/java/org/mozilla/javascript/FunctionObject.java
+++ b/rhino/src/main/java/org/mozilla/javascript/FunctionObject.java
@@ -78,10 +78,10 @@ public class FunctionObject extends BaseFunction {
      */
     public FunctionObject(String name, Member methodOrConstructor, Scriptable scope) {
         if (methodOrConstructor instanceof Constructor) {
-            member = new MemberBox(getDeclarationScope(), (Constructor<?>) methodOrConstructor);
+            member = new MemberBox(scope, (Constructor<?>) methodOrConstructor);
             isStatic = true; // well, doesn't take a 'this'
         } else {
-            member = new MemberBox(getDeclarationScope(), (Method) methodOrConstructor);
+            member = new MemberBox(scope, (Method) methodOrConstructor);
             isStatic = member.isStatic();
         }
         String methodName = member.getName();
--- a/rhino/src/main/java/org/mozilla/javascript/NativeGlobal.java
+++ b/rhino/src/main/java/org/mozilla/javascript/NativeGlobal.java
@@ -23,7 +23,7 @@ import org.mozilla.javascript.xml.XMLLib;
 public class NativeGlobal implements Serializable {
     static final long serialVersionUID = 6080442165748707530L;
 
-    public static void init(Context cx, Scriptable scope, boolean sealed) {
+    public static void init(Context cx, TopLevel scope, boolean sealed) {
         defineGlobalFunction(scope, sealed, "decodeURI", 1, NativeGlobal::js_decodeURI);
         defineGlobalFunction(
                 scope, sealed, "decodeURIComponent", 1, NativeGlobal::js_decodeURIComponent);
@@ -46,7 +46,7 @@ public class NativeGlobal implements Serializable {
                 scope, "Infinity", Double.POSITIVE_INFINITY, READONLY | DONTENUM | PERMANENT);
         ScriptableObject.defineProperty(
                 scope, "undefined", Undefined.instance, READONLY | DONTENUM | PERMANENT);
-        var globalThis = scope instanceof TopLevel ? ((TopLevel) scope).getGlobalThis() : scope;
+        var globalThis = scope.getGlobalThis();
 
         var obj = (Scriptable) scope.get("Object", scope);
         var objProto = (Scriptable) obj.get("prototype", obj);
--- a/rhino/src/main/java/org/mozilla/javascript/NativeGlobal.java
+++ b/rhino/src/main/java/org/mozilla/javascript/NativeGlobal.java
@@ -46,7 +46,13 @@ public class NativeGlobal implements Serializable {
                 scope, "Infinity", Double.POSITIVE_INFINITY, READONLY | DONTENUM | PERMANENT);
         ScriptableObject.defineProperty(
                 scope, "undefined", Undefined.instance, READONLY | DONTENUM | PERMANENT);
-        ScriptableObject.defineProperty(scope, "globalThis", scope, DONTENUM);
+        var globalThis = scope instanceof TopLevel ? ((TopLevel) scope).getGlobalThis() : scope;
+
+        var obj = (Scriptable) scope.get("Object", scope);
+        var objProto = (Scriptable) obj.get("prototype", obj);
+        globalThis.setPrototype(objProto);
+
+        ScriptableObject.defineProperty(scope, "globalThis", globalThis, DONTENUM);
 
         /*
             Each error constructor gets its own Error object as a prototype,
--- a/rhino/src/main/java/org/mozilla/javascript/NativeIterator.java
+++ b/rhino/src/main/java/org/mozilla/javascript/NativeIterator.java
@@ -21,7 +21,7 @@ public final class NativeIterator extends ScriptableObject {
 
     private Object objectIterator;
 
-    static void init(Context cx, ScriptableObject scope, boolean sealed) {
+    static void init(Context cx, TopLevel scope, boolean sealed) {
         LambdaConstructor constructor =
                 new LambdaConstructor(
                         scope,
@@ -63,6 +63,7 @@ public final class NativeIterator extends ScriptableObject {
         // throw StopIteration even if the property of the global
         // scope is replaced or deleted.
         scope.associateValue(ITERATOR_TAG, obj);
+        scope.getGlobalThis().associateValue(ITERATOR_TAG, obj);
     }
 
     /** Only for constructing the prototype object. */
--- a/rhino/src/main/java/org/mozilla/javascript/NativeIterator.java
+++ b/rhino/src/main/java/org/mozilla/javascript/NativeIterator.java
@@ -63,7 +63,6 @@ public final class NativeIterator extends ScriptableObject {
         // throw StopIteration even if the property of the global
         // scope is replaced or deleted.
         scope.associateValue(ITERATOR_TAG, obj);
-        scope.getGlobalThis().associateValue(ITERATOR_TAG, obj);
     }
 
     /** Only for constructing the prototype object. */
--- a/rhino/src/main/java/org/mozilla/javascript/ScriptableObject.java
+++ b/rhino/src/main/java/org/mozilla/javascript/ScriptableObject.java
@@ -2398,11 +2398,11 @@ public abstract class ScriptableObject extends SlotMapOwner
      * @param obj a JavaScript object
      * @return the corresponding global scope
      */
-    public static Scriptable getTopLevelScope(Scriptable obj) {
+    public static TopLevel getTopLevelScope(Scriptable obj) {
         for (; ; ) {
             Scriptable parent = obj.getParentScope();
             if (parent == null) {
-                return obj;
+                return (TopLevel) obj;
             }
             obj = parent;
         }
@@ -2950,20 +2950,8 @@ public abstract class ScriptableObject extends SlotMapOwner
      * @see #getAssociatedValue(Object key)
      */
     public static Object getTopScopeValue(Scriptable scope, Object key) {
-        scope = ScriptableObject.getTopLevelScope(scope);
-        for (; ; ) {
-            if (scope instanceof ScriptableObject) {
-                ScriptableObject so = (ScriptableObject) scope;
-                Object value = so.getAssociatedValue(key);
-                if (value != null) {
-                    return value;
-                }
-            }
-            scope = scope.getPrototype();
-            if (scope == null) {
-                return null;
-            }
-        }
+        var topScope = ScriptableObject.getTopLevelScope(scope);
+        return topScope.getAssociatedValue(key);
     }
 
     /**
--- a/rhino/src/main/java/org/mozilla/javascript/ScriptableObject.java
+++ b/rhino/src/main/java/org/mozilla/javascript/ScriptableObject.java
@@ -700,14 +700,14 @@ public abstract class ScriptableObject extends SlotMapOwner
      *     to regular property access.
      * @since 1.7.6
      */
-    public void setExternalArrayData(ExternalArrayData array) {
+    public void setExternalArrayData(Scriptable scope, ExternalArrayData array) {
         externalData = array;
 
         if (array == null) {
             delete("length");
         } else {
             // Define "length" to return whatever length the List gives us.
-            defineProperty("length", null, GET_ARRAY_LENGTH, null, READONLY | DONTENUM);
+            defineProperty(scope, "length", null, GET_ARRAY_LENGTH, null, READONLY | DONTENUM);
         }
     }
 
@@ -1228,7 +1228,7 @@ public abstract class ScriptableObject extends SlotMapOwner
                         ScriptableObject.PERMANENT
                                 | ScriptableObject.DONTENUM
                                 | (setter != null ? 0 : ScriptableObject.READONLY);
-                ((ScriptableObject) proto).defineProperty(name, null, method, setter, attr);
+                ((ScriptableObject) proto).defineProperty(scope, name, null, method, setter, attr);
                 continue;
             }
 
@@ -1468,7 +1468,8 @@ public abstract class ScriptableObject extends SlotMapOwner
      * @param attributes the attributes of the JavaScript property
      * @see org.mozilla.javascript.Scriptable#put(String, Scriptable, Object)
      */
-    public void defineProperty(String propertyName, Class<?> clazz, int attributes) {
+    public void defineProperty(
+            Scriptable scope, String propertyName, Class<?> clazz, int attributes) {
         int length = propertyName.length();
         if (length == 0) throw new IllegalArgumentException();
         char[] buf = new char[3 + length];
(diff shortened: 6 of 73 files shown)


aardvark179:
Okay, I've almost got this sorted. There remain a small set of failing tests due to nested scopes and associated values, which think I can tackle. This has required an API change in how we create nested scopes from a common sealed parent. I'm going to try and distill that down to a simple API, maybe something like `public Toplevel createIsolate()` and `public TopLevel.createIsolate(ScriptableObject customGlobal)` or something like that. Since this would change our documented API we would need to make a change to the docs as well at some point.

andreabergia:
One concern I have is that it's not great how in so many places you have to do:

```java
        if (scope instanceof TopLevel) {
            scope = ((TopLevel) scope).getGlobalThis();
        }
```
I am unsure on how we could improve it, but it's really diffused basically everywhere we have `Function.call`. Feels easy to forget in some place, especially for new code added in the future.

I am not sure why you are doing all the delegation in `TopLevel` to the `globalThis`. I have some ideas, but if you could draw a diagram of the scope structure, it would be very helpful for me.

aardvark179:
@andreabergia I agree that I don't really like the tests for top level that are now scattered through several bits of initialisation code. The original reason to not force the top scope to be a `TopLevel` was that we had historically allowed custom objects to be used, but I think the separation here allows users to have a custom object as their global this and still force the top level scope to be an actual `TopLevel`. This would certainly help make behaviour consistent with respect to generator functions and other objects whose prototypes are never created as normal properties of the global scope.

I've prototyped this by doing it in stages
1. Making the object we create if `null` is provided as the scope.
2. Throwing an error if we are passed something that isn't a `TopLevel` as the scope
3. Then changing the type.

I'm currently on step 2. and have had make small adjustments to four tests, and have another four which seem to need similar fixes. @gbrail how would you feel about this API change?

aardvark179:
Okay, I've moved all the initialisation stuff over to requiring a `TopLevel`, and it wasn't a big change really. I still have one failing test case due to serialisation and class caches. I'll investigate that over the weekend.

There are still couple of explicit checks for `TopLevel` which I'm also going to try and work through to see which can reasonably be removed without causing too much disruption.

aardvark179:
Well, that went well. 😆 so it turns out that even if we force the creation of all top level scopes to be actual `TopLevel`s that isn’t enough because of two things
1. Objects that are created to be scopes but have their parent scope set to null.
2. Places where we manage to swap the scope and the global this In our code.

The serialisation problem is similar in that the relationship between the top level scope and the global this has been fully set up at the point when something is trying to rely on that.

I’m setting this back to draft until I can track down those problems.

aardvark179:
Had a burst of inspiration this morning and got this down to about 30 failing test cases. I think all the remaining cases are actual bugs where we ended up with `TypeInfoFactory` cached against random objects that hadn't had their scopes set up yet. There's a couple of APIs around defining properties which need fixing, it turns out I even did those fixes on my prototype branch for scopes, so I'll try and bring those across to here.

The serialisation problem is occurring because of the way I handled associated values on the top level scope. To enable the creation of isolates I put the values on the global object rather than the scope, and relied on being able to reach that object _before_ it has been associated with top level scope, so the lookup breaks.

I think the thing to do here is to delegate associated values explicitly when creating isolates. I think two top level scopes created from a sealed shared scope should be guaranteed to share the same type cache.

gbrail:
I feel like I'm falling behind here, and apologies -- but it's a testament to the great work you're all doing.

I want to think more carefully about what we're doing to backward compatibility with all this stuff. I've never been super happy about the whole "parent scope" thing, but it's a fundamental part of Rhino and creating a "sealed shared scope" is used by a lot of examples and docs and probably real projects.

Can we document somewhere what we're proposing that will require people to change their current code, and then make a plan to stage it in upcoming releases?

Between this and other things we're working on, as well as the increased velocity of work on the project, it may indeed be time for a 1.9.0 or even a 2.0 with some reasonable changes to backward compatibility in favor of spec compliance.

p-bakker:
Just a possible heads-up when doing stuff in the area of sealed/shared scopes: #1085

aardvark179:
@gbrail Strongly agree. I think we're approaching the point where we want to do a release prior to any significant API change, and then work to get all those API changes into a 2.0.0 release. I think it might be worth it to create a 2.0 feature branch which we use for this and regularly rebase it on top of `master`.

I understand why people want to create a sealed parent object as it is a reasonably simple way to avoid repeated initialisation and prevent accidental modification, and we should absolutely ensure there is an API for doing so. I think that can be done, ad we can make it make it easier to use than the current `setPrototype`/`setParentScope` dance. That would give people a simple way to do this which we could encourage them to move to, and would hide most of the specifics allowing us to refactor things internally.

aardvark179:
Okay. Looks like there is a small change required for some Android tests and a little tidy up required. I'll try and do that over the weekend.

rbri:
Puh, this is huge (and i have to attend a workshop the next days). Will try it next week - hope that is ok

And a question - my top level scope is the window (or the globalworker)
* Window extends EventTarget
* EventTarget extends HtmlUnitScriptable
* HtmlUnitScriptable extends ScriptableObject

Do i have to make Window extending TopLevel?

aardvark179:
Your `Window` should become the `globalThis` of a TopLevel, so you should be able to do something like `var scope = new TopLevel(new Window)` or use the `createIsolate` API if you want to make a sealed parent.

aardvark179:
While preparing a refactoring of where we resolve `this` during function calls I'm seeing a couple of test failures involving isolates and `globalThis`. I don't think this was a problem introduced by this change, but I am going to try and debug this and work out the best way to resolve it.

aardvark179:
Okay. The test failures I saw appear to be long term bugs in our code base which I've fixed as part of my refactoring and nothing introduced by this change.

aardvark179:
So, reviewing my own change…

I don't like the way we build isolates with a custom Java class, but I didn't like the old way either. They both involved messing with the prototype chain making that custom class unusable as a normal object (or at least really weird since it exposed all of global this in its prototype chain) *and* it required special handling in `FunctionObject`.

I think I have a better model of how to do this, which works by using bound functions and avoids any messing with prototype chains.

aardvark179:
After some experimentation I've gone for cleaning up `ModuleScope`, and have got a fairly clean special case in `FunctionObject` on my follow-up clean up of `this` handling. I still don't like the way we handle isolates with custom classes in `RequireTest.customGlobal`, but we appear to have specific tests that we aren't implementing this using bound functions or similar tricks, so I guess the people who use this fearure actually depend on that behaviour.

gbrail:
This a big one, with the implication that many programs written with Rhino will need to change (although arguably only initialization code, and only if you don't use "var" there I think).

@aardvark179 has diligently kept it rebased. Almost every change requires another rebase, so we need to decide when is a good time to take this, so whatever we do we affect everyone's workflow one way or another.

Are any of the other regular contributors interested in reviewing this separately? Let me know, but at this point I'm fairly convinced that we should merge this and keep moving forward!

aardvark179:
Ideally I'd like an approval from @rbri as I assume they are using some sort of `Window` as their global object.

Rebasing this is pretty simple, as is the scope separation work that sits on top of it, so don't worry too much on that front.

gbrail:
Thanks -- I am OK with the direction of this big change but I would like to see verification from @rbri and his product, and maybe others, first!

rbri:
> Ideally I'd like an approval from @rbri as I assume they are using some sort of `Window` as their global object.

Will try this during the week, hope there is enought time for this.

rhino/src/main/java/org/mozilla/javascript/ScriptableObject.java
```diff
@@ -2399,11 +2416,17 @@ public static Scriptable getClassPrototype(Scriptable scope, String className) {
...
-    public static Scriptable getTopLevelScope(Scriptable obj) {
+    public static TopLevel getTopLevelScope(Scriptable obj) {
         for (; ; ) {
             Scriptable parent = obj.getParentScope();
             if (parent == null) {
-                return obj;
+                if (obj instanceof TopLevel) {
+                    return (TopLevel) obj;
+                } else {
+                    var err = new Error(obj.getClass().getName());
```
rbri: i think we have to create a more meaningfull error text here, only having the class name is a bit too minimal

rhino/src/main/java/org/mozilla/javascript/ScriptableObject.java
```diff
@@ -2399,11 +2416,17 @@ public static Scriptable getClassPrototype(Scriptable scope, String className) {
...
      */
-    public static Scriptable getTopLevelScope(Scriptable obj) {
+    public static TopLevel getTopLevelScope(Scriptable obj) {
         for (; ; ) {
             Scriptable parent = obj.getParentScope();
             if (parent == null) {
-                return obj;
+                if (obj instanceof TopLevel) {
+                    return (TopLevel) obj;
+                } else {
```
rbri: no need for else

rhino/src/main/java/org/mozilla/javascript/ScriptableObject.java
aardvark179: Fixed.

rhino/src/main/java/org/mozilla/javascript/ScriptableObject.java
aardvark179: Fixed.

rbri:
@aardvark179 have started a test migration to this, after some hours of work still not the whole test suite works.
The pain is sourced by the htmlunit code, that is not well prepared for this separation. I like to docuement it here a bit to have an idea what kind of pain others might face.

There are severals things that make this hard:
* the current htmlunit code does not distinguish between scope and globalThis, it was really hard to add this to the prototypes setup code
* TopLevel is a Scripable, in fact this means only the constuctor requires a TopLevel, all the places where the top level is accessed (and a call to globalThis has to be added) are hidden and throwing at runtime (if any)
* found some places in htmlunit where the topLevelScope is checked with instanceof - these places have to change also
* if you call getParentScope - you can get a TopLevel or another Scriptable; if it is a TopLevel, you have to call GlobalThis in addition

Will try to get the whole test suite working and report more during the next days.

aardvark179:
Yeah. Making TopLevel not be a scriptable really requires doing the full separation of the two types. I have a test branch that does go all the way, but that might be more change than you really want to take on at once, though I did find it helped find the last few stragglers where I needed to correct the types.

rbri:
@aardvark179 thanks for rebasing, had a hard time to sort out some blocker and now can work again on this one

rbri:
slowly made some progress, currently investigation a problem with eval(). More details as soon as i understand the problem better.

rbri:
@aardvark179 i think i found the reason for some of my failing tests

org.mozilla.javascript.ScriptRuntime#callSpecial

    if (callType == Node.SPECIALCALL_EVAL) {
        if (thisObj.getParentScope() == null && NativeGlobal.isEvalFunction(fun)) {
            return evalSpecial(cx, scope, callerThis, args, filename, lineNumber);
        }
    } else if (callType == Node.SPECIALCALL_WITH) {
    ....

checks the thisObj.getParentScope() for null.

In my case the thisObj is the window (the globalThis from the TopLevel). But my window has a parentScope the TopLevel object itself.

Question: is is correct that the globalThis might have a parentScope?
Should the check be something like thisObj.getParentScope() instanceof TopLevel?

aardvark179:
Do you have the JS code that is triggering this? I noticed we mis-compile some things as direct evals, and want to check the context in your call. See #2304 for the case eval problem I found.

aardvark179:
@rbti Looking at the `createIsolate` implementations I explicitly set the parent scope of the `globalThis` to be `null`, I think you probably be doing the same for your global window object. Maybe I should add a check in the constructor that the `globalThis` parent scope actually is null, but parent scope is mutable so this won't completely eliminate that problem.

rbri:
@aardvark179 still not there but i found something else :-)

    org.htmlunit.corejs.javascript.NativeConsole.init(Scriptable, boolean, ConsolePrinter)

only gets the sope in, but the last line

    ScriptableObject.defineProperty(scope, "console", obj, ScriptableObject.DONTENUM);

should not use the scope but instead an scriptable provided from the outside.
Hopefully my understanding is correct - still learning more scope magic.

rbri:
next one: ES6Iterator

    protected static void init(
            ScriptableObject scope, boolean sealed, ScriptableObject prototype, String tag) {
        if (scope != null) {
            prototype.setParentScope(scope);
            prototype.setPrototype(getObjectPrototype(scope));
        }

        // Define prototype methods using LambdaFunction
        LambdaFunction next = new LambdaFunction(scope, NEXT_METHOD, 0, ES6Iterator::js_next);
        ScriptableObject.defineProperty(prototype, NEXT_METHOD, next, DONTENUM);

        LambdaFunction iterator =
                new LambdaFunction(scope, "[Symbol.iterator]", 1, ES6Iterator::js_iterator);
        prototype.defineProperty(SymbolKey.ITERATOR, iterator, DONTENUM);

        prototype.defineProperty(
                SymbolKey.TO_STRING_TAG, prototype.getClassName(), DONTENUM | READONLY);

        if (sealed) {
            prototype.sealObject();
        }

        // Need to access Iterator prototype when constructing
        // Iterator instances, but don't have a iterator constructor
        // to use to find the prototype. Use the "associateValue"
        // approach instead.
        if (scope != null) {
            scope.associateValue(tag, prototype);
        }
    }

The method requires a scope as input but this time the scope has to be a ScriptableObject. This is inconsistent.

What about having a separate PR that changes the parameter to Scriptable and replacing

        if (scope != null) {
            scope.associateValue(tag, prototype);
        }
        
with

       if (scope instanceof ScriptableObject) {
            ((ScriptableObject)scope).associateValue(tag, prototype);
        }

aardvark179:
It sounds like your init-ing things with the window as the scope, rather than the TopLevel. I do have a later commit that forces the issue, and the ultimate goal of separating the types _really_ forces the issue.

Can you point me at your WIP commit so I can take a look?

rbri:
@aardvark179 forget the last one regarding ES6 Iterator, this referes to the head code not your one.

Because the step seems to be too big for me, i started to do some separation of scope and globalThis in my code before trying again to use your PR. see (https://github.com/HtmlUnit/htmlunit/commit/5681707522b02bad81d4aec39eda021942f336aa). Hopefully this brings me closer to a final migration strategie.

rbri:
@aardvark179 maybe you can have a look at the NativeConsole thing, i guess this is a real one

aardvark179:
So
```java
ScriptableObject.defineProperty(scope, "console", obj, ScriptableObject.DONTENUM);
```
is fine for defining something in a scope. In my test branch actually made `ScriptableObject.defineProperty` generic across scopes and objects because it is so widely used in our code. Looking at your init code I think you should be doing something like `new TopLevel(new Window())` and then doing init on that. Do you have a branches with your attempt to get things working (rhino and htmlunit I guess)? I can try doing building stuff this evening and see if I can understand where things aren't working for you.

rbri:
@aardvark179 can you pls reabase this

rbri:
@aardvark179 thanks for rebasing, will work on this again the next days

rhino/src/main/java/org/mozilla/javascript/ScriptableObject.java
0xe, line 2335: Missing doc for the new scope argument.

rhino-engine/src/main/java/org/mozilla/javascript/engine/Builtins.java
```diff
@@ -47,7 +47,7 @@ void register(Context cx, ScriptableObject scope, ScriptContext sc) {
 
     private static Object print(Context cx, Scriptable scope, Scriptable thisObj, Object[] args) {
         try {
-            Builtins self = getSelf(thisObj);
+            Builtins self = getSelf(scope);
```
0xe, line 50: Maybe a small comment here that `Builtins` are stored as `associatedValue` on the _scope_ and not `thisObj` which is now just the GlobalThis object? Not a big deal.

rhino/src/main/java/org/mozilla/javascript/NativeIterator.java
```diff
@@ -63,6 +63,7 @@ static void init(Context cx, ScriptableObject scope, boolean sealed) {
         // throw StopIteration even if the property of the global
         // scope is replaced or deleted.
         scope.associateValue(ITERATOR_TAG, obj);
+        scope.getGlobalThis().associateValue(ITERATOR_TAG, obj);
```
0xe: Could use a clarification on why this needs to be associated to both the scope and globalThis.

rhino/src/main/java/org/mozilla/javascript/ScriptableObject.java
0xe: Should we add a deprecated version of the previous method but falling back to `cx.topCallScope` for the first argument?

rhino/src/main/java/org/mozilla/javascript/TopLevel.java
```diff
@@ -264,4 +336,91 @@ public Scriptable getBuiltinPrototype(Builtins type) {
...
+    @Override
+    public int getAttributes(String name) {
+        if (super.get(name, this) != NOT_FOUND) {
+            return super.getAttributes(name);
+        } else {
+            return globalThis.getAttributes(name);
+        }
+    }
+
+    // Technically this is wrong, but there are currently tests that
```
0xe, line 402: Could use a reference to the spec, I believe it's (EcmaScript 2026, [9.1.1.1 Declarative Environment Records](https://tc39.es/ecma262/multipage/executable-code-and-execution-contexts.html#sec-declarative-environment-records)).

tests/testsrc/doctests/xmlOptions.doctest
```diff
@@ -9,11 +9,11 @@ js> x
     <baz>1</baz>
   </bar>
 </foo>
-js> var xmlLib = org.mozilla.javascript.xml.XMLLib.extractFromScope(this);
+js> var xmlLib = __xml_lib__;
```
0xe, line 12: Should we make `extractFromScope` work with `globalThis` as it used to (it is a public API) as a fallback? 

```java
public static XMLLib extractFromScope(Scriptable scope) {
      XMLLib lib = extractFromScopeOrNull(scope);
      if (lib != null) {
          return lib;
      }

      // "scope" could be `globalThis`
      lib = extractFromPrototypeChain(scope);
      if (lib != null) {
          return lib;
      }

      String msg = ScriptRuntime.getMessageById("msg.XML.not.available");
      throw Context.reportRuntimeError(msg);
  }

private static XMLLib extractFromPrototypeChain(Scriptable obj) {
      Scriptable proto = obj.getPrototype();
      while (proto != null) {
          XMLLib lib = extractFromScopeOrNull(proto);
          if (lib != null) {
              return lib;
          }
          proto = proto.getPrototype();
      }
      return null;
  }
```

 `__xml_lib__` seems to be a workaround.

rhino/src/main/java/org/mozilla/javascript/SlotMapOwner.java
```diff
@@ -332,7 +332,7 @@ protected static SlotMap createSlotMap(int initialSize) {
         }
     }
 
-    final SlotMap getMap() {
+    SlotMap getMap() {
```
0xe, line 335: Unintentional?

rhino/src/main/java/org/mozilla/javascript/NativeIterator.java
0xe: Callers of `getStopIterationObject` pass scope already. This is probably not needed?

Image

0xe (approved):
I think these are well-thoughtout fundamental changes that brings us close to the spec. Only a few minor comments and suggestions.

rbri:
Still think we have a issue with NativeConsole.init(Scriptable scope, boolean sealed, ConsolePrinter printer)

Using the provided scope for all the calls looks good to me, but the last line 

    ScriptableObject.defineProperty(scope, "console", obj, ScriptableObject.DONTENUM);

should define the property on the globalThis. Therefore i expect to see also the destinationObject for the new property in the method parameter list.

rbri:
During the last weeks i tried to apply this change to HtmlUnit. It was a hard fight, because there was no clear separation between the globalThis and the TopScope. I think i'm nearly done and fine with this - means: happy to see this merged!

But we have to have in mind, how big this change might be for other applications using Rhino.

Maybe it will be helpfuly to have a clear design sample document including some pictures and sample code that shows how to use Rhino as scripting language for an application.

Including
* the TopLevel / Scope idea
* the globalThis object
* the parentScope hirachy (including the parentScope setup for the globalThis object)
* the helper methods to walk to the TopLevel
* how to add you own implementation of
  * objects/prototypes
  * object properties
  * object functions
* if a function creates a new object - how to setup
  * the scope
  * the prototype (this includes a description how to find the prototype for all user defined objects)

And maybe more (hope the AI can help us a bit to make a shiny document here). Maybe we can start small and improve this step by step

What do you think?

aardvark179:
So, any property defined on TopLevel is simply defined on the associated globalThis object. That should not remain true in the long term (because consts, classes, and some other things shouldn't end up there, but for now it's true, and is why it does not matter that scope is being passed in that property definition call rather than globalThis. Clearly I need to find a good way to communicate that.

@rbri would you be okay with the architectural docs and examples being a follow up PR so we can merge this and start to work through the PR stack? I think I agree with your list, but I think it's also going to need a little bit of a primer on what realms are, what intrinsic are, and a couple of other spec related things.

P.s. Sorry, that came over grumpier than I intended. My jet lagged brain clearly isn't doing tone well today.

> means: happy to see this merged!

🎉 Ship it @gbrail! 😄

aardvark179:
Oh wait, I need to rebase this before we can merge it.

rbri:
> @rbri would you be okay with the architectural docs and examples being a follow up PR so we can merge this and start to work through the PR stack? I think I agree with your list, but I think it's also going to need a little bit of a primer on what realms are, what intrinsic are, and a couple of other spec related things.

That was exactly the reason for my last post, i think we have to go this way and i no longer like to be something like the blocker for this and all the other cool things on top.

rbri:
> P.s. Sorry, that came over grumpier than I intended. My jet lagged brain clearly isn't doing tone well today.

@aardvark179 at first, my english is not good enought to even notice this ;-) And i fear sometimes my writings have the same 'problem'

And second: we are 'fighting' for the best solution - thats at least for me the fun part of all this

rhino/src/main/java/org/mozilla/javascript/TopLevel.java
```diff
@@ -264,4 +336,91 @@ public Scriptable getBuiltinPrototype(Builtins type) {
         Object proto = func != null ? func.getPrototypeProperty() : null;
         return proto instanceof Scriptable ? (Scriptable) proto : null;
     }
+
+    public ScriptableObject getGlobalThis() {
+        return globalThis;
+    }
+
+    @Override
+    public Object get(String name, Scriptable start) {
```
gbrail, line 345: What breaks if this class overrides get, put, etc, but does not override getIds? Are there cases in which some code will need to enumerate the properties of the global object?

Also is there a reason to override get(int) and put(int)?

rhino/src/main/java/org/mozilla/javascript/FunctionObject.java
```diff
@@ -79,10 +78,10 @@ public class FunctionObject extends BaseFunction {
      */
     public FunctionObject(String name, Member methodOrConstructor, Scriptable scope) {
         if (methodOrConstructor instanceof Constructor) {
-            member = new MemberBox((Constructor<?>) methodOrConstructor);
+            member = new MemberBox(getDeclarationScope(), (Constructor<?>) methodOrConstructor);
```
gbrail: Have not verified this myself but by AI friend tells me that at this point in the constructor, getDeclarationScope() will always return null. That looks accurate to me. Is that going to cause a problem later with the Java embedding?

rhino/src/main/java/org/mozilla/javascript/ArrayLikeAbstractOperations.java
```diff
@@ -137,10 +137,10 @@ public static Object coercibleIterativeMethod(
         Object callbackArg = args.length > 0 ? args[0] : Undefined.instance;
 
         Function f = getCallbackArg(cx, callbackArg);
-        Scriptable parent = ScriptableObject.getTopLevelScope(f);
+        TopLevel parent = ScriptableObject.getTopLevelScope(f);
```
gbrail, line 140: Just a minor thing but in my new Java code I have been getting more judicious about using "var" in cases like this and if we had done so before then perhaps this would not have been necessary to make this particular change.

gbrail (commented):
Let's get this thing moving! It's an important next step in our evolution, and since it will affect every Rhino embedder it's time to rip the band aid off and work on this.

With that said, I have a few small suggestions but if other's are generally OK I think that we should move forward soon. Thanks!

rhino/src/main/java/org/mozilla/javascript/ArrayLikeAbstractOperations.java
aardvark179: Yes, I've been using `var` a lot and it would have reduced a bunch of these changes, but I've had enough push back on changing explicit types to `var` that I'm wary of doing it as part of a complex behavioural change.

rhino/src/main/java/org/mozilla/javascript/TopLevel.java
aardvark179: There's enough annoying edge cases with things like `__parent__` that I thought it best to override everything.

rhino/src/main/java/org/mozilla/javascript/FunctionObject.java
aardvark179: Good catch. This was an artefact of re-ordering a bunch of commits a while ago.

tests/testsrc/doctests/xmlOptions.doctest
aardvark179: It is indeed a workaround, but the whole use of the Java API in that test was always a workaround. I don't think the use of e4x is common enough to make this something to really worry about, and I didn't find any real uses of the API outside of test code when I looked. I don't think extractFromScope should altered to extract from an object, it says it deals with scopes, and that is the only thing it should be dealing with.

rhino/src/main/java/org/mozilla/javascript/NativeIterator.java
aardvark179: I think this was a hangover form some early work, we only need to associate this with the scope.

rhino/src/main/java/org/mozilla/javascript/TopLevel.java
aardvark179: It's not that section, `TopLevel` represents a [Global Environment Record](https://tc39.es/ecma262/#sec-global-environment-records), but you really have to read a whole bunch of sections to understand how `const x` is interpreted, so I chose to keep it short and simple.

rhino/src/main/java/org/mozilla/javascript/TopLevel.java
gbrail: Sorry I was going to fast - why NOT override those two as well? Same with getIds?

gbrail:
@rbri this is the first of many that will now start to cause backward compatibility to break, but we've all been talking about this for long enough and we agree it's time. Are you OK? I want to start moving faster on Duncan's big rewrite!

rbri:
> Are you OK? I want to start moving faster on Duncan's big rewrite!

@gbrail all fine for me - let'S move on

gbrail:
OK -- let's do it and move to the next one. Thanks, all!

aardvark179:
Cool. Next in the queue should be the changes to function this resolution ( #2328 ), I'll get that prepped when I'm get back home tomorrow.

Oh, wait, I have WiFi. I've done the rebase and made that PR ready for review.

rhino/src/main/java/org/mozilla/javascript/NativeGlobal.java
```diff
@@ -46,7 +46,13 @@ public static void init(Context cx, Scriptable scope, boolean sealed) {
                 scope, "Infinity", Double.POSITIVE_INFINITY, READONLY | DONTENUM | PERMANENT);
         ScriptableObject.defineProperty(
                 scope, "undefined", Undefined.instance, READONLY | DONTENUM | PERMANENT);
-        ScriptableObject.defineProperty(scope, "globalThis", scope, DONTENUM);
+        var globalThis = scope.getGlobalThis();
+
+        var obj = (Scriptable) scope.get("Object", scope);
+        var objProto = (Scriptable) obj.get("prototype", obj);
+        globalThis.setPrototype(objProto);
```
andreabergia, line 53: Why is this wiring done here, and not somewhere else like in the initialization of `TopLevel` or at the end of `initStandardObject`?

rhino/src/main/java/org/mozilla/javascript/TopLevel.java
```diff
@@ -99,12 +102,72 @@ enum NativeErrors {
...
+
+    public TopLevel() {
+        this(new GlobalThis());
+    }
+
+    public TopLevel(ScriptableObject customGlobal) {
+        globalThis = customGlobal;
+    }
+
+    public static TopLevel createIsolate(TopLevel parent) {
```
andreabergia, line 135: Why the name? I assume it has something to do with v8 isolates (which I don't know well), but since it actually creates a relationship with the existing toplevel via the prototype and builtins, it's a bit confusing.

PR #2162, opened 2025-11-19
Rewrite `Symbol.keyFor` to avoid iterator.

aardvark179:
I've seen a rare concurrent modification exception due to one thread manipulating the global symbol table while the iterator in `js_keyfor` was traversing it. Apparently the synchronised map could not completely protect against this. I've rewritten that code since it can now use the `SymbolKey`' s internal name to retrieve the symbol and check its identity. This avoids the iterator and should be more performant.


gbrail:
Looks good, thanks!

PR #2165, opened 2025-11-21
Migrated `NativeIterator` away from `IdScriptableObject`

andreabergia:
I've had an LLM do most of the changes, but I've reviewed them carefully and it looks good.

Doesn't seem to have an impact on test262.

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/NativeIterator.java
+++ b/rhino/src/main/java/org/mozilla/javascript/NativeIterator.java
@@ -198,7 +198,7 @@ public final class NativeIterator extends ScriptableObject {
 
     private Object next(Context cx, Scriptable scope) {
         Boolean b = ScriptRuntime.enumNext(this.objectIterator, cx);
-        if (!b.booleanValue()) {
+        if (!b) {
             // Out of values. Throw StopIteration.
             throw new JavaScriptException(NativeIterator.getStopIterationObject(scope), null, 0);
         }


rhino/src/main/java/org/mozilla/javascript/NativeIterator.java
```diff
@@ -249,30 +241,5 @@ public Object __iterator__(boolean b) {
...
-        }
-        return id;
-    }
-
-    private static final int Id_constructor = 1,
-            Id_next = 2,
-            Id___iterator__ = 3,
-            MAX_PROTOTYPE_ID = 3;
-
     private Object objectIterator;
```
rbri: Please move the declaration to the top, i think we did that for all other modernized classes

andreabergia: Done

gbrail:
Don't know if I'm impressed by the LLM or proud that we are consistent enough that we can have an LLM do these things but either way it looks good to me. Maybe the LLM can add class support too!

andreabergia:
> Don't know if I'm impressed by the LLM or proud that we are consistent enough that we can have an LLM do these things but either way it looks good to me. 

If anyone is curious, I've been using this prompt, which seems to work really well:

> We have converted various `NativeXXX` classes from `extends IdScriptableObject` to using `LambdaConstructor.` For example, see commit 72adb24a or 01141e1d or 030acb59 or 0d1ca694.
> Do the same for `NativeConsole`. Do not do any other refactoring or changes; focus only on the migration.

PR #2171, opened 2025-11-25
Migrated `NativeConsole` away from `IdScriptableObject`

andreabergia:
Another small step

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/NativeConsole.java
+++ b/rhino/src/main/java/org/mozilla/javascript/NativeConsole.java
@@ -52,42 +52,27 @@ public class NativeConsole extends ScriptableObject {
         obj.setParentScope(scope);
 
         obj.defineProperty(
-                "toSource",
-                new LambdaFunction(scope, "toSource", 0, NativeConsole::js_toSource),
-                DONTENUM);
-        obj.defineProperty(
-                "trace", new LambdaFunction(scope, "trace", 1, NativeConsole::js_trace), DONTENUM);
-        obj.defineProperty(
-                "debug", new LambdaFunction(scope, "debug", 1, NativeConsole::js_debug), DONTENUM);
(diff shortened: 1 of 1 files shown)


rhino/src/main/java/org/mozilla/javascript/NativeConsole.java
```diff
@@ -48,9 +48,120 @@ void print(
...
+                        (cx, s, thisObj, args) ->
+                                realThis(thisObj, "assert").js_assert(cx, s, args)),
+                DONTENUM);
+        obj.defineProperty(
+                "count",
+                new LambdaFunction(
+                        scope,
+                        "count",
+                        1,
+                        (cx, s, thisObj, args) -> realThis(thisObj, "count").js_count(cx, s, args)),
```
ZZZank: I think there's too much nesting: `<lambda> -> js_count() -> count()`. Merging the first two methods like this:

```java
private static Object js_count(Context cx, Scriptable scope, Scriptable thisObj, Object[] args) {
    realThis(thisObj, "count").count(cx, s, args);
    return Undefined.instance;
}
```

and use `NativeConsole::js_count` in LambdaFunction will be better in my opinion

andreabergia: Yeah that's probably better. Done.

aardvark179: This has the weird repetition of the name. I'd pull this logic out into a static method like we have in `NativeObject`. I think there is a strong reason to want this stuff to fit on single lines that makes it easy to see, "these are the method names, and this is the method references they link to."

andreabergia: I've changed it, I think this looks better.

gbrail:
Looks good!

PR #2173, opened 2025-11-26
Implement Symbol.toStringTag for Generators

0xe:
Stacked on top of https://github.com/mozilla/rhino/pull/2176. https://github.com/mozilla/rhino/pull/2173/commits/748752e52ef1a375c670352af129b7455f786fa9 is the only new change.

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/ES6Generator.java
+++ b/rhino/src/main/java/org/mozilla/javascript/ES6Generator.java
@@ -40,6 +40,8 @@ public final class ES6Generator extends ScriptableObject {
                 new LambdaFunction(scope, "[Symbol.iterator]", 0, ES6Generator::js_iterator);
         prototype.defineProperty(SymbolKey.ITERATOR, iterator, DONTENUM);
 
+        prototype.defineProperty(SymbolKey.TO_STRING_TAG, "Generator", DONTENUM | READONLY);
+
         if (sealed) {
             prototype.sealObject();
         }
--- a/tests/testsrc/test262.properties
+++ b/tests/testsrc/test262.properties
@@ -1070,14 +1070,13 @@ built-ins/GeneratorFunction 4/23 (17.39%)
     name.js
     proto-from-ctor-realm-prototype.js
 
-built-ins/GeneratorPrototype 7/61 (11.48%)
+built-ins/GeneratorPrototype 6/61 (9.84%)
     next/from-state-executing.js
     return/from-state-completed.js
     return/from-state-executing.js
     return/from-state-suspended-start.js
     throw/from-state-executing.js
     constructor.js
-    Symbol.toStringTag.js
 
 built-ins/Infinity 0/6 (0.0%)
 


rhino/src/main/java/org/mozilla/javascript/BaseFunction.java
```diff
@@ -329,6 +329,11 @@ protected boolean isGeneratorFunction() {
         return isGeneratorFunction;
     }
 
+    /** Package-private setter for generator function flag. Used during function initialization. */
```
aardvark179: I don't think this is needed. This is something that cannot change about a function, and it's on the descriptor (`isES6Generator()`).

rhino/src/main/java/org/mozilla/javascript/optimizer/OptRuntime.java
```diff
@@ -267,10 +268,40 @@ public static void throwStopIteration(Object scope, Object genState) {
         throw new JavaScriptException(si, "", 0);
     }
 
+    /**
+     * Evaluate default parameters for a generator function before creating the generator object.
+     * Ref: Ecma 2026 14.4.10, FunctionDeclarationInstantiation
+     */
+    public static void evaluateGeneratorDefaultParams(
+            JSFunction funObj, Scriptable activationScope) {
+        Context cx = Context.getCurrentContext();
```
rbri: see my comments regarding the context param in your other PR

rhino/src/main/java/org/mozilla/javascript/ES6Generator.java
```diff
@@ -40,6 +40,9 @@ static ES6Generator init(ScriptableObject scope, boolean sealed) {
                 new LambdaFunction(scope, "[Symbol.iterator]", 0, ES6Generator::js_iterator);
         prototype.defineProperty(SymbolKey.ITERATOR, iterator, DONTENUM);
 
+        prototype.defineProperty(
+                SymbolKey.TO_STRING_TAG, prototype.getClassName(), DONTENUM | READONLY);
```
aardvark179: So, it's clever using `prototype.getClassName()` to get the string `"Generator"`, but the prototype isn't specified as being a generator, so this would need to change as you fix up things like that. The specification states:

> The GeneratorFunction [constructor](https://tc39.es/ecma262/#constructor):
> 
> * Is a standard built-in [function object](https://tc39.es/ecma262/#function-object) that inherits from the Function [constructor](https://tc39.es/ecma262/#constructor).
> * has a [[Prototype]] internal slot whose value is [%Function%](https://tc39.es/ecma262/#sec-function-constructor).
> * has a "length" property whose value is 1𝔽.
> * has a "name" property whose value is "GeneratorFunction".
> * has the following properties:…

I'd just make it a literal `"Generator"`.

Edit: Sorry, referenced the wrong bit of the spec.

> The %GeneratorPrototype% object:
> 
> * is %GeneratorFunction.prototype.prototype%.
> * s an [ordinary object](https://tc39.es/ecma262/#ordinary-object).
> * is not a Generator instance and does not have a [[GeneratorState]] internal slot.
> * has a [[Prototype]] internal slot whose value is [%Iterator.prototype%](https://tc39.es/ecma262/#sec-%iterator.prototype%-object).
> * has properties that are indirectly inherited by all Generator instances.

rhino/src/main/java/org/mozilla/javascript/ScriptRuntime.java
```diff
@@ -5349,6 +5349,34 @@ public static void initFunction(
         }
     }
 
+    /**
+     * Evaluate default parameters for a generator function before creating the generator object.
+     * Ref: Ecma 2026 14.4.10, FunctionDeclarationInstantiation
+     */
+    public static void evaluateGeneratorDefaultParams(
+            JSFunction funObj, Scriptable activationScope, Context cx) {
```
rbri: Sorry, but usually the context is before the scope in most of our functions..
And maybe the funObj should be after the scope. Maybe you can have a look at oll the others and try to stay in line

rhino/src/main/java/org/mozilla/javascript/optimizer/OptRuntime.java
```diff
@@ -268,38 +267,13 @@ public static void throwStopIteration(Object scope, Object genState) {
...
-                                activationScope, sourceCode, "<generator-default-param>", 1, null);
-                ScriptableObject.putProperty(activationScope, paramName, defaultValue);
-            }
-        }
-    }
-
     public static Scriptable createNativeGenerator(
-            JSFunction funObj, Scriptable scope, Scriptable thisObj, int maxLocals, int maxStack) {
-        Context cx = Context.getCurrentContext();
+            JSFunction funObj,
```
rbri: see my last comment

rhino/src/main/java/org/mozilla/javascript/ScriptRuntime.java
0xe: Will reorder them.

rbri (approved):

gbrail:
Thanks!

PR #2175, opened 2025-11-26
simplify class NativeCall

rbri:
simplify class NativeCall
* remove not used fields
* merge two if conditions and
* sort boolean checks to the left for performance


rbri:
@andreabergia @aardvark179 can you please have a short look - if you agree i will merge that

andreabergia (approved):

PR #2176, opened 2025-11-27
Fix default parameter evaluation ordering for generators

0xe:
Fixes a few things in generators for spec compliance:

1. Default parameters for generators should be evaluated after the activation is created but before any other declarations are added.

In [FunctionDeclarationInstatiation](https://tc39.es/ecma262/multipage/ordinary-and-exotic-objects-behaviours.html#sec-functiondeclarationinstantiation), the ordering is

- Create arguments (22)
- Init args and substitute by evaluating default expressions (24-26, can refer to arguments)
- Create function declarations (36)

This is called from [EvaluateGeneratorBody](https://tc39.es/ecma262/multipage/ecmascript-language-functions-and-classes.html#sec-runtime-semantics-evaluategeneratorbody).

2. Prototype for generator functions should be `generatorProto`.

From [InstantiateGeneratorFunctionExpression](https://tc39.es/ecma262/multipage/ecmascript-language-functions-and-classes.html#sec-runtime-semantics-instantiategeneratorfunctionexpression)

> 7. Let prototype be OrdinaryObjectCreate(%GeneratorPrototype%).

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/Interpreter.java
+++ b/rhino/src/main/java/org/mozilla/javascript/Interpreter.java
@@ -296,15 +296,11 @@ public final class Interpreter extends Icode implements Evaluator {
                 ScriptRuntime.initScript(fnOrScript, thisObj, cx, scope, desc.isEvalFunction());
             }
 
-            // For generators with default parameters, evaluate them after activation scope
-            // is created but BEFORE nested function declarations are added to the scope.
-            // This ensures that references to 'arguments' in default params get the arguments
-            // object, not a function declaration that might shadow it.
-            if (fnOrScript instanceof JSFunction) {
-                ScriptRuntime.evaluateGeneratorDefaultParams(cx, scope, (JSFunction) fnOrScript);
-            }
+            // Defer default parameters and nested function declarations until activation scope
+            // creation
+            // Ref: Ecma 2026, 10.2.11, FunctionDeclarationInstantiation
 
-            if (desc.getFunctionCount() != 0) {
+            if (desc.getFunctionCount() != 0 && !desc.isES6Generator()) {
                 if (desc.getFunctionType() != 0 && !desc.requiresActivationFrame()) Kit.codeBug();
                 for (int i = 0; i < desc.getFunctionCount(); i++) {
                     JSDescriptor fdesc = desc.getFunction(i);
--- a/rhino/src/main/java/org/mozilla/javascript/CodeGenUtils.java
+++ b/rhino/src/main/java/org/mozilla/javascript/CodeGenUtils.java
@@ -1,12 +1,9 @@
 package org.mozilla.javascript;
 
-import java.util.ArrayList;
-import java.util.List;
 import org.mozilla.javascript.ast.AstNode;
 import org.mozilla.javascript.ast.AstRoot;
 import org.mozilla.javascript.ast.Block;
 import org.mozilla.javascript.ast.FunctionNode;
-import org.mozilla.javascript.ast.InfixExpression;
 import org.mozilla.javascript.ast.Scope;
 import org.mozilla.javascript.ast.ScriptNode;
 
@@ -111,34 +108,6 @@ public class CodeGenUtils {
         builder.hasRestArg = scriptOrFn.hasRestParameter();
         builder.hasDefaultParameters = scriptOrFn.getDefaultParams() != null;
 
-        // For generators, store default param source for runtime evaluation
-        // (but not if they contain super - those are compiled as bytecode)
-        if (scriptOrFn instanceof FunctionNode) {
-            FunctionNode fn = (FunctionNode) scriptOrFn;
-            if (fn.isGenerator() && scriptOrFn.getDefaultParams() != null) {
-                List<Object> defaultParams = scriptOrFn.getDefaultParams();
-                ArrayList<String> sourceList = new ArrayList<>();
-                boolean hasSuper = false;
-                for (int i = 0; i < defaultParams.size() - 1; i += 2) {
-                    if (defaultParams.get(i) instanceof String
-                            && defaultParams.get(i + 1) instanceof AstNode) {
-                        String paramName = (String) defaultParams.get(i);
-                        AstNode expr = (AstNode) defaultParams.get(i + 1);
-                        if (containsSuper(expr)) {
-                            hasSuper = true;
-                            break;
-                        }
-                        sourceList.add(paramName);
-                        sourceList.add(expr.toSource());
-                    }
-                }
-                // Only store if no super (super cases are handled as bytecode)
-                if (!hasSuper && !sourceList.isEmpty()) {
-                    builder.generatorDefaultParams = sourceList.toArray(new String[0]);
-                }
-            }
-        }
-
         // Calculate arity (function.length) - count params before first default
         builder.arity = FunctionNode.calculateFunctionArity(scriptOrFn);
 
@@ -160,44 +129,5 @@ public class CodeGenUtils {
         } else {
             builder.constructor = new JSCode.NullBuilder<T>();
         }
-    }
-
-    /**
-     * Check if an AST node contains any super references.
-     *
-     * @param node the AST node to check
-     * @return true if the node or any of its children contain super references
-     */
-    private static boolean containsSuper(AstNode node) {
-        int type = node.getType();
-        // Check if this node is a super-related token
-        if (type == Token.SUPER
-                || type == Token.GETPROP_SUPER
-                || type == Token.GETPROPNOWARN_SUPER
-                || type == Token.SETPROP_SUPER
-                || type == Token.GETELEM_SUPER
-                || type == Token.SETELEM_SUPER) {
-            return true;
-        }
-
-        // For InfixExpression nodes (like PropertyGet), check left and right fields
-        if (node instanceof InfixExpression) {
-            InfixExpression infix = (InfixExpression) node;
-            if (infix.getLeft() != null && containsSuper(infix.getLeft())) {
-                return true;
-            }
-            if (infix.getRight() != null && containsSuper(infix.getRight())) {
-                return true;
-            }
-        }
-
-        // Recursively check all children
-        for (Node child = node.getFirstChild(); child != null; child = child.getNext()) {
-            if (child instanceof AstNode && containsSuper((AstNode) child)) {
-                return true;
-            }
-        }
-
-        return false;
     }
 }
--- a/rhino/src/main/java/org/mozilla/javascript/CodeGenerator.java
+++ b/rhino/src/main/java/org/mozilla/javascript/CodeGenerator.java
@@ -102,6 +102,32 @@ class CodeGenerator<T extends ScriptOrFn<T>> extends Icode {
         CodeGenUtils.setConstructor(builder, theFunction);
 
         if (theFunction.isGenerator()) {
+            // For generators with default parameters, generate parameter initialization
+            // bytecode BEFORE Icode_GENERATOR so defaults are evaluated before generator
+            // object creation. Ref: Ecma 2026, 10.2.11 FunctionDeclarationInstantiation
+            Node paramInitBlock = theFunction.getGeneratorParamInitBlock();
+            if (paramInitBlock != null) {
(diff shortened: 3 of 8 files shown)


rhino/src/main/java/org/mozilla/javascript/optimizer/OptRuntime.java
```diff
@@ -267,10 +268,40 @@ public static void throwStopIteration(Object scope, Object genState) {
         throw new JavaScriptException(si, "", 0);
     }
 
+    /**
+     * Evaluate default parameters for a generator function before creating the generator object.
+     * Ref: Ecma 2026 14.4.10, FunctionDeclarationInstantiation
+     */
+    public static void evaluateGeneratorDefaultParams(
+            JSFunction funObj, Scriptable activationScope) {
+        Context cx = Context.getCurrentContext();
```
rbri: Why not providing the context as param? Looks like you have the context already in hand when calling this

rhino/src/main/java/org/mozilla/javascript/optimizer/OptRuntime.java
```diff
@@ -267,10 +268,40 @@ public static void throwStopIteration(Object scope, Object genState) {
...
+            if (Undefined.isUndefined(paramValue)) {
+                Object defaultValue =
+                        cx.evaluateString(
+                                activationScope, sourceCode, "<generator-default-param>", 1, null);
+                ScriptableObject.putProperty(activationScope, paramName, defaultValue);
+            }
+        }
+    }
+
     public static Scriptable createNativeGenerator(
```
rbri, line 270: Same here regarding the context, i think this is a more or less internal function, we cahn change the params without having to take care of backward compatibility. Again we can provide the context from the outside.

rhino/src/main/java/org/mozilla/javascript/Interpreter.java
```diff
@@ -296,6 +296,35 @@ void initializeArgs(
...
+            // For generators with default parameters, evaluate them after activation scope
+            // is created but BEFORE nested function declarations are added to the scope.
+            // This ensures that references to 'arguments' in default params get the arguments
+            // object, not a function declaration that might shadow it.
+            if (fnOrScript instanceof JSFunction) {
+                JSFunction jsFunc = (JSFunction) fnOrScript;
+                String[] generatorDefaultParams =
+                        jsFunc.getDescriptor().getGeneratorDefaultParams();
+                if (generatorDefaultParams != null && generatorDefaultParams.length > 0) {
+                    for (int i = 0; i < generatorDefaultParams.length; i += 2) {
```
rbri: This seems to be a copy of wha we are doing in the opt runtime in this case. Maybe we have to have it only once

0xe: Yeah, will refactor it out to ScriptRuntime and just call it from the interpreter. It's the exact same code. I intended to do that and forgot. Thanks for calling it out!

rhino/src/main/java/org/mozilla/javascript/optimizer/OptRuntime.java
```diff
@@ -268,38 +267,13 @@ public static void throwStopIteration(Object scope, Object genState) {
...
-                                activationScope, sourceCode, "<generator-default-param>", 1, null);
-                ScriptableObject.putProperty(activationScope, paramName, defaultValue);
-            }
-        }
-    }
-
     public static Scriptable createNativeGenerator(
-            JSFunction funObj, Scriptable scope, Scriptable thisObj, int maxLocals, int maxStack) {
-        Context cx = Context.getCurrentContext();
+            JSFunction funObj,
```
rbri: again see the comments on the other pr

andreabergia (changes requested):
If I understand this correctly, for:

```js
function *f(a = 1) { yield 2; }
```

what you are doing is storing in the `JSDescriptor` for `f` an array of string which contains `["a", "1"]` and then you evaluate the string with `1` at runtime.
This feels really horrible for performances. 

Why aren't we generating some code that does:

```js
function *f(a) {
  {
  var a = arguments.length > 0 && arguments[0] !== undefined ? arguments[0] : 'dEFUALT';
  }
  yield 2;
}
```

like babel or swc do? I get that this is pretty complex for generators, given that they have their state, but dynamically `eval`-ing strings at every single call doesn't seem good to me.

0xe:
I agree that evaluating at runtime is pretty terrible for perf. However, I think, we cannot use Babel's approach of putting the initialization inside the generator body.

According to spec, [EvaluateGeneratorBody](https://tc39.es/ecma262/multipage/ecmascript-language-functions-and-classes.html#sec-runtime-semantics-evaluategeneratorbody), generator evaluation must:

- Step 1: Perform FunctionDeclarationInstantiation (which evaluates default parameters)
- Step 2: OrdinaryCreateFromConstructor (calls `GetPrototypeFromConstructor` to read  `functionObject.prototype` to determine the generator instance's [[Prototype]])

Default parameter evaluation happens in [FunctionDeclarationInstantiation](https://tc39.es/ecma262/multipage/ordinary-and-exotic-objects-behaviours.html#sec-functiondeclarationinstantiation),
  - Step 22: Create the arguments object
  - Step 28: evaluates default parameters with access to arguments (perform IteratorBindingInitialization of formals)
  - Step 37: Instantiate function declarations

There's a corresponding [262](https://raw.githubusercontent.com/tc39/test262/main/test/language/statements/generators/generator-created-after-decl-inst.js) case.
 
For example:

```javascript
function* g(a = (g.prototype = null)) {}
var oldProto = g.prototype;
var it = g();
Object.getPrototypeOf(it) === oldProto; // Should be FALSE per spec
```

The default param mutates `g.prototype` during `FunctionDeclarationInstantiation` (step 1), then the generator instance is created in step 2 using the mutated value. If we evaluated the default after returning the generator, this test would fail.

While `jsc`, `engine262` and `rhino` evaluate this correctly, v8 fails this.

It appears that v8 fixed [this](https://issues.chromium.org/issues/42211055) reasonably recently (Apr 2025) and their [rationale](https://chromium-review.googlesource.com/c/v8/v8/+/6462591) seems to be coherent with the above.

Given that we need to evaluate the arguments at runtime, to improve the performance perhaps I can store the `AstNodes` instead in `JSDescriptor` to avoid re-parsing. Thoughts? I think with the different modes we have it gets pretty complicated fast.

0xe:
I've changed the implementation to just generate the bytecode directly instead of storing the source for expressions evaluation for defaults, so for the 262 test case above:

```javascript
function* g(a = (g.prototype = null)) {}
var oldProto = g.prototype;
var it = g();
Object.getPrototypeOf(it) === oldProto; // => false
```

```
 [0] REG_STR_C0 "a"
 [1] NAME
 [2] UNDEF
 [3] SHEQ
 [4] IFNE 20 // jump as no default substitution needed
 [7] LINE : 1
 [10] REG_STR_C0 "a"
 [11] BINDNAME
 [12] REG_STR_C1 "g"
 [13] NAME
 [14] NULL
 [15] REG_STR_C2 "prototype"
 [16] SETPROP
 [17] REG_STR_C0 "a"
 [18] SETNAME
 [19] POP
 [20] GENERATOR : 1 // create generator
 [23] GENERATOR_END : 1
```

and for compiled code:

```
      21: invokedynamic #209,  0            // InvokeDynamic #0:"NAME:GET:a":(Lorg/mozilla/javascript/Scriptable;Lorg/mozilla/javascript/Context;)Ljava/lang/Object;
      26: getstatic     #48                 // Field org/mozilla/javascript/Undefined.instance:Ljava/lang/Object;
      29: invokedynamic #143,  0            // InvokeDynamic #0:"MATH:SHALLOWEQ":(Ljava/lang/Object;Ljava/lang/Object;)Z
      34: ifne          40
      37: goto          70 // jump as no default substitution needed
      40: aload_3
      41: aload_0
      42: invokedynamic #212,  0            // InvokeDynamic #0:"NAME:BIND:a":(Lorg/mozilla/javascript/Scriptable;Lorg/mozilla/javascript/Context;)Lorg/mozilla/javascript/Scriptable;
      47: aload_3
      48: aload_0
      49: invokedynamic #63,  0             // InvokeDynamic #0:"NAME:GET:g":(Lorg/mozilla/javascript/Scriptable;Lorg/mozilla/javascript/Context;)Ljava/lang/Object;
      54: aconst_null
      55: aload_0
      56: aload_3
      57: invokedynamic #216,  0            // InvokeDynamic #0:"PROP:SET:prototype":(Ljava/lang/Object;Ljava/lang/Object;Lorg/mozilla/javascript/Context;Lorg/mozilla/javascript/Scriptable;)Ljava/lang/Object;
      62: aload_0
      63: aload_3
      64: invokedynamic #219,  0            // InvokeDynamic #0:"NAME:SET:a":(Lorg/mozilla/javascript/Scriptable;Ljava/lang/Object;Lorg/mozilla/javascript/Context;Lorg/mozilla/javascript/Scriptable;)Ljava/lang/Object;
      69: pop
      70: aload_0
      71: aload_3
      72: aload         4
      74: aload_1
      75: iconst_0
      76: iconst_0
      77: invokestatic  #223                // Method org/mozilla/javascript/optimizer/OptRuntime.createNativeGenerator:(Lorg/mozilla/javascript/Context;Lorg/mozilla/javascript/Scriptable;Lorg/mozilla/javascript/Scriptable;Lorg/mozilla/javascript/JSFunction;II)Lorg/mozilla/javascript/Scriptable;
```

0xe:
This follows what `jsc` is doing:

```
bb#1
Predecessors: [ ]
[   0] enter
[   1] to_this            srcDst:this, ecmaMode:NotStrictMode, valueProfile:1
[   6] create_lexical_environment dst:loc7, scope:loc4, symbolTable:Cell: 0x101044cc8 (0x300004870:[0x4870/18544, SymbolTable, (0/0, 0/0){}, NonArray, Unknown, Leaf]), StructureID: 18544(const1), initialValue:<JSValue()>(const2)
[  11] mov                dst:loc4, src:loc7
[  14] get_argument       dst:loc8, index:1, valueProfile:2
[  18] stricteq           dst:loc9, lhs:loc8, rhs:Undefined(const3)
[  22] jfalse             condition:loc9, targetLabel:28(->50)  // JUMP to generator creation if default value needn't be evaluated
Successors: [ #3 #2 ]

bb#2
Predecessors: [ #1 ]
[  25] resolve_scope      dst:loc10, scope:loc4, var:0, resolveType:GlobalProperty, localScopeDepth:1
[  32] get_from_scope     dst:loc11, scope:loc10, var:0, getPutInfo:2048<ThrowIfNotFound|GlobalProperty|NotInitialization|NotStrictMode>, localScopeDepth:1, offset:0, valueProfile:3
[  41] mov                dst:loc8, src:Null(const4)
[  44] put_by_id          base:loc11, property:1, value:loc8, flags:
Successors: [ #3 ]

bb#3
Predecessors: [ #1 #2 ]
[  50] put_to_scope       scope:loc7, var:2, value:loc8, getPutInfo:1048580<DoNotThrowIfNotFound|ResolvedClosureVar|Initialization|NotStrictMode>, symbolTableOrScopeDepth:1, offset:0
[  58] create_lexical_environment dst:loc5, scope:loc4, symbolTable:Cell: 0x101044bd8 (0x300004870:[0x4870/18544, SymbolTable, (0/0, 0/0){}, NonArray, Unknown, Leaf]), StructureID: 18544(const0), initialValue:Undefined(const3)
[  63] mov                dst:loc4, src:loc5
[  66] create_generator   dst:loc6, callee:callee
[  70] new_func_exp       dst:loc8, scope:loc4, functionDecl:0
[  74] put_internal_field base:loc6, index:2, value:loc8
[  78] put_internal_field base:loc6, index:3, value:this
[  82] ret                value:loc6
Successors: [ ]


Identifiers:
  id0 = g
  id1 = prototype
  id2 = a
```

andreabergia:
> I've changed the implementation to just generate the bytecode directly instead of storing the source for expressions evaluation for defaults

This looks _way_ better! Thanks for doing it!
I guess we can, relatively easily, extend this to _all_ functions and not just generators, right? But that's worth a separate PR.

0xe:
Closing this as https://github.com/mozilla/rhino/pull/2173 which was stacked on this is already merged.

Issue #2174, opened 2025-11-26
Things that don't work with the debugger.


aardvark179:
For various reasons internally we want to check that code behaves correctly under the debugger, so we have been looking at running the test 262 suite with a no-op debugger attached. Unfortunately this reveals 12 failing tests, all of which are concerned with blocks in the language.
```
Test262SuiteTest > test262Case(String, TestMode, boolean, Test262Case, boolean) > [101246] language/block-scope/leave/verify-context-in-labelled-block.js, DEBUGGER_INTERPRETED, false, org.mozilla.javascript.tests.Test262SuiteTest$Test262Case@3aabb193, false FAILED
    java.lang.AssertionError at Test262SuiteTest.java:582

Test262SuiteTest > test262Case(String, TestMode, boolean, Test262Case, boolean) > [101264] language/block-scope/leave/nested-block-let-declaration-only-shadows-outer-parameter-value-1.js, DEBUGGER_INTERPRETED, false, org.mozilla.javascript.tests.Test262SuiteTest$Test262Case@d4e67af, false FAILED
    java.lang.AssertionError at Test262SuiteTest.java:582

Test262SuiteTest > test262Case(String, TestMode, boolean, Test262Case, boolean) > [101253] language/block-scope/leave/for-loop-block-let-declaration-only-shadows-outer-parameter-value-2.js, DEBUGGER_INTERPRETED, true, org.mozilla.javascript.tests.Test262SuiteTest$Test262Case@642df9b0, false FAILED
    java.lang.AssertionError at Test262SuiteTest.java:582

Test262SuiteTest > test262Case(String, TestMode, boolean, Test262Case, boolean) > [101252] language/block-scope/leave/for-loop-block-let-declaration-only-shadows-outer-parameter-value-2.js, DEBUGGER_INTERPRETED, false, org.mozilla.javascript.tests.Test262SuiteTest$Test262Case@642df9b0, false FAILED
    java.lang.AssertionError at Test262SuiteTest.java:582

Test262SuiteTest > test262Case(String, TestMode, boolean, Test262Case, boolean) > [101265] language/block-scope/leave/nested-block-let-declaration-only-shadows-outer-parameter-value-1.js, DEBUGGER_INTERPRETED, true, org.mozilla.javascript.tests.Test262SuiteTest$Test262Case@d4e67af, false FAILED
    java.lang.AssertionError at Test262SuiteTest.java:582

[...]

PR #2177, opened 2025-11-27
Disambiguate local variable names to fix debugger issues.

aardvark179:
This fixes the problems seen with blocks and the debugger in #2174.

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/CodeGenUtils.java
+++ b/rhino/src/main/java/org/mozilla/javascript/CodeGenUtils.java
@@ -138,6 +138,13 @@ public class CodeGenUtils {
      * Disambiguate all variable names after the parameters. We avoid disambiguating the parameters
      * themselves because those names can be repeated and are not unique even if an activation frame
      * is used.
+     *
+     * <p>This is only required when an interpreted function might have duplicate local variable
+     * names and might be run under the debugger. This can only occur if the function has nested
+     * blocks, uses `let` with a variable name in that nested block which is also used outside that
+     * block, and does not require the use of an activation frame for any other reason. When running
+     * under the debugger variables are accessed by name rather than by index, and hence those names
+     * must not include duplicates.
      */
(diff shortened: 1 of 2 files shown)


rhino/src/main/java/org/mozilla/javascript/CodeGenUtils.java
```diff
@@ -130,4 +133,24 @@ public static <T extends ScriptOrFn<T>> void setConstructor(
...
         }
     }
+
+    /**
+     * Disambiguate all variable names after the parameters. We avoid disambiguating the parameters
+     * themselves because those names can be repeated and are not unique even if an activation frame
+     * is used.
+     */
+    private static String[] disambiguateNames(String[] names, int paramCount) {
+        String[] result = new String[names.length];
```
rPraml, line 150: How hot is this method?
Would it make sense to add a fast exit like `if (names.length <= paramCount) return names` to avoid some object creation

aardvark179: It's called once per method during compilation, so not very hot.

aardvark179:
Thanks to @andreabergia for adding a specific test around this≥

rhino/src/main/java/org/mozilla/javascript/Interpreter.java
```diff
@@ -3840,6 +3846,9 @@ NewState execute(Context cx, CallFrame frame, InterpreterState state, int op) {
                     varDbls[state.indexReg] = frame.sDbl[state.stackTop];
                 }
             } else {
+                // This only occurs if we are using the deubgger, in
```
gbrail: Could you please fix this (repeated) typo? Thanks!

aardvark179: Oops. Fixed.

gbrail:
This is a nice fix -- thanks!

Issue #535, opened 2019-04-22
Number.toLocaleString() works like toString()


vnarayananptc:
var num = 2046430; 
var loc = "en-US";
var result=Number(num).toLocaleString(loc);

Throws error - "illegal radix 0".

When same is called with - var result=Number(num).toLocaleString(undefined); - it works.

Seeing this in the Rhino source code -

org.mozilla.javascript.NativeNumber.java (Rhino 1.7.10)
`
          case Id_toLocaleString:
            {
                // toLocaleString is just an alias for toString for now
                int base = (args.length == 0 || args[0] == Undefined.instance)
                    ? 10 : ScriptRuntime.toInt32(args[0]);
                return ScriptRuntime.numberToString(value, base);
            }
`

It assumes the first arg to be the base, not the locale code. If any locale code is passed in, the toInt32() throws the invalid radix error.

If we pass in "undefined" it sets the base to 10.

FractalCodeRicardo:
Same problem here, any advice is appreciated.

Thanks

JohannesHeld:
I face the same problem. I call this function in a HighCharts dataLabels formatter.

geppoz-it:
still not working

rbri:
Have a customized version in HtmlUnit, will have a look if i can use it for this.

aardvark179:
> still not working

Well, you can either fix this by coming up with a plan to support Ecma 402 (the spec for internationalisation in ecmascript) or you can come up with a patch to not use those arguments. Either option may break some existing users as they would both change the behaviour of code so might need a feature flag or something.

The main spec does say how it should work if 402 support is not implemented (https://tc39.es/ecma262/multipage/numbers-and-dates.html#sec-number.prototype.tolocalestring).

aardvark179:
> Have a customized version in HtmlUnit, will have a look if i can use it for this.

That would be awesome.

PR #2178, opened 2025-11-27
improve the Number.toLocaleString() implementation

rbri:
* if 402 is not enabled for the context, ignore the given parameters
* if 402 is enabled, support string locales

this should at least avoid the exceptions described in #535

### DIFF
--- a/tests/testsrc/test262.properties
+++ b/tests/testsrc/test262.properties
@@ -2937,249 +2937,190 @@ built-ins/parseInt 1/55 (1.82%)
 
 built-ins/undefined 0/8 (0.0%)
 
-intl402/Array 2/2 (100.0%)
-
-intl402/BigInt 7/11 (63.64%)
-    prototype/toLocaleString/de-DE.js
-    prototype/toLocaleString/default-options-object-prototype.js
-    prototype/toLocaleString/en-US.js
-    prototype/toLocaleString/returns-same-results-as-NumberFormat.js
-    prototype/toLocaleString/taint-Intl-NumberFormat.js
-    prototype/toLocaleString/this-value-invalid.js
-    prototype/toLocaleString/throws-same-exceptions-as-NumberFormat.js
-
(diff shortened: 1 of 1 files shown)


rbri:
@aardvark179 please have a look at lets discuss....

aardvark179:
Okay. I've done an exhaustive search of our internal code, and all the uses I can find pass 0 arguments, so should be good with this change. The only bug report I can find internally is that somebody wanted to pass a locale and couldn't, so this is probably fine - especially as it's neat gated behind the 402 feature.

It might be worth updating the 262 tests (I think we'd need to enable 402 mode in a context factory, remove the leading `~` from the intl402 line in the properties file and update then update the properties. I think we maybe pass about 20 tests from that entire group, but it's a start.

rbri:
@aardvark179 ok, will go forward with this, but will be busy the next days with other stuff

tests/testsrc/test262.properties
```diff
@@ -2938,192 +2937,190 @@ built-ins/parseInt 1/55 (1.82%)
 
 built-ins/undefined 0/8 (0.0%)
 
-~intl402
-
```
andreabergia, line 2942: Seems like you accidentally removed this directory

andreabergia: Well, I misclicked. I meant the one after:

```
language/arguments-object 183/263 (69.58%)
```

rbri:
@andreabergia this is in draft because i had a hard time enabling the intl402 tests.
The setup for this is different from the other test folders and or regenertator fails on this (adds wrong entries). Spend already some time over the weekend to understand the generator and try to fix it. But i faild more or less and then tried to reimplement the generator using a different approach but i'm not 100% done.

The plan is to split this into two parts, fix only the problem with some unit tests and then create a second PR for the 262.properties handling. But you have to be a bit patient because of the workshop this week and next week i have to do a talk on a conference...

PR #2182, opened 2025-11-29
Implement ES2024 Resizable ArrayBuffer

anivar:
Implements ES2024 Resizable ArrayBuffer and ES2025 auto-length TypedArray views.

## ES2024 Resizable ArrayBuffer

Constructor accepts maxByteLength option:
```js
new ArrayBuffer(8, { maxByteLength: 16 })
```

Resizable buffers can grow/shrink in-place:
```js
buffer.resize(12)  // grows to 12 bytes
buffer.resize(4)   // shrinks to 4 bytes
```

New properties:
- `resizable` - true if buffer has maxByteLength
- `maxByteLength` - maximum allowed size

## ES2025 Auto-length TypedArrays

TypedArrays on resizable buffers without explicit length track buffer size dynamically:
```js
const rab = new ArrayBuffer(16, { maxByteLength: 32 });
const ta = new Uint8Array(rab);  // auto-length view
console.log(ta.length);  // 16

rab.resize(24);
console.log(ta.length);  // 24
```

Fixed-length views throw TypeError when buffer resizes out of bounds:
```js
const fixedView = new Uint8Array(rab, 0, 4);
rab.resize(2);
fixedView.at(0);  // TypeError: TypedArray is out of bounds
```

## Implementation

**NativeArrayBuffer:**
- Uses `int` primitive for maxByteLength (-1 for fixed-length)
- Lambda pattern for new methods
- Data preserved during resize via System.arraycopy, new bytes zeroed

**NativeTypedArrayView:**
- `isAutoLength` flag for auto-length views
- `updateLength()` recalculates length for auto-length views
- `isTypedArrayOutOfBounds()` validates bounds on resized buffers
- All methods call `validateAndGetLength()` before access

**Errors:**
- TypeError if not resizable, detached, or out of bounds
- RangeError if resize exceeds maxByteLength

## Test262 Results

**59 tests now passing** (227→169 TypedArray failures)

All `return-abrupt-from-this-out-of-bounds` tests now pass across TypedArray methods: at, copyWithin, entries, every, fill, filter, find, findIndex, findLast, findLastIndex, forEach, includes, indexOf, join, keys, lastIndexOf, map, reduce, reduceRight, reverse, set, slice, some, sort, toLocaleString, values.

Fixed-length property tests (byteLength, byteOffset, length) on resizable buffers also passing.

### DIFF
--- a/tests/testsrc/test262.properties
+++ b/tests/testsrc/test262.properties
@@ -2343,44 +2343,29 @@ built-ins/ThrowTypeError 2/14 (14.29%)
     unique-per-realm-function-proto.js
     unique-per-realm-non-simple.js non-strict
 
-built-ins/TypedArray 227/1434 (15.83%)
+built-ins/TypedArray 169/1434 (11.79%)
     from/iterated-array-changed-by-tonumber.js
-    prototype/at/BigInt/return-abrupt-from-this-out-of-bounds.js
     prototype/at/coerced-index-resize.js
     prototype/at/resizable-buffer.js
-    prototype/at/return-abrupt-from-this-out-of-bounds.js
     prototype/byteLength/BigInt/resizable-array-buffer-auto.js
-    prototype/byteLength/BigInt/resizable-array-buffer-fixed.js
     prototype/byteLength/resizable-array-buffer-auto.js
-    prototype/byteLength/resizable-array-buffer-fixed.js
     prototype/byteLength/resizable-buffer-assorted.js
     prototype/byteLength/resized-out-of-bounds-1.js
     prototype/byteLength/resized-out-of-bounds-2.js
     prototype/byteOffset/BigInt/resizable-array-buffer-auto.js
-    prototype/byteOffset/BigInt/resizable-array-buffer-fixed.js
     prototype/byteOffset/resizable-array-buffer-auto.js
-    prototype/byteOffset/resizable-array-buffer-fixed.js
     prototype/byteOffset/resized-out-of-bounds.js
-    prototype/copyWithin/BigInt/return-abrupt-from-this-out-of-bounds.js
     prototype/copyWithin/coerced-target-start-end-shrink.js
     prototype/copyWithin/coerced-target-start-grow.js
     prototype/copyWithin/resizable-buffer.js
-    prototype/copyWithin/return-abrupt-from-this-out-of-bounds.js
-    prototype/entries/BigInt/return-abrupt-from-this-out-of-bounds.js
     prototype/entries/resizable-buffer.js
     prototype/entries/resizable-buffer-grow-mid-iteration.js
     prototype/entries/resizable-buffer-shrink-mid-iteration.js
-    prototype/entries/return-abrupt-from-this-out-of-bounds.js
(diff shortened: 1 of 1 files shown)


rhino/src/main/java/org/mozilla/javascript/typedarrays/NativeArrayBuffer.java
```diff
@@ -29,6 +29,8 @@ public class NativeArrayBuffer extends ScriptableObject {
     private static final byte[] EMPTY_BUF = new byte[0];
 
     byte[] buffer;
+    // ES2024: maxByteLength for resizable buffers (null = fixed-length)
+    private Integer maxByteLength;
```
gbrail: Typically we use primitives and not boxed objects as class members -- is there a reason why this can't be an "int"?

tests/testsrc/test262.properties
```diff
@@ -455,89 +455,89 @@ built-ins/AggregateError 2/25 (8.0%)
 built-ins/Array 257/3077 (8.35%)
     fromAsync 95/95 (100.0%)
     length/define-own-prop-length-coercion-order-set.js
-    prototype/at/coerced-index-resize.js {unsupported: [resizable-arraybuffer]}
-    prototype/at/typed-array-resizable-buffer.js {unsupported: [resizable-arraybuffer]}
+    prototype/at/coerced-index-resize.js
```
gbrail, line 458: In general -- it looks like we implemented the guts of the feature (I mean, I see the code in there) but from these changes it looks like none of the tests have been made to work. Can you investigate why? I would expect that by implementing this we would unblock many, many tests (which would be well worth it). What else is missing that causes all these tests to fail?

gbrail (commented):
Thanks for working on this and for your patience as we ask lots of questions...

rhino/src/main/java/org/mozilla/javascript/typedarrays/NativeArrayBuffer.java
anivar: Good catch! I've changed maxByteLength from Integer to int using -1 as the sentinel value for fixed-length buffers. This avoids the boxing overhead.

tests/testsrc/test262.properties
anivar: Looking more closely, these tests are for TypedArray operations on resizable buffers (auto-length tracking, out-of-bounds behavior). My PR only implements the ArrayBuffer side (constructor, resize(), getters).

  I need to update the TypedArray implementation to properly handle resizable buffers. Let me investigate what's needed and update the PR.

gbrail:
Update - thanks for your work on this! A few things:

* I want to make sure that we make the right decision on implement this, so I opened a discussion:
https://github.com/mozilla/rhino/discussions/2271
* There are a lot of failing tests but a great deal of them are because we're missing features like classes and block-scope const

I really want to get this right (and use it in another project I'm working on) so I'd like to make this PR a basis for a new PR that gets this as close to the spec as we can get.

Thanks!

gbrail:
I've opened https://github.com/mozilla/rhino/pull/2273 which includes all the changes in this PR plus additional fixes and tests. Please take a look at that one -- thanks!

PR #2183, opened 2025-11-29
Implement ES2025 RegExp.escape static method

anivar:
Implements the ES2025 RegExp.escape() static method for safely escaping special regex characters in strings.

## Implementation
- Uses LambdaConstructor pattern for modern Rhino static method definition
- Escapes syntax characters (. * + ? ^ $ | etc.) with backslash prefix
- Escapes control characters (\t \n \v \f \r) with special sequences
- Escapes other punctuators (,-=<>#&!%:;@~'\`") using \xNN hex notation
- Escapes initial digits/letters to prevent ambiguity with legacy sequences
- Throws TypeError for non-string arguments (no automatic coercion)
- Uses explicit hex codes (0x002c) for clarity and maintainability

## Testing
- 15 Java unit tests created, all passing
- test262 compliance: 19/20 tests passing (95%)
- Overall RegExp test failures reduced from 975 to 956

## Impact
Provides standardized, spec-compliant solution for escaping user input in regular expressions, improving code safety and developer experience.

### DIFF
--- a/tests/testsrc/test262.properties
+++ b/tests/testsrc/test262.properties
@@ -2101,7 +2101,7 @@ built-ins/Reflect 12/153 (7.84%)
     set/return-false-if-receiver-is-not-writable.js
     set/return-false-if-target-is-not-writable.js
 
-built-ins/RegExp 956/1868 (51.18%)
+built-ins/RegExp 955/1868 (51.12%)
     CharacterClassEscapes 12/12 (100.0%)
     dotall 4/4 (100.0%)
     escape/escaped-otherpunctuators.js


rhino/src/main/java/org/mozilla/javascript/regexp/NativeRegExpCtor.java
```diff
@@ -113,4 +120,155 @@ private static RegExpImpl getImpl() {
...
+
+            // SPEC STEP 6: Return code point as-is (including underscore, other letters, digits)
+            result.append(c);
+        }
+
+        return result.toString();
+    }
+
+    /** Check if character is a decimal digit (0-9) */
+    private static boolean isDecimalDigit(char c) {
```
gbrail, line 218: I'm sure that we already have implementations of these checks either elsewhere in the codebase, or in the Java Character class. Can you please check first to see if we can reuse those or make them more generic rather than encode this here? I won't be surprised if we need a few new ones but checks like isWhiteSpace and isDecimalDigit are certainly not unique to this method. Thanks!

Issue #2184, opened 2025-12-01
Console functions not require a this object


rbri:
var x = console.error;
    x('hello');

This is a regression from https://github.com/mozilla/rhino/commit/88adf9937bde6a1d7108e6ac74e57ea8c688fcf7
Found this while adapting the latest changes to HtmlUnit.

From Claude

_Great question! Yes, console.error (and other console methods) are somewhat special in modern JavaScript engines.
The code works because modern browsers implement console methods in a way that doesn't strictly depend on this. They're designed to be more forgiving and flexible._

And the spec

<img width="475" height="138" alt="Image" src="https://github.com/user-attachments/assets/bfecac15-12b0-47fa-a481-39eb4d6d9648" />

Looks like the we have to be more careful with calling realThis() for various methods.

rbri:
@andreabergia do you like to work on this or should i do?

andreabergia:
I'll take care of this. And add some tests, because we don't have enough for the `console` 🙂

rbri:
@andreabergia great, have to attend a workshop the next days, will try to have a look at this from time to time.
Will also try to think about more test for that on the HtmlUnit side, maybe we come up with something more complet.

As a wise man once said: There is always one test case missing.

aardvark179:
Here is the [spec](https://console.spec.whatwg.org/#log), in case anybody else would like a little more context than a screen shot of one clause. 🙂

PR #2186, opened 2025-12-01
Fix: allow console methods to be called without a this

andreabergia:
This fixes a regression introduced in https://github.com/mozilla/rhino/commit/88adf9937bde6a1d7108e6ac74e57ea8c688fcf7, and allows code like this:

```js
var fn = console.error;
fn('hello');
```

Closes https://github.com/mozilla/rhino/issues/2184


rbri:
LGTM

aardvark179:
If you take a slightly different approach to the refactoring then the declaration block can remain compact. Maybe something like [this](https://github.com/aardvark179/rhino/commit/a2f9aee69a62765ffe967d28837cc0265fcd6a5f)?

andreabergia:
> If you take a slightly different approach to the refactoring then the declaration block can remain compact. Maybe something like [this](https://github.com/aardvark179/rhino/commit/a2f9aee69a62765ffe967d28837cc0265fcd6a5f)?

Oh yeah, way better. I was thinking something like should be possible, but my fried brain didn't manage to come up with it yesterday. 🙂 
I've added your commit to the PR, thanks

rbri:
great, thanks @andreabergia for the fast fix - will merge it as soon the ci is happy

Issue #2187, opened 2025-12-02
Flapper test


andreabergia:
We have a flapper test! In the CI run for my https://github.com/mozilla/rhino/pull/2186 PR, I have seen the test [`tests/ecma/Date/15.9.5.8.js`](https://github.com/mozilla/rhino/blob/master/tests/testsrc/tests/ecma/Date/15.9.5.8.js) ran by `MozillaSuiteTest` fail: see this [run](https://github.com/mozilla/rhino/actions/runs/19851039740#artifacts) and [step](https://github.com/mozilla/rhino/actions/runs/19851039740/job/56877721100#step:5:278).
However, I then submitted an [empty commit](https://github.com/mozilla/rhino/pull/2186/commits/177f10b7b6c62c502e125c07e29512bc888eb818) in my PR, and the test passed.

---

Update: there's at least another one from this [run](https://github.com/mozilla/rhino/actions/runs/19864734129/job/56924144927?pr=2190): `testsrc/tests/ecma/Date/15.9.5.8.js`

rbri:
@andreabergia you can also simply rerun the build from the browser

andreabergia:
> [@andreabergia](https://github.com/andreabergia) you can also simply rerun the build from the browser

I couldn't find a button. Perhaps only maintainers can?

PR #2188, opened 2025-12-02
Removed a test case about dates that uses the current instant

andreabergia:
It's an inconsistent behavior that has caused a flaky test failure [in the CI](https://github.com/mozilla/rhino/actions/runs/19851039740/job/56877721100#step:5:278).

I've only removed the test case with the "current instant", not the whole test, although really all of these seem kinda obsolete and superset by test262, frankly. They also contain JS implementation of date methods, like this [LocalTime](
https://github.com/mozilla/rhino/blob/e3d6b16aa233e0a427bf1dfe6b4ba3645213b924/tests/testsrc/tests/ecma/shell.js#L304) function, which are pretty complex.
According to Claude at least, there's likely a discrepancy between the daylight savings time behavior, but I didn't investigate much further.

Fixes https://github.com/mozilla/rhino/issues/2187

### DIFF
--- a/tests/testsrc/tests/ecma/Date/15.9.5.8.js
+++ b/tests/testsrc/tests/ecma/Date/15.9.5.8.js
@@ -25,7 +25,7 @@ var TITLE   = "Date.prototype.getMonth()";
 
 writeHeaderToLog( SECTION + " "+ TITLE);
 
-addTestCase( TIME_NOW );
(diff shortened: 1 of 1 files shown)


andreabergia:
Apparently there are other tests flapping today, see https://github.com/mozilla/rhino/actions/runs/19864734129/job/56924144927?pr=2190

So I don't think this is a good solution.

Issue #2180, opened 2025-11-28
ReferenceError: "a" is not defined for `let` in switch default


Rhython:
**Describe the bug**
After commit 2ba6e57, the following script throws an error in Rhino:

```
(function () {
    switch (0) {
        default:
            let a = 0;
            if (a == 0) {
                a++;
            }
            return a;
    }
})();
```

**Error Message**
```
ReferenceError: "a" is not defined.
```

**Steps to Reproduce**
1. Run the above script in Rhino after commit 2ba6e57.
2. Observe the ReferenceError.

**Expected behavior**
The script should execute without errors and return `1`.

**Additional context**
This code works as expected before commit 2ba6e57. The error started occurring after this commit.

aardvark179:
I can't reproduce this. Could you give a little more information on your environment (language version, interpreted or compiled mode, whether you are using the debugger…). That would be really helpful. I've been trying this by using
```
./gradlew run -q --console=plain --args="-version 200 -opt -1"
```

And running your example in the shell, which _shouldn't_ have changed anything.

Rhython:
Ok, I use jdk21 on windows, interpret mode, es6 in my project, and some of my script got error after I pull latest version of rhino. Here is the minimal reproduce code in my project:

```java
import org.mozilla.javascript.*;
import org.mozilla.javascript.ContextFactory.Listener;

public class TestRhino3 {

    static {
        ContextFactory.getGlobal().addListener(new Listener() {
            @Override
            public void contextCreated(Context cx) {
                cx.setLanguageVersion(Context.VERSION_ES6);
                cx.setInterpretedMode(true);
                cx.getWrapFactory().setJavaPrimitiveWrap(false);
            }

            @Override
            public void contextReleased(Context cx) {

            }
        });
    }

    public static void main(String[] args) {
//        System.setProperty("rhino.printTrees", "true");
        String source = "(function () {\n" +
                        "    switch (0) {\n" +
                        "        default:\n" +
                        "            let a = 0;\n" +
                        "            if (a == 0) {\n" +
                        "                a++;\n" +
                        "            }\n" +
                        "            return a;\n" +
                        "    }\n" +
                        "})();\n";
        try (Context cx = Context.enter()) {
            var scope = cx.initSafeStandardObjects();
            Script script = cx.compileString(source, "test", 1, null);
            Object exec = script.exec(cx, scope);
            System.out.println("exec: " + exec);
        }

    }

}
```

[...]

PR #2189, opened 2025-12-02
Fix variable resolution in switch statements with let declarations

andreabergia:
Fixed the regression introduced by 2ba6e57, which made `switch` a proper `Scope`.

The issue was that `Scope.splitScope()` creates an outer block scope and moves the symbol table from the switch to the block, but doesn't update the switch's `parentScope` pointer. This breaks the scope chain used for variable resolution, causing it to skip the block where symbols are stored.

Fixed by manually fixing the scope chain, and follow the same pattern as `transformForLoop()` for manually managing `currentScope` instead of using `pushScope`/`popScope`, since the scope hierarchy is modified after parsing.

Fixes https://github.com/mozilla/rhino/issues/2180


aardvark179 (approved):
LGTM!

rbri:
@gbrail i like to see this merged because i pla another htmlunit release and i like to have this fix included

gbrail:
Thanks -- looks good.

Issue #1922, opened 2025-05-29
java.lang.IllegalStateException: FAILED ASSERTION


yjh1534:
I've got illegalStateException with following js input:

for(var {};;)debugger

with following stack traces:
	at org.mozilla.javascript.Kit.codeBug(Kit.java:417)
	at org.mozilla.javascript.NodeTransformer.transformCompilationUnit_r(NodeTransformer.java:309)
	at org.mozilla.javascript.NodeTransformer.transformCompilationUnit_r(NodeTransformer.java:409)
	at org.mozilla.javascript.NodeTransformer.transformCompilationUnit(NodeTransformer.java:67)
	at org.mozilla.javascript.NodeTransformer.transform(NodeTransformer.java:45)
	at org.mozilla.javascript.NodeTransformer.transform(NodeTransformer.java:33)
	at org.mozilla.javascript.CodeGenerator.compile(CodeGenerator.java:61)
	at org.mozilla.javascript.Interpreter.compile(Interpreter.java:291)
	at org.mozilla.javascript.Context.compileImpl(Context.java:2559)
	at org.mozilla.javascript.Context.compileReader(Context.java:1476)
	at org.mozilla.javascript.Context.compileReader(Context.java:1448)
	at org.mozilla.javascript.Context.evaluateReader(Context.java:1290)

Is it appropriate execution result?

PR #2190, opened 2025-12-02
Handle incorrect destructuring in for loops at parse time

andreabergia:
Code like this:

```js
for (var {};;) {}
```

was parsed correctly, but then failed during the code generation with various exceptions because the destructuring expression after `var` doesn't have a right-hand side.

This change implements a check at parse time, to throw a `SyntaxError`, coherently with what `node` or `spidermonkey` or `quickjs` do.

I'm a bit surprised that this doesn't seem to show any improvement in test262 coverage. Perhaps a test there might be warranted?

Fixes https://github.com/mozilla/rhino/issues/1922

### DIFF
--- a/tests/testsrc/jstests/harmony/destructuring.js
+++ b/tests/testsrc/jstests/harmony/destructuring.js
@@ -31,5 +31,7 @@ assertThrows("(1 ? {} : 490) = 1", SyntaxError);
 assertThrows("(1 ? [] : 490) = 1", SyntaxError);
 assertThrows("for (var {};;) {}", SyntaxError);
 assertThrows("for (let {};;) {}", SyntaxError);
+assertThrows("for (var {a};;) {}", SyntaxError);
+assertThrows("for (let {a};;) {}", SyntaxError);
 
 "success";


aardvark179 (commented):
~~Looks good~~Actually, seems to be causing some failures, and does make me wonder, are there other places that might need similar checks?

aardvark179:
Ah, scratch that, it's a problem with a `Date` test.

andreabergia:
> are there other places that might need similar checks?

From what I can tell, this problem should only happen in `for` loops currently. Variable declarations were already fine, function arguments _do_ allow that syntax, `catch ({a})` should allow it but it's not supported (created https://github.com/mozilla/rhino/issues/2194). For what it's worth, Claude also thinks we don't need new places for this fix.

aardvark179:
Okay looking at the spec and MDN I agree that the only places that allow destructuring binding are loops, functions, and catch blocks.

tests/src/test/java/org/mozilla/javascript/tests/ParserTest.java
```diff
@@ -1490,6 +1490,13 @@ public void oomOnInvalidInput() {
         expectParseErrors("`\\u{8", new String[] {"syntax error"});
     }
 
+    @Test
+    public void errorOnInvalidDestructuringDeclaration() {
+        expectParseErrors(
+                "for(var {};;) {}",
+                new String[] {"Missing = in destructuring declaration", "syntax error"});
```
rbri, line 1497: As a reader of error messages i like this form more - not sure why the french version has the ' and the english not.

    Missing '=' in destructuring declaration

gbrail:
Great, thanks!

PR #2193, opened 2025-12-03
Convert Chinese characters to Unicode escapes to ensure ResourceBundle compatibility

Rhython:
Converted all Chinese characters in Messages_zh_CN.properties into \uXXXX Unicode escape sequences.

ResourceBundle does not support reading UTF-8 encoded .properties files, resulting in incorrect character rendering in Chinese environments. Using Unicode escapes ensures proper loading of localized messages and resolves the garbled text issue.


ZZZank:
Do you have image showing the broken formatting? Rhino now targets Java 11 and newer, and starting from Java9, ResourceBundle will use UTF-8 by default. So this is likely to be unnecessary

Rhython:
You're right, I apologize for submitting this PR without a thorough investigation. I did encounter a garbled character issue, but after careful investigation, I found it to be a Gradle problem. I will close this.

Issue #2194, opened 2025-12-03
Support destructuring in catch


andreabergia:
The following should work and print `test`:

```js
try {
  throw new SyntaxError('test');
} catch ( {message} ) {
  console.log(message);
}
```

Rhino currently fails with "invalid catch block condition".

PR #2195, opened 2025-12-03
Implement support for destructuring in `catch`

andreabergia:
Extend the parser and IR to support things like `catch ({message})` or `catch ([a, b])`.

Note that, while Rhino supports a non-standard extension for conditions in catch clauses, i.e. `catch (a if b)`, I have opted to _not_ support it if we have a destructuring expression.

The implementation does this:

- at parse time, we rely on pre-existing functions
- the AST node for `CatchClause` was made more generic, and will now contain either the name or the destructuring expression
- during the IR generation, from:

```js
catch ( {message} ) { 
  body
}
```

we generate:

```js
catch ( $tempname ) { 
  let {message} = $tempname;
  body
}
```

---

Fixes https://github.com/mozilla/rhino/issues/2194

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/IRFactory.java
+++ b/rhino/src/main/java/org/mozilla/javascript/IRFactory.java
@@ -1210,22 +1210,61 @@ public final class IRFactory {
 
         Node catchBlocks = new Block();
         for (CatchClause cc : node.getCatchClauses()) {
-            Name varName = cc.getVarName();
+            AstNode varName = cc.getVarName();
             Node catchCond = null;
             Node varNameNode = null;
+            Scope catchBody = cc.getBody();
 
             if (varName != null) {
-                varNameNode = parser.createName(varName.getIdentifier());
+                if (varName instanceof Name) {
+                    // Simple identifier
+                    varNameNode = parser.createName(((Name) varName).getIdentifier());
 
-                AstNode ccc = cc.getCatchCondition();
-                if (ccc != null) {
-                    catchCond = transform(ccc);
-                } else {
+                    AstNode ccc = cc.getCatchCondition();
+                    if (ccc != null) {
+                        catchCond = transform(ccc);
+                    } else {
+                        catchCond = new EmptyExpression();
+                    }
+                } else if (varName instanceof org.mozilla.javascript.ast.ArrayLiteral
+                        || varName instanceof org.mozilla.javascript.ast.ObjectLiteral) {
+                    // Destructuring pattern. We basically replace:
+                    //   catch ( {message} ) { body }
+                    // into:
+                    //   catch ( $tempname ) { let {message} = $tempname; body }
+
+                    // The exception will be stored in the temp name
+                    String tempVarName = parser.currentScriptOrFn.getNextTempName();
+                    varNameNode = parser.createName(tempVarName);
+
+                    // The let statement will be used to do the destructuring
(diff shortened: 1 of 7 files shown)


rbri:
Wow 50 mor passing test...

rhino/src/main/java/org/mozilla/javascript/ast/CatchClause.java
```diff
@@ -12,11 +12,13 @@
...
  * <pre><i>CatchClause</i> :
- *        <b>catch</b> ( <i><b>Identifier</b></i> [<b>if</b> Expression] ) Block</pre>
+ *        <b>catch</b> ( <i><b>Identifier</b></i> [<b>if</b> Expression] ) Block
+ *        <b>catch</b> ( <i><b>ObjectPattern</b></i> ) Block
+ *        <b>catch</b> ( <i><b>ArrayPattern</b></i> ) Block</pre>
  */
 public class CatchClause extends AstNode {
 
-    private Name varName;
+    private AstNode varName;
```
rbri, line 21: at least the name of the var and the name of the getter/setter does no longer match the value

rhino/src/main/java/org/mozilla/javascript/IRFactory.java
```diff
@@ -1210,22 +1210,61 @@ private Node transformTry(TryStatement node) {
...
+                if (varName instanceof Name) {
+                    // Simple identifier
+                    varNameNode = parser.createName(((Name) varName).getIdentifier());
 
-                AstNode ccc = cc.getCatchCondition();
-                if (ccc != null) {
-                    catchCond = transform(ccc);
-                } else {
+                    AstNode ccc = cc.getCatchCondition();
+                    if (ccc != null) {
```
rbri, line 1224: depressive if - maybe you can swap the branches

gbrail:
Thanks!

rhino/src/main/java/org/mozilla/javascript/ast/CatchClause.java
andreabergia: I'm not sure what you mean here.

rhino/src/main/java/org/mozilla/javascript/IRFactory.java
andreabergia: I disagree, or rather, I don't particularly see the current version being particularly better or worse than the alternative. This to me reads about the same even if we invert the `if`.

```java
                    AstNode ccc = cc.getCatchCondition();
                    if (ccc != null) {
                        catchCond = transform(ccc);
                    } else {
                        catchCond = new EmptyExpression();
                    }
```

rhino/src/main/java/org/mozilla/javascript/ast/CatchClause.java
rbri: varName sounds like this is a Name but now it is a AstNode but maybe it is more clear to stay with the current naming.

rhino/src/main/java/org/mozilla/javascript/IRFactory.java
rbri: Personally i think this is easier to read but i can live with the 'depressive' version also

    AstNode ccc = cc.getCatchCondition();
    if (ccc == null) {
        catchCond = new EmptyExpression();
    } else {
        catchCond = transform(ccc);
    }

Issue #575, opened 2019-06-19
An error should be thrown when right-hand side of 'instanceof' is not callable


YaoHouyou:
Rhino version: 1.7.11, 1.7.10, 1.7.9
**Testcase:**
````
var foo = function () {
	try {
		throw SyntaxError();
	} catch (e) {
		var value = (e instanceof SyntaxError());
	}
};
 
foo();
````
**Command:**
````
java -jar rhino/rhino-1.7.11.jar -debug -version 200 testcase.js
````
-----
According to [ES5.1](http://www.ecma-international.org/ecma-262/5.1/index.html#sec-11.8.6), If right-value of `instanceof` does not have a internal method, throw a TypeError exception.  Therefore, I think rhino should throw an exception on line 5.

**Comment by ty5491003** [comment]
Please take a look at this example. @gbrail

**Comment by andreabergia** [comment]
This is obsolete, it now correctly throws

```
TypeError: Target of ``instanceof`` must be callable or have ``[Symbol.hasInstance]`` method.
```

The correct behavior was implemented in https://github.com/mozilla/rhino/pull/1751 and https://github.com/mozilla/rhino/pull/1872 afaik.

rbri:
Maybe we can add a test case before closing it.

rbri:
have added a simple test case in https://github.com/mozilla/rhino/pull/2196

rbri:
ok, test case adde, CI build is happy, will close this.

Thanks to all making this working.

PR #2196, opened 2025-12-04
add testcase for #575 that proofs this already works

rbri:


andreabergia (commented):
We already had the class `FunctionPrototypeSymbolHasInstanceTest`.

rbri:
@andreabergia will check that later this week

rbri:
> We already had the class `FunctionPrototypeSymbolHasInstanceTest`.

I think the class FunctionPrototypeSymbolHasInstanceTest is more about the left hand side of instanceof, this one is about the right hand side.

PR #2197, opened 2025-12-04
Don't create activations to track strictness

balajirrao:
This has 2 changes

- Remove the use of activations to keep track of strictness
- Remove `isTopLevelStrict` from `Context`. All tests pass, however I'm not a 100% confident on this one would appreciate a careful review.

This applies only to functions that have a 'use strict' directive in the body. Functions that are in strict top-levels don't use activations.

### DIFF
--- a/rhino/src/test/java/org/mozilla/javascript/tests/Bug782363Test.java
+++ b/rhino/src/test/java/org/mozilla/javascript/tests/Bug782363Test.java
@@ -11,6 +11,7 @@ import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertSame;
 import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
 
 import java.io.IOException;
 import java.util.HashSet;
@@ -133,6 +134,7 @@ public class Bug782363Test {
         test(338);
         try {
             test(339);
+            fail("test(339) did not throw");
         } catch (EvaluatorException e) {
             // may fail with 'out of locals' exception
         }


0xe (approved):

rhino/src/main/java/org/mozilla/javascript/Arguments.java
```diff
@@ -155,8 +155,7 @@ public Object get(int index, Scriptable start) {
     }
 
     private boolean sharedWithActivation(int index) {
-        Context cx = Context.getContext();
-        if (cx.isStrictMode()) {
+        if (activation.isStrict) {
```
andreabergia: Why are you looking at the activation here, and not `Context::isStrictMode`? I understand that, if we have arguments, we always have an activation, but it feels like we'd like to get rid of the "strict" flag in activations completely given this PR, no?
And I fear this might create problems with the debugger, since we _always_ inject activations when we have the debugger turned on.

rhino/src/main/java/org/mozilla/javascript/optimizer/BodyCodegen.java
```diff
@@ -196,6 +196,7 @@ private void initBodyGeneration() {
         enterAreaStartLabel = -1;
         generatorStateLocal = -1;
         savedHomeObjectLocal = -1;
+        parentStrictnessLocal = -1;
```
andreabergia, line 211: Nit: IMHO this could be a local variable, rather than a field.

rhino/src/main/java/org/mozilla/javascript/Interpreter.java
```diff
@@ -4828,6 +4830,10 @@ private static void exitFrame(Context cx, CallFrame frame, Object throwable) {
             ScriptRuntime.exitActivationFunction(cx);
         }
 
+        if (frame.fnOrScript.getDescriptor().getFunctionType() != 0) {
```
andreabergia, line 4982: Why aren't you exiting for scripts? Imagine (in our ServiceNow codebase) a script include inclusion - we _should_ restore the flag once we are done including it!

rhino/src/test/java/org/mozilla/javascript/tests/Bug782363Test.java
```diff
@@ -127,9 +127,9 @@ public void testConst() {
 
     @Test
     public void maxLocals() throws IOException {
-        test(339);
+        test(338);
```
andreabergia, line 134: A comment here explaining why you had to lower the number would be nice

rhino/src/main/java/org/mozilla/javascript/Arguments.java
aardvark179: > And I fear this might create problems with the debugger, since we always inject activations when we have the debugger turned on. 

Oh, right. Guess I'd better make a PR from my scratch branch… #2205 

Also, if we are interested in whether the function is strict, then the function descriptor carries that information.

rhino/src/main/java/org/mozilla/javascript/optimizer/BodyCodegen.java
balajirrao: I did try it out and it doesn't seem like an improvement.

rhino/src/main/java/org/mozilla/javascript/Interpreter.java
balajirrao: Good point! I looked into this a bit and it turns out strictness handling around this is buggy - activations weren't created for strict mode scripts at all. I'll fix it.

rhino/src/main/java/org/mozilla/javascript/Arguments.java
balajirrao: Right. We can't get rid of the strict flag in activations yet - there are (deprecated) public methods that depend on it. I'll deprecate the flag and change the usage sites to use function descriptors.

0xe:
@balajirrao, We have introduced new failures in script-decl-func-err-non-extensible.js and  script-decl-var-err.js in strict mode, only with the optimizing compiler. I think we should address those.

balajirrao:
> @balajirrao, We have introduced new failures in script-decl-func-err-non-extensible.js and script-decl-var-err.js in strict mode, only with the optimizing compiler. I think we should address those.

@0xe yeah, on it. I swear I thought I saw them all pass before pushing 😅

rhino/src/main/java/org/mozilla/javascript/Interpreter.java
balajirrao: @andreabergia I realise my last comment didn't make a lot of sense. 

We never created activations for scripts. This PR doesn't change this behaviour.

I took your comment to mean that not creating activations for scripts was probably a bug, so I tried to fix it. However, I found that it broke 2 test262 tests that had been passing. Upon close investigation, those tests had nothing to do with strictness - they just happened to rely on the "bug"

Now I'm not sure what the semantics of this kind of "script inclusion" should be. However, I found that `require()` also doesn't honor the script's strictness and this definitely seems to be a bug.

balajirrao:
The tests are passing now. See https://github.com/mozilla/rhino/pull/2197#discussion_r2721841596 for details.

rhino/src/main/java/org/mozilla/javascript/NativeCall.java
```diff
@@ -148,7 +148,7 @@ public Scriptable getHomeObject() {
 
     final JSFunction function;
     final Object[] originalArgs;
-    final boolean isStrict;
+    @Deprecated final boolean isStrict;
```
andreabergia: Why deprecated?

Can we remove it? Otherwise, this needs a comment imho.

balajirrao: @andreabergia there's a `NativeCall` constructor that sets this field - I forgot to mark that deprecated. 
I've pushed a commit with the change, please take a look!

aardvark179: The constructor isn't public. Don't deprecate it, get rid of it entirely, and the field.

andreabergia (approved):

rhino/src/main/java/org/mozilla/javascript/NativeCall.java
balajirrao: @aardvark179 The deprecated `createFunctionActivation` and `createArrowFunctionActivation` methods in `ScriptRuntime` call this constructor. I suppose these methods can be removed in 2.0 ?

aardvark179: I think those are only public because compiled classes are in a different package. I can't think of a good reason to create a `NativeCall` except as part of actually executing a function. [Not many references to this on GH](https://github.com/search?q=createFunctionActivation+language%3AJava+NOT+path%3A**%2FScriptRuntime.java+NOT+path%3A**%2FInterpreter.java+NOT+path%3A**%2F*CodeGen.java&type=code) once I exclude forks of our code.

balajirrao: Removed the deprecated methods and got rid of the `isStrict` field from `NativeCall`.

balajirrao:
@gbrail, would appreciate a look when you get a chance!

rhino/src/test/java/org/mozilla/javascript/tests/Bug782363Test.java
aardvark179: Does this still hold true? Like, did the test fail after this change. I know I've made changes to class compilation which have affected the number of locals and didn't touch this test. 😇

balajirrao: Yes, it is test failure that made me change this. I've pushed a commit tightening the assertions https://github.com/mozilla/rhino/pull/2197/changes/ca9cc6a8b94ce13aae09e41c4c77508a50ba3493.

gbrail:
I think this is good -- I've tried it and threw a few test cases at it generated by AI after a code review and it covers them all. I'm happy to merge this.

aardvark179:
Awesome. It definitely makes some other things easier down the line.

PR #2201, opened 2025-12-06
polish the `TypeInfo` system

ZZZank:
Current `TypeInfo` system works good enough in common use cases, but there still some rough corners / lack of design / overengineering that need to be polished.

All changes are documented in commit messages. Notable changes are:
- `TypeInfoFactory.NO_CACHE` is removed, because it's unable to handle self referencing type variables, e.g. `E extends Enum<E>`
- Support for `Outer<T1>.Inner` style parameterized type, including correct formatting and consolidation mapping generation.
- `TypeFormatContext` is redesigned to be more consistent and easy to use.

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/lc/type/TypeFormatContext.java
+++ b/rhino/src/main/java/org/mozilla/javascript/lc/type/TypeFormatContext.java
@@ -80,7 +80,9 @@ public interface TypeFormatContext {
      *
      * @param builder Formatted string of {@code type} will be pushed to this builder
      * @param type The type to be formatted
-     * @param declaring {@code true} if the context should format the result as if the type is being declared instead of being used. For example, in {@code E extends Enum<T>}, this param is {@code true} for the first E, and {@code false} for the second, nested E
+     * @param declaring {@code true} if the context should format the result as if the type is being
+     *     declared instead of being used. For example, in {@code E extends Enum<T>}, this param is
+     *     {@code true} for the first E, and {@code false} for the second, nested E
      */
     default void append(StringBuilder builder, TypeInfo type, boolean declaring) {
         if (type == TypeInfo.NONE) {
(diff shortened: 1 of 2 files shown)


rhino/src/main/java/org/mozilla/javascript/lc/type/TypeFormatContext.java
```diff
@@ -1,40 +1,118 @@
...
+    /// Full feature formatting context
+    ///
+    /// | Type | Full representation | Representation using this context |
+    /// | - | - | - |
+    /// | Class | java.lang.String | java.lang.String |
+    /// | Nested Class | Entry (in java.util.Map) | java.util.Map$Entry |
+    /// | Array | java.lang.String[] | java.lang.String[] |
+    /// | Parameterized | java.lang.List<java.lang.String> | java.util.List<java.lang.String> |
+    /// | Variable | T extends java.lang.String | T extends java.lang.String |
+    /// | [TypeInfo#NONE] | (No standard representation) | ? |
```
ZZZank: Check me: This Javadoc is written in Markdown style, because it's clearer than HTML table when displayed in source file, but this also makes it inconsistent with other Javadoc. Should I replace it with old style Javadoc (aka table via HTML tag) ?

rhino/build.gradle
```diff
@@ -102,7 +102,5 @@ decycle {
     // currently accepted cycles (may be resolved, when it is clear, how to separate the liveconnect stuff)
     ignoring from: "org.mozilla.javascript.lc.type.*", to: "org.mozilla.javascript.(Scriptable|ScriptableObject)"
     ignoring from: "org.mozilla.javascript.lc.type.TypeInfoFactory", to: "org.mozilla.javascript.lc.type.impl.factory.*"
-    ignoring from: "org.mozilla.javascript.lc.type.TypeInfo", to: "org.mozilla.javascript.lc.type.impl.*"
-    ignoring from: "org.mozilla.javascript.lc.type.TypeFormatContext", to: "org.mozilla.javascript.lc.type.impl.ClassSignatureFormatContext"
-
+    ignoring from: "org.mozilla.javascript.lc.type.*", to: "org.mozilla.javascript.lc.type.impl.*"
```
aardvark179, line 105: I think it's worth taking a look at the overall recycle rules for our base package `lc.type` as we keep having to fix them during development. If the intention is that they reflect the future separation of the `lc` components from the main Rhino module then `lc` should have access to whatever is exported from the main module, and we really shouldn't care about anything more precise than that.

gbrail:
Thanks for this, seems like a good improvement!

PR #2202, opened 2025-12-08
cleanup ContextFactory in test

ZZZank:
by using Utils helper instead of creating yet another new ContextFactory class for each test


rbri:
Run './gradlew spotlessApply' to fix all violations.

ZZZank:
It seems I accidentally enabled one buggy test, and good news is that this is the only failing test in CI

gbrail:
Great cleanup, and thanks!

I would love to see a similar usability improvement to the whole codebase -- the whole inheritance-based ContextFactory is one of those things from 1996 that I think Rhino could do without!

PR #2203, opened 2025-12-08
Intl402

rbri:
This enables the intl402 test for the 262 test suite and also provides a first fix for https://github.com/mozilla/rhino/issues/535

This is based on a rewritten test262.properties generator because the current impl has some problem and I was not able to fix that in the existing code. Problems are mainly occure when enabling the intl402 test; different from the other test this part has test files at the folder root also. The stats for exluded folders now always available and the order of excluded folders is a bit more natural.

The new version is much slower but i think this is not a real problem with this suite.

Not so many intl402 tests are passing so far but this is a beginning (hopefully)

### DIFF
--- a/tests/src/test/java/org/mozilla/javascript/tests/es6/NativeNumber2Test.java
+++ b/tests/src/test/java/org/mozilla/javascript/tests/es6/NativeNumber2Test.java
@@ -4,13 +4,10 @@
 
 package org.mozilla.javascript.tests.es6;
 
-import static org.junit.Assert.assertEquals;
-
 import java.util.Locale;
 import org.junit.Test;
 import org.mozilla.javascript.Context;
 import org.mozilla.javascript.EcmaError;
-import org.mozilla.javascript.Scriptable;
 import org.mozilla.javascript.testutils.Utils;
 
 public class NativeNumber2Test {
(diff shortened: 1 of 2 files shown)


gbrail:
Well the new test262 framework works, and that's pretty cool. Let me take some time to read it more carefully (I'd be happy if someone else does too) and make sure it still supports the old use cases!

aardvark179:
I'll try and take a look at this tomorrow.

gbrail:
I think this is good and I appreciate the forward progress. Interested if anyone else wants to look, otherwise I'm happy with it!

aardvark179:
I think it looks good, I've got some changes to run 262 cases in debug mode which would need some to rebase but it doesn't look too bad.

tests/src/test/java/org/mozilla/javascript/tests/es6/NativeNumber2Test.java
```diff
@@ -0,0 +1,262 @@
...
+        assertWithAllModes402("\u0660", "1e-10.toLocaleString('ar-SA')");
+
+        assertWithAllModes402("\u0660\u066b\u0660\u0661", "1e-2.toLocaleString('ar-SA')");
+
+        assertWithAllModes402(
+                "\u0644\u064a\u0633\u00a0\u0631\u0642\u0645", "NaN.toLocaleString('ar-SA')");
+        assertWithAllModes402("\u221e", "Infinity.toLocaleString('ar-SA')");
+    }
+
+    private void assertWithAllModes402(final Object expected, final String script) {
```
andreabergia: Should these two go in `Utils`?

andreabergia:
> The new version is much slower but i think this is not a real problem with this suite.

What do you mean? It already takes well over two minutes for me, so "much slower" is really worrying 😅

rbri:
> > The new version is much slower but i think this is not a real problem with this suite.
> 
> What do you mean? It already takes well over two minutes for me, so "much slower" is really worrying 😅

:hourglass: 
Ok, have done a test - rewriting the file requires 2s with the new writer. I think it is ok compared to the time you wait
* for Santa or 
* the test suite itself to finish

aardvark179:
Yeah, seems fine to me locally, even after enabling the debugger tests.

tests/src/test/java/org/mozilla/javascript/tests/es6/NativeNumber2Test.java
rbri: Done

gbrail:
Thanks! I've had a chance to look at this and try it and it looks good. I'm not going to "squash" it because it touches a few different things -- it probably would have been a good idea to separate the Test262 rewrite with the other changes, but I'm good. Thanks!

PR #2204, opened 2025-12-08
Initialise `CallSite` against top level scope.

aardvark179:
`NativeCallsite.init` is currently called with the scope being the `Error` constructor, but we still end up with `NativeCallSite`'s constructor in the top level scope under the name `CallSite`. This seems wrong, and fixing this doesn't appear to break anything, but I thought it was worth its own PR because it is an odd change, and deserves some scrutiny.


rbri:
do we need a test for that?

aardvark179:
We already have tests that it's present in top scope. I was hoping @gbrail might have a good idea about what Error functionality it is involved in so we can write a test for its behaviour, rather than its presence.

gbrail:
This is going back a bit, but it was in there so that we could emulate some of the non-standard Error stack information like Node and V8 do.  I don't believe that it requires that the CallSite constructor be in the top level scope, though, and I see that in Node it is not. I am fine with this change and since you reviewed it I think we should take it!

PR #2205, opened 2025-12-09
Avoid the need for activation frames when using the debugger.

aardvark179:
The requirement to use activation frames when the debugger is in operation can cause a significant slowdown, and has previously been a way in which behavioural differences can creep in. This changes refactors things to use a proxy scriptable object on the assumption that debug operations are less common than interpreter operations.

V8 benchmarks run with the debugger before this change:

```
Benchmark                  (interpreted)  Mode  Cnt       Score      Error  Units
V8Benchmark.boyer                   true  avgt    5  273172.154 ± 4991.749  us/op
V8Benchmark.cryptoDecrypt           true  avgt    5  480227.320 ± 8768.637  us/op
V8Benchmark.cryptoEncrpyt           true  avgt    5   26817.109 ± 1563.891  us/op
V8Benchmark.deltaBlue               true  avgt    5   20075.138 ±  513.126  us/op
V8Benchmark.earley                  true  avgt    5   16331.189 ±  349.191  us/op
V8Benchmark.rayTrace                true  avgt    5   86539.945 ± 2660.865  us/op
V8Benchmark.regExp                  true  avgt    5  152616.481 ± 4618.534  us/op
V8Benchmark.richards                true  avgt    5   11844.159 ±  315.925  us/op
V8Benchmark.splay                   true  avgt    5    4837.759 ±  392.140  us/op
```
The same benchmarks run after the change:
```
Benchmark                  (interpreted)  Mode  Cnt       Score       Error  Units
V8Benchmark.boyer                   true  avgt    5  214829.091 ±  1482.869  us/op
V8Benchmark.cryptoDecrypt           true  avgt    5  312138.059 ± 11349.310  us/op
V8Benchmark.cryptoEncrpyt           true  avgt    5   16899.571 ±  2247.409  us/op
V8Benchmark.deltaBlue               true  avgt    5   18807.295 ±   146.184  us/op
V8Benchmark.earley                  true  avgt    5   14496.497 ±    48.760  us/op
V8Benchmark.rayTrace                true  avgt    5   72628.804 ±   280.494  us/op
V8Benchmark.regExp                  true  avgt    5  150013.411 ±  3014.196  us/op
V8Benchmark.richards                true  avgt    5   10218.532 ±   460.681  us/op
V8Benchmark.splay                   true  avgt    5    4230.112 ±   848.120  us/op
```

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/Interpreter.java
+++ b/rhino/src/main/java/org/mozilla/javascript/Interpreter.java
@@ -492,14 +492,28 @@ public final class Interpreter extends Icode implements Evaluator {
         }
     }
 
+    /**
+     * This class is intended as proxy to give {@link DebugFrame} access to the contents of local
+     * variables. We take this approach rather than forcing the interpreter to introduce activation
+     * frames because it is faster (assuming local variable manipulation by the interpreter is more
+     * common than inspection by the debugger) and it reduces the chance that programs might
+     * evexcute differently in debug mode.
+     */
     private static class DebugScope implements Scriptable {
         private final CallFrame frame;
         private volatile Map<String, Integer> offsets;
 
+        /** Create a new debug scope associated with a particular call frame. */
         private DebugScope(CallFrame frame) {
             this.frame = frame;
         }
 
+        /**
+         * Populate the map associating names to variable indices. Most names should have been made
+         * unique as part of the compilation process, but arguments with duplicate names will not
+         * have been.The map is build so that duplicate argument names resolve to the last index as
+         * this is also what the compiler does - at least once we are past setting default values.
+         */
         private Map<String, Integer> getOffsets() {
             if (offsets == null) {
                 offsets = buildOffsets(frame);


rhino/src/main/java/org/mozilla/javascript/Interpreter.java
```diff
@@ -471,6 +472,130 @@ && compareDescs(fnOrScript.getDescriptor(), other.fnOrScript.getDescriptor())
...
+            if (value instanceof Double && Double.isFinite((Double) value)) {
+                stack[offset] = DOUBLE_MARK;
+                sDbl[offset] = ((Double) value);
+            } else {
+                stack[offset] = value;
+            }
+        }
+    }
+
+    private static class DebugScope implements Scriptable {
```
rbri, line 502: This looks promising...

I like to suggest to add some jdoc about the motivation behind this, otherwise it will be hard to know why this is here in some months from now.

aardvark179: Yeah, that seems reasonable. I'll try and get that done in the next few days.

rbri (approved):

gbrail:
Thanks, this looks good!

PR #2207, opened 2025-12-12
Support for rest parameters in destructuring

0xe:
Adds support for rest parameters in destructuring patterns (Object & Array). 

The remaining failures in test262 from enabling "object-rest" are due to pending items with:

- proxy support
- async iterators
- let & const

### DIFF
--- a/tests/testsrc/test262.properties
+++ b/tests/testsrc/test262.properties
@@ -3896,7 +3896,7 @@ language/expressions/addition 0/48 (0.0%)
 
 language/expressions/array 0/52 (0.0%)
 
-language/expressions/arrow-function 63/343 (18.37%)
+language/expressions/arrow-function 61/343 (17.78%)
     dstr/ary-ptrn-elem-ary-elision-init.js
     dstr/ary-ptrn-elem-id-init-fn-name-arrow.js
     dstr/ary-ptrn-elem-id-init-fn-name-class.js
@@ -3924,7 +3924,6 @@ language/expressions/arrow-function 63/343 (18.37%)
     dstr/dflt-obj-ptrn-id-init-fn-name-cover.js
     dstr/dflt-obj-ptrn-id-init-fn-name-fn.js
     dstr/dflt-obj-ptrn-id-init-fn-name-gen.js
-    dstr/dflt-obj-ptrn-prop-eval-err.js
     dstr/obj-init-null.js
     dstr/obj-init-undefined.js
     dstr/obj-ptrn-id-init-fn-name-arrow.js
@@ -3936,7 +3935,6 @@ language/expressions/arrow-function 63/343 (18.37%)
     dstr/obj-ptrn-id-init-throws.js
     dstr/obj-ptrn-id-init-unresolvable.js
     dstr/obj-ptrn-list-err.js
-    dstr/obj-ptrn-prop-eval-err.js
     syntax/arrowparameters-bindingidentifier-yield.js non-strict
     syntax/arrowparameters-cover-formalparameters-yield.js non-strict
     syntax/arrowparameters-cover-initialize-2.js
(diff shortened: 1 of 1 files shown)


gbrail:
This is great progress -- when you all get back to work, let me know if someone has time for a code review or if they're already OK with this.

rhino/src/main/java/org/mozilla/javascript/Interpreter.java
```diff
@@ -989,6 +989,27 @@ static <T extends ScriptOrFn<T>> void dumpICode(
                         out.println(tname + " \"" + str + '"');
                         break;
                     }
+                case Icode_OBJECT_REST:
```
aardvark179, line 992: I'm not sure I like the way this byte code is structured, it's the first time we've had a code whose length depend on its data in this way, but I understand how you got there. How big a change would it be to restructure the two complex literal types in `InterpreterData` (regexes and template literals) as a single `Object[]` and put your data structure in that?

**Edit:** Ah, I see the class compiled version is doing the same sort of thing. Maybe this is the point to start moving this sort of complex data onto the `JSDesccriptor` and see if that helps tackle the big `TODO` in the `ScriptRuntime` change.

rhino/src/main/java/org/mozilla/javascript/ScriptRuntime.java
```diff
@@ -899,6 +899,115 @@ public static Object[] padAndRestArguments(
...
+     * Helper for object rest properties in destructuring. Creates a shallow copy of the source
+     * object excluding the specified keys.
+     *
+     * @param cx the current Context
+     * @param scope the current scope
+     * @param source the source object
+     * @param excludeKeys array of property names to exclude
+     * @return a new object with all properties except the excluded ones
+     */
+    // TODO - simplify:
```
aardvark179: That's a pretty big `TODO` list. 🙂

0xe:
> Overall this looks good, but I there's a big TODO list on the `ScriptRuntime` change, and byte code / class compilation seems complex in some ways we might avoid. I think it's worth seeing if we improve those areas.

Thanks for the review. I'll work on addressing the items in the TODO list. I kinda left it for another day as this PR was already getting big. But, happy to address them. :)

rhino/src/main/java/org/mozilla/javascript/ScriptRuntime.java
0xe: Thanks, addressed the problems in the TODO list. Will work on restructuring bytecode next.

rhino/src/main/java/org/mozilla/javascript/Interpreter.java
0xe: This was a bug with `dumpIcode`. Earlier I was directly storing computed keys and static keys in arrays[] and their counts. @andreabergia had suggested me to use `itsliteralIds` instead. I'd forgotten to update branch in `dumpIcode` along with those changes.

This is how the bytecodes emitted looks now. For:

```javascript
  const obj = {a: 1, b: 2, c: 3};
  const {a, ...rest} = obj;
```

```
...
      [43] BINDNAME
      [44] REG_STR_C2 "$0"
      [45] NAME
      [46] REG_IND_C2
      [47] OBJECT_REST excluding [static: "a"; computed: 0]
// [48] would be just 0 byte for computed key count
      [49] REG_STR_C3 "rest"
      [50] SETCONST
...
```

0xe:
> > Overall this looks good, but I there's a big TODO list on the `ScriptRuntime` change, and byte code / class compilation seems complex in some ways we might avoid. I think it's worth seeing if we improve those areas.
> 
> Thanks for the review. I'll work on addressing the items in the TODO list. I kinda left it for another day as this PR was already getting big. But, happy to address them. :)

@aardvark179, I've addressed the items in the TODO. Please let me know if this looks okay.

gbrail:
This is great -- thanks for all the work! Merging it manually because of the small conflict.

PR #2210, opened 2025-12-12
Created AGENTS.md and CLAUDE.md

andreabergia:
I've created an AGENTS.md file with some basic rules and information about the project. I've tried to keep it very synthetic, according to the best practices I've been reading about.

I've also created a CLAUDE.md because apparently Claude Code [doesn't (yet?) support](https://github.com/anthropics/claude-code/issues/6235) the standard AGENTS.md file. According to what I've been [reading](https://www.reddit.com/r/GithubCopilot/comments/1nee01w/agentsmd_vs_claudemd/), it seems the best thing to do is to either symlink it, or create a one-line CLAUDE.md that references the standard AGENTS.md - which is what I have done.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,52 @@
+# Rhino
+
+Rhino is an open source (MPL 2.0) JavaScript engine, implemented in Java 11. The build system used is gradle, via the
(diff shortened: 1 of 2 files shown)


gbrail:
Looks good to me, maybe others experienced with agentic coding can suggest changes as we go!

PR #2215, opened 2025-12-16
Use function descriptors for `NativeBoolean` and `NativeObject` methods.

aardvark179:
This was a change I prototyped a while ago but am resurrecting now to resolve problems found while separating scopes and objects.

## Original motivation

The original idea behind this was to reduce the number of different function implementations and to enable faster initialisation in the future. The idea is to build a static `ClassDescriptor` once and then initialise `JSFunction`s from that when we init methods. At the moment this doesn't really save us any time, but does put us in a position where we could create prototypes of a known shape in the future and thus save significant time initialisation, or introduce lazy initialisation of individual function objects.

@rbri I considered making this annotation based, and using `LambdaMetaFactory` to create lambda's, but decided against it because it isn't supported on Android, but I could build some additional methods for building a descriptor using that method. I think you mentioned that you have a lot of annotated methods for different browser modes, and maybe this could be supported by building multiple class descriptors and then choosing the one to initialise with based on context options or something like that?

## Reason to resurrect this now.

The problem we have is that our lambdas do not pass the function object itself, so to handle creation of objects within built in function we set the scope to be `LambdaFunction`s declaration scope (the callee's scope) in the `LambdaFunction#call` method. This is fine for all standard built in functions, but isn't ideal for `importClass` and `importPackage` which operate on the caller's scope. We dodged this issue because we resolved `this` incorrectly on the caller's side (non-strict functions turn `undefined` into `globalThis`) but  as soon as we start resolving `this` as part of the `call` on the callee we hit a problem.

The internal `JSCodeExec` lambdas used by `JSFunction`s however are passed the original function object, and if marked as strict will not touch the `this` argument during the calling process, and so are a much closer fit for built in functions, and allow the import functions to do their non-standard things.

## Why `NativeBoolean`?

It's a relatively small class, with differing behaviour on the constructor depending on whether it's called as a constructor or a function so it helped verify all the important behaviour., and the changes for `ImporterTopLevel` are built on top of the the separation of top level scopes and `globalThis`.

## `NativeObject` addition

I've added a conversion of`NativeObject` to use descriptors as this was useful for finding bugs in the implementation and other potential problems in converting classes acrosss, as `Object` requires more features than `Boolean`. I think there are still a couple of areas where we will need a slightly richer API (Array unscopeables springs to mind) but I think those can be covered in follow up PRs.

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/NativeObject.java
+++ b/rhino/src/main/java/org/mozilla/javascript/NativeObject.java
@@ -33,98 +33,65 @@ public class NativeObject extends ScriptableObject implements Map {
     public static final String PROTO_PROPERTY = "__proto__";
     public static final String PARENT_PROPERTY = "__parent__";
 
-    static LambdaConstructor init(Context cx, Scriptable s, boolean sealed) {
-        LambdaConstructor ctor =
-                new LambdaConstructor(
-                        s,
+    private static ClassDescriptor LEGACY_DESCRIPTOR;
+    private static ClassDescriptor ES6_DESCRIPTOR;
+
+    static {
+        var builder =
+                new ClassDescriptor.Builder(
                         CLASS_NAME,
                         1,
                         NativeObject::js_constructorCall,
-                        NativeObject::js_constructor) {
-                    @Override
-                    public Scriptable construct(Context cx, Scriptable scope, Object[] args) {
-                        return js_constructor(cx, scope, args);
-                    }
-                };
+                        NativeObject::js_constructor);
 
-        var proto = new NativeObject();
-        proto.setParentScope(s);
-        ctor.setPrototypeProperty(proto);
-        proto.defineProperty("constructor", ctor, DONTENUM);
+        builder.withCtorMethod("getPrototypeOf", 1, NativeObject::js_getPrototypeOf);
+        builder.withCtorMethod("keys", 1, NativeObject::js_keys);
+        builder.withCtorMethod("getOwnPropertyNames", 1, NativeObject::js_getOwnPropertyNames);
+        builder.withCtorMethod("getOwnPropertySymbols", 1, NativeObject::js_getOwnPropertySymbols);
+        builder.withCtorMethod("getOwnPropertyDescriptor", 2, NativeObject::js_getOwnPropDesc);
+        builder.withCtorMethod("getOwnPropertyDescriptors", 1, NativeObject::js_getOwnPropDescs);
+        builder.withCtorMethod("defineProperty", 3, NativeObject::js_defineProperty);
+        builder.withCtorMethod("isExtensible", 1, NativeObject::js_isExtensible);
+        builder.withCtorMethod("preventExtensions", 1, NativeObject::js_preventExtensions);
+        builder.withCtorMethod("defineProperties", 2, NativeObject::js_defineProperties);
+        builder.withCtorMethod("create", 2, NativeObject::js_create);
+        builder.withCtorMethod("isSealed", 1, NativeObject::js_isSealed);
+        builder.withCtorMethod("isFrozen", 1, NativeObject::js_isFrozen);
+        builder.withCtorMethod("seal", 1, NativeObject::js_seal);
+        builder.withCtorMethod("freeze", 1, NativeObject::js_freeze);
+        builder.withCtorMethod("assign", 2, NativeObject::js_assign);
+        builder.withCtorMethod("is", 2, NativeObject::js_is);
+        builder.withCtorMethod("groupBy", 2, NativeObject::js_groupBy);
+        builder.withProtoMethod("toString", 0, NativeObject::js_toString);
+        builder.withProtoMethod("toLocaleString", 0, NativeObject::js_toLocaleString);
+        builder.withProtoMethod("__lookupGetter__", 1, NativeObject::js_lookupGetter);
+        builder.withProtoMethod("__lookupSetter__", 1, NativeObject::js_lookupSetter);
+        builder.withProtoMethod("__defineGetter__", 2, NativeObject::js_defineGetter);
+        builder.withProtoMethod("__defineSetter__", 2, NativeObject::js_defineSetter);
+        builder.withProtoMethod("hasOwnProperty", 1, NativeObject::js_hasOwnProperty);
+        builder.withProtoMethod("propertyIsEnumerable", 1, NativeObject::js_propertyIsEnumerable);
+        builder.withProtoMethod("valueOf", 0, NativeObject::js_valueOf);
+        builder.withProtoMethod("isPrototypeOf", 1, NativeObject::js_isPrototypeOf);
+        builder.withProtoMethod("toSource", 0, ScriptRuntime::defaultObjectToSource);
 
-        defOnCtor(ctor, s, "getPrototypeOf", 1, NativeObject::js_getPrototypeOf);
-        if (Context.getCurrentContext().version >= Context.VERSION_ES6) {
-            defOnCtor(ctor, s, "setPrototypeOf", 2, NativeObject::js_setPrototypeOf);
-            defOnCtor(ctor, s, "entries", 1, NativeObject::js_entries);
-            defOnCtor(ctor, s, "fromEntries", 1, NativeObject::js_fromEntries);
-            defOnCtor(ctor, s, "values", 1, NativeObject::js_values);
-            defOnCtor(ctor, s, "hasOwn", 1, NativeObject::js_hasOwn);
-        }
-        defOnCtor(ctor, s, "keys", 1, NativeObject::js_keys);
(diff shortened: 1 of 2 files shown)


aardvark179:
Apparently I've fixed one thing around constructors and reflection as part of this which makes `isConstructor` in the 262 test harness pass its test.

rhino/src/main/java/org/mozilla/javascript/AbstractEcmaObjectOperations.java
andreabergia: Do we still need these special cases, considering the `if` you added below?

rhino/src/main/java/org/mozilla/javascript/ClassDescriptor.java
```diff
@@ -0,0 +1,165 @@
...
+
+/**
+ * This class represents a descriptor for a constructor and its prototype methods which can then be
+ * used to build a real constructor function and prototype during standard object initialisation.
+ */
+public class ClassDescriptor {
+
+    private final JSDescriptor<JSFunction> ctorDesc;
+    private final Map<String, JSDescriptor<JSFunction>> ctorDescs;
+    private final Map<String, JSDescriptor<JSFunction>> protoDescs;
```
andreabergia: Should we support symbols as well?

rhino/src/main/java/org/mozilla/javascript/NativeBoolean.java
```diff
@@ -16,23 +16,25 @@ final class NativeBoolean extends ScriptableObject {
...
-                        NativeBoolean::js_constructor);
-        constructor.setPrototypePropertyAttributes(DONTENUM | READONLY | PERMANENT);
         // Boolean is an unusual object in that the prototype is itself a Boolean
-        constructor.setPrototypeScriptable(new NativeBoolean(false));
-
-        constructor.definePrototypeMethod(scope, "toString", 0, NativeBoolean::js_toString);
-        constructor.definePrototypeMethod(scope, "toSource", 0, NativeBoolean::js_toSource);
-        constructor.definePrototypeMethod(scope, "valueOf", 0, NativeBoolean::js_valueOf);
+        var constructor = DESCRIPTOR.buildConstructor(scope, new NativeBoolean(false));
+        constructor.setPrototypePropertyAttributes(DONTENUM | READONLY | PERMANENT);
```
andreabergia: Is this something common for all builtins? Should we move it to the `ClassDescriptor`?

rhino/src/main/java/org/mozilla/javascript/NativeBoolean.java
```diff
@@ -63,28 +65,32 @@ public Object getDefaultValue(Class<?> typeHint) {
...
     }
 
-    private static NativeBoolean js_constructor(Context cx, Scriptable scope, Object[] args) {
+    private static NativeBoolean js_constructor(
+            Context cx, JSFunction fun, Object nt, Scriptable s, Object thisObj, Object[] args) {
         boolean b = ScriptRuntime.toBoolean(args.length > 0 ? args[0] : Undefined.instance);
-        return new NativeBoolean(b);
+        var res = new NativeBoolean(b);
+        res.setPrototype((Scriptable) fun.getPrototypeProperty());
+        res.setParentScope(s);
```
andreabergia, line 73: I assume `nt` is the `new.target` and I wonder if - in the future - either of this should be taken from the `nt` rather than the fun...

(I really didn't think too much about it, I'm sure I could answer this myself, but it's an interesting question I think)

rhino/src/main/java/org/mozilla/javascript/ClassDescriptor.java
```diff
@@ -0,0 +1,165 @@
...
+
+        private JSDescriptor<JSFunction> buildDescriptor(
+                String name, int length, JSCode<JSFunction> code) {
+            return buildDescriptor(name, length, code, null);
+        }
+
+        private JSDescriptor<JSFunction> buildDescriptor(
+                String name, int length, JSCode<JSFunction> call, JSCode<JSFunction> construct) {
+            var builder = new JSDescriptor.Builder<JSFunction>();
+            builder.code = identitiyBuilder(call);
```
andreabergia: Typo: 


```suggestion
            builder.code = identityBuilder(call);
```

rhino/src/main/java/org/mozilla/javascript/ClassDescriptor.java
aardvark179: Yeah, we do need to. `Boolean.prototype` is unusual in that it doesn't have any symbol properties.

rhino/src/main/java/org/mozilla/javascript/AbstractEcmaObjectOperations.java
aardvark179: Given I also found in another PR that we need another case for proxy functions I'm tempted to say this belongs as part of the `Function` interface.

rhino/src/main/java/org/mozilla/javascript/NativeBoolean.java
aardvark179: Yeah, I think it's common to everything. I'll move it.

rhino/src/main/java/org/mozilla/javascript/NativeBoolean.java
aardvark179: For ordinary objects setting the prototype is specified as part of [OrdinaryCreateFromConstructor](https://tc39.es/ecma262/#sec-ordinarycreatefromconstructor), but it's not defined as part of [BuiltinCallOrConstruct](https://tc39.es/ecma262/#sec-builtincallorconstruct), instead the behaviour is defined on the individual built in constructors, and I'm not sure if there are special cases hiding in there.

rhino/src/main/java/org/mozilla/javascript/ClassDescriptor.java
aardvark179: Fixed.

rhino/src/main/java/org/mozilla/javascript/NativeBoolean.java
aardvark179: Done.

gbrail:
I like the ClassDescriptor concept and I can see where it will take us, and I think we should go ahead and finish this work. There's one more thing I'd like to see and a question.

First of all, it seems like this could be the foundation of our next-generation way to write JS functionality in Java, as it'll be a lot more readable and require less boilerplate than our raw LambdaConstructor approach. It'll also support future optimizations I can see like object shapes.

Given all that, it needs comprehensive JavaDoc at the class and function level that explains how to do it.

Second, I see that a lot of the property attributes like DONTENUM, etc are hard-coded. That's probably fine if we only use it to define our own native types, but I am sure that as we use it for more things we'll need to add in ways to override those defaults. Ats the very least, right now we should document them.

Thanks!

aardvark179:
Okay. I'll work on Java docs for this.

I'd also like to explore if we can do this in an efficient way using annotations with an optional user provided handler passed in (I think this is all possible in Android API v26 and upwards as they added enough MethodHandle support). This might further reduce the boiler plate while providing a simple model for @rbri to transition away from the old annotation processing model (it would likely be useful for us internally as well).

rbri:
@aardvark179 there was no time to have a really detailed look at this... but maybe there was a misunderstanding. HtmlUnit does NOT use the annotation based java class mapping as provided by Rhino (JSConstructor/JSFunction/JSGetter/JSSetter/JSStaticFunction).

Instead i have my own annotations (https://github.com/HtmlUnit/htmlunit/tree/master/src/main/java/org/htmlunit/javascript/configuration - the classes prefixed by 'Jsx'). When starting HtmlUnit a fixed list of classes (https://github.com/HtmlUnit/htmlunit/blob/master/src/main/java/org/htmlunit/javascript/configuration/JavaScriptConfiguration.java) is parsed for these annotations and a configuration object (tree) is created in memory. There is is https://github.com/HtmlUnit/htmlunit/blob/master/src/main/java/org/htmlunit/javascript/configuration/ClassConfiguration.java for each supported js class for each supported browser. Every time the window content gets refreshed these configuration (the one for the currently simulated browser) is used to setup all the prototypes, ctors and so on.

I guess this approach s similar to your one. For HtmlUnit i have no real nead for enhancing Rhino in regard for this. I guess i can replace (maybe stepwise) the current setup based on MemberBox/reflection with lambdas.

aardvark179:
Okay, I'll go with just tidying up and documenting this approach. I'll likely add a reflection based way to build the descriptor because some of our code base is structured like that, but it can be a follow up PR.

aardvark179:
I've cleaned up the API a little, added Javadocs, and added a conversion of `NativeObject` which helped tease out some issues.

gbrail:
I think this is a great simplification. I'm OK with it. Does anyone else have concerns?

Otherwise, please rebase again (sorry) and we can move forward.

gbrail:
Thanks!

PR #2216, opened 2025-12-17
make NativeCall attributes final

rbri:
small change but hopefully makes the code a bit more clear and helps the JIT


rbri:
strange somthing is wrong here, will close this and try it again

PR #2217, opened 2025-12-17
make NativeCall attributes final

rbri:
mall change but hopefully makes the code a bit more clear and helps the JIT


gbrail:
Looks good, thanks!

PR #2222, opened 2025-12-18
make arguments a function prototype property

rbri:
This is the second step in the arguments refactoring to bring the arguments handling closer to the spec.

* we have some test cases containg // todo; these todo marks the real world expectation

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/BoundFunction.java
+++ b/rhino/src/main/java/org/mozilla/javascript/BoundFunction.java
@@ -48,13 +48,13 @@ public class BoundFunction extends BaseFunction {
 
     @Override
     Object getArguments() {
-        throw ScriptRuntime.typeErrorById("msg.op.not.allowed");
+        ScriptRuntime.ThrowTypeError.throwNotAllowed();
+        return null;
     }
 
     @Override
     void setArguments(Object caller) {
-        // todo
-        throw ScriptRuntime.typeErrorById("msg.op.not.allowed");
+        ScriptRuntime.ThrowTypeError.throwNotAllowed();
     }
 
     @Override
--- a/rhino/src/main/java/org/mozilla/javascript/BaseFunction.java
+++ b/rhino/src/main/java/org/mozilla/javascript/BaseFunction.java
@@ -751,7 +751,7 @@ public class BaseFunction extends ScriptableObject implements Function {
             return null;
         }
         if (activation.isStrict && cx.getLanguageVersion() >= Context.VERSION_ES6) {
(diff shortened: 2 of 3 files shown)


aardvark179:
This confused me for a minute because `Function.prototype.arguments` is one thing that really isn't in the spec, but I see what you're doing, and I'm happy to see things move in this direction.

rbri:
@aardvark179 as mentioned in our meeting I spend my whole time hunting for this regression https://github.com/HtmlUnit/htmlunit/issues/1064. My current assumption is a problem with the caller property being visible to some functions because it is NOT defined on the function prototype but instead on every function.
Had a hard time to understand, that only the move of the 'argument' property is more or less simple. For the 'caller' we have to rethink the activation support and other stuff. Therefore I decide to let this in core-js for the moment.

The good news: after the move of the 'arguments' property we are a bit more closer to the browsers. And after I have applied this PR to my fork, I was able to change the way the caller property is implemented also (https://github.com/HtmlUnit/htmlunit-rhino-fork/commit/88a343c89e1131e2c6f2f2cf76818f45eef20806).

And with this changes, I got a detailed test suite of 'caller' tests working https://github.com/HtmlUnit/htmlunit-core-js/commit/8769e0ed429471e945257f47b6b22451fbb18c95; also some more plain HtmlUnit tests are now passing.

Next step is to check with the jenkins folks, if this solves the problem.... but this is something for the next days.

rhino/src/main/java/org/mozilla/javascript/BoundFunction.java
```diff
@@ -40,8 +40,21 @@ public BoundFunction(
...
+    }
+
+    @Override
+    Object getArguments() {
+        throw ScriptRuntime.typeErrorById("msg.op.not.allowed");
+    }
+
+    @Override
+    void setArguments(Object caller) {
+        // todo
```
andreabergia: What is supposed to happen here, that there is a `todo` for it, I wonder...? 🙂

rbri: throw ScriptRuntime.typeErrorById("msg.op.not.allowed");

is called far to many times (i guess), i have to check if we can have it as static function

rbri:
Jenkins is happy with this - https://github.com/HtmlUnit/htmlunit/issues/1064#issuecomment-3692832879

gbrail:
This looks good to me and I lost track of how old it is. Thanks!

gbrail:
Thanks for doing this, I think it moves things forward!

rbri:
> This looks good to me and I lost track of how old it is. Thanks!

it's from last year :-D - thanks for merging

PR #2225, opened 2025-12-18
Run test262 suite with debugger.

aardvark179:
Emable running the 262 tests in debug mode to ensure we don't regress in behavioural difference.

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -70,14 +70,22 @@ jobs:
             test-java-version: "21"
             name: "Java 21 Build and Test"
             multithread: false
+            debug: false
           - java-version: "25"
             test-java-version: "25"
             name: "Java 25 Build and Test"
             multithread: false
+            debug: false
           - java-version: "25"
             test-java-version: "25"
             name: "Java 25 Multithreaded"
             multithread: true
+            debug: false
+          - java-version: "25"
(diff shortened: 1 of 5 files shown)


rbri (commented):
looks like another step forward for our test suite - great

rbri (approved):

.github/workflows/gradle.yml
```diff
@@ -95,7 +103,7 @@ jobs:
       env:
         RHINO_TEST_JAVA_VERSION: ${{ matrix.test-java-version }}
       run: >-
-        ./gradlew check ${{ matrix.multithread && '-Drhino.useThreadSafeObjectsByDefault=true' || '' }}
+        ./gradlew check ${{ matrix.multithread && '-Drhino.useThreadSafeObjectsByDefault=true' || '' }} ${{ matrix.debug && '-DgrunTest262Debug' || '' }}
```
gbrail: Are you sure it's "grunTest262Debug"? I don't think that's right!

gbrail (changes requested):

.github/workflows/gradle.yml
aardvark179: 🤦 I also realised that I needed more careful conditionals to avoid a null system property getting turned into a blank system property.

gbrail:
Thanks!

Issue #2192, opened 2025-12-03
NativeJavaList iterator unavailable in SafeStandardObject after commit b5e7f859


Rhython:
**Describe the bug**
After the changes introduced by commit b5e7f859, the iterator for NativeJavaList is not available in the SafeStandardObject environment. 

**To Reproduce**
Here is a code example that demonstrates the issue:
```java
try (Context cx = Context.enter()) {
    Scriptable scope = cx.initSafeStandardObjects();
    scope.put("value", scope, List.of("a", "b", "c"));
    Object result = cx.evaluateString(scope, "let res = ''; for (let x of value) { res += x; } res;", "test", 1, null);
    System.out.println(Context.toString(result));
}
```

This code produces the following confusing error:
```
org.mozilla.javascript.EcmaError: TypeError: next is not a function, it is object. (test.js#1)
```

**Expected behavior**
This code runs as expected in release 1.8.0.

**Additional context**
This appears to be a regression introduced by commit b5e7f859. Please investigate if iterators support for NativeJavaList should be restored in the SafeStandardObject environment.

rbri:
@Rhython i think this works as it should. initSafeStandardObjects() does  not install all the java bridging stuff - because that is unsafe.
If you like to have the java objects support you have to use initStandardObjects() instead.

Rhython:
@rbri I don't think so. NativeJavaList was designed to allow Lists to be used as JsArrays, and everything else works perfectly, except for the for-of loop. It's counterintuitive that `for in` works correctly while `for of` doesn't. Currently we can do this:
```
try (Context cx = Context.enter()) {
    Scriptable scope = cx.initSafeStandardObjects();
    scope.put("value", scope, List.of("a", "b", "c"));
    Object result = cx.evaluateString(scope, "let res = ''; for (let x in value) { res += value[x]; } res;", "test", 1, null); 
    System.out.println(Context.toString(result));
}
```
and it works fine. 
The `for...of` loop is supported by `JavaIterableIterator`. Only objects that implement the `java.lang.Iterable` interface can work in the normal loop manner. There is no reflection involved, and it is completely controllable. I don't think this introduces any security issues.

rbri:
@Rhython thanks for the detailed description, will have a closer look....

rbri:
From the description of initSafeStandardObjects()

_Initialize the standard objects, leaving out those that offer access directly to Java
classes. This sets up "scope" to have access to all the standard JavaScript classes, but does
[...]

PR #2226, opened 2025-12-19
revert the change from b5e7f859396f968a079b48cb83f1fc1bad31d33a and add some tests

rbri:
this is related to issue #2192


gbrail:
Thanks, I agree too!

Issue #2227, opened 2025-12-19
RegExp performance regression for 1.8.0++


gbrail:
I have been doing some historical benchmark runs and I see that the following change to the String / regexp code slowed down performance on the V8 "regexp" benchmark by about 15-20%. I didn't see any obvious thing that could be wrong, but it could be something to consider looking at for upcoming releases.

https://github.com/mozilla/rhino/commit/f6e5590bd082b1e40ebdb5f51ba602e744b4be93

@andreabergia thanks!

aardvark179:
I know the reasons, and I know @balajirrao has had a fix in progress for a while, I'll check on it tomorrow.

aardvark179:
Sorry, been busy today with some other urgent things that came up.

gbrail:
I don't think this is critical for the release, rather that we take our
time and do it right, I was just trying to record it!

rbri:
#2246 is merged, can we close this?

aardvark179:
I think so, yes.

rbri:
Ok, closing this as completed

PR #2233, opened 2025-12-21
Regexp performance regression fixes.

aardvark179:
@gbrail here is the stuff @balajirrao and I had worked on a while ago, but never made it to a PR due to other commitments. I believe this should resolve almost all of the regression mentioned in #2227. The main idea is to take a fairly optimised approach for the common cases, while still being able to take a slower approach where needed for spec compliance.


gbrail:
Thanks! This is great and I will take a look -- but I also think that we should go ahead and publish 1.9.0 without it, since I think we're in a good place, and have lots of other big changes coming. Let me know if you think differently!

aardvark179:
I'll let @balajirrao create a new PR for this since he has done the work to get these changes organised into meaningful commits.

PR #2234, opened 2025-12-22
modern initialization code for several LiveConnect classes

ZZZank:
Changes in this PR:
- initialization for `JavaAdapter`, `JavaImporter`, and `NativeJavaTopPackage` is now done via LazilyLoadedCtor+lambda instead of LazilyLoadedCtor+reflection
- migrate NativeJavaTopPackage away from IdFunctionCall: https://github.com/mozilla/rhino/commit/f7e9a7ce08f4380fa504d08bec3fcf5a71d0773e
- some readibility improvement, for example: https://github.com/mozilla/rhino/commit/ee31cf0b0919c445ef3337c4e9ec93de3426e051
- Initialization of top level java packages will now reuse the LazyLoadedCtor for 'Packages'. In legacy initialization code, for each top level package, as well as "getClass" and "Packages", there's a corresponding LazilyLoadedCtor, which means `NativeJavaTopPackage.init()` can be invoked `2 + topPackageNames.length` times at most.

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/NativeJavaTopPackage.java
+++ b/rhino/src/main/java/org/mozilla/javascript/NativeJavaTopPackage.java
@@ -6,6 +6,8 @@
 
 package org.mozilla.javascript;
 
+import java.util.Scanner;
+
 /**
  * This class reflects Java packages into the JavaScript environment. We lazily reflect classes and
  * subpackages, and use a caching/sharing system to ensure that members reflected into one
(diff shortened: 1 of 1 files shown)


aardvark179:
We seem to be combining a couple of different things in here around the `init` changes, and they are going to conflict with #2161, and I think they'll be more churn in this area round the separation of scopes and objects. Would you mind if we put that part off and come up a consistent pattern to apply more widely? Replacing reflection with lambdas seems good though, and I'm happy to see that go forward.

PR #2235, opened 2025-12-22
Implement BigInt64 and BigUint64 DataView methods.

aardvark179:
This PR includes just the big int related changes from #2218.

### DIFF
--- a/tests/testsrc/jstests/es6/dataview.js
+++ b/tests/testsrc/jstests/es6/dataview.js
@@ -19,4 +19,33 @@ load("testsrc/assert.js");
   assertEquals(true, DataView.prototype.hasOwnProperty(Symbol.toStringTag));
 })();
 
+(function TestBigInt64GetAndSet() {
+  function t(origVal, expectedSigned, expectedUnsigned) {
+    const buf = new ArrayBuffer(8);
+    const v = new DataView(buf);
+    v.setBigInt64(0, origVal);
+    const newSignedVal = v.getBigInt64(0);
+    assertEquals(expectedSigned, newSignedVal);
+    v.setBigUint64(0, origVal);
+    const newUnsignedVal = v.getBigUint64(0);
+    assertEquals(expectedUnsigned, newUnsignedVal);
+  }
+
+  t(0n, 0n, 0n);
+  t(1n, 1n, 1n);
+
+  // A large value that still fits in a signed 64-bit value
+  t(1234567890123456789n, 1234567890123456789n, 1234567890123456789n);
+
+  // Just over a 64-bit signed value, wraps around to -1, fits in unsigned
+  t(18446744073709551615n, -1n, 18446744073709551615n);
+
+  // Very much too large for 64 bits, wraps around
+  t(9223372036854775807n + 1n, -9223372036854775808n,  9223372036854775808n);
+
+  // Wraps around in the other direction,
+  // when unsigned simply has less precision
+  t(-9223372036854775808n - 1n, 9223372036854775807n,  9223372036854775807n);
+})();
+
 "success";


rhino/src/main/java/org/mozilla/javascript/typedarrays/NativeDataView.java
```diff
@@ -407,6 +413,85 @@ private void js_setFloat(Context cx, Scriptable scope, int bytes, Object[] args)
...
+        return realThis.js_getBigInt(true, args);
+    }
+
+    private static Object js_getBigUint64(
+            Context cx, Scriptable scope, Scriptable thisObj, Object[] args) {
+        NativeDataView realThis = realThis(thisObj);
+        return realThis.js_getBigInt(false, args);
+    }
+
+    private Object js_getBigInt(boolean signed, Object[] args) {
```
rbri, line 429: i think this shoud not be namend js_...; methods starting with js_ are usually mappend to js lambdas, same for the setter

rhino/src/main/java/org/mozilla/javascript/typedarrays/NativeDataView.java
```diff
@@ -407,6 +413,85 @@ private void js_setFloat(Context cx, Scriptable scope, int bytes, Object[] args)
...
+        long base = ByteIo.readUint64Primitive(arrayBuffer.buffer, offset + pos, littleEndian);
+
+        if (signed) {
+            // Interpret as signed 64-bit integer
+            return java.math.BigInteger.valueOf(base);
+        } else {
+            // Interpret as unsigned 64-bit integer
+            if ((base & 0x8000000000000000L) == 0) {
+                return java.math.BigInteger.valueOf(base);
+            } else {
```
rbri, line 452: minor/personal taste: no need for else here

rhino/src/main/java/org/mozilla/javascript/typedarrays/NativeDataView.java
```diff
@@ -407,6 +413,85 @@ private void js_setFloat(Context cx, Scriptable scope, int bytes, Object[] args)
...
+    }
+
+    private static Object js_setBigUint64(
+            Context cx, Scriptable scope, Scriptable thisObj, Object[] args) {
+        NativeDataView realThis = realThis(thisObj);
+        realThis.js_setBigInt(false, args);
+        return Undefined.instance;
+    }
+
+    private void js_setBigInt(boolean signed, Object[] args) {
```
rbri: why is the 'signed' parameter not used at all (maybe i'm blind?)

rhino/src/main/java/org/mozilla/javascript/typedarrays/NativeDataView.java
aardvark179: I agree, but it's consistent with the naming of everything else in this file.

rhino/src/main/java/org/mozilla/javascript/typedarrays/NativeDataView.java
aardvark179: I'm not sure. I think I might rewrite that function.

aardvark179: Okay, so `longValue` did what was needed providing it never saw a signed big int with 64 or more bits, but that seems fragile and slightly obscure so I've made the sign handling explicit.

rhino/src/main/java/org/mozilla/javascript/typedarrays/NativeDataView.java
```diff
@@ -488,7 +488,13 @@ private void js_setBigInt(boolean signed, Object[] args) {
             throw ScriptRuntime.rangeErrorById("msg.dataview.offset.range");
         }
 
-        long base = val.longValue();
+        // base will be the 64 LSB of the unsigned big integer.
+        long base = val.abs().longValue();
```
gbrail: My AI code review tool (Gemini CLI this time) doesn't like this last change -- and indeed, it doesn't wrap properly, and the output doesn't match Node.js (Not that Node is always right but they do have a lot of people to do QA.) Output does match if I take this bit out and go back to "base=val.longValue()".

I'll see if I can attach the test case it generated -- I rather like it.

Do you mind if I make a version of this PR with those tests in it? Then I really want to get to the release!

gbrail: FWIW:

  The original code was simple and correct:
  long base = val.longValue();

  This correctly truncates any BigInteger to its lower 64 bits, which is
  the exact behavior required by the ECMAScript specification for both
  signed (setBigInt64) and unsigned (setBigUint64) operations. It properly
  handles two's complement for signed values and performs the modulo 2^64
  operation for unsigned values.

  The new code is complex and flawed:

   1 long base = val.abs().longValue();
   2 if (signed) {
   3     base = (base & 0x7fffffffffffffffL) * val.signum();
   4 }

  This logic fails in several key cases:

   * For `setBigInt64` (signed): It fails to correctly write negative
     numbers, most notably Long.MIN_VALUE. The combination of abs(),
     masking with 0x7fffffffffffffffL, and multiplying by the signum
     results in incorrect values. For example, trying to write
     Long.MIN_VALUE results in 0 being written to the buffer.
   * For `setBigUint64` (unsigned): It fails when the JavaScript BigInt is
     negative. The spec requires taking the value modulo 2^64. For
     instance, -1n should wrap around to the largest unsigned 64-bit
     integer (0xFF...FF), which is represented by -1L in Java. The new
     code takes the absolute value, causing 1L to be written instead.*
**

rhino/src/main/java/org/mozilla/javascript/typedarrays/NativeDataView.java
```diff
@@ -407,6 +414,90 @@ private void js_setFloat(Context cx, Scriptable scope, int bytes, Object[] args)
...
+            Context cx, Scriptable scope, Scriptable thisObj, Object[] args) {
+        NativeDataView realThis = realThis(thisObj);
+        realThis.js_setBigInt(false, args);
+        return Undefined.instance;
+    }
+
+    private void js_setBigInt(boolean signed, Object[] args) {
+        int pos = ScriptRuntime.toIndex(isArg(args, 0) ? args[0] : Undefined.instance);
+
+        BigInteger val = ScriptRuntime.toBigInt(isArg(args, 1) ? args[1] : Undefined.instance);
```
gbrail, line 481: Our implementation of this doesn't match the spec -- it should fail if the argument is a Number that's not a BigInt.

ScriptRuntime.toBigInt should impement the abstract ECMAScript operation "ToBigInt", in section 7.1.13 that fails on a Number.

The BigInt constructor should instead call "NumberToBigInt" in section 21.2, which is what we are doing in the constructor.

gbrail:
Given that this is ES2026 stuff, I'd like to wait until we get all the details right of this feature and have more comprehensive tests.

BTW I have had a lot of success by using the "mjsunit" test suite in V8 to test for things like this -- in this case, it caught some things that test262 did not.

aardvark179:
Okay, I've gone and checked the spec, it seems to jump through a bunch of hoops that make the conversions to `BigInt64` and `BigUint64` appear different, but they wind up doing _precisely_ the same thing (it's just that in a section in the middle one number is treated as unsigned and the other as two's complement). So we can do this with a simple `longValue` call, but it definitely needs a comment explaining this

gbrail:
Yep, I confirmed a few different ways that "longValue" matches the correct result in all cases, and compared output to Node.js as well. I'll attach a small test program.

[t.js](https://github.com/user-attachments/files/24403180/t.js)

zhurvla (approved):
Looks good. 
A few else after return can be avoided for readability.
https://github.com/mozilla/rhino/pull/2235/files#diff-a25991e9dccfaf928a5a19d200dd5bc22e2f563b9eacd80576115bf952f5189cR448

gbrail:
I'm good with this now. Thanks for all the work!

PR #2238, opened 2025-12-23
Update build to support Java 25

gbrail:
Update dependencies and tools to support building on Java 25:

* Update Gradle to latest
* Update ErrorProne to support latest
* Update list of ErrorProne exclusions -- lots of opportunities for future fixes
* Fix a Date test that seems inconsistent in current form TBH
* Update GitHub Actions to run with Java 25

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -8,6 +8,9 @@ on:
 
 permissions: read-all
 
+env:
+  GRADLE_OPTS: "--enable-native-access=ALL-UNNAMED"
+
 jobs:
   test:
     runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


aardvark179:
Do we plan to continue supporting Java 11?

gbrail:
No I think we should drop Java 11 but I'm trying to do this one step at a time, the first step IMO is to make it work at all with Java 25.

gbrail:
If anyone has an idea why that one date / time zone tests differs in behavior between 21 and 25 I'd love to hear ideas -- otherwise either we need to track it down or disable that test!

aardvark179:
I'm not sure why it has changed, but maybe we should just assert on the start of the string rather than the name of the timezone?

PR #2240, opened 2025-12-24
simplify serialization of `MemberBox`

ZZZank:
1. There's a boolean at the start of serialized data, representing whether the `member` is present. This PR removed this because `MemberBox` is an internal class and never has its `member` set to `null`
2. The writing/reading of parameter types is rewritten to read/write method descriptor string, which can be converted from/to parameter types array via `MethodType`.
3. `serialVersionUID` updated to reflect serialization logic change


gbrail:
Looks good to me -- thanks for the simplification!

Issue #2239, opened 2025-12-23
Rhino 1.9 and `core-js` compat tests


zloirock:
I found that Rhino 1.9 has been released and tried to update `core-js` compat data.

Since [binaries are no longer attached to Rhino releases](https://github.com/mozilla/rhino/releases/tag/Rhino1_9_0_Release), [this doc](https://github.com/mozilla/rhino/blob/f3123bf5a2a47695da8467d26bc8113322cb4f73/RELEASE-STEPS.md#update-core-js-compat) should be updated,
```
npm run compat-rhino YOUR_PATH_TO_RHINO
```

So, I tried to run `core-js` on Rhino 1.9 and found a strange error:

<img width="1435" height="605" alt="Image" src="https://github.com/user-attachments/assets/1ba23e95-32b0-4c71-9bac-ba075b73ae54" />

The compat data tests runner gets [an object with tests](https://github.com/zloirock/core-js/blob/400f00eac72f04438843d4f5059cffe839b55718/tests/compat/tests.js#L357), enumerates properties of this object, and runs them.

On the third property of this object, Rhino throws an error, no matter what it contains.

Rhino 1.8 (and all the rest JS engines) works fine here.

aardvark179:
Okay. I think we need to update the docs, maybe they should point to instructions in the core-js repo since it can change between releases.

Running the tests with `npm run compat-rhino /Users/duncan.macgregor/git/mozilla/rhino/rhino-all/build/libs/rhino-all-1.9.0.jar` is producing the error you see. I can fix _that_ but then we immediately fail with a different error.

@gbrail do we have a procedure to follow before doing the release, these feel like things we should have caught.

zloirock:
> maybe they should point to instructions in the core-js repo 

https://github.com/zloirock/core-js/blob/master/CONTRIBUTING.md#how-to-update-core-js-compat-data

aardvark179:
Okay, the other problem is that at some point `global` has stopped being set in whatever context this is run in. I'll bisect that later to work out when it broke. Fixing the script for that does allow me to run the script on 2.0 snapshot and shows the following changes

```
es.error.is-error                            | not required
es.array.concat                              | not required
es.array.filter                              | not required
es.array.map                                 | not required
es.array.slice                               | not required
es.array-buffer.is-view                      | not required
es.array-buffer.detached                     | not required
es.array-buffer.transfer                     | not required
es.array-buffer.transfer-to-fixed-length     | not required
[...]

PR #2241, opened 2025-12-24
Avoid illegal characters in compiled method names.

aardvark179:
Ensure that functions names that reach the class compiler are turned into legal JVM method names.

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/optimizer/Codegen.java
+++ b/rhino/src/main/java/org/mozilla/javascript/optimizer/Codegen.java
@@ -17,6 +17,7 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.regex.Pattern;
 import org.mozilla.classfile.ByteCode;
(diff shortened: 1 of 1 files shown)


aardvark179:
This should fix part of #2239.

rhino/src/main/java/org/mozilla/javascript/optimizer/Codegen.java
```diff
@@ -873,7 +880,7 @@ String cleanName(final ScriptNode n) {
         } else {
             result = "script";
         }
-        return result;
+        return illegalChars.matcher(result).replaceAll("_");
```
rbri, line 883: Not sure how the method name construction works. But maybe this simple replacement might lead to duplicate names if the source names are only differ in invalid cars.

aardvark179: Luckily they get a unique numerical component added.

gbrail:
OK, makes sense to me, thanks!

gausie:
Could we get a 1.9.1 with this fix?

PR #2242, opened 2025-12-24
Ensure `global` is present in top level.

aardvark179:
It's not in the spec, and it's not their In all browsers, but compatibility checking does look for it.


aardvark179:
This should fix part of #2239.

gbrail:
Thanks -- if we can verify that this makes the core-js work again, then we should consider whether we need a 9.1.0 or a 9.0.0.1 with this fix and one or two more.

PR #2244, opened 2026-01-05
Add a LiveConnect variant of MemberBox

ZZZank:
This PR is made for solving some cycle references between LiveConnect and the rest of Rhino, by splitting MemberBox into two classes:
1. MemberBox, for usages in non-LiveConnect classes like `FunctionObject` and `ScriptableObject`. MemberBox uses `Class<?>` instead of `TypeInfo` for storing types, thus not requiring `TypeInfo` or `TypeInfoFactory`.

2. ExecutableBox, for LiveConnect usage. Features like arg nullability detector, the ability to create getter/setter function, and serialization support is removed, because they didn't to be used in any aspect of LiveConnect

|Features | original MemberBox | MemberBox | ExecutableBox |
| - | - | - | - |
| wrap Java `Member` usage | ✅ | ✅ | ✅ |
| Java Generics Support | ✅ | ❌ | ✅ |
| Serialization | ✅ | ✅ | ❌ |
| create and cache getter/setter `BaseFunction` | ✅ | ✅ | ❌ |
| `NullabilityDetector` | ✅ | ✅ | ❌ |
| Search accessible method after failure | ✅ | ✅ | ❌ |

### DIFF
--- a/tests/src/test/java/org/mozilla/javascript/tests/type_info/CustomTypeInfoFactoryTest.java
+++ b/tests/src/test/java/org/mozilla/javascript/tests/type_info/CustomTypeInfoFactoryTest.java
@@ -1,8 +1,6 @@
 package org.mozilla.javascript.tests.type_info;
 
 import java.io.*;
-import java.lang.reflect.Modifier;
-import java.util.Arrays;
 import org.junit.jupiter.api.Assertions;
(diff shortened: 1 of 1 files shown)


gbrail:
This looks like a good refactoring and works for me. Thanks!

PR #2245, opened 2026-01-07
Respect custom [Symbol.hasInstance] implementations

0xe:
Fixes a couple of bugs with [Symbol.hasInstance] support that was implemented in https://github.com/mozilla/rhino/commit/4cea728ab.

1. Custom [Symbol.hasInstance] implementation should be respected
2. Pass the right argument to non-function objects with custom [Symbol.hasInstance] implementations

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/BaseFunction.java
+++ b/rhino/src/main/java/org/mozilla/javascript/BaseFunction.java
@@ -373,10 +373,14 @@ public class BaseFunction extends ScriptableObject implements Function {
 
         // Attempt to call custom Symbol.hasInstance implementation if present
         Object hasInstanceMethod = ScriptRuntime.getObjectElem(this, SymbolKey.HAS_INSTANCE, cx);
-        if (hasInstanceMethod instanceof Callable) {
+        if (hasInstanceMethod instanceof Function) {
             return ScriptRuntime.toBoolean(
-                    ((Callable) hasInstanceMethod)
-                            .call(cx, getParentScope(), this, new Object[] {instance}));
+                    ((Function) hasInstanceMethod)
+                            .call(
+                                    cx,
+                                    ((Function) hasInstanceMethod).getParentScope(),
+                                    this,
+                                    new Object[] {instance}));
         }
 
         Object protoProp = ScriptableObject.getProperty(this, PROTOTYPE_PROPERTY_NAME);
--- a/rhino/src/main/java/org/mozilla/javascript/ScriptableObject.java
+++ b/rhino/src/main/java/org/mozilla/javascript/ScriptableObject.java
@@ -857,8 +857,7 @@ public abstract class ScriptableObject extends SlotMapOwner
     @Override
     public boolean hasInstance(Scriptable instance) {
         // Default for JS objects (other than Function) is to do prototype
-        // chasing. This will be overridden in NativeFunction and non-JS
-        // objects.
+        // chasing. Overridden in BaseFunction, E4X, and Java interop classes.
 
         Context cx = Context.getCurrentContext();
         Object hasInstance = ScriptRuntime.getObjectElem(this, SymbolKey.HAS_INSTANCE, cx);


0xe:
@aardvark179 pointed out that we _may_ have to walk the prototype chain to find [Symbol.hasInstance] implementation. Marking the pull as draft until that's fixed.

rhino/src/main/java/org/mozilla/javascript/BaseFunction.java
```diff
@@ -358,6 +358,17 @@ public String getTypeOf() {
      */
     @Override
     public boolean hasInstance(Scriptable instance) {
+        Context cx = Context.getCurrentContext();
+        // Attempt to call custom Symbol.hasInstance implementation if present
+        if (has(SymbolKey.HAS_INSTANCE, this)) {
+            Object hasInstanceMethod = get(SymbolKey.HAS_INSTANCE, this);
+            if (hasInstanceMethod instanceof Callable) {
```
andreabergia: I haven't checked the spec, but trying in node to set the symbol value to something that's not a callable throws, whereas this implementation ignores it silently:

```
> o = { [Symbol.hasInstance]: 42 }
{ [Symbol(Symbol.hasInstance)]: 42 }
> o instanceof o
Uncaught TypeError: number 42 is not a function
```

0xe: Thanks, added a test.

rhino/src/test/java/org/mozilla/javascript/SymbolHasInstanceTest.java
```diff
@@ -0,0 +1,86 @@
...
+                        + "function SubFunc() {}\n"
+                        + "Object.setPrototypeOf(SubFunc, BaseFunc);\n"
+                        + "var result = ({}) instanceof SubFunc;\n"
+                        + "called && result;";
+        Utils.assertWithAllModes_ES6(true, script);
+    }
+
+    /** errors thrown by custom Symbol.hasInstance should be propagated */
+    @Test
+    public void testSymbolHasInstanceErrorPropagated() {
```
rbri, line 57: assertEcmaError... or another method from Utils should do the job here

0xe: Thanks, fixed it.

rhino/src/main/java/org/mozilla/javascript/BaseFunction.java
andreabergia: I actually don't see the test...

aardvark179:
Do those test cases depend on `class`? It would certainly be worth understanding why no extra tests are passing.

rhino/src/main/java/org/mozilla/javascript/BaseFunction.java
```diff
@@ -404,6 +417,15 @@ private static Object js_hasInstance(
             return false; // NOT_FOUND, null etc.
         }
 
+        // E4X XML constructors have invalid prototype properties.
```
andreabergia: I don't think there's any tests for the e4x code.

0xe: There are some in MozillaSuiteTest and `rhino-xml/src/test/` and a few in the main code base as well (StringifyXmlTest).

rhino/src/main/java/org/mozilla/javascript/BaseFunction.java
0xe: Sorry, added it now.

0xe:
> Do those test cases depend on `class`? It would certainly be worth understanding why no extra tests are passing.

We have all the tests that are present in [262](https://github.com/tc39/test262/tree/master/test/built-ins/Function/prototype/Symbol.hasInstance) for [Symbol.hasInstance]. Out of those we have `Symbol.hasInstance/this-val-poisoned-prototype.js` is failing for a different problem.

0xe:
> LGTM but I'm a bit surprised this doesn't improve any test262 cases. Maybe we should add something to test262?

Yeah, I think some new tests need to be contributed to test262.

aardvark179 (approved):
I think this can be merged. We have internal test cases covering this so do not need to wait for test262 cases to be created.

rhino/src/main/java/org/mozilla/javascript/BaseFunction.java
```diff
@@ -404,6 +417,15 @@ private static Object js_hasInstance(
...
+        // Return false instead of throwing for E4X compatibility (ECMA-357 13.4.3.10).
+        if (thisObj instanceof IdFunctionObject) {
+            Object tag = ((IdFunctionObject) thisObj).getTag();
+            if ("XMLObject".equals(tag)) {
```
rbri: why we have tow different ways to determine XMLObject - above we use

    if (hasInstanceMethod instanceof Callable && !(instance instanceof XMLObject)) {

aardvark179: That one is subtle, and probably deserves a better comment. The case further up is saying, "if the instance being checked is not an XML object then do this…", whereas this check is concerned with whether we are checking if something is an instance of an XML object.

I think.

But looking at the E4X spec section mentioned I see

> 3.4.3.10 [[HasInstance]] ( V )
> Overview
> The XML constructor has a more elaborate [[HasInstance]] method than other function objects. The XML constructor [[HasInstance]] method is defined such that given an XML object or an XMLList object x, the expression x instance of XML will return true. This treatment intentionally blurs the distinction between a single XML object and an XMLList containing only one value[…]

If the XML constructor has a separate `Symbol.hasInstace` function then why not give it a separate `Symbol.hasInstance` function and get this stuff out of the basic `Function.prototype` version?

aardvark179: I've been going through classes converting them to descriptors, and the XML hasinstance stuff is _fun_. The implementation of hasinstance on XML returns true if the argument is an instance of XML of XMLList, but XMLList has the default hasInstance.

That took a while for me to understand.

0xe: I've pulled this logic out of BaseFunction into XMLCtor.java. I needed a way to distinguish between whether it was XML or XMLList constructor that was called. Please let me know if this makes sense.

gbrail:
I'd love to hear from @rbri just to sanity check the last commit.

rbri:
will do some review tomorrow - currently i'm a bit overloaded with the work on htmlunit 5 (jdk17, jetty 9 to 12 migration, websocket api changes and hopefully http/2 support)

rhino/src/main/java/org/mozilla/javascript/BaseFunction.java
```diff
@@ -358,6 +358,16 @@ public String getTypeOf() {
      */
     @Override
     public boolean hasInstance(Scriptable instance) {
+        Context cx = Context.getCurrentContext();
```
rbri, line 372: Personally i think we have far too many places, where Context cx = Context.getCurrentContext(); is called.
In the past we did something like
* create another methodn that gets also the context as parameter
* deprecate this method and call the new one with Context.getCurrentContext()
* change all calling places in our code base to call the new one

At least this method gets called from some places that already have the context at hand. It's a shame to throw it away and later on get it from a thread local...

rhino/src/main/java/org/mozilla/javascript/ScriptableObject.java
rbri, line 859: i think we have to adjust the comment here, at least this is overwritten by BaseFunction also

rhino/src/main/java/org/mozilla/javascript/BaseFunction.java
```diff
@@ -358,6 +358,16 @@ public String getTypeOf() {
...
+
+        // Attempt to call custom Symbol.hasInstance implementation if present
+        Object hasInstanceMethod = ScriptRuntime.getObjectElem(this, SymbolKey.HAS_INSTANCE, cx);
```
rbri, line 375: there are differences to the code in ScriptableObject, we check for Callable here and for Function there. And the scope retrieval is also different.
Maybe that is correct but then i like to have a comment here why this is like this.

rhino-xml/src/main/java/org/mozilla/javascript/xmlimpl/XMLCtor.java
```diff
@@ -189,6 +214,12 @@ private static void readSettings(XMLCtor thisObj, Scriptable source) {
     /** hasInstance for XML objects works differently than other objects; see ECMA357 13.4.3.10. */
     @Override
     public boolean hasInstance(Scriptable instance) {
-        return (instance instanceof XML || instance instanceof XMLList);
+        if (isXMLType) {
```
rbri, line 217: In the good old days i had expected a abstract method two overwritten implementations returning true/false. But today i guess instanceof is similar fast and i use it for myself at many places.

rbri (changes requested):
did also some smoke test with this changes in HtmlUnit. Works fine so far.

Sorry for the bunch of comments, but you asked for a review ;-)

0xe:
> did also some smoke test with this changes in HtmlUnit. Works fine so far.
> 
> Sorry for the bunch of comments, but you asked for a review ;-)

Thanks! Will address them.

rhino/src/main/java/org/mozilla/javascript/BaseFunction.java
0xe: Aligned it with ScriptableObject which does feel more accurate.

rhino/src/main/java/org/mozilla/javascript/ScriptableObject.java
0xe: Updated to mention all the overrides.

rhino/src/main/java/org/mozilla/javascript/BaseFunction.java
0xe: Yeah, many of the callers (eg: ScriptRuntime.instanceOf, XMLCtor) have Context available and fetching it again from the thread local again is a waste.

The problem is that `hasInstance(Scriptable instance)` is part of `Scriptable` interface, and changing it and deprecating the existing one would be an API change. Should we track that as a issue and address separately instead?

rbri: agree, maybe we have to use a discussion to decide how to maintain this kind of changes for 2.0. An issue will be great to have a starting point.

At lease i think we have similar problems with the various propery access method requirering a retun value to finalize eg. the proxy impl. Maybe we find a cool solution for all that.

gbrail:
This looks good to me too, thanks!

PR #2247, opened 2026-01-07
Add a default `isConstructor` to `Function` and refactor to use it.

aardvark179:
I kept tripping over this in `AbstractEcmaObjectOperations` so I thought it was time to tidy this up.


gbrail:
Looks good and nice cleanup -- thanks!

PR #2249, opened 2026-01-09
Simplify serialisation of `NativeJavaObject` and `JavaMembers`

### LINKED ISSUE #2248: Potential `JavaMembers` deserialisation problem.
Because `NativeJavaObject` recreates its `JavaMembers` on deserialisation, and because `JavaMembers` itself can lookup function prototypes dynamically, there is the potential for it all to go very wrong by causing that part of an object graph to be deserialised before enough of the top level scope has been recreated for the prototype to actually be looked up.

I stumbled across this in my scopes refactoring, and I think the only way to resolve this properly is by serialising most of java members so that nothing is looked up dynamically in scopes.

Have I ever mentioned that I hate Java serialisation?

aardvark179:
This fixes #2248. It turned out to be much simpler to fix than I had feared as almost everything was already serialisable.

### DIFF
--- a/rhino/src/test/java/org/mozilla/javascript/tests/ContinuationsApiTest.java
+++ b/rhino/src/test/java/org/mozilla/javascript/tests/ContinuationsApiTest.java
@@ -277,7 +277,10 @@ public class ContinuationsApiTest {
                 cx.setInterpretedMode(true); // must use interpreter mode
                 cx.evaluateString(
                         globalScope,
-                        "function f(a) { Number.prototype.blargh = function() {return 'foo';}; var k = myObject.f(a); var t = []; return new Number(8).blargh(); }",
+                        "function f(a) { Number.prototype.blargh = "
+                                + "function() {return 'foo';}; var k = myObject.f(a); "
+                                + "var t = []; return new Number(8).blargh(); }; "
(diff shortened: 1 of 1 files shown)


ZZZank:
A bit amazed by the fact that this PR didn't cause any additional error. I was expecting some `NativeJavaField` related errors since NativeJavaField is not serializable

aardvark179:
Good point. I'll make that one serialisable too, and add some more to the tests.

ZZZank:
I think serializing each NativeJavaXXX object separately is a bad idea, especially considering the raw member objects (`Method`, `Field`, etc) are not serializable, we will have to record their information and recreate the raw member manually, causing a huge amount of ''declaring class'' in serialized data.

Actually I think nothing in JavaMembers should be serializable at all, other than the name of the class that this JavaMembers reflects

aardvark179:
We **have** to serialise those things, they can become reachable via multiple routes by assigning them to other variables.

I'd be happy to see a rewrite of JavaMembers that separates the raw members out and never serialises those, but then you'll have to make sure all the `NativeJavaMethod/Constructor/etc` stuff references things through that, and that none of those mutable JS objects are themselves stored inside `JavaMembers`

PR #2252, opened 2026-01-14
drop various versions

rbri:
As we are on the road to version 2.0 I like to suggest to drop the support for some really old versions.

This PR should give an impression what happens if we do that.


rbri:
Maybe it is a bad idea to simply drop the test from the mozilla suite. Another option might be to adjust the version we use for running the test (jsref.js) and only disable the ones testing for incompatible behaviour

aardvark179:
For compatibility we expose three versions internally, so I'd like to check those and have a think about how large a diff we would end up with. I agree we shouldn't try to maintain the range of versions we currently do,

gbrail:
I'm supportive of this and wonder if others have had a chance to think about it!

rbri:
after some thinking about this i will give it antother try, will try to remove the same versions but preserve more of the existing tests. I have the feeling the grandfathers of this did something similar long long ago with the first version.
In the hope everybody agrees with removing this old stuff i will start next week and provide an alternative to this.

rbri:
> I'm more worried about the big removal of tests

Yes that is exactly what i like to address in the new version of this PR. Stay tuned but i have an idea what to do and will try it out.

rbri:
replaced by https://github.com/mozilla/rhino/pull/2281

PR #2253, opened 2026-01-14
Fix debugger eval for Script

nicolas-albert:
## Summary
- use Script.exec(...) for debugger eval instead of casting to Callable

## Problem
In 1.9.0, Context.compileString returns Script (JSScript) which does not implement Callable; Watch/Evaluate in the Swing debugger throws a ClassCastException.

## Repro
1. Run debugger, hit breakpoint
2. Evaluate `1+1` in Watch/Evaluate
3. `class org.mozilla.javascript.JSScript cannot be cast to class org.mozilla.javascript.Callable`

## Fix
Use Script.exec(cx, scope, thisObj) for evaluation to preserve `this` semantics.

### DIFF
--- a/rhino-tools/src/main/java/org/mozilla/javascript/tools/debugger/Dim.java
+++ b/rhino-tools/src/main/java/org/mozilla/javascript/tools/debugger/Dim.java
@@ -745,8 +745,15 @@ public class Dim {
         cx.setInterpretedMode(false);
         cx.setGeneratingDebug(false);
         try {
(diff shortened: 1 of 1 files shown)


gbrail:
Thanks -- this is a nice fix. If you have ideas how to automatically test the debugger, we should get them implemented, since this is clearly a miss in our test coverage!

aardvark179:
Catching this case requires not just testing the debugger interface but testing this client. I'll think about how to plug this gap.

PR #2255, opened 2026-01-14
use `ClassValue` for caching class-based TypeInfo

ZZZank:
Changes due to migrating to `ClassValue`:
- `ConcurrentFactory` now holds a weak, instead of strong, reference to the resolved class. In other words, resolving a class using `ConcurrentFactory` will not prevent JVM from reclaiming the resolved class now.
- Better performance. `ClassValue` is surprisingly fast, even faster than `ConcurrentHashMap`.

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -14,7 +14,9 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        api-level: [ 26, 28, 30, 33 ]
+        # oldest supported API level: API 26 (Android 8)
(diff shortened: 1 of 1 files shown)


aardvark179:
`ClassValue` only became available in Android API version 34, so you may need a different approach if it's not available. I have no idea why it took so long to introduce given it landed in Java 7.

gbrail:
This is fair, and I appreciate Android compatibility. Now that we have a little mini Android test suite, would you consider adding a test?

gbrail:
Thanks for all those updates, this looks good to me now!

PR #2256, opened 2026-01-15
Add a method to temporarily suspend microtask processing

gbrail:
This makes it possible for a framework to stop Promises from being resolved until other scripts have run. This can be helpful when trying to emulate the behavior of some browsers and web framework that may depend on this.

Take a look at the tests to see how this affects resolution of Promises.

I ran in to this while trying to run the test framework for the WHATWG standards to run in Rhino:

https://github.com/web-platform-tests/wpt

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/Context.java
+++ b/rhino/src/main/java/org/mozilla/javascript/Context.java
@@ -2512,11 +2512,9 @@ public class Context implements Closeable {
      * @see #suspendMicrotaskProcessing()
      */
     public void processMicrotasks() {
-        if (microtaskSuspendCount == 0) {
-            Runnable head;
-            while ((head = microtasks.poll()) != null) {
-                head.run();
-            }
+        Runnable head;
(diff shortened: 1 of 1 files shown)


gbrail:
This one is pretty innocuous unless you really need it, but I'm not hearing much, and will likely merge it soon.

rhino/src/main/java/org/mozilla/javascript/Context.java
```diff
@@ -2506,15 +2506,46 @@ public void enqueueMicrotask(Runnable task) {
...
+     * <p>Nothing will happen if suspendMicrotaskProcessing was called.
+     *
+     * @see #suspendMicrotaskProcessing()
      */
     public void processMicrotasks() {
-        Runnable head;
-        do {
-            head = microtasks.poll();
-            if (head != null) {
+        if (microtaskSuspendCount == 0) {
```
aardvark179: Should the check be done here, or as part of the `while` clause?

gbrail: Yes indeed it should ... changed it in the next commit

PR #2263, opened 2026-01-18
prevent Field/Method/Constructor to be directly accessible in JS code

ZZZank:
```javascript
/** @type {java.lang.Class} */
let c = ...;

// these operations below will now always return `undefined`
c.getMethod("name", argTypes)
c.getField("name", type)
c.getConstructor(types)
c.getMethods()[i]
```

`Class` itself is not denied, because I believe after this PR, `Class` is more or less a safe type.
`ClassLoader` also not denied, because it seems that the only potentially dangerous usage is `classLoader.loadClass("some.forbidden.clazz.Here")`, which just gives a `Class`.

I believe directly using reflection in JS code is an incredibly rare use case, but technically this is still a breaking change.

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/WrapFactory.java
+++ b/rhino/src/main/java/org/mozilla/javascript/WrapFactory.java
@@ -8,6 +8,7 @@
 
 package org.mozilla.javascript;
 
+import java.lang.reflect.AccessibleObject;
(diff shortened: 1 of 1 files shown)


ZZZank:
Marking as draft because some LiveConnect tests themself are using reflection directly

PR #2264, opened 2026-01-19
Support reserved words in XML attribute names.

### LINKED ISSUE #2258: E4X: Can not access XML attribute with name "class"
When creating a XML object that has an attribute "class", then accessing this attribute causes an exception: 

```
>var p = <p class="clazz" abc="abc">Hello</p>;
>p.@abc
abc
>p.@class
org.mozilla.javascript.EvaluatorException: missing name after .@
>p.toXMLString();
<p abc="abc" class="clazz">Hello</p>
```

This used to work in 1.7.x and breaks from 1.8 on.

aardvark179:
This updates the XML attribute access parsing code to match the main property access parsing code. This has the same behaviour regarding reserved words as property names. This should hopefully fix #2258.


gbrail:
This seems OK to me -- it is an XML-specific fix. Can anyone else think of any unfriendly side effects?

PR #2272, opened 2026-01-28
list Trireme as separate entry and link to Apigee

rbri:


gbrail:
Sure, although I doubt at this point whether anyone will go back and get Trireme working on the latest Rhino again but it sure was fun. Apigee doesn't use it any more though!

PR #2273, opened 2026-01-29
Support resizable typed arrays, ArrayBuffers, and DataViews

gbrail:
This starts with the work that @anivar did in https://github.com/mozilla/rhino/pull/2182 and adds some additional tests and fixes to make the tests work. The test262 tests for this feature are unhelpful because they depend on features that we don't implement like classes and block-scoped "const." Other tests from the V8 suite cover the same ground and I was able to transpile them with Babel so that they run. The result matches the spec quite well.

This implementation just implements resizing by allocating a new byte[] and copying. Once we are done with this one, and now that we have tests, I'll open a new one with a different optimization mode for resizing


tests/src/test/java/org/mozilla/javascript/tests/es2024/DataViewResizableArrayBufferTest.java
```diff
@@ -0,0 +1,10 @@
+package org.mozilla.javascript.tests.harmony;
+
+import org.mozilla.javascript.Context;
+import org.mozilla.javascript.drivers.LanguageVersion;
+import org.mozilla.javascript.drivers.RhinoTest;
+import org.mozilla.javascript.drivers.ScriptTestsBase;
+
+@RhinoTest("testsrc/jstests/harmony/dataview-resizablearraybuffer.js")
+@LanguageVersion(Context.VERSION_ES6)
```
rbri, line 9: as this test and TypedArrayResizableArrayBufferTest running with version es6 - should we place them in the es6 folder instead of harmony

tests/testsrc/jstests/es2024/dataview-resizablearraybuffer.js
```diff
@@ -0,0 +1,499 @@
+// Copyright 2021 the V8 project authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+// Flags: --allow-natives-syntax --js-staging
+
+// This file was imported from the V8 project and processed using Babel
```
rbri, line 7: do we need a similar comment at the top of typedarray-helpers.js and typedarray-resizablearraybuffer.js?

gbrail:
Makes sense to align these tests with the right language versions -- updated the test locations and added those comments.

gbrail:
Hey all -- I'd like to merge this, LMK if you'd like to try more things!

rbri:
LGTM

aardvark179:
Looks good to me, though I did get tripped up on the giant test file. 😄

PR #2276, opened 2026-01-30
Fix double-evaluation of Date constructor arguments

jimmyhmiller:
The date_msecFromArgs function was calling toNumber() and then toInteger() on the same argument, which caused the argument's toString/valueOf to be called twice. Fixed by using the already-computed number value when converting to integer.

Fixes 2 test262 tests:
- built-ins/Date/UTC/coercion-order.js
- built-ins/Date/coercion-order.js


gbrail:
Looks good -- thanks!

PR #2281, opened 2026-02-01
Drop versions

rbri:
This replaces https://github.com/mozilla/rhino/pull/2252

This time no test deleted. I simply followed some patterns already used in the test code base.

@andreabergia @gbrail @aardvark179 @0xe @balajirrao @rPraml i'm not sure what to do with version 1.4 - it is not so much used at all, only regexp and octal string parsing. Should I try to remove that also?

### DIFF
--- a/tests/testsrc/tests/shell.js
+++ b/tests/testsrc/tests/shell.js
@@ -99,7 +99,7 @@ function startTest() {
     version ( "140" );
   }
   else if ( VERSION == "JS_1.2"  || gTestsuite == 'js1_2') {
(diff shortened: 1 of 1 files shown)


aardvark179:
I'm not sure we care about 1.4, we basically maintain old non-ES semantics, ES5, and latest.

gbrail:
I think this is great. In the interest of super pedanticness I see that Copilot identified a typo. Can we fix that, and then we'll merge this and get rid of 1.4 as well? 

Thanks!

rbri:
@gbrail see the linked PR

rbri:
As @gbrail agrees to drop 1.4 also i close this for #2282

PR #2282, opened 2026-02-02
drop version 1.0 - 1.4 support

rbri:
This drops support for version 1.0 - 1.4.

All test from the (old) mozilla test suite are preserved, only the configuration is changed to use version 1.5 for run them. Some tests are updated to reflect the spec changes. All the test expectation changes are done in a way to be backward compatible (means you can still run them with the original version in case your fork still supports this versions).


gbrail:
I think this is great and I'm going to merge it. Thanks!

PR #2283, opened 2026-02-02
Fix JSON.parse to preserve negative zero

jimmyhmiller:


rhino/src/main/java/org/mozilla/javascript/json/JsonParser.java
```diff
@@ -311,7 +311,7 @@ private Number readNumber(char c) throws ParseException {
         String num = src.substring(numberStart, pos);
         final double dval = Double.parseDouble(num);
         final int ival = (int) dval;
-        if (ival == dval) {
+        if (ival == dval && Double.compare(dval, -0.0) != 0) {
```
rbri: maybe we can do the Double.compare(dval, -0.0)check before the int cast and simply retrun the Double value

jimmyhmiller: Done

jimmyhmiller:
Failure appears to be environmental. A rerun should clear it up. (Passed fine on my fork)

rbri:
> Failure appears to be environmental. A rerun should clear it up. (Passed fine on my fork)

Yes this happens from time to time, restarted...

gbrail:
Looks good to me too -- thanks!

PR #2284, opened 2026-02-02
Optimize regex backtracking for quantifiers before end-of-line anchor

balajirrao:
I've been wanting to enable the Unicode property escapes test262 tests for a while, but they timed out on CI (and failed with OOM on my machine). The culprit turned out to be excessive allocation for backtrack and regexp state stack.

Taking inspiration from [.NET runtime's regex optimizations](https://github.com/dotnet/runtime/issues/1349), this PR adds atomic quantifier optimization for patterns ending with `$`. The key insight is that when a greedy quantifier like `\w+` is followed by `$`, backtracking is futile - once the quantifier has consumed all it can, trying shorter matches won't help the anchor succeed. More generally, backtracking can be skipped when there's no overlap between what the quantifier matches and what follows it. The optimization is conservatively enabled for now - more optimizations to follow.

I've added a benchmark to test this optimization - `RegexpAtomicQuantifierBenchmark`:
    Before: 336,143 us/op
    After:  180,384 us/op
    Speedup: 1.86x (46% faster)

With the Unicode property escape tests enabled, RegExp test262 failures drop from 52% to 37.5%

These new tests make the test262 suite take longer to run - from under 2 minutes to a bit more than 3 minutes on my machine.

### DIFF
--- a/rhino/src/test/java/org/mozilla/javascript/tests/NativeRegExpTest.java
+++ b/rhino/src/test/java/org/mozilla/javascript/tests/NativeRegExpTest.java
@@ -1131,89 +1131,114 @@ public class NativeRegExpTest {
     }
 
     @Test
-    public void atomicQuantifierOptimization() {
-        // Sanity check: non-atomic patterns still work
-        Utils.assertWithAllModes_ES6("non-atomic *b", "b", "/a*b/.exec('b')[0]");
-
-        // Patterns ending with $ have greedy quantifiers marked as atomic.
-        // The optimization only applies to greedy quantifiers.
-
-        // a+ requires at least one 'a'
-        Utils.assertWithAllModes_ES6("match +$", "aaa", "/a+$/.exec('aaa')[0]");
-        Utils.assertWithAllModes_ES6("no match +$", null, "/a+$/.exec('aaab')");
-
-        // a* can match zero chars - matches empty string at end
-        Utils.assertWithAllModes_ES6("match *$", "aaa", "/a*$/.exec('aaa')[0]");
-        Utils.assertWithAllModes_ES6("*$ matches empty at end", "", "/a*$/.exec('b')[0]");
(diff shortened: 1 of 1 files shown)


rbri:
270 pasing tests ❤️

aardvark179:
Awesome! I had looked at turning on the Unicode tests but couldn't for the same reasons you found.

rhino/src/test/java/org/mozilla/javascript/tests/NativeRegExpTest.java
rbri: just a minor one, these multi tests are a real pain if you like to debug this. You have to comment out all the tests before in some way to not reach your breakpoint by all the passing tests before.

If possible i like to avoid this and have many small test methods.....

balajirrao: Yeah, I agree. Will fix!

rbri (approved):

rhino/src/test/java/org/mozilla/javascript/tests/NativeRegExpTest.java
```diff
@@ -1129,4 +1129,116 @@ public void testUnicodePropertyEscapeZanabazarSquare() {
                 "\uD806\uDE45",
                 "'\\u{11A45}'.match(/\\p{sc=Zanabazar_Square}/u)[0]");
     }
+
+    @Test
+    public void atomicQuantifierOptimization_greedyQuantifiers() {
```
andreabergia, line 1134: Refactoring this to JUnit 5's parametrized tests, with the `@CSVSource` annotation, will probably make this easier to maintain.

andreabergia (approved):

gbrail:
Thanks, everyone, for looking at this -- looks good to me!

PR #2286, opened 2026-02-05
Fix IdFunctionObject.isConstructor() to return useCallAsConstructor

jimmyhmiller:
Built-in methods like Date.prototype.getDate, RegExp.prototype.exec, etc. were incorrectly reporting as constructors even though calling them with 'new' would throw TypeError. This caused 65 test262 "not-a-constructor" tests to fail.

The fix adds an isConstructor() override that returns the existing useCallAsConstructor field, which is only set to true for functions explicitly marked as constructors via markAsConstructor().

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/IdFunctionObject.java
+++ b/rhino/src/main/java/org/mozilla/javascript/IdFunctionObject.java
@@ -108,6 +108,11 @@ public class IdFunctionObject extends BaseFunction {
     }
 
     @Override
+    public boolean isConstructor() {
+        return useCallAsConstructor;
+    }
+
+    @Override
     public String getFunctionName() {
         return (functionName == null) ? "" : functionName;
     }
--- a/tests/testsrc/test262.properties
+++ b/tests/testsrc/test262.properties
@@ -804,10 +804,8 @@ built-ins/Date 35/594 (5.89%)
 
 ~built-ins/DisposableStack 91/91 (100.0%)
 
-built-ins/Error 7/53 (13.21%)
+built-ins/Error 5/53 (9.43%)
     isError/error-subclass.js {unsupported: [class]}
-    isError/is-a-constructor.js
-    prototype/toString/not-a-constructor.js
     prototype/no-error-data.js
     prototype/S15.11.4_A2.js
     cause_abrupt.js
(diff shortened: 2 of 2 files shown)


aardvark179:
Yeah, that tracks with the test cases that started working on my branch when I converted those classes to be based on descriptors.

gbrail:
I haven't looked at this yet, but in general we have been fixing these "for
free" when we convert the native objects from the old IdScriptableObject
style to the new-ish LambdaConstructor style, and new to the new descriptor
style. I'd rather do that conversion for this class first, as it gives us
other benefits like better performance and optimization opportunities. (And
it seems like this kind of conversion is probably well within the
capabiltiies of opencode or Claude at this point...)

aardvark179:
I actually have those conversions ready on a test branch. The conversions themselves went very smoothly, but filling the gaps in the descriptor API were a little more tricky.

I was waiting to get #2161 and its first successor (fixing our this resolution) in before landing the descriptor work, but could pull it forward I think.

jimmyhmiller:
Feel free to close if you don't want it. If figure the 5 line diff to 65 tests was a pretty good win. But if we want to wait for the conversion to the new style no worries.

jimmyhmiller:
@gbrail Do you want me to close this in favor of waiting on @aardvark179's refactor?

andreabergia:
> I haven't looked at this yet, but in general we have been fixing these "for free" when we convert the native objects from the old IdScriptableObject style to the new-ish LambdaConstructor style, and new to the new descriptor style. I'd rather do that conversion for this class first, as it gives us other benefits like better performance and optimization opportunities. 

I'm not sure what you mean with "this class", because Jimmy's fix is in the base `IdFunctionObject`.

IMHO this is worth merging; if we ever complete the migration away from `IdScriptableObject` we might get rid of that class, but until then, we should fix its problems because it's still pretty commonly used. And it's part of the public API - there might be applications using Rhino that create classes that extend it. This is a tiny, very targeted code change.

gbrail:
OK, that makes sense. I'm happy to merge this. Can you please resolve the conflicts? Thanks!

(All we need is for you to "git pull origin master" or whatever you do, resolve the conflicts, and push the result back to your branch -- I can handle a rebase later using GitHub.)

jimmyhmiller:
@gbrail I like keeping my master rebased with upstream anyways. All rebased and ready to merge whenever you want.

gbrail:
Agreed -- thanks!

PR #2287, opened 2026-02-06
`IdScriptableObject` conversion to descriptors.

aardvark179:
Keeping this as a draft for now, because there are some more bits of API cleanup I'd like to pull in, and I'm trying to decide if we should do this in one go, or class by class.

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/ArrayLikeAbstractOperations.java
+++ b/rhino/src/main/java/org/mozilla/javascript/ArrayLikeAbstractOperations.java
@@ -39,7 +39,7 @@ public class ArrayLikeAbstractOperations {
             Context cx,
             IterativeOperation operation,
             Scriptable scope,
-            Scriptable thisObj,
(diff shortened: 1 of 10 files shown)


aardvark179:
On my experimental branch I've now got all the commits reordered so all the API changes come first, and then the conversion of built-ins to descriptors. One thing this has made clear is there is a lot of repetition because of the tow axes of prototype–constructor and String–Symbol. We aren't going to get rid of that entirely but I'm going to try and at least simplify this internally, and that should hopefully avoid any lingering bugs on a single branch. I'm also going to take a look at boiling down a couple of related APIs I added as I found corner cases into simpler unified ones.

aardvark179:
Okay, done most of the API cleanup, still need to tidy up JavaDocs, and a couple of things, but all the other changes stack up on this nicely.

Given the relative size of the commits I'll definitely do the `IdScriptable` conversions one at a time, and turn this first PR into just the API cleanup.

PR #2293, opened 2026-02-07
Refactor & tidy up API based on experience with converting built ins.

aardvark179:
This is the full tidy up of the AIP. I'll stack the conversions of `IdScriptableObject`s on top of this.

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/ArrayLikeAbstractOperations.java
+++ b/rhino/src/main/java/org/mozilla/javascript/ArrayLikeAbstractOperations.java
@@ -39,7 +39,7 @@ public class ArrayLikeAbstractOperations {
             Context cx,
             IterativeOperation operation,
             Scriptable scope,
-            Scriptable thisObj,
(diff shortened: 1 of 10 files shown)


aardvark179:
I've created PRs for most of the `IdScriptableObject` classes I have converted to descriptors. I'll get the XML related classes done tomorrow—they have a couple of tricky interactions with the scope refactoring that I haven't got sorted out yet—and `NativeContinuation` should be easy after that as should be `NativeGenerator`.

`NativeCall` has no practical use (instances are not accessible as an except maybe via `__parent__` and constructor isn't useful), and in my scopes work is no longer a `Scriptable` at all, so is trivial to convert, or can be left until after the scopes work.

`NativeCallSite` should probably be looked at alongside `NativeError`, but… I think it should actually be looked at in terms of a wider refactoring of how we handle error formatting. We aren't really using it for its purpose as described in the V8 stack API.

gbrail:
Thanks -- looks good to me!

PR #2294, opened 2026-02-07
Convert `NativeDate` to descriptor.

aardvark179:
This stacks on top of #2293 and converts `NativeDate` from `IdScriptableObject` to a descriptor. This change does cause the reflection based `subclassing` 262 test to start failing. This is because we had special handling for `IdScriptableObject` to fake up just enough `new.target` handling for it to work, but we can't extend that more widely without causing other test failures.

I propose to let these test fail during the conversion of built ins to descriptors, and then implement proper support for `new.target` once the transition is complete and we can utilise the additional arguments in the `JSCode` interface.


gbrail:
Looks good as well, thanks!

PR #2295, opened 2026-02-07
Convert `NativeError` to descriptor

aardvark179:
Stacks on top of #2294 .

### DIFF
--- a/tests/testsrc/test262.properties
+++ b/tests/testsrc/test262.properties
@@ -804,12 +804,8 @@ built-ins/Date 35/594 (5.89%)
 
 ~built-ins/DisposableStack 91/91 (100.0%)
 
-built-ins/Error 7/53 (13.21%)
+built-ins/Error 3/53 (5.66%)
     isError/error-subclass.js {unsupported: [class]}
-    isError/is-a-constructor.js
-    prototype/toString/not-a-constructor.js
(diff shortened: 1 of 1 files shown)


rhino/src/main/java/org/mozilla/javascript/NativeError.java
```diff
@@ -131,31 +188,31 @@ static void installCause(NativeObject options, NativeError obj) {
...
-                0);
-        ctor.defineProperty(
-                "prepareStackTrace",
-                protoProps::getPrepareStackTrace,
-                protoProps::setPrepareStackTrace,
-                0);
-
-        super.fillConstructorProperties(ctor);
-    }
+    // @Override
```
gbrail: Can you please take out this big commented-out bit before we merge?

aardvark179:
Okay, we do at least have some tests for `prepareStackTrace`, even though I'm not sure the feature as implemented really achieves V8's goal with this API, but it will do for now.

rhino/src/main/java/org/mozilla/javascript/NativeError.java
aardvark179: Commented section has been removed.

gbrail:
Thanks! Rebased and merged it manually.

PR #2296, opened 2026-02-07
Convert regexp objects to descriptors.

aardvark179:
This converts the regexp classes to uses descriptors. I'm not sure the `RegExpCtor` class even really needs to continue to exist, I think at this point it could simply be folded into the main `NativeRegExp`, but I haven't done that yet.

### DIFF
--- a/tests/src/test/java/org/mozilla/javascript/tests/Test262SuiteTest.java
+++ b/tests/src/test/java/org/mozilla/javascript/tests/Test262SuiteTest.java
@@ -109,7 +109,6 @@ public class Test262SuiteTest {
                             "default-arg",
                             "new.target",
                             "object-rest",
-                            "regexp-dotall",
                             "resizable-arraybuffer",
                             "SharedArrayBuffer",
                             "tail-call-optimization",
--- a/tests/testsrc/test262.properties
+++ b/tests/testsrc/test262.properties
@@ -16,7 +16,7 @@ annexB/built-ins 54/241 (22.41%)
     RegExp/prototype/compile/pattern-regexp-flags-defined.js
     RegExp/prototype/compile/this-cross-realm-instance.js
     RegExp/prototype/compile/this-subclass-instance.js {unsupported: [class]}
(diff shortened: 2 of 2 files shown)


gbrail:
This looks good to me but since they're all stacked I'm going to wait for changes to NativeError before taking this. Thanks!

aardvark179:
> This looks good to me but since they're all stacked I'm going to wait for changes to NativeError before taking this. Thanks!

I've unstacked this one, since now the base is in and the rest is independent. I want to take a slightly longer look at the error case to make sure we're actually testing the things we claim to support.

gbrail:
Thanks! This one looks good too and I'm going to merge it manually as well.

rhino/src/main/java/org/mozilla/javascript/regexp/NativeRegExp.java
```diff
@@ -147,39 +150,143 @@ public class NativeRegExp extends IdScriptableObject {
...
+                        .withMethod(PROTO, SymbolKey.SEARCH, 1, NativeRegExp::js_search)
+                        .withMethod(PROTO, SymbolKey.REPLACE, 2, NativeRegExp::js_replace)
+                        .withMethod(PROTO, SymbolKey.SPLIT, 2, NativeRegExp::js_split)
+                        .withProp(CTOR, SymbolKey.SPECIES, ScriptRuntimeES6::symbolSpecies)
+                        .withProp(
+                                PROTO,
+                                "source",
+                                (c) -> new String(realThis(c, "get source").re.source),
+                                null,
+                                DONTENUM | READONLY | PERMANENT)
```
rbri, line 176: @aardvark179 why is this permanent, at least in real browsers this is configurable

aardvark179: Good point. I preserved our existing behaviour, but it is wrong, and fixing it does make two further tests pass.

rbri: ok, will have a look at my tests and mybe make a pr, thanks @aardvark179 and have a nice weekend

PR #2298, opened 2026-02-08
Make `JavaMembers` immutable

ZZZank:
Thr goal of this PR (and possibility some follow-up PRs) is to make JavaMembers an info-only and immutable data holder. To achieve this, the mutable `NativeJavaMethod` object is replaced by an new immutable class `ExecutableOverload`.

There's also some cleanup and code logic simplification.

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/JavaMembers.java
+++ b/rhino/src/main/java/org/mozilla/javascript/JavaMembers.java
@@ -445,30 +445,23 @@ class JavaMembers {
 
     private void reflect(
             Context cx, Scriptable scope, boolean includeProtected, boolean includePrivate) {
-        // We reflect methods first, because we want overloaded field/method
-        // names to be allocated to the NativeJavaMethod before the field
-        // gets in the way.
         var typeFactory = TypeInfoFactory.get(scope);
 
         var accessibleMethods = discoverAccessibleMethods(cl, includeProtected, includePrivate);
         var accessibleFields = getAccessibleFields(includeProtected, includePrivate);
 
+        // We reflect methods first, because we want overloaded field/method
+        // names to be allocated to the NativeJavaMethod before the field
+        // gets in the way.
         for (int cursor = 0; cursor < 2; cursor++) {
             var isStatic = (cursor == 0);
 
             collectMethods(accessibleMethods, isStatic, typeFactory);
 
             collectFields(accessibleFields, isStatic, typeFactory);
-        }
 
-        // Create bean properties from corresponding get/set methods first for
-        // static members and then for instance members
-        for (int tableCursor = 0; tableCursor != 2; ++tableCursor) {
-            boolean isStatic = (tableCursor == 0);
-            Map<String, Object> ht = isStatic ? staticMembers : members;
-
-            // Add the new bean properties.
-            ht.putAll(extractBeaning(ht, isStatic, includePrivate));
+            var table = isStatic ? staticMembers : members;
+            table.putAll(extractBeaning(table, isStatic, includePrivate));
         }
 
         // Reflect constructors
@@ -595,7 +588,12 @@ class JavaMembers {
         return nameComponent;
     }
 
-    /** at this stage, {@code members} includes {@link NativeJavaMethod} and {@link Field} */
+    /**
+     * Create bean properties from corresponding get/set methods
+     *
+     * <p>see {@link #collectFields(Collection, boolean, TypeInfoFactory)} for possible values in
+     * member table at this stage
+     */
     private static Map<String, BeanProperty> extractBeaning(
             Map<String, Object> members, boolean isStatic, boolean includePrivate) {
         var beans = new HashMap<String, BeanProperty>();


aardvark179:
I'll try and review this more thoroughly tomorrow, but it definitely looks like a good direction. One thing I think is worth thinking about is what types are actually being stored in `members` and `staticMembers`? At the moment the code is quite complex because those maps can contain any sort of `Object`, but if you can start to get that down to a set of classes you control and can fit into a type hierarchy then Ithat will really help make the code easier to understand.

ZZZank:
> get that down to a set of classes you control and can fit into a type hierarchy

Surely I will, my plan is to introduce an interface like this:

```java
interface JavaMember {
    /// `name` in `obj.name` and `obj['name']`. Not guaranteed to be the same as [Member#getName()]
    String name();
}
```

and make all possible values in `members` implement this interface. This is also the reason why I will usually introduce a seemingly useless `private final String name;` field when refactoring existed classes

rhino/src/main/java/org/mozilla/javascript/JavaMembers.java
```diff
@@ -495,13 +505,12 @@ private void reflect(
...
-                                    scope, method.methods, new NativeJavaField(field, typeFactory));
-                    Map<String, FieldAndMethods> fmht =
-                            isStatic ? staticFieldAndMethods : fieldAndMethods;
+                } else if (member instanceof ExecutableOverload) {
+                    var withField =
+                            new ExecutableOverload.WithField(
+                                    (ExecutableOverload) member,
+                                    new NativeJavaField(field, typeFactory));
+                    var fmht = isStatic ? staticFieldAndMethods : fieldAndMethods;
                     if (fmht == null) {
```
aardvark179: This, and the similar cases which check for the hast tables being `null` are just making the code harder to read and understand. The `members` and `staticMembers` maps are initialised in the constructor so can be made null, and I think it's reasonable to make `fieldAndMethods` and ``staticFieldAndMethods` initialised in the constructor as well. Then all those fields can be made final and a bunch of null checks can be removed.

rhino/src/main/java/org/mozilla/javascript/JavaMembers.java
```diff
@@ -782,14 +791,14 @@ private static ExecutableBox extractSetMethod(ExecutableBox[] methods, boolean i
...
-        Map<String, FieldAndMethods> result = new HashMap<>(len);
-        for (FieldAndMethods fam : ht.values()) {
-            FieldAndMethods famNew = new FieldAndMethods(scope, fam.methods, fam.field);
-            famNew.javaObject = javaObject;
-            result.put(fam.field.raw().getName(), famNew);
+        Map<String, FieldAndMethods> result = new HashMap<>((int) (len / 0.75 + 1));
+        for (var withField : ht.values()) {
+            var fieldAndMethods = new FieldAndMethods(scope, withField);
+            fieldAndMethods.javaObject = javaObject;
+            result.put(withField.name, fieldAndMethods);
```
aardvark179: Is `withField.name` always the same as the key from the map? I think it is, but not using the entry set key or a simple mapping here makes me wonder if I'm wrong.

rhino/src/main/java/org/mozilla/javascript/JavaMembers.java
```diff
@@ -782,14 +791,14 @@ private static ExecutableBox extractSetMethod(ExecutableBox[] methods, boolean i
...
-        Map<String, FieldAndMethods> ht = isStatic ? staticFieldAndMethods : fieldAndMethods;
+        var ht = isStatic ? staticFieldAndMethods : fieldAndMethods;
         if (ht == null) return null;
         int len = ht.size();
-        Map<String, FieldAndMethods> result = new HashMap<>(len);
-        for (FieldAndMethods fam : ht.values()) {
-            FieldAndMethods famNew = new FieldAndMethods(scope, fam.methods, fam.field);
-            famNew.javaObject = javaObject;
-            result.put(fam.field.raw().getName(), famNew);
+        Map<String, FieldAndMethods> result = new HashMap<>((int) (len / 0.75 + 1));
```
aardvark179: You're mapping the all the entries from one map to another, so I would suggest just make the initial capacity the old size. It might even be worth writing this in a streams style like

```Java
return (isStatic ? staticFieldAndMethods : fieldAndMethods).entrySet().stream().collect(
    Collectors.toMap(k -> k, withField -> {
        var fAndM = new FieldAndMethods(scope, withField);
        fAndM.javaObject = javaObject;
        return fAndM; });
```

rhino/src/main/java/org/mozilla/javascript/JavaMembers.java
```diff
@@ -431,8 +446,7 @@ private void reflect(
         // gets in the way.
         var typeFactory = TypeInfoFactory.get(scope);
 
-        Method[] methods = discoverAccessibleMethods(cl, includeProtected, includePrivate);
-        for (Method method : methods) {
+        for (Method method : discoverAccessibleMethods(cl, includeProtected, includePrivate)) {
```
aardvark179: Could we maybe refactor this giant `reflect` method into some more self contained things? The comments definitely suggest that it really is five distinct stages, could we actually extract these? Getting rid of of the `methods` variable is a great start on this because it makes it easier to see it's not used later, but it's still a bit of a monster.

aardvark179 (commented):
This looks good, but I think there are a couple of things that could be done to help simplify things in this class as part of this work.

rhino/src/main/java/org/mozilla/javascript/JavaMembers.java
ZZZank: Yes it is, and if the interface design in https://github.com/mozilla/rhino/pull/2298#issuecomment-3869901194 actually came true, then all `JavaMember#name()` will be the same as the corresponding key in map.

Though `withField.name` is indeed confusing, I will change that

rhino/src/main/java/org/mozilla/javascript/JavaMembers.java
ZZZank: the `new HashMap(int)` constructor is strange, because the provided integer doesn't actually represent the max amount of elements before resizing. So I used `(int) (len / 0.75 + 1)` to compute the actual expected capacity. Sadly `HashMap.newHashMap(expectedSize)` is not available before Java 19

gbrail:
@aardvark179 since you're engaged in a few of these, can you give me an "approved" when you're happy with them?

@ZZZank there are a bunch of PRs in the queue from you with a variety of levels of comments. I'd like to get to a point where at least the two of you are happy with them and I might come back and ask if there's an error that you need.

aardvark179 (approved):
Looks good, there's probably more things we can tidy up here, but we don't have to do it all in one go.

aardvark179:
@ZZZank it would be great to talk through where we want to go with all the Java interop stuff. Want to get on the slack so we can talk about this, and/or have a zoom meeting to talk through this stuff?

ZZZank:
@aardvark179 Sure, how can I join Slack channel?

Also I got "Slack not available in your country" when signing up for Slack and ended up used a VPN. So I might be randomly unavailable when my VPN provider is malfunctioning

aardvark179:
@gbrail could we get this merged soon? It will be needed for #2330 to be merged without breaking contiuations.

gbrail:
Yes, this looks good too, thanks!

Issue #2265, opened 2026-01-20
Custom WrapFactory to handle byte[] no longer working in 1.9.0


CardContact:
In OpenSCDP we use a lot of binary data and the Global Platform Scripting Spec that we implement
has a dedicated [ByteString](https://www.openscdp.org/scsh3/bytestring.html) and [ByteBuffer](https://www.openscdp.org/scsh3/bytebuffer.html) host class.

In our embedding we use a custom WrapFactory that converts a byte[] returned from native 
Java invocations into ByteString instances:

```
>var bs = java.lang.reflect.Array.newInstance(java.lang.Byte.TYPE, 10);
>bs instanceof ByteString
true
```

With 1.9 this stops working, as WrapFactory.wrap() is no longer called for byte[] objects. Instead this version seems to generate NativeJavaArray objects.

Before we rewrite the host objects to support NativeJavaArray, I want to make sure that there is no other way to restore the previous logic.

aardvark179:
`WrapFactory` has certainly seen some work, but I believe `wrap` should still be called in most cases, but I see @ZZZank short circuited it and called `wrapAsJavaObject` directly in `InterfaceAdapter` and `NativeJavaObject`. I'd suggest moving your customisation their as that has been a public method for a long time, and documented as being called by `wrap`.

I think you may be doing some fairly uncommon things in your embedding, so it would be really helpful if you can turn those into unit tests so we have a better idea of the behaviour you are depending on, and if we change it then it should be more obvious from the changes in the tests.

aardvark179:
Actually, I see [quite a few custom wrap factories](https://github.com/search?q=org.mozilla.javascript+%22extends+WrapFactory%22+NOT+path%3A**%2FWrapFactory.java+language%3AJava&type=code), we may need look at how much behaviour we are covering for custom wrap factories and make sure we aren't causing accidental regressions.

CardContact:
I replaced `wrap` with `wrapAsJavaObject` and that is working fine.

ZZZank:
Cna you elaborate how did you customize the `WrapFactory`? In the provided code example, `byte[]` is created via `Array.newInstance(...)` method call, which will indeed call `wrap` instead of `wrapAsJavaObject` to wrap the result:

https://github.com/mozilla/rhino/blob/c9b182a468e16f189b2b00efc84afff9f2e8e5fa/rhino/src/main/java/org/mozilla/javascript/NativeJavaMethod.java#L209-L213

(see Line 211)

[...]

PR #2299, opened 2026-02-11
Enforce `WrapFactory` users to override new `wrap(...)` and `wrapAsJavaObject(...)` method

ZZZank:
- Javadoc for old methods are moved to new methods. Old javadoc now provides a link to new method
- `@since` for new methods
- old methods that are not supposed to be overridden now are marked as `final`.

Hopefully these changes can properly notify users about `WrapFactory` changes, and prevent issues similar to #2265


gbrail:
Thanks -- looks good!

PR #2302, opened 2026-02-16
Fix for setting prototype and parent scopes.

aardvark179:
@rbri noticed that after the change to descriptors we had stopped setting the patent scope on some objects.

I'm trying to get the API right now, so that as we introduce support for `new.target` we'll have one place where we need iterate on things, hence I'm making sure we're passing in everything we need to make those decisions correctly. I'm passing in new target and the built in because those correspond to what [OrdinaryCreateFromConstructor](https://tc39.es/ecma262/#sec-ordinarycreatefromconstructor) uses.

### DIFF
--- a/tests/src/test/java/org/mozilla/javascript/tests/BuiltinCtorTest.java
+++ b/tests/src/test/java/org/mozilla/javascript/tests/BuiltinCtorTest.java
@@ -1,0 +1,41 @@
+package org.mozilla.javascript.tests;
+
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
+
+import org.junit.Test;
+import org.mozilla.javascript.Context;
+import org.mozilla.javascript.Scriptable;
+import org.mozilla.javascript.TopLevel;
+
+public class BuiltinCtorTest {
+
+    @Test
+    public void checkCtorArity() {
+        try (Context cx = Context.enter()) {
+            TopLevel tl = new TopLevel();
+            cx.initSafeStandardObjects(tl);
+            for (var bi : TopLevel.Builtins.values()) {
+                var ctor = tl.getBuiltinCtor(bi);
+                if (ctor.isConstructor()) {
+
+                    Object res = null;
+                    try {
+                        res = ctor.construct(cx, tl, new Object[0]);
+                    } catch (Exception e) {
+                        // Some constructors do throw a type error.
+                    }
+                    if (res == null) {
+                        continue;
+                    }
+                    assertTrue(
+                            String.format("Expected %s() to return an object", bi.name()),
+                            res instanceof Scriptable);
+                    assertNotNull(
+                            String.format("Expected %s() to set parent scope", bi.name()),
+                            ((Scriptable) res).getParentScope());
+                }
+            }
+        }
+    }
+}


rbri:
@aardvark179 thanks for the fast fix - do you see any option for some unit test?

rbri:
first smoke test was successful

aardvark179:
> @aardvark179 thanks for the fast fix - do you see any option for some unit test?

I'm trying to work out what the tests should be. `__parent__` is not specified, and it feels like there are a lot of possible corner cases. Should built in constructors set the parent scope to the caller's scope, or constructor's scope? For JS functions do the same rules apply? What if the constructor builds an object itself and returns that—should we respect its parent scope, or reset it explicitly? 

Do you happen to know what bits of behaviour are important to htmlunit?

rbri:
> Do you happen to know what bits of behaviour are important to htmlunit?

so far I faced only one place where this comes into play
* the URLSearchParam ctor (https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/URLSearchParams); supports an iterator

My code looks like this

        // if params is a sequence
        if (params instanceof Scriptable paramsScriptable && hasProperty(paramsScriptable, SymbolKey.ITERATOR)) {

            final Context cx = Context.getCurrentContext();

            final List<NameValuePair> nameValuePairs = new ArrayList<>();

            try (IteratorLikeIterable itr = buildIteratorLikeIterable(cx, paramsScriptable)) {
                for (final Object nameValue : itr) {

and the buildIteratorLikeIterable()

    private static IteratorLikeIterable buildIteratorLikeIterable(final Context cx, final Scriptable iterable) {
        final Object iterator = ScriptRuntime.callIterator(iterable, cx, iterable.getParentScope());
        return new IteratorLikeIterable(cx, iterable.getParentScope(), iterator);
    }

And there is the problem, the iterable is the params NativeArray with null parentScope
Finally you end up with the NPE

    Caused by: java.lang.NullPointerException: Cannot invoke "org.htmlunit.corejs.javascript.Scriptable.getParentScope()" because "obj" is null
	at org.htmlunit.corejs.javascript.ScriptableObject.getTopLevelScope(ScriptableObject.java:2404)
	at org.htmlunit.corejs.javascript.ES6Iterator.<init>(ES6Iterator.java:60)
	at org.htmlunit.corejs.javascript.NativeArrayIterator.<init>(NativeArrayIterator.java:33)
	at org.htmlunit.corejs.javascript.NativeArray.js_values(NativeArray.java:2073)
	at org.htmlunit.corejs.javascript.ClassDescriptor$BuiltInJSCode.execute(ClassDescriptor.java:743)
	at org.htmlunit.corejs.javascript.ClassDescriptor$BuiltInJSCode.execute(ClassDescriptor.java:723)
	at org.htmlunit.corejs.javascript.JSFunction.call(JSFunction.java:146)
	at org.htmlunit.corejs.javascript.ScriptRuntime.callIterator(ScriptRuntime.java:3328)
	at org.htmlunit.javascript.host.URLSearchParams.buildIteratorLikeIterable(URLSearchParams.java:270)
	at org.htmlunit.javascript.host.URLSearchParams.resolveParams(URLSearchParams.java:186)
	at org.htmlunit.javascript.host.URLSearchParams.jsConstructor(URLSearchParams.java:165)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)

aardvark179:
Okay. I think the real bug is that the iterator should be created based on the declaration scope of `Array.prototype.values`, not the scope of the parent scope of the caller (which we don't have so fill in with the array's parent scope).

aardvark179:
Okay, after thinking about this for a while I've added a test that every built in constructor that can be called as a constructor with zero args produces an object with a non-null parent scope. Four of them return errors—`Symbol` and `BigInt` because they are specified to be constructors but to throw when called as such, and `Promise` and `DataView` because they have a required argument.

gbrail:
Yes, this makes sense to me too, thanks!

PR #2304, opened 2026-02-19
Avoid compiling `a.eval()` as a direct eval.

aardvark179:
Compiling `other.eval()` as a direct eval causes problems if `other` is a different realm as we end up evaluating the argument with the wrong parent scope. This caused one test to incorrectly pass (we got the wrong generator function prototype in the eval, but also set the wrong prototype when doing `Reflect.construct`).


gbrail:
Thanks -- looks good!

PR #2305, opened 2026-02-20
fix some jdoc errors

rbri:


gbrail:
Thanks! This reduces the number of warnings on publishing, although "TypeFormatContext.java" still has a bunch.

PR #2306, opened 2026-02-20
Convert XML objects from `IdScriptableObject`s to descriptors

aardvark179:
Now all the `hasSymbol` work has landed this now rebased quite easily from my test branch.

### DIFF
--- a/rhino-xml/src/main/java/org/mozilla/javascript/xmlimpl/QName.java
+++ b/rhino-xml/src/main/java/org/mozilla/javascript/xmlimpl/QName.java
@@ -6,19 +6,58 @@
 
 package org.mozilla.javascript.xmlimpl;
 
-import java.util.Objects;
+import static org.mozilla.javascript.ClassDescriptor.Destination.PROTO;
+
+import org.mozilla.javascript.ClassDescriptor;
 import org.mozilla.javascript.Context;
-import org.mozilla.javascript.IdFunctionObject;
-import org.mozilla.javascript.IdScriptableObject;
+import org.mozilla.javascript.JSFunction;
 import org.mozilla.javascript.ScriptRuntime;
 import org.mozilla.javascript.Scriptable;
+import org.mozilla.javascript.ScriptableObject;
 import org.mozilla.javascript.Undefined;
 
 /** Class QName */
-final class QName extends IdScriptableObject {
+final class QName extends ScriptableObject {
     static final long serialVersionUID = 416745167693026750L;
 
-    private static final Object QNAME_TAG = "QName";
+    private static final String QNAME_TAG = "QName";
+
+    private static final ClassDescriptor DESCRIPTOR;
+
+    static {
+        DESCRIPTOR =
+                new ClassDescriptor.Builder(
+                                QNAME_TAG, 0, QName::js_constructorCall, QName::js_constructor)
+                        .withMethod(PROTO, "toString", 0, QName::js_toString)
+                        .withMethod(PROTO, "toSource", 0, QName::js_toSource)
+                        .build();
+    }
+
+    static void init(Context cx, Scriptable scope, ScriptableObject proto, boolean sealed) {
+        DESCRIPTOR.buildConstructor(cx, scope, proto, sealed);
+    }
+
+    private static Object js_constructor(
+            Context cx, JSFunction f, Object nt, Scriptable s, Object thisObj, Object[] args) {
+        var realThis = realThis(f.getPrototypeProperty(), f);
+        return realThis.jsConstructor(cx, true, args);
+    }
+
+    private static Object js_constructorCall(
+            Context cx, JSFunction f, Object nt, Scriptable s, Object thisObj, Object[] args) {
+        var realThis = realThis(f.getPrototypeProperty(), f);
+        return realThis.jsConstructor(cx, false, args);
+    }
+
+    private static Object js_toString(
+            Context cx, JSFunction f, Object nt, Scriptable s, Object thisObj, Object[] args) {
+        return realThis(thisObj, f).toString();
+    }
+
+    private static Object js_toSource(
+            Context cx, JSFunction f, Object nt, Scriptable s, Object thisObj, Object[] args) {
+        return realThis(thisObj, f).js_toSource();
+    }
 
     private XMLLibImpl lib;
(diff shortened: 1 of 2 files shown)


gbrail:
Thanks! Can you please get rid of the unused constructor:

/home/gjb/src/rhino/rhino-xml/src/main/java/org/mozilla/javascript/xmlimpl/XMLCtor.java:23: warning: [UnusedMethod] Constructor 'XMLCtor' is never used.
    private XMLCtor(Context cx, Scriptable scope, XML xml, Object tag, int id, int arity) {
            ^
    (see https://errorprone.info/bugpattern/UnusedMethod)

aardvark179:
I will. I'd actually like to remove the entire `XMLCtor` class since it is no longer really needed, but I'll save that for a follow up PR because it requires a little bit of delicate surgery round access to the lib.

aardvark179:
I've stacked the actor class removal on top of this at https://github.com/aardvark179/rhino/tree/aardvark179-ctor-cleanup, I think that's a big enough set of change to be its own PR.

rhino-xml/src/main/java/org/mozilla/javascript/xmlimpl/Namespace.java
```diff
@@ -6,19 +6,39 @@
...
+    private static final String NAMESPACE_TAG = "Namespace";
+
+    private static final ClassDescriptor DESCRIPTOR;
+
+    static {
+        DESCRIPTOR =
+                new ClassDescriptor.Builder(
+                                NAMESPACE_TAG,
+                                2,
+                                Namespace::js_constructor,
```
gbrail: The ClassDescriptor.Builder signature has "call" and then "construct" and that seems to propagate within the ClassDescriptor code. It looks like we're passing the constructor and "call" arguments in reverse order here, which doesn't seem to show up on tests if it's a bug.

rhino-xml/src/main/java/org/mozilla/javascript/xmlimpl/QName.java
```diff
@@ -6,19 +6,58 @@
...
 
-    private static final Object QNAME_TAG = "QName";
+    private static final String QNAME_TAG = "QName";
+
+    private static final ClassDescriptor DESCRIPTOR;
+
+    static {
+        DESCRIPTOR =
+                new ClassDescriptor.Builder(
+                                QNAME_TAG, 0, QName::js_constructor, QName::js_constructorCall)
```
gbrail: Same question with "call" and "construct"  -- seems like the descriptor class expects them the other way around.

rhino-xml/src/main/java/org/mozilla/javascript/xmlimpl/XMLObjectImpl.java
```diff
@@ -26,10 +27,66 @@
...
     private XMLLibImpl lib;
     private boolean prototypeFlag;
 
+    private static final ClassDescriptor DESCRIPTOR;
+
+    static {
+        DESCRIPTOR = new ClassDescriptor.Builder(XMLOBJECT_TAG).build();
+    }
+
+    public static ClassDescriptor.Builder populatePrototypeDescrriptor(
```
gbrail: Typo? (Or pirates wrote this code, not sure about all the rrrs)

rhino-xml/src/main/java/org/mozilla/javascript/xmlimpl/XML.java
```diff
@@ -17,6 +24,96 @@ class XML extends XMLObjectImpl {
...
+                                        "XML", 1, XML::js_constructorCall, XML::js_constructor))
+                        .withMethod(CTOR, SymbolKey.HAS_INSTANCE, 1, XML::js_hasInstance)
+                        .build();
+    }
+
+    public static void init(
+            Context cx,
+            ScriptableObject scope,
+            XMLObjectImpl proto,
+            boolean sealdd,
```
gbrail: Typo and should be "sealed"?

rhino-xml/src/main/java/org/mozilla/javascript/xmlimpl/XML.java
```diff
@@ -17,6 +24,96 @@ class XML extends XMLObjectImpl {
...
+        var xmlListProto =
+                xmlListCtor instanceof JSFunction
+                        ? ((JSFunction) xmlListCtor).getPrototypeProperty()
+                        : null;
+
+        if (!(xmlProto instanceof XML)) {
+            throw ScriptRuntime.typeErrorById("msg.arg.not.object", ScriptRuntime.typeof(xmlProto));
+        }
+        if (!(xmlListProto instanceof XMLList)) {
+            throw ScriptRuntime.typeErrorById("msg.arg.not.object", ScriptRuntime.typeof(xmlProto));
```
gbrail: Should this be "ScriptRuntime.typeof(xmlListProto)"?

gbrail (changes requested):
In trying to have all the PRs reviewed by a decent AI it found some bugs and it looks like there's a good one in here, so please take a look. Thanks!

rhino-xml/src/main/java/org/mozilla/javascript/xmlimpl/Namespace.java
aardvark179: Fixed.

rhino-xml/src/main/java/org/mozilla/javascript/xmlimpl/QName.java
aardvark179: Fixed.

rhino-xml/src/main/java/org/mozilla/javascript/xmlimpl/XMLObjectImpl.java
aardvark179: Fixed (after hitting the LSP server hard enough to make it _actually_ change the name).

rhino-xml/src/main/java/org/mozilla/javascript/xmlimpl/XML.java
aardvark179: I think so, yes. Fixed.

gbrail:
This is great, thanks!

PR #2307, opened 2026-02-20
Make `RegExp.prototype.source` configurable.

aardvark179:
Thanks to @rbri for pointing out this was wrong. It was wrong in the `IdScriptbaleObject` implementation, and I copied the wrongness over. 🙂


rbri:
Thanks @aardvark179

PR #2308, opened 2026-02-20
Fix generator .return() to include value when in completed state

jimmyhmiller:
When .return(value) is called on a completed generator, set the value property on the result object instead of returning undefined.


gbrail:
Yes, this is a nice clean fix. Thanks!

PR #2310, opened 2026-02-21
Add OSGi bundle metadata

pavelhoral:
This is a different attempt on #1278.

* I have added metadata for rhino, rhino-engine and rhino-tools as that is what I need. I am not sure if the same should be done for rhino-xml.
* The imports and exports to some degree mirror what is being declared in module-info.java.
* I took inspiration from Groovy metadata and Apache ServiceMix bundle for old RhinoJS. However I am not 100 % sure everything is needed (especially the Service Loader stuff for the `RegExpLoader`).

Disclamer: I got help from Claude Code, but I have verified that the bundles correctly load and work in my OSGi environment. I have tested the OSGi bundles on `rhino_1_9_1` branch - would be nice to get this backported as OSGi does not like major version changes).

### DIFF
--- a/tests/build.gradle
+++ b/tests/build.gradle
@@ -4,6 +4,19 @@ plugins {
     id 'jacoco-report-aggregation'
 }
 
+sourceSets {
+    integrationTest {
+        java.srcDir 'src/integrationTest/java'
+        compileClasspath += sourceSets.main.output
+        runtimeClasspath += sourceSets.main.output
+    }
+}
+
+configurations {
+    integrationTestImplementation.extendsFrom testImplementation
+    integrationTestRuntimeOnly.extendsFrom testRuntimeOnly
+}
+
 dependencies {
     implementation project(':rhino-engine')
     implementation project(':examples')
@@ -19,6 +32,8 @@ dependencies {
     testImplementation "org.yaml:snakeyaml:2.6"
     testImplementation "javax.xml.soap:javax.xml.soap-api:1.4.0"
     testImplementation "javax.activation:javax.activation-api:1.2.0"
+
+    integrationTestImplementation "org.apache.felix:org.apache.felix.framework:7.0.5"
 }
 
 tasks.withType(JavaCompile) {
@@ -28,6 +43,23 @@ tasks.withType(JavaCompile) {
             '-Xlint:-deprecation,unchecked'
     ]
 }
+
+tasks.register('integrationTest', Test) {
+    description = 'Runs JPMS and OSGi integration tests against the built JARs.'
+    group = 'verification'
+    testClassesDirs = sourceSets.integrationTest.output.classesDirs
+    classpath = sourceSets.integrationTest.runtimeClasspath
+
+    useJUnitPlatform()
+
+    dependsOn ':rhino:jar', ':rhino-engine:jar', ':rhino-tools:jar', ':rhino-xml:jar', ':rhino-kotlin:jar'
+
+    [':rhino', ':rhino-engine', ':rhino-tools', ':rhino-xml', ':rhino-kotlin'].each { proj ->
+        systemProperty "${proj.substring(1)}.jar", project(proj).jar.archiveFile.get().asFile.path
+    }
+}
+
+check.dependsOn integrationTest
 
 test {
     maxHeapSize = "1g"
--- a/tests/src/integrationTest/java/org/mozilla/javascript/tests/JpmsModuleTest.java
+++ b/tests/src/integrationTest/java/org/mozilla/javascript/tests/JpmsModuleTest.java
@@ -1,0 +1,133 @@
+package org.mozilla.javascript.tests;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import java.lang.module.Configuration;
+import java.lang.module.ModuleFinder;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.List;
+import java.util.ServiceLoader;
+import java.util.Set;
+import java.util.concurrent.Callable;
+import javax.script.ScriptEngineFactory;
+import org.junit.jupiter.api.Test;
+
+/**
+ * Tests that Rhino modules work correctly under JPMS by creating a ModuleLayer from the built JARs
+ * and verifying real module behavior: resolution, service loading, and composition.
+ */
+public class JpmsModuleTest {
+
+    @Test
+    void rhinoExecutesJavaScriptInModuleLayer() throws Exception {
+        ModuleLayer layer = createLayer(List.of("rhino"), Set.of("org.mozilla.rhino"));
+
+        ClassLoader loader = layer.findLoader("org.mozilla.rhino");
+        Object result = withContextClassLoader(loader, () -> evaluateWithRhino(loader, "1 + 2"));
+        assertEquals(3, ((Number) result).intValue());
+    }
+
+    @Test
+    void scriptEngineDiscoveredViaServiceLoader() throws Exception {
+        ModuleLayer layer =
+                createLayer(
+                        List.of("rhino", "rhino-engine"),
+                        Set.of("org.mozilla.rhino", "org.mozilla.rhino.engine"));
+
+        ClassLoader loader = layer.findLoader("org.mozilla.rhino.engine");
+
+        boolean found =
+                withContextClassLoader(
+                        loader,
+                        () -> {
+                            for (ScriptEngineFactory factory :
+                                    ServiceLoader.load(ScriptEngineFactory.class, loader)) {
+                                if (factory.getEngineName().toLowerCase().contains("rhino")) {
+                                    Object result = factory.getScriptEngine().eval("40 + 2");
+                                    assertEquals(42, ((Number) result).intValue());
+                                    return true;
+                                }
+                            }
+                            return false;
+                        });
+        assertTrue(found, "Rhino ScriptEngineFactory not discovered via ServiceLoader");
+    }
+
+    @Test
+    void xmlModuleWorksInModuleLayer() throws Exception {
+        ModuleLayer layer =
+                createLayer(
+                        List.of("rhino", "rhino-xml"),
+                        Set.of("org.mozilla.rhino", "org.mozilla.javascript.xml"));
+
+        ClassLoader loader = layer.findLoader("org.mozilla.rhino");
+        Object result =
+                withContextClassLoader(loader, () -> evaluateWithRhino(loader, "typeof XML"));
+        assertEquals("function", result);
+    }
+
+    @Test
+    void xmlModuleNotAvailableWithoutXmlInLayer() throws Exception {
+        ModuleLayer layer = createLayer(List.of("rhino"), Set.of("org.mozilla.rhino"));
+
+        ClassLoader loader = layer.findLoader("org.mozilla.rhino");
+        Object result =
+                withContextClassLoader(loader, () -> evaluateWithRhino(loader, "typeof XML"));
+        assertEquals("undefined", result);
+    }
+
+    private ModuleLayer createLayer(List<String> jarProperties, Set<String> moduleNames) {
+        Path[] jarPaths =
+                jarProperties.stream()
+                        .map(
+                                name -> {
+                                    String path = System.getProperty(name + ".jar");
+                                    assertNotNull(
+                                            path,
+                                            "System property '"
+                                                    + name
+                                                    + ".jar' not set."
+                                                    + " Run tests via Gradle.");
+                                    return Paths.get(path);
+                                })
+                        .toArray(Path[]::new);
+
+        ModuleFinder finder = ModuleFinder.of(jarPaths);
+        ModuleLayer parent = ModuleLayer.boot();
+        Configuration parentConfig = parent.configuration();
+        Configuration config = parentConfig.resolve(finder, ModuleFinder.of(), moduleNames);
+
+        return parent.defineModulesWithOneLoader(config, ClassLoader.getPlatformClassLoader());
+    }
+
+    private <T> T withContextClassLoader(ClassLoader loader, Callable<T> action) throws Exception {
+        ClassLoader original = Thread.currentThread().getContextClassLoader();
+        Thread.currentThread().setContextClassLoader(loader);
+        try {
+            return action.call();
+        } finally {
+            Thread.currentThread().setContextClassLoader(original);
+        }
+    }
+
+    private Object evaluateWithRhino(ClassLoader loader, String script) throws Exception {
+        Class<?> ctxClass = loader.loadClass("org.mozilla.javascript.Context");
+        Object cx = ctxClass.getMethod("enter").invoke(null);
+        try {
+            Object scope = ctxClass.getMethod("initStandardObjects").invoke(cx);
+            return ctxClass.getMethod(
+                            "evaluateString",
+                            loader.loadClass("org.mozilla.javascript.VarScope"),
+                            String.class,
+                            String.class,
+                            int.class,
+                            Object.class)
+                    .invoke(cx, scope, script, "test", 1, null);
+        } finally {
+            ctxClass.getMethod("exit").invoke(null);
+        }
+    }
+}
--- a/tests/src/integrationTest/java/org/mozilla/javascript/tests/OsgiModuleTest.java
+++ b/tests/src/integrationTest/java/org/mozilla/javascript/tests/OsgiModuleTest.java
@@ -1,0 +1,216 @@
+package org.mozilla.javascript.tests;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.HashMap;
+import java.util.Map;
+import org.junit.jupiter.api.Test;
+import org.osgi.framework.Bundle;
+import org.osgi.framework.BundleContext;
+import org.osgi.framework.Constants;
+import org.osgi.framework.launch.Framework;
+import org.osgi.framework.launch.FrameworkFactory;
+
+/**
+ * Tests that Rhino bundles work correctly in an OSGi environment by embedding an Apache Felix
+ * framework, installing the built JARs as bundles, and verifying real behavior.
+ */
+public class OsgiModuleTest {
+
+    /**
+     * Provide osgi.serviceloader capabilities via framework properties so bundles resolve without
+     * requiring SPI Fly. Also provide osgi.serviceloader for services that bundles require.
+     */
+    private static final String EXTRA_CAPABILITIES =
+            String.join(
+                    ",",
+                    "osgi.extender;osgi.extender=osgi.serviceloader.processor;version:Version=1.0",
+                    "osgi.extender;osgi.extender=osgi.serviceloader.registrar;version:Version=1.0");
+
+    @Test
+    void rhinoBundleResolvesAndStarts() throws Exception {
+        withFramework(
+                ctx -> {
+                    Bundle rhino = installBundle(ctx, "rhino");
+                    rhino.start();
+                    assertEquals(Bundle.ACTIVE, rhino.getState());
+                });
+    }
+
+    @Test
+    void rhinoExecutesJavaScriptInOsgi() throws Exception {
+        withFramework(
+                ctx -> {
+                    Bundle rhino = installBundle(ctx, "rhino");
+                    rhino.start();
+
+                    Object result = evaluateInBundle(rhino, "1 + 2");
+                    assertEquals(3, ((Number) result).intValue());
+                });
+    }
+
+    @Test
+    void rhinoEngineBundleResolvesAndStarts() throws Exception {
+        withFramework(
+                ctx -> {
+                    Bundle rhino = installBundle(ctx, "rhino");
+                    Bundle engine = installBundle(ctx, "rhino-engine");
+                    rhino.start();
+                    engine.start();
+                    assertEquals(Bundle.ACTIVE, engine.getState());
+
+                    // Verify the ScriptEngineFactory class is loadable from the bundle
+                    Class<?> factory =
+                            engine.loadClass(
+                                    "org.mozilla.javascript.engine.RhinoScriptEngineFactory");
+                    assertNotNull(factory);
+                });
+    }
+
+    @Test
+    void rhinoToolsBundleResolvesAndStarts() throws Exception {
+        withFramework(
+                ctx -> {
+                    Bundle rhino = installBundle(ctx, "rhino");
+                    Bundle tools = installBundle(ctx, "rhino-tools");
+                    rhino.start();
+                    tools.start();
+                    assertEquals(Bundle.ACTIVE, tools.getState());
+                });
+    }
+
+    @Test
+    void rhinoXmlBundleResolvesAndStarts() throws Exception {
+        withFramework(
+                ctx -> {
+                    Bundle rhino = installBundle(ctx, "rhino");
+                    Bundle xml = installBundle(ctx, "rhino-xml");
+                    rhino.start();
+                    xml.start();
+                    assertEquals(Bundle.ACTIVE, xml.getState());
+                });
+    }
+
+    @Test
+    void allBundlesResolveAndStartTogether() throws Exception {
+        withFramework(
+                ctx -> {
+                    Bundle rhino = installBundle(ctx, "rhino");
+                    Bundle engine = installBundle(ctx, "rhino-engine");
+                    Bundle tools = installBundle(ctx, "rhino-tools");
+                    Bundle xml = installBundle(ctx, "rhino-xml");
+
+                    rhino.start();
+                    engine.start();
+                    tools.start();
+                    xml.start();
+
+                    assertEquals(Bundle.ACTIVE, rhino.getState(), "rhino");
+                    assertEquals(Bundle.ACTIVE, engine.getState(), "rhino-engine");
+                    assertEquals(Bundle.ACTIVE, tools.getState(), "rhino-tools");
+                    assertEquals(Bundle.ACTIVE, xml.getState(), "rhino-xml");
+                });
+    }
+
+    /**
+     * Verifies that rhino-xml's exported class is accessible from the bundle classloader, i.e. the
+     * OSGi wiring between rhino and rhino-xml is correct. Full cross-bundle ServiceLoader discovery
+     * (making {@code typeof XML === "function"}) requires an OSGi service loader mediator like SPI
+     * Fly, which is beyond the scope of this test.
+     */
+    @Test
+    void xmlBundleExportsAreAccessible() throws Exception {
+        withFramework(
+                ctx -> {
+                    Bundle rhino = installBundle(ctx, "rhino");
+                    Bundle xml = installBundle(ctx, "rhino-xml");
+                    rhino.start();
+                    xml.start();
+
+                    Class<?> loaderImpl =
+                            xml.loadClass("org.mozilla.javascript.xmlimpl.XMLLoaderImpl");
+                    assertNotNull(loaderImpl);
+                    Class<?> loaderInterface =
+                            rhino.loadClass("org.mozilla.javascript.xml.XMLLoader");
+                    assertTrue(
+                            loaderInterface.isAssignableFrom(loaderImpl),
+                            "XMLLoaderImpl should implement XMLLoader");
+                });
+    }
+
+    /**
+     * Evaluates a JavaScript expression using Rhino classes loaded from the given bundle. Sets the
+     * thread context classloader to the bundle's classloader to ensure ServiceLoader resolves
+     * services from the OSGi environment, not the test classpath.
+     */
+    private Object evaluateInBundle(Bundle bundle, String script) throws Exception {
+        ClassLoader bundleClassLoader =
+                bundle.adapt(org.osgi.framework.wiring.BundleWiring.class).getClassLoader();
+        ClassLoader original = Thread.currentThread().getContextClassLoader();
+        Thread.currentThread().setContextClassLoader(bundleClassLoader);
+        try {
+            Class<?> ctxClass = bundle.loadClass("org.mozilla.javascript.Context");
+            Object cx = ctxClass.getMethod("enter").invoke(null);
+            try {
+                Object scope = ctxClass.getMethod("initStandardObjects").invoke(cx);
+                return ctxClass.getMethod(
+                                "evaluateString",
+                                bundle.loadClass("org.mozilla.javascript.VarScope"),
+                                String.class,
+                                String.class,
+                                int.class,
+                                Object.class)
+                        .invoke(cx, scope, script, "test", 1, null);
+            } finally {
+                ctxClass.getMethod("exit").invoke(null);
+            }
+        } finally {
+            Thread.currentThread().setContextClassLoader(original);
+        }
+    }
+
+    // --- Helpers ---
+
+    private void withFramework(FrameworkConsumer action) throws Exception {
+        Path storageDir = Files.createTempDirectory("osgi-test");
+        Map<String, String> config = new HashMap<>();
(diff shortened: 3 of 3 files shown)


aardvark179:
So, if we want to export OSGi meta-data alongside our JPMS module
Info it feels like we should either ensure that the OSGi data is generated from the JPMS data, or that we have a test ensuring the two remain in sync.

I thought OSGi had even added support for JPMS, but I could be mistaken-remembering, it had been a while since I messed with this stuff.

pavelhoral:
Trying to generate anything would mean a bigger change in the build scripts, probably switching to https://plugins.gradle.org/plugin/biz.aQute.bnd.builder. I can try doing that, if it has a chance of being potentially approved.

pavelhoral:
Played with BND a bit and I am pretty convinced that the only way to keep module-info.java and OSGi metadata in sync is to actually let BND generate both and ideally using its annotation system (https://bnd.bndtools.org/chapters/230-manifest-annotations.html).

pavelhoral:
I have added metadata for rhino-xml and fixed cardinality and resolution attributes.

I have also wrote unit test that checks what can be checked. Ideally this should be accompanied with `bnd verify` to make sure the imports are complete, but that is a task for another day (and maybe you will find more convenient to create the metadata via bnd).

pavelhoral:
I have prepared feature branch for backporting this to 1.9 - https://github.com/pavelhoral/rhino/tree/feat-osgi-1_9.

gbrail:
Does anyone else have experience with BND? If it's a straightforward Gradle plugin, and it lets us generate both Java module config and Osgi metadata from a single source, I'm all for it if you're willing to put together a PR.

pavelhoral:
I have (with help from Claude Code and a lot of double-checking) switched OSGi metadata and JPMS module info build to BND plugin.

<details>
<summary>Generated change summary</summary>

> This commit replaces hand-written OSGi manifest attributes and `module-info.java` files with the [bnd Gradle plugin](https://bnd.bndtools.org/) (`biz.aQute.bnd.builder`), which generates both OSGi and JPMS metadata from a single source of truth in each module's `build.gradle`.
> 
> ## What changed
> 
> **Added bnd plugin** (`biz.aQute.bnd.gradle:7.2.1`) to `buildSrc/build.gradle` and applied `biz.aQute.bnd.builder` to all 5 published modules: `rhino`, `rhino-engine`, `rhino-tools`, `rhino-xml`, `rhino-kotlin`.
> 
> **Deleted all `module-info.java` files** (8 files total -- main and test source sets):
> - `rhino/src/main/java/module-info.java`
> - `rhino-engine/src/main/java/module-info.java` + test
> - `rhino-tools/src/main/java/module-info.java` + test
> - `rhino-xml/src/main/java/module-info.java` + test
> - `rhino-kotlin/src/main/java/module-info.java`
> - `testutils/src/main/java/module-info.java`
> 
> **Deleted `OsgiMetadataTest.java`** -- no longer needed since bnd generates correct metadata by construction.
> 
> **Removed `osgiVersion` helper and `javaModuleVersion`** from `rhino.library-conventions.gradle` -- bnd derives `Bundle-Version` automatically from the project version.
> 
> ## How bnd generates metadata
> 
> Each module's `build.gradle` now contains a `jar { bundle { bnd(...) } }` block with:
> - **OSGi headers**: `Bundle-SymbolicName`, `-exportcontents`, `Provide-Capability`, `Require-Capability`, etc.
> - **JPMS instructions**: `-jpms-module-info`, `-jpms-module-info-options`, `-jpms-module-info-ee` -- bnd uses these plus the OSGi capability model to generate `module-info.class` directly in the JAR.
> - **`Import-Package`** is auto-calculated by bnd from bytecode analysis, with explicit overrides only where needed (e.g., `resolution:=optional` for `android.os`, `java.beans`, `jdk.dynalink*`, `javax.swing*`, `java.awt*`, `org.jline.*`).
> 
> ## Key design decisions
> 
> - **`-exportcontents`** (not `Export-Package`) is used so bnd doesn't filter jar contents -- it only sets the header.
> - **`access=0`** prevents bnd from generating `open` modules (which it defaults to when `osgi.extender` capabilities are present).
> - **`register:=` directives** on `Provide-Capability` tell bnd which implementation class provides each service, enabling `provides ... with ...` in the generated module-info.
> - **Attribute-based `Require-Capability`** format (e.g., `osgi.serviceloader;osgi.serviceloader=X`) instead of filter-based format -- required by bnd's JPMS plugin to generate correct `uses` directives.
> 
> ## Improvements over the previous approach
> 
> - **Single source of truth**: OSGi and JPMS metadata are derived from the same bnd instructions, eliminating drift between manifest and module-info.
> - **Correct `Import-Package`**: bnd calculates imports from bytecode, catching packages the hand-written manifests missed or got wrong.
> - **Optional imports**: `android.os`, `java.beans`, `jdk.dynalink*` (rhino), `javax.swing*`, `java.awt*` (rhino-tools) are now correctly marked `resolution:=optional` in OSGi and `requires static` in JPMS, improving support for headless/minimal JREs.
> - **`rhino-kotlin` OSGi support**: Previously had no OSGi metadata; now fully configured with bnd.
> - **Bug fix**: `rhino-xml` had `Bundle-SymbolicName: org.mozilla.rhino.engine` (copy-paste error) -- fixed to `org.mozilla.rhino.xml`.
> 
> ## Known differences from previous metadata
> 
> - **`requires java.base`** no longer shows `mandated` flag -- this is cosmetic; bnd cannot set this flag, but `java.base` is always implicitly required regardless.
> - **`contains` directives** appear in JPMS output for non-exported packages -- bnd lists all packages, while javac only listed compiled ones. These are informational only and have no runtime effect.
> - **`java.desktop` changed from `transitive` to `static`** in both `rhino` and `rhino-tools` -- this makes `java.desktop` optional at runtime, allowing Rhino to load on minimal JREs without the desktop module (as long as no AWT/Swing code paths are hit).
> 
> ## Verification
> 
> Helper scripts are included in `scripts/` for comparing metadata before and after:
> - `scripts/dump-jpms.sh` -- dumps JPMS module descriptors via `jar --describe-module`
> - `scripts/dump-osgi.sh` -- dumps OSGi headers via `bnd print`
</details>


I have intentionally added this change as a separate commit from my previous changes that only added OSGi metadata and did not touch `module-info.java`. That way you can compare the approaches and decide what you like more.

Generated script for extracting JPMS and OSGi metadata:

* [dump-jpms.sh](https://github.com/user-attachments/files/25803054/dump-jpms.sh)
* [dump-osgi.sh](https://github.com/user-attachments/files/25804033/dump-osgi.sh)

<details>
<summary>JPMS differences</summary>

```diff
diff --git a/jpms-before.txt b/jpms-after.txt
index 0622aa2..dd50cca 100644
--- a/jpms-before.txt
+++ b/jpms-after.txt
@@ -1,5 +1,5 @@
 === rhino ===
-org.mozilla.rhino
+org.mozilla.rhino@2.0.0.SNAPSHOT
 exports org.mozilla.classfile
 exports org.mozilla.javascript
 exports org.mozilla.javascript.annotations
@@ -13,52 +13,62 @@ exports org.mozilla.javascript.optimizer
 exports org.mozilla.javascript.serialize
 exports org.mozilla.javascript.typedarrays
 exports org.mozilla.javascript.xml
-requires java.base mandated
+requires java.base
 requires java.compiler
-requires java.desktop transitive
-requires jdk.dynalink
+requires java.desktop static
+requires jdk.dynalink static
 uses org.mozilla.javascript.NullabilityDetector
 uses org.mozilla.javascript.RegExpLoader
 uses org.mozilla.javascript.config.RhinoPropertiesLoader
 uses org.mozilla.javascript.xml.XMLLoader
 provides org.mozilla.javascript.RegExpLoader with org.mozilla.javascript.regexp.RegExpLoaderImpl
+contains org.mozilla.javascript.dtoa
+contains org.mozilla.javascript.json
+contains org.mozilla.javascript.lc
+contains org.mozilla.javascript.lc.member
+contains org.mozilla.javascript.lc.type.impl
+contains org.mozilla.javascript.lc.type.impl.factory
 contains org.mozilla.javascript.regexp
+contains org.mozilla.javascript.resources
+contains org.mozilla.javascript.v8dtoa
 
 === rhino-engine ===
-org.mozilla.rhino.engine@2.0.0-SNAPSHOT
+org.mozilla.rhino.engine@2.0.0.SNAPSHOT
 exports org.mozilla.javascript.engine
-requires java.base mandated
+requires java.base
 requires java.scripting transitive
 requires org.mozilla.rhino transitive
 provides javax.script.ScriptEngineFactory with org.mozilla.javascript.engine.RhinoScriptEngineFactory
 
 === rhino-tools ===
-org.mozilla.rhino.tools@2.0.0-SNAPSHOT
+org.mozilla.rhino.tools@2.0.0.SNAPSHOT
 exports org.mozilla.javascript.tools
 exports org.mozilla.javascript.tools.debugger
 exports org.mozilla.javascript.tools.jsc
 exports org.mozilla.javascript.tools.shell
-requires java.base mandated
-requires java.desktop transitive
+requires java.base
+requires java.desktop static
 requires org.jline.reader static
 requires org.jline.terminal static
 requires org.mozilla.rhino transitive
 uses org.mozilla.javascript.tools.ConsoleProvider
 provides org.mozilla.javascript.tools.ConsoleProvider with org.mozilla.javascript.tools.shell.BasicConsoleProvider org.mozilla.javascript.tools.shell.JLineConsoleProvider
+contains org.mozilla.javascript.tools.debugger.treetable
+contains org.mozilla.javascript.tools.resources
 main-class org.mozilla.javascript.tools.shell.Main
 
 === rhino-xml ===
-org.mozilla.javascript.xml@2.0.0-SNAPSHOT
+org.mozilla.javascript.xml@2.0.0.SNAPSHOT
 exports org.mozilla.javascript.xmlimpl
-requires java.base mandated
+requires java.base
 requires java.xml transitive
 requires org.mozilla.rhino transitive
 provides org.mozilla.javascript.xml.XMLLoader with org.mozilla.javascript.xmlimpl.XMLLoaderImpl
 
 === rhino-kotlin ===
-org.mozilla.rhino.kotlin@2.0.0-SNAPSHOT
+org.mozilla.rhino.kotlin@2.0.0.SNAPSHOT
 exports org.mozilla.kotlin
-requires java.base mandated
+requires java.base
 requires kotlin.metadata.jvm
 requires kotlin.stdlib
 requires org.mozilla.rhino

```
</details>

<details>
<summary>OSGi differences (from what my previous commit was generating)</summary>

```diff
diff --git a/osgi-before.txt b/osgi-after.txt
index 9f5e77b..4a72443 100644
--- a/osgi-before.txt
+++ b/osgi-after.txt
@@ -1,7 +1,36 @@
 === rhino ===
 Bundle-SymbolicName                     org.mozilla.rhino
-Bundle-Version                          2.0.0
+Bundle-Version                          2.0.0.SNAPSHOT
 DynamicImport-Package                   *
+Import-Package
+  android.os                             {resolution:=optional}
+  java.beans                             {resolution:=optional}
+  java.io                                
+  java.lang                              
+  java.lang.annotation                   
+  java.lang.invoke                       
+  java.lang.ref                          
+  java.lang.reflect                      
+  java.math                              
+  java.net                               
+  java.nio.charset                       
+  java.security                          
+  java.text                              
+  java.time                              
+  java.time.format                       
+  java.time.temporal                     
+  java.util                              
+  java.util.concurrent                   
+  java.util.concurrent.atomic            
+  java.util.concurrent.locks             
+  java.util.function                     
+  java.util.jar                          
+  java.util.regex                        
+  java.util.stream                       
+  jdk.dynalink                           {resolution:=optional}
+  jdk.dynalink.linker                    {resolution:=optional}
+  jdk.dynalink.linker.support            {resolution:=optional}
+  jdk.dynalink.support                   {resolution:=optional}
 Export-Package
   org.mozilla.classfile                  {version=2.0.0}
   org.mozilla.javascript                 {version=2.0.0}
@@ -17,70 +46,130 @@ Export-Package
   org.mozilla.javascript.typedarrays     {version=2.0.0}
   org.mozilla.javascript.xml             {version=2.0.0}
 Provide-Capability
-  osgi.serviceloader                     {osgi.serviceloader=org.mozilla.javascript.RegExpLoader}
+  osgi.serviceloader                     {osgi.serviceloader=org.mozilla.javascript.RegExpLoader, register:=org.mozilla.javascript.regexp.RegExpLoaderImpl}
 Require-Capability
   osgi.ee                                {filter:=(&(osgi.ee=JavaSE)(version>=11))}
   osgi.extender                          {filter:=(osgi.extender=osgi.serviceloader.processor)}
   osgi.extender                          {filter:=(osgi.extender=osgi.serviceloader.registrar)}
-  osgi.serviceloader                     {filter:=(osgi.serviceloader=org.mozilla.javascript.RegExpLoader)}
-  osgi.serviceloader                     {filter:=(osgi.serviceloader=org.mozilla.javascript.NullabilityDetector), resolution:=optional}
-  osgi.serviceloader                     {filter:=(osgi.serviceloader=org.mozilla.javascript.config.RhinoPropertiesLoader), resolution:=optional}
-  osgi.serviceloader                     {filter:=(osgi.serviceloader=org.mozilla.javascript.xml.XMLLoader), resolution:=optional}
+  osgi.serviceloader                     {osgi.serviceloader=org.mozilla.javascript.RegExpLoader}
+  osgi.serviceloader                     {osgi.serviceloader=org.mozilla.javascript.NullabilityDetector, resolution:=optional}
+  osgi.serviceloader                     {osgi.serviceloader=org.mozilla.javascript.config.RhinoPropertiesLoader, resolution:=optional}
+  osgi.serviceloader                     {osgi.serviceloader=org.mozilla.javascript.xml.XMLLoader, resolution:=optional}
 
 === rhino-engine ===
 Bundle-SymbolicName                     org.mozilla.rhino.engine
-Bundle-Version                          2.0.0
+Bundle-Version                          2.0.0.SNAPSHOT
 Import-Package
+  java.io                                
+  java.lang                              
+  java.lang.invoke                       
+  java.lang.reflect                      
+  java.nio.charset                       
+  java.util                              
   javax.script                           
-  org.mozilla.javascript                 {version=2.0.0}
+  org.mozilla.javascript                 {version=[2.0,3)}
 Export-Package
   org.mozilla.javascript.engine          {version=2.0.0}
 Provide-Capability
-  osgi.serviceloader                     {osgi.serviceloader=javax.script.ScriptEngineFactory}
+  osgi.serviceloader                     {osgi.serviceloader=javax.script.ScriptEngineFactory, register:=org.mozilla.javascript.engine.RhinoScriptEngineFactory}
 Require-Capability
   osgi.ee                                {filter:=(&(osgi.ee=JavaSE)(version>=11))}
   osgi.extender                          {filter:=(osgi.extender=osgi.serviceloader.registrar)}
 
 === rhino-tools ===
 Bundle-SymbolicName                     org.mozilla.rhino.tools
-Bundle-Version                          2.0.0
+Bundle-Version                          2.0.0.SNAPSHOT
 Import-Package
-  org.jline.reader                       {resolution:=optional}
-  org.jline.terminal                     {resolution:=optional}
-  org.mozilla.javascript                 {version=2.0.0}
-  org.mozilla.javascript.commonjs.module {version=2.0.0}
-  org.mozilla.javascript.commonjs.module.provider {version=2.0.0}
-  org.mozilla.javascript.config          {version=2.0.0}
-  org.mozilla.javascript.debug           {version=2.0.0}
-  org.mozilla.javascript.optimizer       {version=2.0.0}
-  org.mozilla.javascript.serialize       {version=2.0.0}
+  java.awt                               {resolution:=optional}
+  java.awt.event                         {resolution:=optional}
+  java.awt.geom                          {resolution:=optional}
+  java.io                                
+  java.lang                              
+  java.lang.invoke                       
+  java.lang.ref                          
+  java.lang.reflect                      
+  java.net                               
+  java.nio.charset                       
+  java.security                          
+  java.security.cert                     
+  java.text                              
+  java.util                              
+  java.util.regex                        
+  javax.swing                            {resolution:=optional}
+  javax.swing.border                     {resolution:=optional}
+  javax.swing.event                      {resolution:=optional}
+  javax.swing.filechooser                {resolution:=optional}
+  javax.swing.table                      {resolution:=optional}
+  javax.swing.text                       {resolution:=optional}
+  javax.swing.tree                       {resolution:=optional}
+  org.jline.reader                       {version=[3.30,4), resolution:=optional}
+  org.jline.terminal                     {version=[3.30,4), resolution:=optional}
+  org.mozilla.javascript                 {version=[2.0,3)}
+  org.mozilla.javascript.commonjs.module {version=[2.0,3)}
+  org.mozilla.javascript.commonjs.module.provider {version=[2.0,3)}
+  org.mozilla.javascript.config          {version=[2.0,3)}
+  org.mozilla.javascript.debug           {version=[2.0,3)}
+  org.mozilla.javascript.optimizer       {version=[2.0,3)}
+  org.mozilla.javascript.serialize       {version=[2.0,3)}
 Export-Package
   org.mozilla.javascript.tools           {version=2.0.0}
   org.mozilla.javascript.tools.debugger  {version=2.0.0}
   org.mozilla.javascript.tools.jsc       {version=2.0.0}
   org.mozilla.javascript.tools.shell     {version=2.0.0}
 Provide-Capability
-  osgi.serviceloader                     {osgi.serviceloader=org.mozilla.javascript.tools.ConsoleProvider}
+  osgi.serviceloader                     {osgi.serviceloader=org.mozilla.javascript.tools.ConsoleProvider, register:=org.mozilla.javascript.tools.shell.BasicConsoleProvider}
+  osgi.serviceloader                     {osgi.serviceloader=org.mozilla.javascript.tools.ConsoleProvider, register:=org.mozilla.javascript.tools.shell.JLineConsoleProvider}
 Require-Capability
   osgi.ee                                {filter:=(&(osgi.ee=JavaSE)(version>=11))}
   osgi.extender                          {filter:=(osgi.extender=osgi.serviceloader.processor)}
   osgi.extender                          {filter:=(osgi.extender=osgi.serviceloader.registrar)}
-  osgi.serviceloader                     {filter:=(osgi.serviceloader=org.mozilla.javascript.tools.ConsoleProvider), cardinality:=multiple}
+  osgi.serviceloader                     {osgi.serviceloader=org.mozilla.javascript.tools.ConsoleProvider, cardinality:=multiple}
 
 === rhino-xml ===
-Bundle-SymbolicName                     org.mozilla.rhino.engine
-Bundle-Version                          2.0.0
+Bundle-SymbolicName                     org.mozilla.rhino.xml
+Bundle-Version                          2.0.0.SNAPSHOT
 Import-Package
-  javax.script                           
-  org.mozilla.javascript                 {version=2.0.0}
-  org.mozilla.javascript.xml             {version=2.0.0}
+  java.io                                
+  java.lang                              
+  java.lang.invoke                       
+  java.util                              
+  java.util.concurrent                   
+  javax.xml.parsers                      
+  javax.xml.transform                    
+  javax.xml.transform.dom                
+  javax.xml.transform.stream             
+  org.mozilla.javascript                 {version=[2.0,3)}
+  org.mozilla.javascript.xml             {version=[2.0,3)}
+  org.w3c.dom                            
+  org.xml.sax                            
 Export-Package
   org.mozilla.javascript.xmlimpl         {version=2.0.0}
 Provide-Capability
-  osgi.serviceloader                     {osgi.serviceloader=org.mozilla.javascript.xml.XMLLoader}
+  osgi.serviceloader                     {osgi.serviceloader=org.mozilla.javascript.xml.XMLLoader, register:=org.mozilla.javascript.xmlimpl.XMLLoaderImpl}
 Require-Capability
   osgi.ee                                {filter:=(&(osgi.ee=JavaSE)(version>=11))}
   osgi.extender                          {filter:=(osgi.extender=osgi.serviceloader.registrar)}
 
 === rhino-kotlin ===
+Bundle-SymbolicName                     org.mozilla.rhino.kotlin
+Bundle-Version                          2.0.0.SNAPSHOT
+Import-Package
+  java.lang                              
+  java.lang.annotation                   
+  java.lang.invoke                       
+  java.lang.reflect                      
+  java.util                              
+  java.util.function                     
+  java.util.stream                       
+  kotlin                                 
+  kotlin.metadata                        
+  kotlin.metadata.jvm                    
+  org.mozilla.javascript                 {version=[2.0,3)}
+Export-Package
+  org.mozilla.kotlin                     {version=2.0.0}
+Provide-Capability
+  osgi.serviceloader                     {osgi.serviceloader=org.mozilla.javascript.NullabilityDetector, register:=org.mozilla.kotlin.KotlinNullabilityDetector}
+Require-Capability
+  osgi.ee                                {filter:=(&(osgi.ee=JavaSE)(version>=11))}
+  osgi.extender                          {filter:=(osgi.extender=osgi.serviceloader.registrar)}
 ```
</details>

**Comment by gbrail** [comment]
Thanks! Two things:

First of all, my AI found two regressions in the module file when I asked it to compare before and after and I think it's right:

```
diff --git a/rhino-tools/build.gradle b/rhino-tools/build.gradle
index 4126ce71d..385029e6e 100644
--- a/rhino-tools/build.gradle
+++ b/rhino-tools/build.gradle
@@ -57,7 +57,7 @@ jar {
             '-jpms-module-info': 'org.mozilla.rhino.tools;access=0',
             '-jpms-module-info-options': [
                 'org.mozilla.rhino;transitive=true;static=false',
-                'java.desktop;transitive=false;static=true',
+                'java.desktop;transitive=true;static=false',
                 'org.jline.terminal;static=true',
                 'org.jline.reader;static=true',
                 'java.base;static=false',
diff --git a/rhino/build.gradle b/rhino/build.gradle
index 2efdfd1d8..6b0189437 100644
--- a/rhino/build.gradle
+++ b/rhino/build.gradle
@@ -47,8 +47,8 @@ jar {
             ].join(','),
             '-jpms-module-info': 'org.mozilla.rhino;access=0;modules="java.compiler,jdk.dynalink"',
             '-jpms-module-info-options': [
-                'java.desktop;transitive=false;static=true',
-                'jdk.dynalink;static=true;transitive=false',
+                'java.desktop;transitive=true;static=false',
+                'jdk.dynalink;static=false;transitive=false',
                 'java.compiler;static=false',
                 'java.base;static=false',
             ].join(','),
```

Second, how do we verify that the OSGi manifests are right?

The way we have it now, if someone makes a mistake in a module specification then tests will fail (because we run many of the tests in a module themselves). If someone makes a mistake in the OSGi bit in the future, how do we catch it? I think you had a previous version of this PR that had some tests.

**Comment by pavelhoral** [comment]
> Thanks! Two things:
> 
> First of all, my AI found two regressions in the module file when I asked it to compare before and after and I think it's right:

This is intentional as described in the summary I posted. I wanted the `java.desktop` and `java.dynalink` to be `optional` in OSGi (to be precise I don't think that packages from those modules should be `required` -> i.e. OSGi bundle won't start unless they are available). Claude warns me that the transitivity/non-static requirement would make the JPMS definition inconsistent with the OSGi metadata... so I would rather ask the opposite question: **Is it really really necessary to have those module defined as `static=false`/`transitive=true` in JPMS?**

> Second, how do we verify that the OSGi manifests are right?

Properly testing would require testing in OSGi environment. I can try to create some Apache Felix test if that is required, but it probably wouldn't be that lightweight.

> The way we have it now, if someone makes a mistake in a module specification then tests will fail (because we run many of the tests in a module themselves).

**Here be careful, because my change removed the modules from tests AFAIK!** There was a intermediate step, where I had both `module-info.java` and `bnd` generating the final `module-info` which felt very strange... I have asked why is that needed and the reason was "for tests as bnd is active only during packaging phase, but those tests can run in classpath mode perfectly fine". So if it was really a hard requirement to have modules in unit tests, I need to change this (I am just not sure if it would be possible to have bnd act in a way that we have no duplicate module definition).

Question here is whether you really want to run all unit tests under JPMS, or if you just want some lightweight integration smoke test that will verify modules work together.

**Comment by pavelhoral** [comment]
Playing with it a bit more -> when the tests are run with JPMS, they fail as the `jdk.dynalink` module was not available. But that is expected if the test explicitly tests the optimization while not adding `jdk.dynalink` to module path. If only a specific feature needs that module it can remain optional (static). I would say this is a bug in the original module definition.

**Comment by pavelhoral** [comment]
I have pushed one additional commit that adds integration tests for JPMS and OSGi. These tests are intentionally separate from the pre-existing unit tests as they depend on the built JAR files.

This setup makes much more sense to me (intentional testing of JPMS and OSGi) than what was done previously -> running all unit tests as JPMS modules.

<details>
<summary>Summary generated by Claude</summary>

# JPMS and OSGi integration tests

Dedicated integration tests verify that Rhino's BND-generated module metadata works correctly in real JPMS and OSGi environments. These tests run against the built JARs (not classes directories) and are separated from regular unit tests so that `./gradlew test` does not require JARs to be built first.

## Running

```
./gradlew integrationTest
```

This is also included in `./gradlew check`.

## Build setup

A dedicated `integrationTest` source set in `tests/build.gradle` keeps integration tests separate from unit tests. The `integrationTest` task depends on all module JARs being built and passes their paths as system properties (e.g. `rhino.jar`, `rhino-engine.jar`).

Apache Felix (`org.apache.felix.framework:7.0.5`) is added as a dependency for OSGi testing.

## JpmsModuleTest

Tests JPMS module behavior in-process using the `ModuleLayer` API. Each test creates a module layer from the built JARs via `ModuleFinder.of()` and `Configuration.resolve()`, then loads Rhino classes from the layer via reflection. The layer's classloader is parented by the platform classloader (not the system classloader) to isolate it from test classpath classes.

Tests:
- **rhinoExecutesJavaScriptInModuleLayer** -- Rhino evaluates JavaScript within a JPMS module layer
- **scriptEngineDiscoveredViaServiceLoader** -- `ScriptEngineFactory` is discovered via JPMS `provides`/`uses` service binding
- **xmlModuleWorksInModuleLayer** -- `typeof XML` is `"function"` when `rhino-xml` is in the module layer
- **xmlModuleNotAvailableWithoutXmlInLayer** -- `typeof XML` is `"undefined"` without `rhino-xml`

## OsgiModuleTest

Tests OSGi bundle behavior by embedding an Apache Felix framework, installing the built JARs as bundles, and verifying resolution, activation, and cross-bundle wiring. The `jdk.dynalink` package is boot-delegated so bundles can access it without an explicit import. Framework-level `osgi.serviceloader` capabilities are provided so bundles resolve without requiring SPI Fly.

Tests:
- **rhinoBundleResolvesAndStarts** -- bundle resolution and activation
- **rhinoExecutesJavaScriptInOsgi** -- JavaScript execution within the bundle
- **rhinoEngineBundleResolvesAndStarts** -- rhino-engine resolves with rhino
- **rhinoToolsBundleResolvesAndStarts** -- rhino-tools resolves with rhino
- **rhinoXmlBundleResolvesAndStarts** -- rhino-xml resolves with rhino
- **allBundlesResolveAndStartTogether** -- all 4 bundles active simultaneously
- **xmlBundleExportsAreAccessible** -- cross-bundle wiring works (`XMLLoaderImpl` implements `XMLLoader` across bundle boundaries)

</details>

File: rhino/build.gradle [hunk 1]
```diff
@@ -1,11 +1,62 @@
...
+    bundle {
+        bnd(
+            'Bundle-SymbolicName': 'org.mozilla.rhino',
+            'Bundle-Vendor': 'Mozilla Foundation',
+            'Bundle-License': 'https://www.mozilla.org/MPL/2.0/',
+            'DynamicImport-Package': '*',
+            'Import-Package': [
+                'android.os;resolution:=optional',
+                'java.beans;resolution:=optional',
+                'jdk.dynalink*;resolution:=optional',
```
pavelhoral, line 19: I might have misjudged `jdk.dynalink` dependency. Would Rhino work without it?

Not sure about `java.beans` as well.

pavelhoral: Checked the source code and dynalink (as part of the optimizer) can be stripped away as it is being loaded dynamically - https://github.com/mozilla/rhino/blob/4bf1df91036c7627f817d037aa768c3f79a9f8d8/rhino/src/main/java/org/mozilla/javascript/Context.java#L2669 . So the optional resolution is correct.

Java Beans is not optional as it is being imported and used in Context. I have removed the optionality for that package.

pavelhoral:
Any chance of moving forward with this?

gbrail:
Thank you for doing what I asked! And now that you did it I can see that using BND, and TBH I found that more confusing, since it means that we had to understand JPMS and OSGi and *also* BND, which I felt like looks more complicated than either.

So, I created a new PR that just cherry-picks your first and third commits. That means managing module configuration in two places, but both are standard and you now have tests to verify correctness.

If that makes sense to you then I will close this PR and move forward with that one!

https://github.com/mozilla/rhino/pull/2357

pavelhoral:
If you want to cherry-pick only the first and last commit, I might need to update them as I think I fixed some issues of the first commit in the second one that you want to drop. I will refactor this PR, so you can safely cherry-pick those commits.

Btw. I still think that having the BND is a good thing to have and forcing yourself to use JPMS even in unit tests is not optimal, however I am not maintainer of this project so your preferences have definitely more wight :).

pavelhoral:
@gbrail  I have updated this PR with the following:

* Rebased on the current master branch.
* Single commit adding BND plugin for building OSGi metadata.
* Single commit adding integration tests in JPMS and OSGi environment.

Using BND is much safer and better way than handling the metadata manually. I have removed stuff that touched JPMS module-info, so those will remain maintained separately. IMO much more important is that the integration tests are passing and not that metadata are 100 % compatible.

Side note: BND correctly detected unrelated issue with the code where `NativeJavaConstructor` has a public constructor that has non-exported (both JPMS and OSGi) type `ExecutableBox`. This should be fixed probably by making the constructor package private (but that is out of scope for this PR).

gbrail:
Great! I think this is the right approach, because we keep the original Java module files around for the people who understand them (like me).

I got this warning on build -- the first seems to be about the OSGi manifest itself -- do we need to fix it?

`/home/gjb/src/rhino/rhino/build/tmp/jar/bnd16346029016373776129.bnd:0: warning: Export org.mozilla.javascript,  has 1,  private references [org.mozilla.javascript.lc.member]
WARNING: A terminally deprecated method in sun.misc.Unsafe has been called
WARNING: sun.misc.Unsafe::staticFieldOffset has been called by org.apache.felix.framework.util.SecureAction (file:/home/gjb/.gradle/caches/modules-2/files-2.1/org.apache.felix/org.apache.felix.framework/7.0.5/dd4cb10c68d9ceb7897e561be7c4c16d25347193/org.apache.felix.framework-7.0.5.jar)
WARNING: Please consider reporting this to the maintainers of class org.apache.felix.framework.util.SecureAction
WARNING: sun.misc.Unsafe::staticFieldOffset will be removed in a future release`

pavelhoral:
> ```/home/gjb/src/rhino/rhino/build/tmp/jar/bnd16346029016373776129.bnd:0:
> warning: Export org.mozilla.javascript, has 1, private references [org.mozilla.javascript.lc.member]
> ```

This is what I was referring to in my last comment:

_Side note: BND correctly detected unrelated issue with the code where NativeJavaConstructor has a public constructor that has non-exported (both JPMS and OSGi) type ExecutableBox. This should be fixed probably by making the constructor package private (but that is out of scope for this PR)._

> ```
> WARNING: A terminally deprecated method in sun.misc.Unsafe has been called
> WARNING: sun.misc.Unsafe::staticFieldOffset has been called by org.apache.felix.framework.util.SecureAction (file:/home/gjb/.gradle/caches/modules-2/files-2.1/org.apache.felix/org.apache.felix.framework/7.0.5/dd4cb10c68d9ceb7897e561be7c4c16d25347193/org.apache.felix.framework-7.0.5.jar)
> WARNING: Please consider reporting this to the maintainers of class org.apache.felix.framework.util.SecureAction
> WARNING: sun.misc.Unsafe::staticFieldOffset will be removed in a future release`
> ```

This one is issue in Apache Felix - https://issues.apache.org/jira/browse/FELIX-6759. Unless they fix it, that warning will be there I guess.

gbrail:
Thanks! I think that this is ready to go, but I have to ask for one more thing, which is to rebase this against the current master. We are in the middle of a giant refactor of some key interfaces and unfortunately that causes this PR to fail. Sorry but thanks and we'll finish this soon!

pavelhoral:
> Thanks! I think that this is ready to go, but I have to ask for one more thing, which is to rebase this against the current master.

Done.

gbrail:
Thanks for all this!

pavelhoral:
@gbrail  Are you open for 1.9.x backport PR as well?

gbrail:
Yes although no immediate plans for a new release -- releases aren't a
terrible amount of work but they're not none, and I don't have a queue of
recently-fixed critical bugs yet. That will change at some point, though!

I created a branch called "rhino_1_9_backports" that you can base it
against so we're ready when the time comes.

PR #2312, opened 2026-02-27
Fix a problem in dumpICode and increase test coverage

andreabergia:
As a follow-up of https://github.com/mozilla/rhino/pull/2129, I started by increasing the code coverage. While doing that, I found a small issue with `Icode_LITERAL_NEW_ARRAY` and I fixed it.


gbrail:
Looks good, thanks!

Issue #2316, opened 2026-02-28
IllegalStateOperation when interpreting complicated things


gbrail:
I've been working on a set of performance tests based on the JetStream2 benchmarks that are more comprehensive and repeatable than what we have now. A few of the tests fail with the IllegalStateException below.

This was introduced by the big refactoring to introduce JSFunction and friends:

https://github.com/mozilla/rhino/commit/8e69a72421c084d757d71014d3fa70b47d59357b

Here's the stack trace:

java.lang.IllegalStateException: FAILED ASSERTION
        at org.mozilla.javascript.Kit.codeBug(Kit.java:352)
        at org.mozilla.javascript.Interpreter.interpret(Interpreter.java:1437)
        at org.mozilla.javascript.InterpreterData.execute(InterpreterData.java:104)
        at org.mozilla.javascript.JSFunction.construct(JSFunction.java:155)
        at org.brail.rhinobenchmarks.BenchmarkRunner.makeRunner(BenchmarkRunner.java:83)
        at org.brail.rhinobenchmarks.BenchmarkRunner.load(BenchmarkRunner.java:51)
        at org.brail.rhinobenchmarks.BenchmarkDriver.loadFile(BenchmarkDriver.java:25)
        at org.brail.rhinobenchmarks.StandardBenchmarks.load(StandardBenchmarks.java:32)
        at org.brail.rhinobenchmarks.Main.call(Main.java:61)
        at org.brail.rhinobenchmarks.Main.call(Main.java:11)
        at picocli.CommandLine.executeUserObject(CommandLine.java:2031)
        at picocli.CommandLine.access$1500(CommandLine.java:148)
        at picocli.CommandLine$RunLast.executeUserObjectOfLastSubcommandWithSameParent(CommandLine.java:2469)
        at picocli.CommandLine$RunLast.handle(CommandLine.java:2461)
        at picocli.CommandLine$RunLast.handle(CommandLine.java:2423)
        at picocli.CommandLine$AbstractParseResultHandler.execute(CommandLine.java:2277)
        at picocli.CommandLine$RunLast.execute(CommandLine.java:2425)
        at picocli.CommandLine.execute(CommandLine.java:2174)
        at org.brail.rhinobenchmarks.Main.main(Main.java:101)

I'm not sure where to start with this but @aardvark179 if you have any ideas we could fix a regression that could potentially affect other things.

FWIW, the "pdfjs" and "regexp" benchmarks from the Octane test suite, and I can supply code to running them if useful.

PR #2317, opened 2026-02-28
Ensure TopLevel is present when functions are constructed

gbrail:
This fix ensures that TopLevel is present in the interpreter in more cases. Without it, my transpiled version of the Octane "pdfjs" test fails with the Kit.codeBug() shows in the issue below:

https://github.com/mozilla/rhino/issues/2316

With this fix, the benchmark runs, and all the other tests pass too.

This bug was discovered and fixed by Gemini 3. And it seems to fix my problem. 

Please take a look!


gbrail:
Thanks for looking at this!

PR #2319, opened 2026-03-03
Refactor iCode debug printing

andreabergia:
This follows up on https://github.com/mozilla/rhino/pull/2129 and https://github.com/mozilla/rhino/pull/2312 and implements the refactoring [suggested by Duncan](https://github.com/mozilla/rhino/pull/2129#issuecomment-3415948888).

In addition to adding some test coverage, I have refactored the code in `Interpreter` that debug prints the bytecode. Now rather than having a giant switch statement, all the logic is in a virtual method of `InstructionClass`, which is therefore capable of being executed or printed.

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/Interpreter.java
+++ b/rhino/src/main/java/org/mozilla/javascript/Interpreter.java
@@ -4378,13 +4378,13 @@ public final class Interpreter extends Icode implements Evaluator {
         @Override
         void dumpICode(int op, String tname, ICodeDumpContext ctx) {
             boolean copyArray = ctx.idata.itsICode[ctx.pc] != 0;
-            ++ctx.pc;
             if (ctx.indexReg < 0) {
                 ctx.out.println(tname + " length: " + (-ctx.indexReg - 1));
             } else {
                 Object[] keys = (Object[]) ctx.idata.literalIds[ctx.indexReg];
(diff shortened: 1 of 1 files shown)


rhino/src/main/java/org/mozilla/javascript/Interpreter.java
```diff
@@ -4448,6 +4405,13 @@ NewState execute(Context cx, CallFrame frame, InterpreterState state, int op) {
             frame.stack[++state.stackTop] = storage;
             return null;
         }
+
+        @Override
+        void dumpICode(int op, String tname, ICodeDumpContext ctx) {
+            int skipIdx = 0xFF & ctx.idata.itsICode[ctx.pc];
+            ++ctx.pc;
```
rbri: usually updating the pc is the last step in all the methods, but this seems to be an exception - maybe we can make this similar to the rest; and the next one alsoe

andreabergia: Ok, you're right, it makes stuff more coherent. Updated.

aardvark179:
This looks good, thanks @andreabergia.

0xe (approved):
Thanks for the missing test for object rest in dumpIcode. (I should've added it in #2207.)

gbrail:
This is great, thanks!

PR #2321, opened 2026-03-06
Fix uint8 overflow in array literal skipIndexesId and sourcePositions

Finomosec:
## Summary
- `CodeGenerator.visitArrayLiteral` encodes `skipIndexesId` and `sourcePositions` as uint8 via `addUint8`, which overflows when a script contains more than 255 object/array literals (causing `literalIds` to exceed 255 entries)
- Changed `addUint8` to `addUint16` for these values in `CodeGenerator` and updated corresponding reads in `Interpreter` (`DoLiteralNewArray`, `DoSpread`) to use `getIndex` (uint16)
- Added `bytecodeSpan` entry for `Icode_LITERAL_NEW_ARRAY` to account for the new 2-byte encoding
- Only affects interpreter mode (`optimizationLevel = -1`)

## Reproduction
A script with 260+ object literals followed by a sparse array literal (e.g. `[1, , 3]`) triggers `IllegalStateException` (from `Kit.codeBug()` in `addUint8`) during compilation in interpreter mode.

## Test plan
- Added `ArrayLiteralOverflowTest` with 4 tests covering both interpreted and compiled modes, with and without spread operators
- Full existing test suite passes without regressions

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,6 +1,6 @@
 rootProject.name=rhino-root
 group=org.mozilla
-version=1.9.0
+version=1.9.1-SNAPSHOT
 mavenReleaseRepo=https://oss.sonatype.org/service/local/staging/deploy/maven2/
 mavenSnapshotRepo=https://oss.sonatype.org/content/repositories/snapshots
 githubPackagesRepo=https://maven.pkg.github.com/mozilla/rhino
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,6 +1,6 @@
 rootProject.name=rhino-root
 group=org.mozilla
-version=1.9.1-SNAPSHOT
+version=1.9.1
 mavenReleaseRepo=https://oss.sonatype.org/service/local/staging/deploy/maven2/
 mavenSnapshotRepo=https://oss.sonatype.org/content/repositories/snapshots
 githubPackagesRepo=https://maven.pkg.github.com/mozilla/rhino
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,6 +1,6 @@
 rootProject.name=rhino-root
 group=org.mozilla
-version=1.9.1
+version=1.9.1-PATCHED
 mavenReleaseRepo=https://oss.sonatype.org/service/local/staging/deploy/maven2/
 mavenSnapshotRepo=https://oss.sonatype.org/content/repositories/snapshots
 githubPackagesRepo=https://maven.pkg.github.com/mozilla/rhino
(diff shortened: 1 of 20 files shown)


gbrail:
I think this is potentially very helpful as I have been trying to run more standard benchmarks and we've run into some of these types of overflows. However I think that your actual change is buried behind a lot of other commits in this PR -- can you try to rebase your change against the current master branch? You should end up with just one commit that's different when you compare in the GitHub UI. Thanks!

gradle.properties
```diff
@@ -1,6 +1,6 @@
 rootProject.name=rhino-root
 group=org.mozilla
-version=1.9.1
+version=1.9.1-PATCHED
```
andreabergia, line 3: This needs to be reverted

rhino/src/test/java/org/mozilla/javascript/tests/ArrayLiteralOverflowTest.java
```diff
@@ -0,0 +1,90 @@
...
+import org.mozilla.javascript.Context;
+import org.mozilla.javascript.ContextFactory;
+import org.mozilla.javascript.Scriptable;
+
+/**
+ * Regression test for addUint8 overflow in CodeGenerator.visitArrayLiteral.
+ * When a script contains more than 255 object/array literals, the literalIds
+ * table grows beyond 255, and skipIndexesId (encoded as uint8) overflows.
+ */
+public class ArrayLiteralOverflowTest {
```
andreabergia, line 15: This test should be modified to use `Utils.runWithAllModes`.

andreabergia:
This looks correct to me. I've put a couple of small comments, but they should be pretty easy to change.

gbrail:
I'm happy to take this but someone (or something) needs to fix the PR so it doesn't have a copy of a whole bunch of other PRs in it. TBH I'm not sure how it got this way but I might suggest just creating a new PR and cherry-pick the one commit that actually changes something.

rbri:
as @Finomosec seems to be from germany, i tried to reach out to him via linkedin - maybe we can align. If not i will extract a simple PR from this one... 🤞

rbri:
reaching out to @Finomosec via linkedin was not successful. Will extract a separate pr from this and we can ref Finomosec in the commit

rbri:
have extracted #2398

Finomosec:
Sorry for the delayed response. The LinkedIn message slipped past me.

Thanks @rbri for extracting a clean PR (#2398) and to @gbrail / @andreabergia for the review feedback. Closing this in favor of #2398.

rbri:
@Finomosec thanks for bringing this up, #2398 #2400 are merged now

Issue #2300, opened 2026-02-12
Possible memory leak in Rhino 1.8.0+: const + cached compileString() script retains Java objects


jgoderbauer:
### Summary
After upgrading to **Rhino 1.8.0**, we observed that under certain conditions **Java objects exposed to Rhino are retained indefinitely** and therefore **not garbage collected**. This looks related to the introduction of `ConstAwareLinker` and only reproduces when:

1. the JavaScript uses a `const` binding (instead of `var`), **and**
2. the script is **compiled once and cached** (using `Context.compileString(...)` and storing the returned `Script`).

I’m unsure whether this is:
- a Rhino bug (unexpected strong reference chain), or
- expected behavior and our embedding/caching strategy is incorrect.

### Expected behavior
When a script execution completes and the **evaluation scope** becomes unreachable, objects introduced for that execution (including Java objects wrapped via `NativeJavaObject`) should be eligible for garbage collection—**even if the compiled `Script` is cached**.

### The problem in detail
When creating Objects on the Java side and introduce them, like this:
```java
MyJavaClass myObj = new MyJavaClass();
NativeJavaObject nativeJavaObject = new NativeJavaObject(evaluationScope, myObj, myObj.getClass());
evaluationScope.put("myJavaObject", evaluationScope, nativeJavaObject);
```

that java object can be used within rhino. E.g:
```javascript
var originator = "Rhino User";
myJavaObject.sayHello(originator);
```

The instance of this `MyJavaClass` is however held endlessly under the already mentioned conditions.
It's relevant that the `const` within the JavaScript code is actually used (GET).

### Reason

The `target` within the `ConstAwareLinker` uses an object. Unlike other linkers, that maintains a hard reference to that `target` object.
A form of Cache holds onto pre compiled `Script`s (and maintains a hard reference there).

That leads to the issue that the instance of `MyJavaClass` is still reachable and not garbage collected.

#### Code to reproduce and understand the issue
This is an example code that simplifies and reduces the issue to a minimum:

```java
public class JsEvaluator {
    // a global cache for compiled scripts
    private static final Map<String, Script> scriptCache = new HashMap<>();

    // Initializes global scope with standard objects
    private static final TopLevel NATIVE_TOP_LEVEL_SCOPE = new TopLevel() {{
        ContextFactory.getGlobal().enterContext();
        try
        {
            ScriptRuntime.initStandardObjects(Context.getCurrentContext(), this, true);
            sealObject();
        }
        finally
        {
[...]

PR #2326, opened 2026-03-08
Use WeakReference to prevent a memory leak for const optimizations

gbrail:
Prevent a memory leak because of our const optimization in compiled mode, which means that the constant value may otherwise be retained even if the scope in which it was created is GCed. This the optimization will be undone when the target object or the actual value is GCed.

Should fix https://github.com/mozilla/rhino/issues/2300

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/optimizer/ConstAwareLinker.java
+++ b/rhino/src/main/java/org/mozilla/javascript/optimizer/ConstAwareLinker.java
@@ -3,6 +3,7 @@ package org.mozilla.javascript.optimizer;
 import java.lang.invoke.MethodHandle;
 import java.lang.invoke.MethodHandles;
 import java.lang.invoke.MethodType;
+import java.lang.ref.WeakReference;
 import jdk.dynalink.StandardNamespace;
 import jdk.dynalink.StandardOperation;
 import jdk.dynalink.linker.GuardedInvocation;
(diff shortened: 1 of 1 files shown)


gbrail:
Ugh. Yes we will just not cache null values... Thanks!

gbrail:
If you look up around line 45 or 46, "getConstValue" is the method that looks at the property to see if it's truly a constant, and we don't go through with the optimization if the resulting value is null. So, I think that we already check for this and won't try to cache a constant if the value is "null" regardless.

aardvark179:
Ah okay, that makes some sort of sense. Do we know where and why we get an advantage from the JVM knowing a JS value is constant?

gbrail:
I can make a few synthetic benchmarks, let me see which one.... but given
that this only triggers when "const" is actually used and we don't support
"const" in all the places then I can se that it has limited utility.

gbrail:
FWIW I agree that this optimization doesn't get triggered much. If I construct a contrived test case I can see it helping about 25% for me:

```'use strict';

var a = 0;
const SOMETHING = 100000;
const ITERATIONS = 10000000;

const start = new Date();
for (a = 0; a < ITERATIONS; a++) {
  if (a === SOMETHING) {
    console.log('A is %d!', a);
  }
  if (a === ITERATIONS) {
    console.log('Got past the end!');
  }
}
const end = new Date();

console.log('All Done: a = %d', a);
console.log('Elapsed: %f seconds', (end - start) / 1000.0);
```

I do believe that this is a pretty simple, targeted optimization, though, and I'd like to keep it with this fix.

aardvark179:
Okay, that seems reasonable. Let's keep the optimisation working for now, and have a chat about whether we could add some more constant analysis in the compilers that might give us the same effect.

Issue #2320, opened 2026-03-04
Two `let` statements in single line accepted without semicolon


thomas-metz-ix:
The following executes in 1.9.1 although it is syntactically not correct as it would require a semicolon between the let statements:

```
java -cp rhino-1.9.1.jar:rhino-tools-1.9.1.jar org.mozilla.javascript.tools.shell.Main
Rhino 1.9.1
js> let a = 4 let b = 5
js> a + b
9
js> 
```

PR #2327, opened 2026-03-09
Fixed automatic semicolon insertion for `let`

andreabergia:
There was a bug that treated `let` differently from `var` and `const` with respect to the automated semicolon insertion, i.e.

```js
let a = 4 let b = 5
```

was incorrectly allowed.

Fixes https://github.com/mozilla/rhino/issues/2320

### DIFF
--- a/tests/src/test/java/org/mozilla/javascript/tests/es5/LetTest.java
+++ b/tests/src/test/java/org/mozilla/javascript/tests/es5/LetTest.java
@@ -112,16 +112,6 @@ class LetTest {
     }
 
     @Test
-    void noSemicolonSameLine() {
-        Utils.assertEvaluatorException_1_8("missing ; before statement", "let a = 4 let b = 5");
-    }
-
-    @Test
-    void autoSemicolonInsertionNewline() {
-        Utils.assertWithAllModes_1_8(9, "let a = 4\nlet b = 5\na + b");
-    }
-
-    @Test
     void letInsideFunctionWhileIf() {
         String script =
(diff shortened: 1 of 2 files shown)


0xe (approved):
Subtle catch!

tests/src/test/java/org/mozilla/javascript/tests/es5/LetTest.java
```diff
@@ -111,6 +111,16 @@ void letInsideFunctionSwitchIf() {
         Utils.assertWithAllModes_1_8(1, script);
     }
 
+    @Test
```
rbri: @andreabergia to you know if we have tests for this constructs for var/const already? Just to be sure we introduce no regression....

andreabergia: Good point. I assumed we do, because we already had an "if", but I'll check better.

rbri: Had the same assumtion but I'll asked better ;-)

andreabergia: I had misread the code and it only affected `let`. Apparently there were no tests for `const` and `var`, but also no bug.
I've ended up moving the new test cases into the new `VariableDeclAsiTest`, and added some tests to `var` and `const` for completeness.

gbrail:
This is great -- thanks!

Issue #2163, opened 2025-11-21
Separation of scopes and objects


aardvark179:
I'm creating this issue to track the things I know of which need to be tackled to separate objects and scopes in our implementation.

# Separating scopes and objects

Future performance optimisations become significantly easier when we can separate the concepts of scopes and objects. Specifically it makes traversal through parents one dimensional rather than two dimensional as scopes do not have prototypes and objects to not have parents (at least not in way that should ever be important when resolving a property).

To separate the concepts of scopes and objects in Rhino without disturbing too many APIs it would be useful to preserve a few things.
1. We would like to keep the existing APIs on `Scriptable` which define property access in terms like `Object get(String name, Scriptable start)`. Crucially the `start` argument when looking up a property on an object must _always_ be an object.
2. We would like our new scopes to offer a similar API like this: `Object get(String name, Scope start)`. Again, the `start` argument when looking up something on a scope should always be a scope.
3. We would like to be able to use the same underlying property / slot map implementation. That means that either this implementation must duplicate methods to provide tow versions of each method, and this must also happen on various slot definitions, or we need to do something with parameterised types in order to preserve our guarantees.
4. Pretty much any solution that satisfies points 1. to 3. prevents us from having objects that are both scopes and objects.

My proposed way to handle this is with the following type hierarchy:
* `SuperType<T extends SuperType<T>>` (`SuperType` is chosen as a deliberately bad name here, we can discuss what it should actually be called at a later date).
  * `Scriptable extends SuperType<Scriptable>`
  * `Scope extends SuperType<Scope>`
  
This allows a large number of uses of property access and definition to remain unchanged, especially in user code, and requires simple modification of our slot map code. It also prevents us form accidentally (or purposefully) passing a scope as an object or an object as a scope, but all of this is at the price of not having objects which are both scopes _and_ objects.

Since Scopes only ever have name bindings we could theoretically make the super type handle those, but we might reasonably use symbol properties for internal data, and the interface is simpler if the two halves of the hierarchy are the same.

There are other small differences between the two halves of the hierarchy. Scopes have parents, while objects have prototypes, but a parent is of a scope is always a scope, and a prototype of an object is always an object, so they are really a common operation that can be written as `T getAncestor()`.

So, what problems do we need to resolve to make this separation achievable?

## `TopLevel` is both a scope and an object

[...]

PR #2328, opened 2026-03-10
Move resolution for undefined this to functions to better handle realms and strictness.

aardvark179:
This PR stacks on top of #2161, and is the next step in resolving #2163.

Previously we handled resolving `globalThis` at the point at which we looked up a function. While working to separate scopes this kept on throwing up issues because the behaviour depends on the strictness of the function, whether the function is built in, and on the realm in which the function is declared (the `globalThis` is different for each realm). Moving this handling to the callee side helped to make a whole bunch of 262 tests pass, and made it much easier to keep them passing while refactoring scopes.

### DIFF
--- a/tests/testsrc/test262.properties
+++ b/tests/testsrc/test262.properties
@@ -448,7 +448,7 @@ built-ins/AggregateError 2/25 (8.0%)
     newtarget-proto-fallback.js
     proto-from-ctor-realm.js
 
-built-ins/Array 253/3077 (8.22%)
+built-ins/Array 240/3077 (7.8%)
     fromAsync 95/95 (100.0%)
     length/define-own-prop-length-coercion-order-set.js
     prototype/at/coerced-index-resize.js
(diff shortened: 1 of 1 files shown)


rhino/src/main/java/org/mozilla/javascript/TopLevel.java
```diff
@@ -136,6 +136,10 @@ public static TopLevel createIsolate(TopLevel parent) {
         var newGlobal = new NativeObject();
         newGlobal.setPrototype(parent.getGlobalThis());
         newGlobal.setParentScope(null);
+        newGlobal.put("globalThis", newGlobal, newGlobal);
+        newGlobal.setAttributes("globalThis", ScriptableObject.DONTENUM);
+        newGlobal.put("global", newGlobal, newGlobal);
+        newGlobal.setAttributes("globalThis", ScriptableObject.DONTENUM);
```
gbrail: Do you mean to set DONTENUM on "global" and not twice on "globalThis?"

aardvark179:
I'd been holding off on fixing that ambiguous method signature purely because changing it is a compatibility change. Maybe renaming the static would be the safest path here as that will fail at compile time and cannot lead to anything unexpected.

gbrail:
That's OK -- we can fix that warning later, but do take a look at the double-setting of DONTENUM above. Thanks!

aardvark179:
Will do.

rhino/src/main/java/org/mozilla/javascript/TopLevel.java
aardvark179: Fixed.

PR #2329, opened 2026-03-10
Introduce `VarScope` as a subclass of `Scriptable`

aardvark179:
This PR stacks on top of #2328 and is the next step after it in resolving #2163.

This change adds `VarScope` as a subclass of `Scriptable`, hoists a number of methods up from `ScriptableObject` to `SlotMapOwner` and introduces a new `ScopeObject` class. This is the first stage of separating objects and scopes, and lays the initial ground work, follow up PRs will change the APIs within Rhino to use `VarScope`, change the use sites to accept only `VarScope`s, and then finally once all those changes are done will we split the hierarchy.

This PR has also been written so that it is possible to split it into the individual commits, let me know @gbrail and @rbri if you would like it split like that.

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/ScopeWrapper.java
+++ b/rhino/src/main/java/org/mozilla/javascript/ScopeWrapper.java
@@ -1,0 +1,45 @@
+package org.mozilla.javascript;
+
+public class ScopeWrapper extends ScriptableObject {
+    private static final long serialVersionUID = -7471457301304454454L;
+
+    private final Scriptable scope;
+
+    public ScopeWrapper(Scriptable scope) {
+        this.scope = scope;
+    }
+
+    @Override
+    public Scriptable getPrototype() {
+        return null;
+    }
+
+    @Override
+    public Object get(String name, Scriptable start) {
+        return scope.get(name, scope);
+    }
+
+    @Override
+    public void put(String name, Scriptable start, Object value) {
+        scope.put(name, scope, value);
+    }
+
+    @Override
+    public boolean has(String name, Scriptable start) {
+        return scope.has(name, scope);
+    }
+
+    @Override
+    public VarScope getParentScope() {
+        return scope.getParentScope();
+    }
+
+    @Override
+    public String getClassName() {
+        return "scope";
+    }
+
+    public Scriptable getScope() {
+        return scope;
+    }
+}
--- a/rhino/src/main/java/org/mozilla/javascript/SpecialRef.java
+++ b/rhino/src/main/java/org/mozilla/javascript/SpecialRef.java
@@ -54,9 +54,20 @@ class SpecialRef extends Ref {
             case SPECIAL_PROTO:
                 return target.getPrototype();
             case SPECIAL_PARENT:
-                return target.getParentScope();
+                return getScriptableForScope(target.getParentScope());
             default:
                 throw Kit.codeBug();
+        }
+    }
+
+    private static Scriptable getScriptableForScope(Scriptable scope) {
+        if (scope instanceof WithScope) {
+            return ((WithScope) scope).getObject();
+        } else if (scope != null) {
+            // Wrap it.
+            return new ScopeWrapper(scope);
+        } else {
+            return null;
         }
     }
 
@@ -72,7 +83,6 @@ class SpecialRef extends Ref {
             case SPECIAL_NONE:
                 return ScriptRuntime.setObjectProp(target, name, value, cx);
             case SPECIAL_PROTO:
-            case SPECIAL_PARENT:
                 {
                     Scriptable obj = ScriptRuntime.toObjectOrNull(cx, value, scope);
                     if (obj != null) {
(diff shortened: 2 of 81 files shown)


aardvark179:
@gbrail, @rbri, I'm breaking this down into smaller, more easily review PRs of the individual commits. First one is #2348.

gbrail:
My code review AI is trying to find ways to trigger those Kit.codeBug exceptions. For example, you can do the below (if you're using an old language level only). 

I'd like to keep moving forward, but I wonder how much old code is going to crash out instead of throwing a JavaScript exception. There might be other places. Do we want to handle those things differently? With the variety of Rhino code there all sorts of things like this may be possible.

```
$ ./gradlew run --args="-version 0"
Rhino 2.0.0-SNAPSHOT (JLine xterm-256color)
js> var w = new With({});
js> w;
java.lang.IllegalStateException: FAILED ASSERTION: Attempt to get default value of scope.
        at org.mozilla.javascript.Kit.codeBug(Kit.java:365)
        at org.mozilla.javascript.VarScope.getDefaultValue(VarScope.java:25)
        at org.mozilla.javascript.ScriptRuntime.toPrimitive(ScriptRuntime.java:4317)
        at org.mozilla.javascript.ScriptRuntime.toString(ScriptRuntime.java:1142)
        at org.mozilla.javascript.Context.toString(Context.java:1732)
        at org.mozilla.javascript.tools.shell.Main.processSource(Main.java:498)
        at org.mozilla.javascript.tools.shell.Main.processFiles(Main.java:187)
        at org.mozilla.javascript.tools.shell.Main$IProxy.run(Main.java:100)
        at org.mozilla.javascript.Context.call(Context.java:570)
        at org.mozilla.javascript.ContextFactory.call(ContextFactory.java:429)
        at org.mozilla.javascript.tools.shell.Main.exec(Main.java:171)
        at org.mozilla.javascript.tools.shell.Main.main(Main.java:146)
Exception in thread "main" java.lang.IllegalStateException: FAILED ASSERTION: Attempt to get default value of scope.
        at org.mozilla.javascript.Kit.codeBug(Kit.java:365)
        at org.mozilla.javascript.VarScope.getDefaultValue(VarScope.java:25)
        at org.mozilla.javascript.ScriptRuntime.toPrimitive(ScriptRuntime.java:4317)
        at org.mozilla.javascript.ScriptRuntime.toString(ScriptRuntime.java:1142)
        at org.mozilla.javascript.Context.toString(Context.java:1732)
        at org.mozilla.javascript.tools.shell.Main.processSource(Main.java:498)
        at org.mozilla.javascript.tools.shell.Main.processFiles(Main.java:187)
        at org.mozilla.javascript.tools.shell.Main$IProxy.run(Main.java:100)
        at org.mozilla.javascript.Context.call(Context.java:570)
        at org.mozilla.javascript.ContextFactory.call(ContextFactory.java:429)
        at org.mozilla.javascript.tools.shell.Main.exec(Main.java:171)
        at org.mozilla.javascript.tools.shell.Main.main(Main.java:146)
```

gbrail:
Interestingly enough, despite my big structure changes, a few side projects I have based on rhino haven't had to be rewritten yet

gbrail:
Let me know what you think of the Kit.codeBug stuff -- I'm still willing to merge this in favor of forward progress but I don't want to leave everyone behind if there is a more JavaScripty way of failing rather than bombing out -- I feel like raw Java exceptions should be reserved for "it is impossible to get here" situations and not just obsolete feature usage.

aardvark179:
The `Kit.codebug` stuff should generally move to being compile time errors once full separation happens. 

I believe `new With({})` was very much a Mozilla extension, and I don't recall finding a real world use of it. If we do want to continue supporting it then it should create a with scope based on the passed in object, and when wrap that in the same way we have to do for `__parent__`.

gbrail:
No, I don't want to support it -- if that's the only real thing we can find I'm fine with this for now and we'll find more as we go along (although so far all my test cases keep working)

gbrail:
This is OK -- we have more opportunities to work out all the implications. Thanks!

rhino/src/main/java/org/mozilla/javascript/ScopeWrapper.java
```diff
@@ -0,0 +1,45 @@
+package org.mozilla.javascript;
+
+public class ScopeWrapper extends ScriptableObject {
```
andreabergia, line 3: This is used to expose a scope as a scriptable object for With and Parent, right?
Should it also delegate the associated values?

rhino/src/main/java/org/mozilla/javascript/SpecialRef.java
```diff
@@ -127,11 +137,13 @@ public Object set(Context cx, Scriptable scope, Object value) {
                         } else {
                             target.setPrototype(obj);
                         }
-                    } else {
-                        target.setParentScope(obj);
                     }
                     return obj;
                 }
+            case SPECIAL_PARENT:
```
andreabergia, line 143: This makes `__parent__` not writable, right?
Or was it already not really writable for some reason, and you just made it more explicit?

aardvark179: I search for cases where people attempted to change `__parent__` and didn't find any. I am not sure what it even means to change it in general.

PR #2330, opened 2026-03-10
Change APIs and use sites to use the new `VarScope` type.

aardvark179:
This PR stacks on top of #2329 and is the next stage in resolving #2163.

This PR changes the `call` and `construct` signatures to make the scopes `VarScope`s, then changes the `JSCode` APIs, and then adds a series of commits that change all the places these are implemented to make them obey the new type. We can do this gradually because `VarScope` is a subtype of `Scriptable` and so method references can be adapted.

This change also has to include @ZZZank's change #2298 to ensure that serialisation of native methods can still be done, without it there are too many ordering issues with deserialisation and the continuations tests will fail.

Like the previous PR in this series each commit involved in this change can be taken separately, and the PR can be split for purposes of review if desired.

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/FunctionObject.java
+++ b/rhino/src/main/java/org/mozilla/javascript/FunctionObject.java
@@ -153,6 +153,7 @@ public class FunctionObject extends BaseFunction {
         if (type == ScriptRuntime.BooleanClass || type == Boolean.TYPE) return JAVA_BOOLEAN_TYPE;
         if (type == ScriptRuntime.DoubleClass || type == Double.TYPE) return JAVA_DOUBLE_TYPE;
         if (ScriptRuntime.ScriptableClass.isAssignableFrom(type)) return JAVA_SCRIPTABLE_TYPE;
+        if (ScriptRuntime.VarScopeClass.isAssignableFrom(type)) return JAVA_VARSCOPE_TYPE;
         if (type == ScriptRuntime.ObjectClass) return JAVA_OBJECT_TYPE;
 
         // Note that the long type is not supported; see the javadoc for
@@ -473,7 +474,7 @@ public class FunctionObject extends BaseFunction {
      * new objects.
      */
     @Override
-    public Scriptable createObject(Context cx, Scriptable scope) {
+    public Scriptable createObject(Context cx, VarScope scope) {
         if (member.isCtor() || parmsLength == VARARGS_CTOR) {
             return null;
         }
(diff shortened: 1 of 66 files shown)


aardvark179:
Okay, we're down to just the one commit in here so we might as well use this PR rather than creating a new number.

rhino/src/main/java/org/mozilla/javascript/commonjs/module/ModuleScope.java
```diff
@@ -38,14 +38,9 @@ public URI getBase() {
         return base;
     }
 
-    @Override
-    public String getClassName() {
-        return "module";
-    }
```
andreabergia, line 44: Why was this removed?

rhino/src/main/java/org/mozilla/javascript/FunctionObject.java
```diff
@@ -528,7 +529,8 @@ private void readObject(ObjectInputStream in) throws IOException, ClassNotFoundE
     public static final int JAVA_BOOLEAN_TYPE = 3;
     public static final int JAVA_DOUBLE_TYPE = 4;
     public static final int JAVA_SCRIPTABLE_TYPE = 5;
-    public static final int JAVA_OBJECT_TYPE = 6;
+    public static final int JAVA_VARSCOPE_TYPE = 6;
+    public static final int JAVA_OBJECT_TYPE = 7;
```
andreabergia: Maybe swap these two, so that `JAVA_OBJECT_TYPE` doesn't change value for old code?
Though I imagine that no old code would compile anyway and thus this is a moot point in practice...

rbri: :-D made the same comment here https://github.com/mozilla/rhino/pull/2331#discussion_r3178267279,
cool to see we both stumbled over the same thing

rbri: my fear was a bit more about de/serialization

aardvark179: I put immediately after `JAVA_SCRIPTABLE_TYPE` because I thought they belonged together, but I could move it back.

rhino/src/main/java/org/mozilla/javascript/commonjs/module/ModuleScope.java
aardvark179: This maybe should have been held until the next PR, but doesn't hurt here. `getClassName` should only be needed for objects that actually are access and might be printed, and *scopes* do not fall into that category. `VarScope` doesn't define that method, it's only on the `Scriptable` side of the hierarchy.

gbrail:
FWIW I don't think we do anything to promise serializability compatibility
between versions. (And TBH if we could eliminate serialization that'll be
great but I don't think we can.)

gbrail:
Thanks!

PR #2331, opened 2026-03-10
Fully separate object and scope types.

### LINKED ISSUE #2163: Separation of scopes and objects

aardvark179:
This is the final stage of separating objects and scopes and stacks on top of #2330, and resolves #2163.

This PR does the final separation of `VarScope` and `Scriptable` making them both subclasses of `PropHolder<T>`, and then removes the old `NativeWith` class as it is no longer used.


jimmyhmiller:
@aardvark179 ~~Any reason these aren't stacked PRs? Might make it a bit easier to review each addition.~~ (Apparently hard to do across forks)

aardvark179:
Okay, almost at the end of the scope separation. I'm marking this as ready for review since I don't propose to spilt it into a bunch of smaller PRs.

rbri:
tested with core-js - works fine

rhino-xml/src/main/java/org/mozilla/javascript/xmlimpl/XMLLibImpl.java
```diff
@@ -626,18 +625,18 @@ public String escapeAttributeValue(Object o) {
...
     @Override
-    public Ref nameRef(Context cx, Object name, Scriptable scope, int memberTypeFlags) {
+    public Ref nameRef(Context cx, Object name, VarScope scope, int memberTypeFlags) {
         if ((memberTypeFlags & Node.ATTRIBUTE_FLAG) == 0) {
             // should only be called for cases like @name or @[expr]
             throw Kit.codeBug();
         }
         XMLName xmlName = toAttributeName(cx, name);
-        return xmlPrimaryReference(cx, xmlName, scope);
+        return xmlPrimaryReference(cx, xmlName, (VarScope) scope);
```
rbri, line 634: this cast i not needed

rhino-xml/src/main/java/org/mozilla/javascript/xmlimpl/XMLLibImpl.java
```diff
@@ -647,6 +646,6 @@ public Ref nameRef(
             }
         }
 
-        return xmlPrimaryReference(cx, xmlName, scope);
+        return xmlPrimaryReference(cx, xmlName, (VarScope) scope);
```
rbri, line 649: not needed cast

rhino/src/main/java/org/mozilla/javascript/FunctionObject.java
```diff
@@ -528,7 +529,8 @@ private void readObject(ObjectInputStream in) throws IOException, ClassNotFoundE
     public static final int JAVA_BOOLEAN_TYPE = 3;
     public static final int JAVA_DOUBLE_TYPE = 4;
     public static final int JAVA_SCRIPTABLE_TYPE = 5;
-    public static final int JAVA_OBJECT_TYPE = 6;
+    public static final int JAVA_VARSCOPE_TYPE = 6;
```
rbri: any reason why not let the java_object_type stay with the old value and add varscope at the end?

rhino/src/main/java/org/mozilla/javascript/Interpreter.java
```diff
@@ -541,24 +542,22 @@ public void delete(String name) {}
         public void delete(int index) {}
 
         @Override
-        public Object get(String name, Scriptable start) {
+        public void delete(Symbol key) {}
+
+        @Override
+        public Object get(String name, VarScope start) {
```
rbri: start -> scope

rhino/src/main/java/org/mozilla/javascript/JavaAdapter.java
```diff
@@ -576,9 +576,9 @@ public static Scriptable runScript(final Script script) {
         return ContextFactory.getGlobal()
                 .call(
                         cx -> {
-                            ScopeObject global = ScriptRuntime.getGlobal(cx);
-                            script.exec(cx, global, global);
-                            return global;
+                            TopLevel global = ScriptRuntime.getGlobal(cx);
```
rbri: the var name 'global' is a bit confunsing - maybe 'top' or 'topLevel'?

rhino/src/main/java/org/mozilla/javascript/NativeGlobal.java
```diff
@@ -487,8 +487,8 @@ private static Object js_unescape(Context cx, VarScope scope, Object thisObj, Ob
      * executed via ScriptRuntime.callSpecial().
      */
     private static Object js_eval(Context cx, VarScope scope, Object[] args) {
-        VarScope global = ScriptableObject.getTopLevelScope(scope);
-        return ScriptRuntime.evalSpecial(cx, global, global, args, "eval code", 1);
+        TopLevel global = ScriptableObject.getTopLevelScope(scope);
```
rbri: global -> topLevel

rhino/src/main/java/org/mozilla/javascript/SpecialRef.java
```diff
@@ -95,50 +95,43 @@ public Object set(Context cx, Scriptable owner, Object value) {
...
+                        if ("function".equals(typeOfTarget)) {
+                            if (value == null) {
+                                target.setPrototype(Undefined.SCRIPTABLE_UNDEFINED);
                                 return value;
                             }
 
                             final String typeOfValue = ScriptRuntime.typeof(value);
-                            if (NativeSymbol.TYPE_NAME.equals(typeOfTarget)) {
-                                return value;
+                            if ("object".equals(typeOfValue) || "function".equals(typeOfValue)) {
```
rbri, line 116: maybe we have to introduce constants of static methods for this checks

rbri (changes requested):
there are still some places pointing to NativeWith
Image

VarScope start -> VarScope scope in method signatures

rhino/src/main/java/org/mozilla/javascript/EqualObjectGraphs.java
```diff
@@ -320,11 +341,11 @@ private static Object[] getSortedIds(final Scriptable s) {
...
 
-    private static Object[] getIds(final Scriptable s) {
-        if (s instanceof ScriptableObject) {
+    private static <T extends PropHolder<T>> Object[] getIds(final PropHolder<T> s) {
+        if (s instanceof SlotMapOwner) {
             // Grabs symbols too
-            try (var map = ((ScriptableObject) s).startCompoundOp(false)) {
-                return ((ScriptableObject) s).getIds(map, true, true);
+            try (var map = ((SlotMapOwner<T>) s).startCompoundOp(false)) {
+                return ((SlotMapOwner<T>) s).getIds(map, true, true);
```
andreabergia: You can simplify this with java 17 now! 🙂

rhino/src/main/java/org/mozilla/javascript/EqualObjectGraphs.java
```diff
@@ -344,4 +365,16 @@ private static Object getValue(final Scriptable s, final Object id) {
             throw new ClassCastException();
         }
     }
+
+    private static Object getValue(final VarScope s, final Object id) {
+        if (id instanceof Symbol) {
+            return s.get((Symbol) id, s);
+        } else if (id instanceof Integer) {
+            return s.get((int) id, s);
+        } else if (id instanceof String) {
```
andreabergia, line 374: This looks like a bug, it should be `CharSequence` or we'll have problems with `ConsString`.

aardvark179: That does look like a bug, but also a bug that was already present in other cases.

aardvark179:
I've addressed almost all of these, and found some other unnecessary casts. I will do a little more tidy up tomorrow.

rhino/src/main/java/org/mozilla/javascript/EqualObjectGraphs.java
aardvark179: I think there about six places in there that appear to have the same bug, and there are whole bunch of places that could be cleaned up nicely with Java 17 features. I think that's worth doing as a separate PR.

rbri: > I think there about six places in there that appear to have the same bug, and there are whole bunch of places that could be cleaned up nicely with Java 17 features. I think that's worth doing as a separate PR.

and maybe we can have test cases for these also then

aardvark179: Yeah. I will try generating some.

rbri:
@aardvark179 i guess you are done with this - right?
from my point of view we can merge this

aardvark179:
Yeah, I'm done.

rbri:
> Yeah, I'm done.

@gbrail will be great if you merge this, i like to update core-js to this and test the next PR's afterwards

gbrail:
Needs one last rebase and fix I'm afraid, as the other PR I merged changed the compilation interface on Context. Then we can merge this and get going on the next thing!

rbri:
@aardvark179 

> Compilation failed; see the compiler output below.
  /home/runner/work/rhino/rhino/rhino/src/main/java/org/mozilla/javascript/Context.java:1201: error: incompatible types: VarScope cannot be converted to Scriptable
              return script.exec(this, scope, scope);

aardvark179:
D'oh! That will teach me to go, "No rebase conflicts, it must be fine."

aardvark179:
@gbrail, @rbri - I think that's all the failures squashed now.

gbrail:
Looks good, congratulations!

PR #2332, opened 2026-03-10
Convert pretty much all built ins to use `ClassDescriptor`s

aardvark179:
This PR stacks on top of #2331 and converts almost all built ins from `LambdaConstructor` to  `ClassDescriptor`. I think the only ones left are `NativeSymbol`, `JavaAdapter`, and some test cases. I'll add those before moving this from draft.

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/BaseFunction.java
+++ b/rhino/src/main/java/org/mozilla/javascript/BaseFunction.java
@@ -6,7 +6,9 @@
 
 package org.mozilla.javascript;
 
+import static org.mozilla.javascript.ClassDescriptor.Builder.value;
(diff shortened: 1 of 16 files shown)


aardvark179:
I've added the conversion of iterators, generators, and generator functions, to this PR. That also required some bug fixes around our iteration handling, and I've made the choice that we should do the spec compliant thing if the language version is greater than 1.8, and preserve the old spider monkey behaviour with older JS language versions.

aardvark179:
This is now down to the last batch. This last batch required making some changes to iteration as fixing the inheritance of some iterators broke legacy behaviour. I changed things to be consistent with the spec in ES6 mode, as this only affected ES6 iterators, and kept the existing behaviour for legacy modes.

gbrail:
This looks great, thanks!

PR #2333, opened 2026-03-11
Remove the separate regexp and XML constructor classes as they are no longer required.

aardvark179:
The constructor classes were needed when regexp and XML were `IdScriptableObject` based, but now that those classes have been moved over to class descriptors we no longer need those separate `Ctor` classes.


gbrail:
This looks good, thanks!

PR #2334, opened 2026-03-11
Make `this` an `Object` not a `Scriptable`, and handle `toObject` for non-strict functions.

aardvark179:
This PR stacks on top of #2332, adding two commits
[Convert thisObj to be Object not Scriptable.](https://github.com/mozilla/rhino/commit/c6c7fc8cf03216af796b681dadf6387c7b19550e)
[Update test 262 properties.](https://github.com/mozilla/rhino/commit/f45037e918d989a261b2adde8450f6c52393fb7c)

This PR converts `this` to be an `Object` rather than a `Scriptable`. We then handle the `toObject` conversion in functions.


aardvark179:
This was merged as part of the `new.target` work.

PR #2338, opened 2026-03-18
Make NativeDate.toLocaleString() behave more like real browsers

rbri:
Co-authored-by: Lai Quang Duong

This PR does the following

Make toLocaleString(), toLocaleDateString(), and toLocaleTimeString() output closer to real browser behavior by building DateTimeFormatter dynamically using the resolved locale instead of static pre-built formatters.

Now the outputs are almost identical to real browser with the exception of some minor cases.

see https://github.com/HtmlUnit/htmlunit-rhino-fork/pull/20

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/NativeDate.java
+++ b/rhino/src/main/java/org/mozilla/javascript/NativeDate.java
@@ -12,7 +12,9 @@ import static org.mozilla.javascript.ClassDescriptor.Destination.PROTO;
 
 import java.time.Instant;
 import java.time.ZoneId;
+import java.time.chrono.IsoChronology;
 import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeFormatterBuilder;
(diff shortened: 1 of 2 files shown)


gbrail:
This looks good to me, and works, although I don't know if it will conflict with one of the upcoming PRs to convert descriptors - at some point we'll want to move Date away from IdFunctionObject. But it looks good to me, thanks!

aardvark179:
Rebasing that should be pretty easy, happy to see this go in.

Edit: wait, Date is already moved to descriptors.

rbri:
@aardvark179 yes its already descriptor based
@gbrail will be great if you can merge this (and hopefully this #2130 also :-))

gbrail:
No worries, I was confused because in the PR I saw a switch statement, but I now see that's called in a helper function. I'll do this soon.

rbri:
Thanks

PR #2341, opened 2026-03-19
Improvements in lexical scoping.

aardvark179:
Improves our handling of `let` and `const`.

This was partially implemented by Claude, but I ended up undoing almost all of its changes as it produced some extremely baroque attempts to check scoping rules. The key to making this all work was coming up with a good way to implement scopes without `with`, and then finding all the things that needed to be fixed up to support that (unwinding during exceptions, breaking out of loops and blocks, etc.).

There is definitely still work to be done here - I don't think `catch` should have its own special scope type, and there's a couple of edge cases no `var` and `function` handling that still need to be cleaned up, but I think the main chunk of work is now ready for review.


aardvark179:
I'll need to do some more work on this. There is a rich seem of horrible annex b behaviour which definitely should not exist in strict mode.

gbrail:
This is great progress -- it'll make a big difference. I found a test that regressed so far (and a few more). Here's  a test program:

`function f() {
  if (true) {
    for (var i = 0; i < 3; ++i) {}
    var i = 0;      // legal: var redeclaration is always allowed
  }
}`

This works with the current master branch but fails with these changes.

aardvark179:
Thanks for pointing that out. That had caused a regression in test 262, but it was in the middle of a bunch of other change and I hadn't spotted it. There are 3 hopefully very minor regressions in the annexB tests, and one regressions round generator arguments and const redeclaration, but that was a negative test which was passing for the wrong reasons. That last regression is because we currently allow:
```javascript
function f(x) { const x = 7; return x; }
```
which should throw an error because we are redeclaring a variable as a `const`. This is, I think, because we are treating parameters differently, and is probably worth fix in a separate PR.

rbri:
@aardvark179 fantastic!! 
Will try to have a look but i fear i will not have the time during the next week. And then i'm on vaction for another two weeks and afterwards, i have to do some public presentations....

Will do my best...

Thanks for all that stuff!

gbrail:
I don't like to paste AI output directly in here because I want to make sure that it is valid, but my code review tool found some new NPEs with this branch that weren't there before, and it has a suggestion for a fix (that might be dubious). The regressions are real, though, and I don't think we should merge this until they are fixed. Once they are, though, this is a great improvement!

I will attach test cases in the next comment...

gbrail:
## B1 — Parse-time NPE crash on labeled / with-body functions in sloppy ES6

Running any of the cases below crashes the shell at parse time (both compiled and interpreter modes) with an unhandled exception instead of either accepting the code or reporting a clean SyntaxError:

```
Exception in thread "main" java.lang.NullPointerException: Cannot invoke
    "org.mozilla.javascript.ast.Symbol.getName()" because "sym" is null
	at org.mozilla.javascript.ast.ScriptNode.hasAnnexBConflict(ScriptNode.java:324)
	at org.mozilla.javascript.ast.ScriptNode.doAnnexBHoisting(ScriptNode.java:310)
	at org.mozilla.javascript.Parser.parse(Parser.java:741)
	at org.mozilla.javascript.Parser.parse(Parser.java:629)
	at org.mozilla.javascript.Context.parse(Context.java:2782)
	at org.mozilla.javascript.Context.compileImpl(Context.java:2715)
	at org.mozilla.javascript.Context.compileScript(Context.java:1561)
	at org.mozilla.javascript.Context.compileString(Context.java:1509)
```

| Case | V8 (spec behavior) | Branch | master |
|---|---|---|---|
| `lbl: function lf() { return 1; }` — labeled function at script top level | valid; `typeof lf === "function"` | **NPE crash** | ok |
| `a: b: function mf() { return 1; }` — multiple labels, script top level | valid | **NPE crash** | ok |
| `function outer() { lbl: function lf() { return 1; } }` — labeled function in function body | valid | **NPE crash** | ok |
| `eval('lbl: function ef() { return 1; }')` | valid | **NPE crash** | ok |
| `new Function('lbl: function f() { return 1; }')` | valid | **NPE crash** | ok |
| `var o = {}; with(o) function f() {}` | SyntaxError (Annex B B.3.2: functions may not be declared as the single statement of a `with` body in non-strict code) | **NPE crash** | accepted (non-compliant, no crash) |
| `if (false) lbl: function f() {}` | SyntaxError (13.6.1: IsLabelledFunction in sole statement position) | **NPE crash** | accepted (non-compliant, no crash) |
| `if (true) ; else lbl: label2: function f() {}` | SyntaxError (same rule, else position) | **NPE crash** | accepted (non-compliant, no crash) |

Repro files:

```js
// case 1 — valid code, crashes
lbl: function lf() { return 1; }
java.lang.System.out.println(typeof lf); // V8/master: "function"

// case 2 — valid code, crashes
a: b: function mf() { return 1; }

// case 3 — valid code, crashes
function outer() { lbl: function lf() { return 1; } }

// case 4/5 — valid code, crashes
eval('lbl: function ef() { return 1; }');
new Function('lbl: function f() { return 1; }');

// case 6 — should be a clean SyntaxError, crashes
var o = {}; with(o) function f() {}

// case 7/8 — should be clean SyntaxErrors, crash
if (false) lbl: function f() {}
if (true) ; else label1: label2: function test262() {}
```

This matches the test262 tests newly recorded as failing in this branch's `test262.properties`:

- `language/statements/if/labelled-fn-stmt-first.js non-strict` — negative test, `if (false) label1: label2: function test262() {} else ;` expects SyntaxError
- `language/statements/if/labelled-fn-stmt-lone.js non-strict` — negative test, `if (false) label1: label2: function test262() {}` expects SyntaxError
- `language/statements/if/labelled-fn-stmt-second.js non-strict` — negative test, `if (true) ; else label1: label2: function test262() {}` expects SyntaxError
- `annexB/language/statements/labeled/function-declaration.js non-strict` — **positive** test, `label: function g() {}` / `label1: label2: function f() {}` at top level must parse and work

### Root cause

In `Parser.function()` (Parser.java:1060-1082), for `FUNCTION_BLOCK_SCOPED` with `currentScope == currentScriptOrFn && !inUseStrictDirective` no symbol is defined:

```java
if (type == FunctionNode.FUNCTION_BLOCK_SCOPED) {
    // Block-scoped function in strict mode: define as let-like binding
    if (inUseStrictDirective || currentScope != currentScriptOrFn) {
        defineSymbol(Token.LET, name.getIdentifier());
    }
}
...
// ES2024, B.3.2.1: register annexB candidate after FunctionNode is created.
if (type == FunctionNode.FUNCTION_BLOCK_SCOPED
        && !inUseStrictDirective
        && name != null
        && name.length() > 0) {
    Symbol letSym = currentScope.getSymbol(name.getIdentifier()); // null in the case above
    currentScriptOrFn.addAnnexBFunction(letSym, fnNode);
}
```

`ScriptNode.doAnnexBHoisting()` (ScriptNode.java:310) then calls `hasAnnexBConflict(letSym)`, which dereferences the null symbol at line 324.

The routing flaw: `statementHelper()` case `FUNCTION` (Parser.java:1546) sends *every* function statement it sees to `FUNCTION_BLOCK_SCOPED` in ES6 mode. Unlabeled functions are intercepted earlier by the special cases in `Parser.parse()` (Parser.java:686) and `parseFunctionBody()` (Parser.java:797), so only **labeled** functions (via `nameOrLabel()`) and **with-body** functions reach the `BLOCK_SCOPED` path at script/function level, where `currentScope == currentScriptOrFn`. Blocks, switches, and catch clauses push their own scopes (so they don't crash), and while/for bodies push Loop scopes (so they cleanly report "let declaration not directly within block").

Suggested fix: when `currentScope == currentScriptOrFn`, don't treat the function as block-scoped — route it as `FUNCTION_STATEMENT` (mirroring the two existing special cases), which gives the spec-correct var-hoisting for the valid top-level cases and lets the existing single-statement-context checks produce the clean SyntaxError for the `with`/`if`-body cases. (A null-guard on the Annex B registration alone would stop the crash but would leave e.g. top-level `lf` unbound — `typeof lf === "undefined"` where V8 says `"function"`.)

---

## B2 — Duplicate block function declarations wrongly rejected in sloppy mode

```js
function outer() {
  { function a() { return 1; } function a() { return 2; } return a(); }
}
outer();
```

| Engine | Result |
|---|---|
| V8 (sloppy) | valid — Annex B **B.3.3.4** explicitly allows duplicate names in a block when the duplicates are only bound by FunctionDeclarations; last wins → `2` |
| V8 (strict) | SyntaxError (duplicate lexical binding) |
| Branch (sloppy) | **SyntaxError: "redeclaration of variable a"** |
| master (sloppy) | valid, last wins → `2` |

Same for switch:

```js
switch (1) { case 1: function h() {} function h() {} }
```

Corroborated by `annexB/language/function-code/function-redeclaration-block.js non-strict` and `function-redeclaration-switch.js non-strict` in the branch's `test262.properties` (both pass on master). The block test's header quotes B.3.3.4: *"It is a Syntax Error if the LexicallyDeclaredNames of StatementList contains any duplicate entries, **unless the source code matching this production is not strict mode code and the duplicate entries are only bound by FunctionDeclarations**."*

### Root cause

Block functions are defined as `LET` symbols even in sloppy mode (Parser.java:1063, `defineSymbol(Token.LET, name.getIdentifier())`), so the second `function a()` trips the let-redeclaration check in `defineSymbol`. Sloppy block functions need var-like (redeclarable) symbol semantics; only strict mode should get let-like bindings.

aardvark179:
I've done some refactoring and fixed up those redeclaration cases. There are still some failing tests around multiple generator declarations, but they weren't passing before and I I think are worth doing in a follow PR (they involve questions about what can be declared in a single statement context, and we've never handled that properly).

gbrail:
This is much closer! I still have one type of script that throws a SyntaxError with this patch that didn't previously -- it was something that came up when trying to run the JetStream benchmarks.

These scripts throw a SyntaxError:

    function f(a){for (var d in c)var a}
    function f(a){while(0)var a}
    function f(a){for (var d in c))var a}

Also I think the number of ErrorProne warnings ticked up a bit with this PR.

Could you please look at those? If I can run those other workloads, it's OK with me if we don't fix everything, but I don't want to merge this if previously-working code breaks. Thanks!

aardvark179:
Okay. I have a fix for those, but I want to add some tests round that before pushing because it doesn't seem to affect any 262 tests.

gbrail:
For me at least, this no longer breaks things that used to work, so from that perspective I think it's ready enough. I have a project that uses Babel, including the "transform-block-scoping" rule, to transform some code from the JetStream benchmark. 

The code still works with that transformation in, but unfortunately it fails if I take it out, and it no longer fails with a syntax error, but instead with an "undefined" where it shouldn't be, so I'm not sure what to make of that. I think it's good progress but we need to keep fixing bugs if we want it to be ready for a release IMO. If I can get all my tests to pass without that Babel plugin I would call that a big victory and we're not quite there yet!

aardvark179:
If you can link me to the failing case I'll take a look.

aardvark179:
Ah, sandboxing and env stuff was conspiring to make my version managed node unreachable. Had to hit the bazel config over the head. Was the problem evident with all benchmarks or a subset?

aardvark179:
Okay, so we have a difference in `Octane/raytrace.js`. I'll see if I can generate some tests for the individual functions to understand where the problem is.

aardvark179:
Okay. Found the problem. We aren't doing quite the right thing on loop compilation when the scopes don't need to be reified. I'll have a think on how to fix that.

gbrail:
Here is one test case that the AI figured out overnight:

```
let f = [undefined, undefined, undefined];
for (let x of [1, 2, 3]) {
  f[x - 1] = function () { return x; };
}
// expected: f[0]()===1, f[1]()===2, f[2]()===3
// Rhino:    f[0]()===3, f[1]()===3, f[2]()===3   <-- bug
```

This is literally the test262 test language/statements/for-of/head-let-fresh-binding-per-iteration.js. The same bug hits C-style for (0,1,2 → 3,3,3) and for-in (a,b → b,b). A plain block { let x; } is not affected — only loop-head bindings are.

aardvark179:
Yeah, way ahead of you. I've got a fix for the consts inside non-reified scopes. The per-iteration lexical environment bug is another little joy in the JS spec which is slightly more awkward to fix. Essentially in in a loop like this
```javascript
for (let x = 0; x < 10; x++) {
  x = x + 2;
}
```
then you do the following.
1. Create a lexical environment.
2. Run `let x = 0;` in that environment.
3. Create another new lexical environment and copy `x` into it and make it the current environment.
4. Run the test `x < 10;` in that environment.
5. Run `x = x + 2;` in the current environment.
6. Create _another_ new lexical environment and copy `x` into it and make it the current environment.
7. Run the iteration step `x++` in the current environment.
8. Go back to 4.

I've mostly got that working for the cases where scopes aren't reified having found some subtle compiler bugs, and that gets `reytrace.js` working without having to rewrite the block scoped variables. It doesn't however fix the reified case like that test 262 example. We need a little more scope machinery to make that work. I'm trying to find a good shape for scope machinery so that it will all be reasonably optimisable.

gbrail:
I don't know, some days I wish we had picked a less complicated language.

aardvark179:
The changes to make per iteration environment work feel extensive enough that I think we might want to split them into additional PRs. We have a couple of things which are interacting badly.
1. `const x = 7; x++` doesn't throw an error correctly (should do so *even in non-strict mode*). Fixing this requires touching the compiler backends and slot map logic in quite a few places, and not fixing this results in 262 tests which used to fail due to parsing issues instead looping infinitely.
2. The main per iteration environment changes require some changes at IR, node transformer, and back ends to work at all, and also some subtle bug fixes round `for of` and `for in` to not cause annoying test failures.
3. I've found some more scope handling bugs round using this stuff outside of a function where we appear to have some other long standing compilation issue.

And we need to make sure we only do this on ES6 because there some old mozilla tests which actually fail if we fix things.

That feels like enough subtle potential breaking changes that they are worth splitting out if that's okay.

gbrail:
It's OK with me -- with this change, we will require transformation of
existing code due to scoping issues but if you think you can get the other
changes made in a fairly short number of weeks then I think we'll be in a
very good place.

My AI came up with one fix which was pretty small, but it basically just
replaced every "const" declaration in a for-in or for-of loop with a "let"
in IRFactory. (I can paste the diff.) It does technically solve the
problem, and basically does what Babel would have done to transform
existing code before these fixes go in, but it's one of those AI-generated
fixes that fixes the problem without actually implementing the feature. If
there are better fixes on the way then I'm OK with waiting.

rbri:
> replaced every "const" declaration in a for-in or for-of loop with a "let"

Thats more or less what i'm doing in core-js to "fix" the const support....

gbrail:
If you all are OK I will go ahead and merge this since after all it doesn't fix every issue but it fixes an awful lot!

rbri:
Fine for me, will be back from vacation on the weekend and then have a closer look. But i agree this is an overall huge improvement thats worth to have it now and improve it from now on.

rbri:
@aardvark179 

> Yeah, way ahead of you. I've got a fix for the consts inside non-reified scopes.

Shold we include this existing fix as you have it already?

aardvark179:
No, because it reveals other long standing bugs that result in infinite loops in some test 262 cases. Can't even mark those as failing because we have the CI job which checks for any new passing cases.

aardvark179:
Okay, getting closer to something that fixes all the let, const, for, for-in, and for-of issues, but it's definitely big enough to be its own PR. I still needs one bug fix in compiled mode and little bit of tidying up, and I'd I think it still needs some refinement to ensure aren't changing ES5 and earlier behaviour. Should have that done by Monday.

I'd say merge this PR @gbrail, the follow up one is about 1K lines of total change so is big enough to be its own thing.

PR #2347, opened 2026-03-22
NativeProxy: fix set trap not passing receiver object

rbri:
Co-authored-by: Lai Quang Duong

This
* passes the receiver to the set trap
* passes the receiver to the get trap
* fixes an StackOverflow exception (became visible after the change for 'passes the receiver to the set trap')
* adds a bunch of spec hints to the implementation of the NativeReflect class
* implements a workaround for the missing put return value in class NativeProxy, this workaround is used by NativeReflect to implement a more correct behaviour
* fixes a check in NativeReflect; now checks for Callable in sync with the spec

This does not make so many more 262 test to pass, but i think is at least a small step forward.

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/NativeProxy.java
+++ b/rhino/src/main/java/org/mozilla/javascript/NativeProxy.java
@@ -537,6 +537,13 @@ class NativeProxy extends ScriptableObject {
      */
     @Override
     public void put(String name, Scriptable start, Object value) {
+        putAndReturn(name, start, value);
+    }
+
+    /**
+     * Workaround for the missing return value of method {@link #put(String, Scriptable, Object)}
+     */
+    boolean putAndReturn(String name, Scriptable start, Object value) {
         /*
          * 1. Assert: IsPropertyKey(P) is true.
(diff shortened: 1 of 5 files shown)


rbri:
will work on the failing tests

rbri:
This is much bigger than expected, i have to fix the NativeReflect impl (inf fact bring it closer to the spec) to fix the StackOverflow.
Hope i have some time tomorrow to work on this.

rbri:
@duonglaiquang sorry for being late on this, but is was harder than expected. My plan is to make a HtmlUnit 5.1.0 release this weekend, i fear this will not be reviewed/merged for the release. Hope this is ok for you...

rbri:
will work on all the sugestions... at least some looking correct ;-)

rbri:
this seems to be a rabbit hole, i have to think about a good way to fix this step by step. Back to Draft...

rbri:
most of this is done in https://github.com/mozilla/rhino/commit/d208d0e96521d9d4e276202903fbbb6be2e964d9
the ctor stuff will be a new PR

PR #2348, opened 2026-03-24
Change `ScriptableObject` to `TopLevel` to avoid churn later.

aardvark179:
This is the first step in introducing `VarScope` as a subtype of `Scriptable`. Breaking it out into its own PR as it is an almost entirely mechanical change which can be reviewed on its own.


gbrail:
This looks like a fine next step and I want us to keep moving so I'm going to merge this soon

gbrail:
FWIW, I'm going to try and keep making steady progress on these, which I'm largely going to be doing by looking at which PR of yours is in non-draft status, so please keep rebasing and keeping that Draft flag up to date -- I will not merge anything that says Draft. Thanks!

PR #2349, opened 2026-03-26
Change `Scriptable` inheritance and add type param in preparation for scopes.

aardvark179:
Introduce a `PropHolder<T extends PropHolder<T>>` superclass of `Scriptable`. This stacks on #2348 and is the next commit broken out of #2329.

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/MemberBox.java
+++ b/rhino/src/main/java/org/mozilla/javascript/MemberBox.java
@@ -36,7 +36,7 @@ final class MemberBox implements Serializable {
     transient Function asSetterFunction;
     transient Object delegateTo;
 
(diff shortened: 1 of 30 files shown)


gbrail:
Thanks for chunking these out so we can assess the impact!

rhino/src/main/java/org/mozilla/javascript/MemberBox.java
```diff
@@ -36,7 +36,7 @@ final class MemberBox implements Serializable {
     transient Function asSetterFunction;
     transient Object delegateTo;
 
-    private final Scriptable scope;
+    final Scriptable scope;
```
andreabergia, line 39: Accident?

PR #2350, opened 2026-03-27
Hoist methods up to `SlotMapOwner`.

aardvark179:
Hoist a bunch of stuff from `ScriptableObject` up to `SlotMapOwner` in preparation for scopes.

This is the next commit broken out of #2329.


aardvark179:
@gbrail could you re-trigger the CI RUN? I think decycle randomly failed.

gbrail:
This looks good to me and makes sense. Thanks!

PR #2351, opened 2026-03-27
Move all tests to Junit 5

gbrail:
This PR used OpenRewrite to convert all our Junit 4 tests to Junit 5. I know this is big and there is never a really good time to do this.

By doing this, we simplify the overall test execution path by getting rid of two test runtimes, and we make it more clear to new developers how to write tests. Also, this will make it easier to explore other optional build systems, which I'm thinking about if it means that we can speed up build and test.

### DIFF
--- a/buildSrc/src/main/groovy/rhino.java-conventions.gradle
+++ b/buildSrc/src/main/groovy/rhino.java-conventions.gradle
@@ -18,12 +18,10 @@ repositories {
 }
 
 dependencies {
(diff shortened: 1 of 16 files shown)


gbrail:
@aardvark179 this isn't holding anything back, but it's also a nice cleanup task. How badly does it break your big string of PRs? I could just merge it and get the pain over with or we can wait until things have calmed down a bit.

aardvark179:
Feel free to merge this whenever, rebasing on top of it is all pretty smooth.

PR #2358, opened 2026-03-28
Start to build our scope hierarchy.

aardvark179:
Create the basic scope classes required to move things over. Next part of #2329.

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/DeclarationScope.java
+++ b/rhino/src/main/java/org/mozilla/javascript/DeclarationScope.java
@@ -1,0 +1,9 @@
+package org.mozilla.javascript;
+
+public class DeclarationScope extends ScopeObject {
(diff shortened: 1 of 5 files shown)


gbrail:
Looks OK to me and the starting point for lots of things of course.

I wonder if at this point now that we're modernizing a lot if it's time to get rid of "Kit.codeBug" and replace it with throwing some sort of assertion error.

gbrail:
I'm going to merge this so that we can keep moving forward, and I really don't have a way to verify that this new hierarchy is correct until we move farther along.

One thing we should consider for an upcoming commit is to rerun "serialver" on this classes so that they have unique IDs. I personally wish we could just delete serialization from Rhino (and in fact from Java) but I know that people rely on it, and that means that the UIDs should probably unique and not cut-and-pasted like they are now.

PR #2361, opened 2026-04-03
Convert call and construct APIs to take `VarScope` scopes

aardvark179:
This is the first two commits of #2330, and will be followed by PRs that convert the use sites to the new type. We can stage this because the when converting method references to functional interfaces the Java compiler allows for widening of types, in this case `VarScope` to `Scriptable`.


aardvark179:
I think the recursive `ConsString` problem is likely an intermittent failure, and I'm not really sure what this test is trying to exercise. If we're interested in the total number of appends then we should simply increase the `ConsString` append limit, or change the strategy of what is being appended. At the moment the four thousand recursive iterations are only really testing that the stack per call is always below a limit along with the test harness parts.

gbrail:
This looks good and it's ready to go!

This is the first PR in this series that will start to break existing code -- I am ready to move forward with it but I wonder what the implications will be on downstream projects and whether we start to get more input from the community once we do this.

gbrail:
It looks like we missed most of the functions defined on LambdaConstructor and LambdaFunction -- that's causing some code I wrote to fail to compile, and I'm not sure how to reconcile that. Do you want to address that in a different PR, and is there a way to find other places where scope should now be a VarScope?

aardvark179:
> It looks like we missed most of the functions defined on LambdaConstructor and LambdaFunction -- that's causing some code I wrote to fail to compile, and I'm not sure how to reconcile that. Do you want to address that in a different PR, and is there a way to find other places where scope should now be a VarScope?

I think they all get fixed up in [the fourth round of migration](https://github.com/mozilla/rhino/pull/2330/changes/53b5b59401cf982b8fd4f1c6077803a6cddeda0c), but it would also be great to get one of your examples which is currently failing, because it suggests we've got a gap in our testing that we should try and close.

gbrail:
Sure -- this is a project that I noodle on from time to time as a way to exercise Rhino, but it never really got off the ground. But it's a good example of a codebase and since I wrote it I use LambdaFunction and LambdaConstructor all over the place. I can't currently build this branch I don't think because I am stuck in between using Scriptable and VarScope.

https://github.com/gbrail/what-js/tree/rhino-update

For example, here's some build output:

`                          ^
/home/gjb/src/what-js/core/src/main/java/org/brail/jwhat/events/Events.java:10: error: incompatible types: Scriptable cannot be converted to VarScope
    AbortSignal.init(cx, scope);
                         ^
/home/gjb/src/what-js/core/src/main/java/org/brail/jwhat/console/Console.java:70: error: incompatible types: incompatible parameter types in lambda expression
            (Context lcx, Scriptable ls, Scriptable thisObj, Object[] args) -> assertImpl(args)),
                                                                                         ^
/home/gjb/src/what-js/core/src/main/java/org/brail/jwhat/console/Console.java:79: error: incompatible types: incompatible parameter types in lambda expression
                logImpl("debug", args)),
                       ^
/home/gjb/src/what-js/core/src/main/java/org/brail/jwhat/console/Console.java:88: error: incompatible types: incompatible parameter types in lambda expression
                logImpl("error", args)),
                       ^
/home/gjb/src/what-js/core/src/main/java/org/brail/jwhat/console/Console.java:97: error: incompatible types: incompatible parameter types in lambda expression
                logImpl("info", args)),
                       ^
/home/gjb/src/what-js/core/src/main/java/org/brail/jwhat/console/Console.java:106: error: incompatible types: incompatible parameter types in lambda expression
                logImpl("log", args)),
                       ^
/home/gjb/src/what-js/core/src/main/java/org/brail/jwhat/console/Console.java:115: error: incompatible types: incompatible parameter types in lambda expression
                logImpl("warn", args)),
                       ^
/home/gjb/src/what-js/core/src/main/java/org/brail/jwhat/console/Console.java:123: error: incompatible types: incompatible parameter types in lambda expression
            (Context lcx, Scriptable ls, Scriptable thisObj, Object[] args) -> count(args)),
                                                                                    ^
/home/gjb/src/what-js/core/src/main/java/org/brail/jwhat/console/Console.java:131: error: incompatible types: incompatible parameter types in lambda expression
            (Context lcx, Scriptable ls, Scriptable thisObj, Object[] args) -> countReset(args)),
                                                                                         ^
/home/gjb/src/what-js/core/src/main/java/org/brail/jwhat/console/Console.java:139: error: incompatible types: incompatible parameter types in lambda expression
            (Context lcx, Scriptable ls, Scriptable thisObj, Object[] args) -> time(args)),
                                                                                   ^
/home/gjb/src/what-js/core/src/main/java/org/brail/jwhat/console/Console.java:147: error: incompatible types: incompatible parameter types in lambda expression
            (Context lcx, Scriptable ls, Scriptable thisObj, Object[] args) -> timeLog(args)),
                                                                                      ^
/home/gjb/src/what-js/core/src/main/java/org/brail/jwhat/console/Console.java:155: error: incompatible types: incompatible parameter types in lambda expression
            (Context lcx, Scriptable ls, Scriptable thisObj, Object[] args) -> timeEnd(args)),
                                                                                      ^
/home/gjb/src/what-js/core/src/main/java/org/brail/jwhat/stream/DefaultReadableController.java:87: error: cannot find symbol
      var sa = PromiseWrapper.wrap(cx, scope, startAlgo.call(cx, scope, sink, ScriptRuntime.emptyArgs));
                                                                        ^
  symbol:   variable sink
  location: class DefaultReadableController
/home/gjb/src/what-js/core/src/main/java/org/brail/jwhat/stream/DefaultReadableController.java:91: error: cannot find symbol
          pullIfNeeded(cx, scope);
          ^
  symbol:   method pullIfNeeded(Context,VarScope)
  location: class DefaultReadableController
/home/gjb/src/what-js/core/src/main/java/org/brail/jwhat/stream/WritableStream.java:54: error: no suitable method found for definePrototypeMethod(VarScope,String,int,(Context l[...]ctor))
    constructor.definePrototypeMethod(
               ^
    method LambdaConstructor.definePrototypeMethod(Scriptable,String,int,SerializableCallable) is not applicable
      (argument mismatch; incompatible parameter types in lambda expression)
    method LambdaConstructor.definePrototypeMethod(Scriptable,SymbolKey,int,SerializableCallable) is not applicable
      (argument mismatch; String cannot be converted to SymbolKey)
/home/gjb/src/what-js/core/src/main/java/org/brail/jwhat/stream/WritableStream.java:59: error: incompatible types: Scriptable cannot be converted to VarScope
            getWriter(lcx, ls, to, writerConstructor));
                           ^
Note: /home/gjb/src/what-js/core/src/main/java/org/brail/jwhat/events/AbortSignal.java uses or overrides a deprecated API.
Note: Recompile with -Xlint:deprecation for details.
Note: Some messages have been simplified; recompile with -Xdiags:verbose to get full output
16 errors`

aardvark179:
Okay, there are two problems in your code. You appear to be part way through converting to `VarScope` which is why you're getting the error in `Events.java`. The lambda failures are happening because the `Call` and `Construct` types have changed, but you're using explicit types in your lambdas. So if you do
```java
new LambdaFunction(
    this,
    "countReset",
    1,
    (Context lcx, Scriptable ls, Scriptable thisObj, Object[] args) -> countReset(args))
```
the types have to match exactly. I however you let the type be inferred like this:
```java
new LambdaFunction(
    this,
    "countReset",
    1,
    (lcx, ls, thisObj, args) -> countReset(args))
```
Then everything will work nicely and if you do a method reference like:
```java
new LambdaFunction(
    this,
    "countReset",
    1,
    MyClass::jsCountReset
```
Then the compiler *will* insert any widening casts required so your method can look like:
```java
public Object jsCountRest(Context cx, Scriptable scope, Scriptable thisObj, Object[] args) {
    return null;
}
```

gbrail:
Yes good point -- I looked more carefully with more time now and I can get all my stuff to build and test. Let's go ahead with this!

rhino/src/main/java/org/mozilla/javascript/Interpreter.java
```diff
@@ -610,6 +613,29 @@ public void setParentScope(Scriptable parent) {
         public void setPrototype(Scriptable prototype) {
             // Do nothing.
         }
+
+        @Override
+        public void defineConst(String name, Scriptable start) {
+            // TODO Auto-generated method stub
```
andreabergia, line 619: Maybe a `throw new UnsupportedOperationException("this should never have been called!")` rather than a TODO comment?

Issue #2237, opened 2025-12-22
Update to Java 25 and change minimum version to 17


gbrail:
Now that we have released 1.9, I'd like to move forward with Java support. Here's what we should do, in order:

1. Upgrade Gradle (or else the rest won't work)
2. Get build working on Java 25, which it looks like will at least need a few dependency updates
3. Ensure CI runs smoothly with Java 25
4. Stop running CI on Java 11
5. Switch target Java release to 17
6. Get rid of Java 11-specific workarounds (if we have any)
7. Take advantage of Java 17 features

So far I tried #1 and it worked fine but the rest is going to take some work!

gbrail:
Update: We have done steps 1-4. I would like to move forward with steps 4-6, which switch the minimum JVM version for running Rhino to 17.

andreabergia:
https://github.com/mozilla/rhino/pull/2364 achieves step 5 and 6, right?

aardvark179:
I believe so, yes.

PR #2364, opened 2026-04-12
Update versions to support Java 17 build

gbrail:
This will be the beginning of removal of support for Java 11.

It is part of the steps for https://github.com/mozilla/rhino/issues/2237

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -13,7 +13,7 @@ env:
   GRADLE_OPTS: "--enable-native-access=ALL-UNNAMED"
 
 jobs:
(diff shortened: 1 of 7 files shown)


gbrail:
Hello all -- this is the next step in our upgrade, which removes Java 11 support. I don't expect it will cause any problems but I will wait a bit before merging just in case.

rbri:
great, core-js is already uses jdk17 and it seems all is fine. +1

aardvark179:
Thanks for doing this. I've just discovered there a whole bunch of 262 tests that are failing on Java 11 because of unicode issues, so I'll be happy to see it go.

PR #2365, opened 2026-04-15
Fix javascript property access to Character field via Setter

jonasPoehler:
While updating from Rhino 1.8.0 to 1.9.1, I found a regression when accessing a Java Character field through it's setter. I provided a Testcase checking for analogous behaviour of field-access vs. method-access as well as a fix for the issue.
Since the javascript code was working with 1.8.0 it might be a possibility of this fix to be ported to the 1.9.x branch. If that is desirable I can provide an additional pull request for that.

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/NativeJavaObject.java
+++ b/rhino/src/main/java/org/mozilla/javascript/NativeJavaObject.java
@@ -340,10 +340,10 @@ public class NativeJavaObject implements Scriptable, SymbolScriptable, Wrapper, 
                     return 1;
                 } else if (to.isInstance(fromObj)) {
                     return 2;
+                } else if (to.isCharacter()) {
+                    return 3;
                 } else if (to.isPrimitive()) {
-                    if (to.isCharacter()) {
-                        return 3;
-                    } else if (!to.isBoolean()) {
+                    if (!to.isBoolean()) {
                         return 4;
                     }
                 }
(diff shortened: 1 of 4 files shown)


ZZZank:
What's the expected behaviour here? [LiveConnect Specification](https://www.oracle.com/java/technologies/javase/liveconnect-docs.html#JS_STRING_VALUES) allows String -> char conversion, but String -> Character conversion is not defined.

jonasPoehler:
Originally I came from the expectation, that assigning a single character javascript string to a java.lang.Character property, either by direct property access or through setter, should be allowed and work identical to assinging a primitive char. This was the case before 1.9.x so we had code which did relied on this. This would still be my prefered behaviour, since javascript does not know its own char type (e.g. charAt returns a String as well).
While reading through the specification you linked I understood, that technically none of the two access patterns should work and that a conversion from a javascript string to java.lang.Character was simply not defined. Since the direct access by public property (field-access) does work in rhino main (this behaviour I didn't fix), I would at least assume that the two would behave the same. Section 2.3 (Data type conversion) starts with "When calling a Java method from JavaScript, setting a Java field, or passing arguments to a constructor", while 2.3.3 defines the string conversion without java.lang.Character.

I'm not too sure how to deal with this. In my opinion rhino isn't consistent with neither the LiveConnect specification nor with itself with regards to handling java.lang.Character properties/fields. My prefered way would be to make both access types work and enhance the user-experience when interacting with java code. The other option would be, to be consistent with the specification (which would take away even more known to work javascript code). What do you think? Am I misunderstanding the specification? Or what is the intended behaviour of rhino with regards to LiveConnect?

PR #2366, opened 2026-04-15
Prefere non-vararg methods over methods with same argument types but with trailing vararg

jonasPoehler:
While updating from Rhino 1.8.0 to 1.9.1, I found a regression where rhino seems to favor the vararg version of two methods with the same base argument types. I provided a Testcase checking for this behaviour as well as a fix for the issue.
Since the javascript code was working with 1.8.0 it might be a possibility of this fix to be ported to the 1.9.x branch. If that is desired I can provide an additional pull request for that.

### DIFF
--- a/tests/src/test/java/org/mozilla/javascript/tests/NativeJavaMethodTest.java
+++ b/tests/src/test/java/org/mozilla/javascript/tests/NativeJavaMethodTest.java
@@ -3,6 +3,7 @@ package org.mozilla.javascript.tests;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
+import java.util.Map;
 import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Test;
 import org.mozilla.javascript.Context;
@@ -31,6 +32,10 @@ public class NativeJavaMethodTest {
             captured.add("1.2");
         }
 
+        public void f1(String s, MethodDummy... varargObj) {
+            captured.add("1.N");
+        }
+
         public void f2(String s1, String s2) {
             captured.add("2");
         }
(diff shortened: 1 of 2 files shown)


rPraml:
PTAL: There's a similar old draft #1433 with som3 comments

ZZZank:
As far as I know, the behaviour that the PR title describes is, when calling `someMethod(1)` with these Java methods:

```java
interface SomeClass {
    void someMethod(int i);
    void someMethod(int i, int... extra)
}
```

, `someMethod(int)` should be preferred. But changes of this PR seems to be solving another problem?

jonasPoehler:
@ZZZank you were right, I mixed in another fix, which I meant to create a separate PR for. I have removed those changes from this pull request so it now only represents what I originally meant to present. I will post the other issue separately

tests/src/test/java/org/mozilla/javascript/tests/NativeJavaMethodTest.java
```diff
@@ -42,6 +43,14 @@ public void f2(String s1, String s2, String... sN) {
         public void fN(String s1, String s2, String... sN) {
             captured.add("N." + sN.length);
         }
+
+        public void f3(Map<String, String> m, String... sN) {
```
ZZZank: Looks like the test here is for another fix? To test Rhino preference on vararg method, you will probably need to manipulate the order of which Rhino discover methods, like so:

```java
public abstract class NonVarArgParent { // dummy class
    public void someMethod(int i) {}
}

public class VarArgAtSubclass extends NonVarArgParent { // test target
    public void someMethod(int i, int... extra) {}
}

public abstract class VarArgParent { // dummy class
    public void someMethod(int i, int... extra) {}
}

public class VarArgAtParent extends VarArgParent { // test target
    public void someMethod(int i) {}
}
```

and then test can be performed by calling `testObj.someMethod(1)` with `testObj` being instances of `VarArgAtParent` and `VarArgAtSubclass`. In all cases, `someMethod(int)` should be preferred.

The reason this will work is that (almost) all `Class#getMethods()` implementation will discover a method declared by the class itself earlier than methods declared by parent classes, probably due to some recursive algorithm.

I also recommend making a new test class for this test, to avoid mixing too many contents

jonasPoehler: Thank you for your feedback! I incorporated it into the test and can confirm that the original issue was still reproducable with this and that the varargs-tiebraker does in fact fix it.

ZZZank (approved):

PR #2369, opened 2026-04-19
First round of conversion of `Scriptable` to `VarScope`.

aardvark179:
Start moving a bunch of internal API uses over to `VarScope`, this is the next commit from #2330.

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/regexp/NativeRegExp.java
+++ b/rhino/src/main/java/org/mozilla/javascript/regexp/NativeRegExp.java
@@ -4304,12 +4304,12 @@ public class NativeRegExp extends ScriptableObject {
     }
 
     private static Object js_compile(
-            Context cx, JSFunction f, Object nt, Scriptable s, Object thisObj, Object[] args) {
+            Context cx, JSFunction f, Object nt, VarScope s, Object thisObj, Object[] args) {
         return realThis(thisObj, f).compile(cx, s, args);
     }
 
     private static Object js_toString(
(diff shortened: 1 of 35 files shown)


rhino/src/main/java/org/mozilla/javascript/regexp/NativeRegExp.java
```diff
@@ -4304,12 +4304,12 @@ private void setLastIndex(Object value) {
     }
 
     private static Object js_compile(
-            Context cx, JSFunction f, Object nt, Scriptable s, Object thisObj, Object[] args) {
+            Context cx, JSFunction f, Object nt, VarScope s, Object thisObj, Object[] args) {
```
rbri, line 4307: What about unification of the parameter names, lets name the scope param always 'scope'

rbri:
Will try this tomorrow

gbrail:
Thanks -- works fine for me, but looking forward to seeing @rbri  check it out

rbri:
Works fine with HtmlUnit also. Hopefully @aardvark179 agrees on replacing the s -> scope parameter names.

And as an overall feedback - this scope thing makes the code in the end much cleaner. Had to sort out the wild mix of scope and scriptable in core-js but it was worth all the work.

aardvark179:
I'll have a look at naming the scopes `scope`. The reason I adopted the short `s` was to stop every function being reformulated to six lines of arguments. Maybe the slightly shorter type name is enough to stop that.

rbri:
ah, ok, but 's' always started some thinking at my side - is this a s-criptable or a s-cope

gbrail:
This is a great step forward, thanks!

PR #2370, opened 2026-04-20
Make OSGi integration test closer to a real-world environment

pavelhoral:
This is an update to how I implemented OSGi tests as I have discovered the actual test differs from what is the real-world usage.

What has been changed:

* tests don't have rhino modules on classpath (which could affect what ServiceLoader see)
* OSGi tests do not "hack" Thread Context Class Loader, but they instead use SPI Fly for proper integration testing


gbrail:
It's been a while since I touched Osgi but it makes sense to me. Thanks!

PR #2375, opened 2026-04-21
first test for Context and VarScope parameter

rbri:
Similar to our arch unit tests this adds another test that works on the source code and checks some parameter names

@gbrail @aardvark179 @0xe @andreabergia if you like this, i will enhance this and our code to make this pass.

### DIFF
--- a/tests/src/test/java/org/mozilla/archunit/CodeStyleTest.java
+++ b/tests/src/test/java/org/mozilla/archunit/CodeStyleTest.java
@@ -1,0 +1,160 @@
+/* -*- Mode: java; tab-width: 8; indent-tabs-mode: nil; c-basic-offset: 4 -*-
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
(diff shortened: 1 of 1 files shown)


gbrail:
I think this is an interesting idea, curious what others think. Would it be more productive to get back to this by adding checks to either Checkstyle or errorprone?

rbri:
The code was created by my new grilfriend Claude.ia - so low effort. Did something similar at work for a huge .Net solution to have the different project setups in sync.

Writing your own rules for checkstyle more or less means you have a separate project for this rules, and then adding the resulting jar to checkstyle. Plan to do this for HtmlUnit.

gbrail:
Yeah, I'm not sure I really have a strong opinion -- what you have here is
simple and has no dependencies, just wondering how to make it scale if we
want more.

rbri:
> Yeah, I'm not sure I really have a strong opinion

same here, thats why i'm asking for feedback

0xe:
I really like the spirit of enforcing style rules. And the approach looks nice as well by using the compiler API. I'm a bit unsure about what would be a good first set of rules and how permissive/restrictive they should be.

andreabergia:
I don't have a strong opinion, but I'm a bit not sure this is worth it, frankly. But I like how simple it is.

PR #2376, opened 2026-04-22
Add an optional Bazel build

gbrail:
Add an optional Bazel build.

The Bazel build:

* Only really works on Linux
* Does not support all features of the Gradle build
* Doesn't cover all the targets yet

The intent of this PR is to offer an optional build system that can potentially run much
faster and speed up key workflows for people working on Rhino.

Bazel's caching means that during "real development," often it can build and test faster than Gradle, although in a cold build both are about the same.

The real win with Bazel is using it with a hosted build service, or on a build farm, which can cut the time to build and test Bazel to only a few minutes.

In progress:

1) Does not work on Windows because Bazel falls back from symlinks to another way to locate files on that platform
2) Huge number of nuisance warnings. Apparently this can be disabled by writing a "toolchain"
which may make it harder to build remotely.
3) No support for the android tests yet
4) Does not run spotless


andreabergia (commented):
Interesting!

I think that, once this is stabilized, part of this such as `TestSource` and the moving of the commonjs/module could be extracted in their own PR first.

gbrail:
Folks -- if no one else is interested in this, that's cool, but merging this is going to make my own life more easier so if there are no objections I'd like to merge it and keep it up-to-date.

PR #2378, opened 2026-04-22
Second round of conversion of `Scriptable` to `VarScope`.

aardvark179:
Move more internal APIs over to `VarScope`. This is the next commit from #2330.

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/NativeObject.java
+++ b/rhino/src/main/java/org/mozilla/javascript/NativeObject.java
@@ -96,7 +96,7 @@ public class NativeObject extends ScriptableObject implements Map {
 
     static JSFunction init(Context cx, Scriptable s, boolean sealed) {
         var desc = cx.version >= Context.VERSION_ES6 ? ES6_DESCRIPTOR : LEGACY_DESCRIPTOR;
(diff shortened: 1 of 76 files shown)


rbri:
Will do the usual tomorrow...

rbri (approved):
LGTM
HtmlUnit test suite passes, and i found another place where i have used the scriptable instead of the scope.
Looking forward to the next step on thsi journey

gbrail:
OK -- we'll keep moving forward!

rhino/src/main/java/org/mozilla/javascript/NativeObject.java
```diff
@@ -239,8 +239,7 @@ private static Object js_protoGetter(Scriptable thisObj) {
         Let O be ? ToObject(this value).
         2. Return ? O.[[GetPrototypeOf]]().
         */
-        ScriptableObject o = (ScriptableObject) ScriptRuntime.toObject(thisObj, thisObj);
-        return o.getPrototype();
+        return thisObj.getPrototype();
```
andreabergia, line 242: This is subtle. Why does it work? `thisObj` is a scriptable, so why does that have a prototype if it's the base class of VarScope?

PR #2379, opened 2026-04-24
Third round of conversion of `Scriptable` to `VarScope`.

aardvark179:
Next round of internal API conversion. This is the next commit from #2330.

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/NativeError.java
+++ b/rhino/src/main/java/org/mozilla/javascript/NativeError.java
@@ -255,12 +255,13 @@ final class NativeError extends ScriptableObject {
 
         // The "prepareStackTrace" function takes an array of CallSite objects.
         for (int i = 0; i < stack.length; i++) {
-            NativeCallSite site = (NativeCallSite) cx.newObject(this, "CallSite");
+            NativeCallSite site =
+                    (NativeCallSite) cx.newObject(prepare.getDeclarationScope(), "CallSite");
             site.setElement(stack[i]);
             elts[i] = site;
         }
 
-        Scriptable eltArray = cx.newArray(this, elts);
+        Scriptable eltArray = cx.newArray(prepare.getDeclarationScope(), elts);
         return prepare.call(cx, prepare.getDeclarationScope(), this, new Object[] {this, eltArray});
     }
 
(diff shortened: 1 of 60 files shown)


rbri:
core-js test will follow soon

rbri:
the related core-js tests are green, let it in :-D and enjoy the weekend

gbrail:
Yes, looking good for me too, thanks!

rhino/src/main/java/org/mozilla/javascript/NativeError.java
```diff
@@ -255,12 +255,13 @@ private Object callPrepareStack(Function prepare, ScriptStackElement[] stack) {
 
         // The "prepareStackTrace" function takes an array of CallSite objects.
         for (int i = 0; i < stack.length; i++) {
-            NativeCallSite site = (NativeCallSite) cx.newObject(this, "CallSite");
+            NativeCallSite site =
+                    (NativeCallSite) cx.newObject(prepare.getDeclarationScope(), "CallSite");
```
andreabergia, line 259: This is just really a fix rather than a pure type migration, right? Before the code would basically find the top scope from the `this`, whereas now it's going to use the correct declaration scope. Correct?

rhino/src/main/java/org/mozilla/javascript/NativeSet.java
```diff
@@ -17,7 +17,8 @@ public class NativeSet extends ScriptableObject {
 
     private boolean instanceOfSet = false;
 
-    static Object init(Context cx, Scriptable scope, boolean sealed) {
+    static Object init(Context cx, VarScope s, boolean sealed) {
+        TopLevel scope = (TopLevel) s;
```
andreabergia, line 21: I assume this changes in a follow-up PR?

rhino/src/main/java/org/mozilla/javascript/NativeError.java
aardvark179: There are a bunch of places like this where the type change exposes these problems. In this commit it is a big fix, but a couple of commits later when the type hierarchies are separated then these changes become essential.

rhino/src/main/java/org/mozilla/javascript/NativeSet.java
aardvark179: Yes.

PR #2380, opened 2026-04-25
Fourth round of conversions of `Scriptable` to `VarScope`.

aardvark179:
Next round of internal API conversion. This is the next commit from https://github.com/mozilla/rhino/pull/2330.

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/JavaMembers.java
+++ b/rhino/src/main/java/org/mozilla/javascript/JavaMembers.java
@@ -146,7 +146,13 @@ class JavaMembers {
         return cx.getWrapFactory().wrap(cx, ScriptableObject.getTopLevelScope(scope), got, type);
     }
 
-    void put(Scriptable scope, String name, Object javaObject, Object value, boolean isStatic) {
+    void put(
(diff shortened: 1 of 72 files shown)


rbri (approved):
LGTM, core-js tests green, let it in

gbrail:
This works for me, thanks!

rhino/src/main/java/org/mozilla/javascript/JavaMembers.java
```diff
@@ -165,18 +171,14 @@ void put(Scriptable scope, String name, Object javaObject, Object value, boolean
             if (bp.setter == null) {
                 throw reportMemberNotFound(name);
             }
-            bp.setter.call(
-                    Context.getContext(),
-                    ScriptableObject.getTopLevelScope(scope),
-                    scope,
-                    new Object[] {value});
+            bp.setter.call(Context.getContext(), scope, obj, new Object[] {value});
```
andreabergia, line 174: This is a significant change - you're not using the top level scope, but the given one! Seems like a bug fix, but I wonder if it has implication in our fork.

PR #2381, opened 2026-04-27
Add abstraction layer for file lookups

gbrail:
Different build systems that we may use in the future handle test files
differently. Introduce an abstraction layer to locate test files (especially
the "testsrc" directory and its contents) that we can extend for these later,
and ensure that all tests use them.

Also fix two class path problems that will cause problems for more complex 
build systems like Bazel.

### DIFF
--- a/rhino/src/test/java/org/mozilla/javascript/tests/Issue176Test.java
+++ b/rhino/src/test/java/org/mozilla/javascript/tests/Issue176Test.java
@@ -21,7 +21,7 @@ public class Issue176Test {
         cx = Context.enter();
         try {
             InputStreamReader in =
-                    new InputStreamReader(Bug482203Test.class.getResourceAsStream("Issue176.js"));
+                    new InputStreamReader(Issue176Test.class.getResourceAsStream("Issue176.js"));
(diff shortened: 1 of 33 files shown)


gbrail:
This is the first step of https://github.com/mozilla/rhino/issues/2382

rbri (approved):

rhino/src/test/java/org/mozilla/javascript/tests/Issue176Test.java
```diff
@@ -21,7 +21,7 @@ public void throwing() throws Exception {
         cx = Context.enter();
         try {
             InputStreamReader in =
-                    new InputStreamReader(Bug482203Test.class.getResourceAsStream("Issue176.js"));
+                    new InputStreamReader(Issue176Test.class.getResourceAsStream("Issue176.js"));
```
andreabergia, line 24: Shouldn't this use TestSource as well?

gbrail:
No actually -- resources still work the same way in Bazel and other tools,
as long as access them through getResourceAsStream and friends. It's just
reading regular files off the disk that can be more complicated. There are
a few places where we do a bit of both and I have yet to make those changes.

PR #2384, opened 2026-04-28
Change `ScriptRuntime` in preparation for full separation.

aardvark179:
Next commit from #2330.

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/ScriptRuntime.java
+++ b/rhino/src/main/java/org/mozilla/javascript/ScriptRuntime.java
@@ -236,7 +236,6 @@ public class ScriptRuntime {
         new LazilyLoadedCtor<>(scope, "Math", sealed, true, NativeMath::init);
         new LazilyLoadedCtor<>(scope, "JSON", sealed, true, NativeJSON::init);
 
-        NativeWith.init(scope, sealed);
         NativeScript.init(cx, scope, sealed);
 
         NativeIterator.init(cx, scope, sealed); // Also initializes NativeGenerator & ES6Generator
@@ -1579,7 +1578,7 @@ public class ScriptRuntime {
     private static final String DEFAULT_NS_TAG = "__default_namespace__";
 
     public static Object setDefaultNamespace(Object namespace, Context cx) {
-        Scriptable scope = cx.currentActivationCall;
+        VarScope scope = cx.currentActivationCall;
         if (scope == null) {
             scope = getTopCallScope(cx);
         }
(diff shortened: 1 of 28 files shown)


rhino/src/main/java/org/mozilla/javascript/ScriptRuntime.java
```diff
@@ -2407,7 +2413,6 @@ private static Object nameOrFunction(
                     result = firstXMLObject.get(name, firstXMLObject);
                 }
                 // For top scope thisObj for functions is always scope itself.
```
rbri: i guess this comment has to go also

rbri (approved):
tested with core-js

rhino/src/main/java/org/mozilla/javascript/ScriptRuntime.java
aardvark179: Good catch! Fixed.

gbrail:
During one of these rounds, can you please run either "./gradlew
publishToMavenLocal" or I guess just the "javadoc" task and fix the
warnings? There are a lot more javadoc warnings now than there once were.

gbrail:
Feel free to just send that last comment to your AI tool of choice and see
if it'll fix it for us too!

aardvark179:
I'll do that tomorrow. I also need to  tidy up some serialisation ids.

aardvark179:
I've tidied up JavaDoc. I'd like to do the serial version cleanup separately because I found other cases where we've copied and pasted ids, and so it might as well be its own separate cleanup PR.

gbrail:
This caused one of my test projects to break in a way that I don't understand yet -- I'm sure it's me but I'll try and look later.

rhino/src/main/java/org/mozilla/javascript/optimizer/BodyCodegen.java
```diff
@@ -1850,8 +1850,7 @@ private void generateYieldPoint(Node node, boolean exprContext) {
                 String name = unnestedYields.get(node);
                 cfw.addALoad(variableObjectLocal);
                 cfw.addALoad(contextLocal);
-                cfw.addALoad(variableObjectLocal);
-                addDynamicInvoke("PROP:GETNOWARN:" + name, Signatures.PROP_GET_NOWARN);
+                addDynamicInvoke("NAME:GET:" + name, Signatures.NAME_GET);
```
andreabergia, line 1853: I'm a bit surprised by this (and by the similar change at line 1873). If this is intentional, can you explain why you're changing from a GETPROP to a NAME_GET? Is it because now you _know_ you are looking up from a scope so you can use the NAME instead of the "generic get prop from a scriptable" only?

aardvark179: Exactly that, yes.

andreabergia: Ok, but I'm wondering if we subtly introduced a bug in the case the name is not found.
Before you had a `NOWARN` which I seem to recall would not error out, now you have a `NAME_GET`, which I imagine throws a `ReferenceError`.
I think `yield not_found` would have errored out before too, but I wanna be sure I'm not missing something :)

gbrail:
I'm OK with this and ready to keep moving forward -- there will still be opportunity to look for bad behaviors introduced by some of these changes, which I think in general are surfacing places where we were not being careful enough with scopes and objects. Hoping that @andreabergia has more areas for us to look at or create new tests for.

PR #2386, opened 2026-04-28
Created request classes for Context.compile and eval

andreabergia:
The various variants of `Context.compile` and `eval` take already a ton of arguments. To implement source maps, we need to add at least one more, so I've started by refactoring them to take two simple record classes, named `FunctionCompileSpec` and `ScriptCompileSpec`.

I've attempted various designs and I think this one, with two separate classes, is the least bad option. :) I've created builder for simpler usage.

Backward compatibility is preserved, and I have not refactored existing tests. We could do that, since the 2.0 upgrade is big enough that refactorings like that make sense. I can do that as a follow-up step if you all think it's worthwhile.

Really only necessary mainly as a prerequisite of https://github.com/mozilla/rhino/pull/2387

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/Context.java
+++ b/rhino/src/main/java/org/mozilla/javascript/Context.java
@@ -1191,6 +1191,26 @@ public class Context implements Closeable {
     }
 
     /**
+     * Compile and execute a script described by {@code spec} in the given scope.
+     *
+     * @return the result of evaluating the source, or null if compilation produced no script
+     */
+    public final Object evaluateScript(ScriptCompileSpec spec, VarScope scope) {
+        Script script = compileScript(spec);
+        if (script != null) {
+            return script.exec(this, scope, scope);
+        }
+        return null;
+    }
+
+    /**
+     * @see #evaluateScript(ScriptCompileSpec, VarScope)
+     */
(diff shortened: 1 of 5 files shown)


aardvark179 (commented):
There is some left over type ambiguity I think we could tighten up as part of this. If you look at something like `Interpreter.compile` you'll see it still returns an `Object` even though internally it creates a `CompilationResult<?>`. It would be nice if we could make it return a `CompilationResult<T extends ScriptOrFn<T>>`. I think these new `CompileSpec` types should actually allow us to thread the approbate `T` through the whole process and make a lot of this code more type safe.

0xe (approved):

andreabergia:
Got rid of `record` given the problems I've documented in https://github.com/mozilla/rhino/issues/2394.

> There is some left over type ambiguity I think we could tighten up as part of this. If you look at something like `Interpreter.compile` you'll see it still returns an `Object` even though internally it creates a `CompilationResult<?>`. It would be nice if we could make it return a `CompilationResult<T extends ScriptOrFn<T>>`. I think these new `CompileSpec` types should actually allow us to thread the approbate `T` through the whole process and make a lot of this code more type safe.

Addressed in https://github.com/mozilla/rhino/pull/2388

gbrail:
Looks good to me too, thanks!

PR #2387, opened 2026-04-28
Implement source map support

andreabergia:
Stacked on top of https://github.com/mozilla/rhino/pull/2388. This is basically the internal implementation that we have had in ServiceNow for some years now, slightly cleaned up.

---

Adds an abstract `SourceMapper` interface. Callers attach a mapper via
`ScriptCompileSpec`/`FunctionCompileSpec` builders; the mapper is held
on `CompilerEnvirons`.

It is consulted during code generation in both backends (`BodyCodegen`
and `CodeGenerator`) so emitted line numbers refer to the original
source - i.e. this means that features such as debugger breakpoints or
stack traces come for free.

Parser error reporting and the debugger source handoff also use the
mapper. Null mappings cause the corresponding line entry to be skipped
in both LineNumberTable and Icode_LINE.

Tests cover both backends where applicable (via
`Utils.runWithAllModes`), although there are some interpreter-only
tests. `InterpreterIcodeCapture` was added to reduces duplication with
`InterpreterBytecodeDumpTest`.

---

The next step will be implementing a fully [spec-compliant](https://tc39.es/ecma426/) source mapper parser that uses the [official test suite](https://github.com/tc39/source-map-tests/).

### DIFF
--- a/rhino/build.gradle
+++ b/rhino/build.gradle
@@ -31,6 +31,7 @@ jar {
                 'org.mozilla.javascript.lc.type',
                 'org.mozilla.javascript.optimizer',
                 'org.mozilla.javascript.serialize',
+                'org.mozilla.javascript.sourcemap',
                 'org.mozilla.javascript.typedarrays',
                 'org.mozilla.javascript.xml',
             ].join(','),
--- a/rhino/src/main/java/org/mozilla/javascript/sourcemap/SourceMapper.java
+++ b/rhino/src/main/java/org/mozilla/javascript/sourcemap/SourceMapper.java
@@ -1,0 +1,42 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.javascript.sourcemap;
+
+/**
+ * Maps positions in a transpiled (target) script back to positions in the original source. Attach
+ * an instance via {@link
+ * org.mozilla.javascript.ScriptCompileSpec.Builder#sourceMapper(SourceMapper)} (or the equivalent
+ * on {@link org.mozilla.javascript.FunctionCompileSpec}) so that line numbers in stack traces, the
+ * debugger source handoff, and parser error messages refer to the original source rather than the
+ * transpiled output.
+ */
+public interface SourceMapper {
+
+    /**
+     * Maps a target {@code (line, column)} position to the corresponding original-source position.
+     *
+     * @param targetLine 1-indexed line in the transpiled source
+     * @param targetColumn 1-indexed column in the transpiled source
+     * @return the original-source position, or {@code null} if no mapping exists for this position
+     */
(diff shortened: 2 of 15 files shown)


rhino/src/main/java/org/mozilla/javascript/sourcemap/SourceMapper.java
```diff
@@ -0,0 +1,42 @@
...
+    String getOriginalSource();
+
+    /**
+     * Returns the text of the given line of the original source. Used to populate parser error
+     * messages with the offending line from the original source.
+     *
+     * @param sourceLine 1-indexed line number in the original source
+     * @return the line text, or {@code null} if the line is out of range or unavailable
+     */
+    String getSourceLineText(int sourceLine);
```
aardvark179, line 41: So, I'm trying to work out if this really makes sense. If you have an error in your original source then that should have been caught during whatever transformation process is being done, and if that transformation process introduced the error then that's the source you probably want to see.

aardvark179 (commented):
I think the code looks good, but could we get a couple of test cases to represent the most common sorts of map?
1. A source map which represents minification, i.e. it has collapsed all source lines to one minified line.
2. A source map which represents transpiration, i.e. it has expanded each line to potentially multiple lines, but not collapsed any lines.

I know we are currently limited by our line maps, and that's a good reason for source maps not being applied lazily, but I would be interested if you think there is a path to not evaluating them until they are really needs for either error reporting or debugging

rhino/src/main/java/org/mozilla/javascript/sourcemap/SourceMapper.java
andreabergia: Yeah, realistically I doubt that you'd get an error in Rhino if your code has gone through the transpiler already. But, this is the logic we have in our fork and I thought it was simple enough to be worth keeping.

0xe (approved):
looks good!

andreabergia:
> 1. A source map which represents minification, i.e. it has collapsed all source lines to one minified line.
> 2. A source map which represents transpiration, i.e. it has expanded each line to potentially multiple lines, but not collapsed any lines.

Added

andreabergia:
The CI failure seems to be just an environmental transitory problem:

> * What went wrong:
Execution failed for task ':rhino-kotlin:decycleMain'.
  > Stream closed

rhino/build.gradle
```diff
@@ -31,6 +31,7 @@ jar {
                 'org.mozilla.javascript.lc.type',
                 'org.mozilla.javascript.optimizer',
                 'org.mozilla.javascript.serialize',
+                'org.mozilla.javascript.sourcemap',
```
gbrail, line 34: I realize I'm the one who did NOT want BND to manage both OSGi modules and Java modules, but I believe that if you're going to export this new package in OSGi, you should probably export it in module-info.java as well unless you're sure it'll never be used outside this module.

andreabergia: Fixed that (and rebased)

gbrail:
OK, thanks!

PR #2388, opened 2026-04-29
Refactor compilation API for full type-safety

andreabergia:
Implements [Duncan's suggestion](https://github.com/mozilla/rhino/pull/2386#pullrequestreview-4191925795). Stacked on top of https://github.com/mozilla/rhino/pull/2386

---

Gets rid of the unchecked casts in `Context` for function or script compilation by completely splitting the API in two in the `Evaluator` interface. The `<T>` gets propagated and makes the code wholly type-safe, at the cost of a couple interfaces more.

Also added a test about the "can't compile jvm class => fallback to interpreter" pre-existing feature.


0xe (approved):

gbrail:
Looks good to me too, thanks!

PR #2391, opened 2026-05-02
Make path lookup more consistent for commonjs tests

gbrail:
In order to work with other build systems like Bazel, we either need to look up resources using the classpath or using the new "TestSource" abstraction, and we can't mix and match.


gbrail:
This one works and should be uncontroversial so I'm merging it!

andreabergia (commented):
LGTM

PR #2392, opened 2026-05-02
Fix register count in ClassCompiler

gbrail:
ClassCompiler does not pass the right number to "stopMethod". The test only
passes because we are running with Jacoco which adds extra stuff to the
bytecode.

Verified this by turning off Jacoco in "publishing-conventions.gradle"
and verifying that CLassCompilerTest fails.


andreabergia (approved):
Looks correct

PR #2393, opened 2026-05-05
Add ECMA-426 source map support

andreabergia:
Implements a compliant [ECMA-426](https://tc39.es/ecma426/) source map parser and validates it with the [official test suite](https://github.com/tc39/source-map-tests/).

Usage:

```java
SourceMapV3 mapper = SourceMapV3.parse(mapJsonString);                                                                   
Script script = cx.compileScript(
      ScriptCompileSpec.fromSource(transpiledSource)                                                                         
          .sourceName("bundle.js")                                                                                           
          .sourceMapper(mapper)
          .build());                                                                                                         
```

The implementation is in the new package `org.mozilla.javascript.sourcemap`, and it relies quite a bit on having Java 17 and `record` support! 🥳 

Some less used variants such as indexed (sectioned) source maps and transitive (chained) source maps are not supported. All the test cases from the official suite pass, except for those that deal with these two variants. 

`SourceMapV3Test` is the equivalent of `Test262SuiteTest` - it has a "known failure" file (equivalent of `test262.properties`) and supports a property to update it (`-DupdateSourceMapTestsExcludelist=true`). See its javadoc for details.

Claude was used to write most of the code, but I had a lot of input in the architecture and I reviewed the code with attention. I think it's pretty solid.

---

One non obvious design choices that is worth discussing is that I have implemented a tiny standalone JSON parser, to avoid having to tweak `NativeJson`. It's a couple hundred lines of code, which isn't great, but it has solid code coverage (about 88%). We can improve on that, decide to refactor NativeJson and extract a standalone parser that we can use in both places, or keep it like this. I'd like to hear opinions on this.


gbrail:
Reading the old comments here... The JSON parser in NativeJson is really old, and I am sure that if we use some of the accumulated knowledge from projects like Jackson we could do better. OTOH we should absolutely NOT introduce a Jackson dependency!

But I think that having one good, fast JSON parser in Rhino would be a solid thing to shoot for.

andreabergia:
Rebased and ready for review

gbrail:
This is cool but when I did the Bazel stuff I didn't anticipate a third test suite that pulls in a directory of third-party files and requires us to find them in various test environments. I thought fixing this would be simple but apparently not. You can try to follow the pattern used in Test262SuiteTest or I can find some time later!

andreabergia:
> This is cool but when I did the Bazel stuff I didn't anticipate a third test suite that pulls in a directory of third-party files and requires us to find them in various test environments. I thought fixing this would be simple but apparently not. You can try to follow the pattern used in Test262SuiteTest or I can find some time later!

I've done the changes, but apparently the CI didn't re-triggered automatically. Can you launch it?

p-bakker:
GitHub action outage

rhino/src/main/java/org/mozilla/javascript/sourcemap/MappingsDecoder.java
```diff
@@ -0,0 +1,183 @@
...
+            }
+            char c = src.charAt(pos++);
+            int digit = (c < BASE64.length) ? BASE64[c] : -1;
+            if (digit < 0) {
+                throw new SourceMapException(
+                        "invalid Base64 character '" + c + "' at line " + line);
+            }
+            continuation = (digit & VLQ_CONTINUATION_BIT) != 0;
+            long chunk = digit & VLQ_BASE_MASK;
+            result |= chunk << shift;
```
gbrail, line 179: I have not verified this myself but my AI is suggesting that there's a possibility of an overflow that doesn't get caught if shift grows beyond 64 bits, since there's no explicit check for it at this point (but there is one later, maybe too late), it'll just roll over to a negative number. Is there something else that we can check here?

rhino/src/main/java/org/mozilla/javascript/sourcemap/SourceMapJsonParser.java
```diff
@@ -0,0 +1,267 @@
...
+            } else if (c < 0x20) {
+                throw error("control character in string");
+            } else {
+                sb.append(c);
+            }
+        }
+        throw error("unterminated string");
+    }
+
+    private Object readNumber() {
```
gbrail, line 190: I feel like there's code here to parse a Number from text that we can either reuse from elsewhere in the codebase, or make this code generic and replace it with this. Probably a good optimization project to look at later.

rhino/src/main/java/org/mozilla/javascript/sourcemap/SourceMapV3.java
```diff
@@ -0,0 +1,315 @@
...
+ */
+public final class SourceMapV3 implements SourceMapper {
+
+    private final String file; // optional, may be null
+    private final List<String> sourcePaths; // resolved, may contain nulls
+    private final List<String> sourcesContent; // parallel; nullable entries; may be null overall
+    private final List<String> ignoreList; // resolved paths
+    private final List<String> names;
+    private final List<List<Segment>> segmentsByLine;
+    private final Map<String, List<String>> lineCache = new ConcurrentHashMap<>();
```
gbrail, line 30: I'm, curious about this line cache -- is it the intention that a single SourceMapper would be installed somewhere and reused by many concurrent scripts at once? And if so, is there a possibility that it will grow without bounds?

gbrail (commented):
This is great and I appreciate the 93% code coverage too.

I have a few comments, they're a bit generic, but wanted to understand a few things better. Otherwise it looks very good and thanks!

rhino/src/main/java/org/mozilla/javascript/sourcemap/SourceMapJsonParser.java
andreabergia: I'm not sure about this - the json grammar and the JS one aren't the same (JS is a bit more flexible). Keeping them distinct seems the best choice here to me.

rhino/src/main/java/org/mozilla/javascript/sourcemap/SourceMapV3.java
andreabergia: Two things:

1) the class is immutable; this cache is the only mutable part. Having a `ConcurrentHashMap` makes this thread-safe. As on the sharing or reusing, I'm not 100% sure it's something someone will realistically do - the idea is really to have one source map, do one code compilation, and throw it away. But nothing in the API prevents other usages, and I think the overhead is low.

2) It's bounded by the content of the original source files.

rhino/src/main/java/org/mozilla/javascript/sourcemap/MappingsDecoder.java
andreabergia: Makes sense; added a test and a guard

gbrail:
This is great -- thanks!

PR #2396, opened 2026-05-06
Second attempt to use record with Android

andreabergia:
Following Claude's suggestion, alternative to https://github.com/mozilla/rhino/pull/2395.

Let's see what happens in the CI first


andreabergia:
This doesn't work either. I give up :-)

Issue #2126, opened 2025-10-16
Object assign does not appear to work for NativeArray


rmmasse40:
The problem looks to be with the AbstractEcmaObjectOperations put method. It uses the ScripttableObject putImpl method which ignores the dense implementation, using the slotMap instead.

aardvark179:
I think this was fixed in d243106d325ff10cd48b7b7ac99b4cb8273a484e last December. Could you either check using a build on master, or provide an explicit test case and the exact version you're using?

rmmasse40:
My original submission was on 1.7.15 which is when we noticed the issue. I have updated to version 1.8.0 (which includes the commit) and still seeing the issue.
Here is a quick test case:

let src = [1,2,4];
let tgt = [4,5,7];
Object.assign(tgt, src);
tgt;

"tgt" result shows as [4,5,7]. i.e. unchanged rather than expected [1,2,4]

aardvark179:
Ah yes. I see it is still not doing quite the right thing. I will investigate more when less jet lagged.

aardvark179:
Okay, I have a fix, and I’ll make a PR. This whole area is a little finicky because all of Rhino’s property access methods almost map to those in the spec, but not quite perfectly. Also there appears to be nothing in test262 for this particular combination. I should probably check if there are any other exotic object types they are missing.

I think they wrote tests for all object types with special `[[Set]]` implementations, but ignored the fact that ordinary objects delegate to `[[DefineOwnProperty]]` and so you also need tests for object types with special versions of that.

aardvark179:
@gbrail the implementation of this whole area is slightly off because `put` doesn't return a boolean, but in the spec `[[Set]]` does. If we're changing method signatures as part of separating out scopes from scriptable objects should we take the opportunity to fix that, and then we could start to tidy up this whole area?

rbri:
@aardvark179 the same for delete - there is also the return code missing. This is a real problem in some areas. Do you have a backward compatible solution in mind?

aardvark179:
> [@aardvark179](https://github.com/aardvark179) the same for delete - there is also the return code missing. This is a real problem in some areas. Do you have a backward compatible solution in mind?

I don't have a backward compatible solution in mind, I tried when prototyping separating scopes and scriptable objects we have too many cases where our uses of `super.put()` and similar resulted in very hard to debug errors.

[...]

PR #2399, opened 2026-05-11
Fix Object.assign with array target losing length update (#2126)

Finomosec:
## Summary

`Object.assign([], src)` returned an empty array instead of a copy of `src`'s
enumerable own properties. The integer-keyed property writes went through
`AbstractEcmaObjectOperations.put` → `ScriptableObject.putOwnProperty` →
`putImpl`, which bypasses `NativeArray.put(int)` and so the array's length
was never updated. The slots got stored but stayed invisible to iteration
and serialization.

Reproducer (failing before this PR):

```js
const arr = ['bla', 'blubb'];
const clone = Object.assign([], arr);
JSON.stringify(clone);      // → "[]"   (expected: ["bla","blubb"])
clone.length;               // →  0     (expected: 2)
```

## Fix

Short-circuit to `targetObj.put(...)` in `NativeObject.js_assign` when the
target is a `NativeArray`, so the array's own put-path (which maintains
`length`) is used. Same special-case for the String-key path, since numeric
string keys like `"0"`, `"1"` are also indexed properties for arrays.

The non-array code path stays exactly as before.

## Tests

New `ObjectAssignArrayTargetTest` (32 cases) covering:
- the original reproducer, length update, indexed access
- pre-filled targets, sparse sources (holes skipped), multiple sources
- mix with `push` / spread / `forEach` / `map` / `reduce` / `splice`
- pre-sized arrays via `new Array(n)` (length must not shrink)
- nested `Object.assign` chains, 100-entry volume test
- object-target baseline (must remain unchanged)
- spread, basic array/object destructuring (the supported subset)

All tests run in both interpreter and compiled mode via
`Utils.assertWithAllModes_ES6`.

## Context

Refs #2126.

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/NativeObject.java
+++ b/rhino/src/main/java/org/mozilla/javascript/NativeObject.java
@@ -704,18 +704,33 @@ public class NativeObject extends ScriptableObject implements Map {
             } else {
                 ids = sourceObj.getIds();
             }
+            // Issue #2126: when the target is a NativeArray,
+            // AbstractEcmaObjectOperations.put() routes through
+            // ScriptableObject.putOwnProperty()/putImpl(), which bypasses
+            // NativeArray.put(int) and thus fails to update the array length.
+            // The slots get written but stay invisible (length stays 0).
+            // Short-circuit to the array's own put() so length is maintained.
+            boolean targetIsArray = targetObj instanceof NativeArray;
             for (Object key : ids) {
                 if (key instanceof Integer) {
                     int intId = (Integer) key;
                     if (sourceObj.has(intId, sourceObj) && isEnumerable(intId, sourceObj)) {
                         Object val = sourceObj.get(intId, sourceObj);
-                        AbstractEcmaObjectOperations.put(cx, targetObj, intId, val, true);
+                        if (targetIsArray) {
+                            targetObj.put(intId, targetObj, val);
+                        } else {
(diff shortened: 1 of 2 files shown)


Finomosec:
For reviewer context: this PR implements exactly the approach @aardvark179 outlined in [#2126 (comment)](https://github.com/mozilla/rhino/issues/2126#issuecomment-3958910836):

> We can't simply subclass `putOwnProperty` for `NativeArray` because that causes infinite recursion due to the way things are currently written. I think the only short term fix for this bug is to special case things for arrays.

Scope is intentionally narrow — only the `NativeArray` target case in `js_assign`. Other exotic object types that may share the same problem (raised earlier by @aardvark179 in the same issue) are left for the broader `put`/`[[Set]]` refactor that's planned alongside the scope split for 2.0.

Also pushed an update that adds the exact reproducer from the original issue (`Object.assign([4,5,7], [1,2,4])` → `[1,2,4]`) as its own test case, so the regression is documented verbatim.

rbri:
@Finomosec pleasr run

./gradlew spotlessApply

to correct the formatting

gbrail:
Looks great, going to give the others a minute to take a look, thanks!

rbri (approved):

gbrail:
I rebased the tests and ran them locally, and no test262 tests failed, but sadly no new tests passed either. I still think we should merge this though!

rbri:
@gbrail did something similar, same result.

gbrail:
Still a good fix though -- thanks!

PR #2398, opened 2026-05-11
Fix uint8 overflow in array literal skipIndexesId and sourcePositions

rbri:
fix: use uint16 instead of uint8 for array literal skipIndexesId and sourcePositions When a script contains more than 255 object/array literals, the literalIds table grows beyond 255. The skipIndexesId (used for sparse array literals) and sourcePositions (used for spread in sparse arrays) were encoded as uint8, causing an IllegalStateException (Kit.codeBug) when the values exceeded 255.

Changed addUint8 to addUint16 in CodeGenerator.visitArrayLiteral and updated the corresponding reads in Interpreter (DoLiteralNewArray, DoSpread) to use getIndex (uint16) instead of single-byte reads.

adapted from the inital work of Finomosec (pr: #2321)

PR #2400, opened 2026-05-11
fix for ignored spreads in array construction

rbri:
found - while working on #2398

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/NewLiteralStorage.java
+++ b/rhino/src/main/java/org/mozilla/javascript/NewLiteralStorage.java
@@ -69,9 +69,14 @@ public abstract class NewLiteralStorage {
         }
 
         if (spreadAdjustments != null) {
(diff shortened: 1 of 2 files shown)


rbri:
@0xe please have a look

aardvark179:
I am a little surprised this didn't cause any 262 cases to pass. I'd like to dig into this a little and compare it to my proof of concept branch.

rbri:
> I am a little surprised this didn't cause any 262 cases to pass

I think we had this with some other fixes before, the 262 is far from complete

aardvark179:
Okay, I've checked this on my branch, and it doesn't even cause any new 262 tests to pass there. I _think_ there is at least one array elision test that would cover this, but we have failures there due to other issues.

gbrail:
I think this is in good shape from the new test case -- thanks!

rbri:
thanks

0xe (commented):
👍

PR #2401, opened 2026-05-12
Tidy up serial version ids.

aardvark179:
This PR updates all non-test serial versuion UIDs to make them unique.

### DIFF
--- a/examples/src/main/java/Counter.java
+++ b/examples/src/main/java/Counter.java
@@ -4,6 +4,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+import java.io.Serial;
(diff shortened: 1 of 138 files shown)


rbri (approved):

gbrail:
Since we're doing this and also moved to Java 17, can we also add the @Serial annotation to all the serialVersionUID fields as well?

gbrail:
You could throw a bunch of tokens at this problem, BTW, or you could use OpenRewrite:

https://docs.openrewrite.org/recipes/staticanalysis/addserialannotationtoserialversionuid

aardvark179:
I used open rewrite, and the experience was mixed. It's easy to add the rules, but it changed other things when adding the import (collapsed 5+ imports from `java.io` to a single `import java.io.*`) which show up as a compiler warnings and are not fixed by our spotless config.

gbrail:
Looks good to me, thanks!

PR #2404, opened 2026-05-12
Tighten up exceptions related to class compilation.

aardvark179:
I discovered that our blanket class file format exception handling on compilation was hiding some bugs in my proof of concept branches. Tightening up the exception types has uncovered these issues, and hasn't shown up any other tests failures.


rbri:
looks reasonable for me

aardvark179:
Well, I ran the basic test suite, and test 262. Guess I should have checked the Mozilla suite as well. I'll investigate tomorrow.

gbrail:
This is cool and I'm good with it, but it turns out that there exist scripts that end up being out of range for our usual opcodes that can still be run in interpreted mode. Here's one

https://github.com/gbrail/rhino-benchmarks/blob/main/threejs/three.js

You can just run that script in the Rhino shell in the "master" branch and it runs to completion, but if you run it with this change it fails with ClassFileFormatException: negative operand

That indicates that overflows aren't necessarily confined to larger values, and that there's some wrapping happening somewhere as well!

aardvark179:
Nice catch. That was a bug of one of the constant pool functions attempting to cast the integer const number to a short, and then back, and it being sign extended. I haven't added a separate specific test for that because constructing something with enough constants which doesn't overflow anything else is proving a little tricky.

gbrail:
Thanks, this is much better!

PR #2406, opened 2026-05-16
First round of moving remaining built in objects to descriptors.

aardvark179:
This is the first chunk of #2332 which moves base functions and native strings over to the new descriptor system.

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/BaseFunction.java
+++ b/rhino/src/main/java/org/mozilla/javascript/BaseFunction.java
@@ -6,6 +6,8 @@
 
 package org.mozilla.javascript;
 
+import static org.mozilla.javascript.ClassDescriptor.Destination.PROTO;
+
 import java.io.Serial;
 import java.util.EnumSet;
 import org.mozilla.javascript.xml.XMLObject;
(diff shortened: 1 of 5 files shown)


rbri:
will smoke test it in HtmlUnit

Btw @aardvark179 i really like this stepwise approach

rhino/src/main/java/org/mozilla/javascript/BaseFunction.java
```diff
@@ -30,94 +32,56 @@ public class BaseFunction extends ScriptableObject implements Function {
...
-                new LambdaConstructor(
-                        scope,
-                        FUNCTION_CLASS,
-                        1,
-                        BaseFunction::js_constructorCall,
-                        BaseFunction::js_constructor);
-
+    private static final ClassDescriptor DESCRIPTOR;
+    private static final ClassDescriptor ES6_DESCRIPTOR;
+    //    private static final ClassDescriptor GENERATOR_DESCRIPTOR;
```
rbri: remove?

rhino/src/main/java/org/mozilla/javascript/BaseFunction.java
```diff
@@ -396,18 +360,34 @@ public boolean hasInstance(Scriptable instance) {
...
+            var desc = ((JSFunction) f).getDescriptor();
+            return desc == APPLY_DESCRIPTOR || desc == CALL_DESCRIPTOR;
+        }
+        return false;
     }
 
     private static Object js_hasInstance(
-            Context cx, VarScope scope, Object thisObj, Object[] args) {
-        if (!(thisObj instanceof Callable)) {
+            Context cx, JSFunction f, Object nt, VarScope s, Object thisObj, Object[] args) {
```
rbri, line 387: what about scp as var name for scope - still think it is easier to tread

what does 'nt' stands for, maybe we can also find a better wording

rhino/src/main/java/org/mozilla/javascript/BaseFunction.java
```diff
@@ -396,18 +360,34 @@ public boolean hasInstance(Scriptable instance) {
             Id_arguments = 5,
             MAX_INSTANCE_ID = 5;
 
-    static boolean isApply(KnownBuiltInFunction f) {
-        return f.getTag() == APPLY_TAG;
+    static boolean isApply(Callable f) {
+        if (f instanceof KnownBuiltInFunction) {
```
rbri: we are at java17 - let us simplify this - 

if (f instanceof KnownBuiltInFunction kf) {

the same at many other places in this pr

rhino/src/main/java/org/mozilla/javascript/NativeScript.java
```diff
@@ -26,7 +26,7 @@ class NativeScript extends BaseFunction {
 
     private static final Object SCRIPT_TAG = "Script";
 
-    static LambdaConstructor init(Context cx, VarScope scope, boolean sealed) {
+    static LambdaConstructor init2(Context cx, VarScope scope, boolean sealed) {
```
rbri: maybe a better name?

rbri:
smoke test was successful

rhino/src/main/java/org/mozilla/javascript/BaseFunction.java
aardvark179: Fixed

rhino/src/main/java/org/mozilla/javascript/BaseFunction.java
aardvark179: Fixed.

rhino/src/main/java/org/mozilla/javascript/NativeScript.java
aardvark179: Changed to `initLambda`

rhino/src/main/java/org/mozilla/javascript/BaseFunction.java
aardvark179: `nt` stands for `new.target` which is mostly just a thing for constructors. The problem we have is that the standard arguments line with indentation looks like:
```Java
            Context cx, JSFunction f, Object nt, VarScope s, Object thisObj, Object[] args) {
```
which is 93 characters. So we don't have a lot of space to play with without it being formatted to six lines.

gbrail:
Looks good to me, it seems like we can start to move forward with more of these. I think that with our existing dynamic linking support (and yet another attempt to try different types of SlotMap) we can significantly speed up access to methods and properties on native objects by optimizing when different objects have the same descriptor.

PR #2408, opened 2026-05-18
Fix typed array constructor to avoid large test 262 regression.

aardvark179:
The update to 262 meant we were exercising the typed array constructor more thoroughly in many tests, and failing. This isn't a perfect fix (I think something better can be done once everything is converted over to descriptors and we can easily check for known iterator methods) but this will do for now.


rbri:
wow > 500 test fixed 🚀

aardvark179:
Laugh it fuzzball, you broke them. 😄

rbri (approved):

gbrail:
Thanks!

PR #2410, opened 2026-05-20
fix: Throw TypeError if spread target has no @@iterator

0xe:
No new enabled 262 tests pass.

### DIFF
--- a/tests/src/test/java/org/mozilla/javascript/tests/ArrayLiteralSpreadTest.java
+++ b/tests/src/test/java/org/mozilla/javascript/tests/ArrayLiteralSpreadTest.java
@@ -283,6 +283,63 @@ class ArrayLiteralSpreadTest {
     }
 
     @Test
+    void testSpreadNonIterableObjectThrowsTypeError() {
+        // Per ecma-262 13.2.5.5, spreading an object without @@iterator throws TypeError.
+        String script =
+                "try {\n"
+                        + "  var result = [...{a: 1, b: 2}];\n"
+                        + "  'should not reach here';\n"
+                        + "} catch (e) {\n"
+                        + "  e instanceof TypeError ? 'TypeError' : 'wrong: ' + e;\n"
+                        + "}";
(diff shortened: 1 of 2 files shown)


rbri:
ask copilot just to see if i overlooked something

rbri:
@0xe maybe copilot has a point here - i think we do not have tests for ...null / ...undefined.

0xe:
> @0xe maybe copilot has a point here - i think we do not have tests for ...null / ...undefined.

Yep, addressed and added tests.

tests/src/test/java/org/mozilla/javascript/tests/ArrayLiteralSpreadTest.java
```diff
@@ -282,6 +282,63 @@ void testSpreadIteratorNextThrows() {
...
+                        + "var saved = Array.prototype[Symbol.iterator];\n"
+                        + "delete Array.prototype[Symbol.iterator];\n"
+                        + "var outcome;\n"
+                        + "try {\n"
+                        + "  var result = [...arr];\n"
+                        + "  outcome = 'should not reach here';\n"
+                        + "} catch (e) {\n"
+                        + "  outcome = e instanceof TypeError ? 'TypeError' : 'wrong: ' + e;\n"
+                        + "} finally {\n"
+                        + "  Array.prototype[Symbol.iterator] = saved;\n"
```
andreabergia, line 312: I think you can ignore the "save/restore"; other tests will run in a new context/scope anyway.

andreabergia (approved):
LGTM

aardvark179:
LGTM.

Something that might be worth thinking about is as a follow up is whether we can come up with a better way to factor out optimised iterator handling. There are over 20 places in the spec which reference `GetIterator` which should all be optimisable for built in types in a pretty much identical way (now we have descriptors this is pretty easy to do by comparing the iterator's descriptor to the known. built in).

gbrail:
Thanks!

PR #2412, opened 2026-05-20
Second round of moving built ins to descriptors.

aardvark179:
Second chunk of #2332, which moves Number, Math, Script, JSON, Global, iterators, and the error classes over to descriptors.


gbrail:
Great progress, thanks!

PR #2413, opened 2026-05-21
Third round of moving built ins to descriptors.

aardvark179:
This is the third chunk of stuff from #2332. This moves typed array related things to descriptors.


gbrail:
This is great -- thanks!

Issue #2367, opened 2026-04-16
TokenStream sets back the cursor (to a value+1) when it shouldn't do that (only when EOF)


jcompagner:
see my comments here on the commit:

https://github.com/mozilla/rhino/commit/eb1df3592bc8b8169d60198e497210a90daacb50#r182597310

i pushed a fix and a description also on our fork:

https://github.com/Servoy/rhino/commit/1d7c55b82c06682bd8800fe81355b000aaaad51b

 i also added here a few test that really test for the tokenStart/End in the 2 scenarios:

https://github.com/Servoy/org.eclipse.dltk.javascript/commit/9f3ba83b5cf73392f174f8dd959ed7492d65403d#diff-97a5ed909916dc014f3b6ead02e4974d9dbf45b569e04000466a4749c1ce596dR2735

rbri:
Hallo @jcompagner,

tried a bit to reproduce your case with the current head. But had no success...
Maybe you can have a look at my tests and give me some hints what your problem is.


    @Test
    public void testArrowRhinoWithLineEndingOnlyThis() {
        FunctionNode arrowFn = parseAndExtractArrowFn("()=>this");

        assertEquals(0, arrowFn.getPosition());
        assertEquals(0, arrowFn.getAbsolutePosition());
        assertEquals(8, arrowFn.getLength());

        Block body = (Block) arrowFn.getBody();
        assertEquals(4, body.getPosition());
        assertEquals(4, body.getAbsolutePosition());
        assertEquals(4, body.getLength());
    }

    @Test
    public void testArrowRhinoWithLineEndingThisWithProperty() {
        FunctionNode arrowFn = parseAndExtractArrowFn("()=>this.xs");

        assertEquals(0, arrowFn.getPosition());
        assertEquals(0, arrowFn.getAbsolutePosition());
        assertEquals(11, arrowFn.getLength());

        Block body = (Block) arrowFn.getBody();
        assertEquals(4, body.getPosition());
        assertEquals(4, body.getAbsolutePosition());
        assertEquals(7, body.getLength());
    }

jcompagner:
ah sorry those test are for us to test at least the behavior you wanted..
The thing is more that if i change the code back those test will still work but a lot of others start to break:

	public void testIdentifierDoc() {
		String source =  " /** @type {JsRecord} */\r\n"
				+ "		this.record = record\r\n"
				+ "		/** @type {RegExp} */\r\n"
				+ "		some_expr = 1\r\n";
[...]

PR #2414, opened 2026-05-21
TokenStream sets back the cursor (to a value+1)

rbri:
TokenStream sets back the cursor (to a value+1) when it shouldn't do that (only when EOF)

fix and testcases provided by @jcompagner; see #2367 for more


gbrail:
Good fix -- thanks!

PR #2419, opened 2026-05-31
Reflect and Proxy get() improvements

rbri:
Reflect#get:
* if receiver is not present, then set receiver to target
* use the receiver when searching for the property
* walk the prototype chain when searching for the property

Proxy#get:
* NativeProxy has to use the correct Receiver when calling the get trap


gbrail:
This seems fine to me and it fixes a test -- I have no problems merging but wanted to give others a chance. Thanks!

gbrail:
Looks good to me. Thanks!

Issue #2416, opened 2026-05-28
Integrate new instruction & operand based interpreter.


aardvark179:
Last year @jimmyhmiller and @camnwalter worked on a new interpreter for Rhino. The design is based on instruction objects combined with operands that describe where and how they get their data, but remains primarily stack based like the existing interpreter. So, where we might currently have instructions like this:

```
REG_INT_C1
LOCAL_LOAD
REG_INT_C2
LOCAL_LOAD
ADD
```

the new interpreter would have a single `Add` instruction with two operands for local 1, and local 2. Although the individual instruction objects are significantly large in footprint the operand objects can be heavily reused and the reduced number of instructions makes the memory overhead considerably smaller than might be expected. Performance on some benchmarks is also significantly higher than old interpreter.

I've been working on a set of PRs to integrate these changes which I hope to start push here soon. These will be roughly:
1. Some refactoring of our test 262 harness to enable running with debug enabled and enabling the addition of new modes ( #2225 ).
2. Refactoring `Context` to use an `EvaluationMethod` enumerator for choosing between interpreter, interpreter v2, and compilation, and conversion of all tests to use this (#2424).
3. Some refactoring of the current interpreter to extract common functionality
4. Introducing the V2 infrastructure (basic interpreter loop, compiler etc.)
5. A series of commits to introduce the operands and instructions in a logical order, and in groups small enough to be easily reviewable against the existing interpreter
6. Enabling the new interpreter

There is also some other potential clean up we could combine with this work where the V2 interpreter has introduced enums instead of lists of static ints for things like function types. Although these changes were originally limited to just the V2 code I'll try to expand any that sense more widely applicable across the codebase. I will also provide full comparative benchmarks.

aardvark179:
At the moment I've got all the pretty much up to date with function descriptors but I'm still tidying up and iterating on things a bit. So step 1 and 2 are completely ready, but steps 3 onwards will likely need to wait till next week.

aardvark179:
Okay, I've mostly got all this rebased, divided into sensible PRs, and tidied up. Here are the benchmark comparisons.
First, V8:
```
Benchmark                   (evalMethod)  Mode  Cnt       Score       Error  Units
V8Benchmark.boyer            Interpreter  avgt    5  236247.007 ± 30089.816  us/op
V8Benchmark.boyer          InterpreterV2  avgt    5  204855.793 ± 56842.097  us/op
[...]

PR #2424, opened 2026-06-03
Introduce `EvaluationMethod` enum to enable the introduction of new interpreter.

aardvark179:
This change introduces an enum that expresses the possible methods by which code can be evaluated and encapsulates the logic to instantiate interpreters or compilers for that method.


benchmarks/build.gradle
```diff
@@ -32,7 +32,7 @@ jmh {
             }
         }
     }
-    jvmArgs = ['-XX:+BackgroundCompilation']
+    jvmArgs = ['-XX:-BackgroundCompilation']
```
andreabergia: As we discussed in our call, this should really be moved to its own PR

gbrail:
Is the idea here that there are multiple options going forward other than two? If so then fine, it's just an intrusive change to an interface we just changed recently.

aardvark179:
> Is the idea here that there are multiple options going forward other than two? If so then fine, it's just an intrusive change to an interface we just changed recently.

Exactly that, this is the second thing on the list for #2416. Now I've got a branch with everything laid out somewhat sensibly I'll create a draft PR for the whole thing with an explanation of the architecture and why we did it.

aardvark179:
I should point out I am not getting rid of the `setInterpreterMode` or the `setOptimizationLevel` APIs. The reason for introducing an enum and changing the tests to use this this is that the new interpreter can be introduced without touching the tests themselves.

rhino/src/test/java/org/mozilla/javascript/tests/ContinuationsApiTest.java
```diff
@@ -102,7 +103,8 @@ public void scriptWithContinuations() {
     public void scriptWithMultipleContinuations() {
         try (Context cx = Context.enter()) {
             try {
-                cx.setInterpretedMode(true); // must use interpreter mode
+                cx.setEvaluationMethod(EvaluationMethod.Interpreter);
+                ; // must use interpreter mode
```
andreabergia: These semicolons on their own seem to be an accident

andreabergia:
> I should point out I am not getting rid of the `setInterpreterMode` or the `setOptimizationLevel` APIs. The reason for introducing an enum and changing the tests to use this this is that the new interpreter can be introduced without touching the tests themselves.

Why not?
Since 2.0 requires anyway various changes from any user, could we take the opportunity to completely drop these API? I think at least the `setInterpreterMode` has been deprecated for a while?

benchmarks/build.gradle
gbrail: Thanks, please take a look at this so we don't leak too much stuff together here.

gbrail:
as for changing "setInterpreterMode," we only changed this interface very recently in Rhino timescale (earlier this year if I recall, or late last year), so we should be careful. Maintaining those few old methods for backward compatibility seems pretty harmless to me at the moment.

benchmarks/build.gradle
aardvark179: Removed. I'll make a separate PR for this.

rhino/src/test/java/org/mozilla/javascript/tests/ContinuationsApiTest.java
aardvark179: Sorry, I thought I'd cleaned that up, but apparently I completely failed to actually push stuff.

gbrail:
This looks good now, thanks!

PR #2425, opened 2026-06-04
some more improvements for Reflect/Proxy

rbri:
* Reflect set uses the correct target when forwarding the call
* NativeProxy support the receiver when calling the set trap
* introduce and use AbstractEcmaObjectOperations ordinarySet() and ordinarySetWithOwnDescriptor() for Reflect
* some fixes to support `with (proxy) {}` statements
* defineOwnProperty() should not throw when called vial Reflect
* improvements for Reflect set() when applying to TypedArrays

Seems like a bunch of tests are missing in test262, no real visible progress here.
But all the unit tests added for this are checked with real browsers for correctness.


rbri:
please squezze

rbri:
successfully tested with HtmlUnit

gbrail:
This is very nice -- thanks for all the testing. We've verified various ways, let's merge it. Thanks!

PR #2426, opened 2026-06-04
New faster interpreter based on instruction objects and operands.

aardvark179:
This change implements a new interpreter based on `Instruction` objects rather than byte codes. Instructions generally have one or two operands which express where they get their data from (literal values, variables, or the stack). Currently all instructions that produce results put those results on the stack, but in the future the operands concept could be extended to  where the result is placed and provide further improvements.

This design provides significant performance improvements over the old interpreter (see benchmarks in #2416 ) in return for a slightly increased memory overhead. This overhead is less than might be expected because operand objects can be heavily reused, fewer instructions are usually required for the V2 interpreter, and the byte code itself is often only a portion of the overall foot print of interpreted functions compared to constants, metadata, etc.

While upstreaming this work I've taken the opportunity to remove a lot of duplication between the two interpreters as inconsistencies between these were the source of a few issues while rebasing this work.

### DIFF
--- a/benchmarks/src/jmh/java/org/mozilla/javascript/benchmarks/BuiltinBenchmark.java
+++ b/benchmarks/src/jmh/java/org/mozilla/javascript/benchmarks/BuiltinBenchmark.java
@@ -22,7 +22,7 @@ public class BuiltinBenchmark {
     @State(Scope.Thread)
     public static class AbstractClassState {
 
-        @Param({"Interpreter", "Compiler"})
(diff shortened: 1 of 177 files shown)


aardvark179:
So the test failure appears to be due to some refactoring in the V2 compiler _just_ pushing it over the edge in terms of stack usage. It passes on my local machine, but increasing the size of the expression a little bit pushes it over the edge. I'm going to try a couple of approaches to refactoring the compiler to avoid this, but from some experimentation it's clear we've always been sailing close to the wind in terms of stack depth and this test.

PR #2427, opened 2026-06-08
Fix cross realm array creation behaviour.

aardvark179:
Apparently this worked (possibly by accident) before we converted arrays away from `IdScriptableObject`. Now works for us internally _and_ passes 262 tests.


gbrail:
Looks great, thanks!

PR #2428, opened 2026-06-09
Refactor away from inheriting from `Icode`.

aardvark179:
Next bit of prep for #2426.

`CodeGenerator` and `Interpreter` both inherited from `Icode` as an old trick to avoid large numbers of static imports. As part of introducing a new interpreter we want to be able to inherit from an abstract class which itself should not inherit from `Icode`.

I chose to fix this by removing all the initial `Icode_` prefixes. This allows all the use sites to change simply from `Icode_THING` to `Icode.THING` and avoids either starred static imports or large numbers of static imports.


andreabergia (approved):

gbrail:
Looks great, thanks!

PR #2429, opened 2026-06-09
Fix enter and exit of frames under generators to be matched pairs.

### LINKED ISSUE #2099: Mismatched calls to `enterFrame` and `exitFrame` when using generators.
While testing refactoring of scope handling I've found that we may have mismatched calls to `enterFrame` and `exitFrame` while using generators. My stack handling in `Context` is like this:

```java
    private Scriptable scope;

    public static class StackHandler {

        private final Context cx;
        private final Scriptable oldScope;
        private final Scriptable currentScope;

        private StackHandler(Context cx, Scriptable oldScope, Scriptable newScope) {
            this.cx = cx;
            this.oldScope = oldScope;
            this.currentScope = newScope;
        }

        public void unwind() {
            assert cx.scope == currentScope;
            cx.scope = oldScope;
        }
    }

    public StackHandler pushScope(Scriptable newScope) {
        var res = new StackHandler(this, this.scope, newScope);
        this.scope = newScope;
        return res;
    }
```

and was done this way specifically to check pushing a scope and unwinding were always done in the right order.

`enterFrame` was altered to push the scope, and `exitFrame` to do the unwind. With this change continuation tests fail (which I kind of expected), but so do pretty much all tests which use generators, which suggests we've got a bug regarding those and enter / exit handling.

aardvark179:
This change fixed the calls of enter and exit frame with a generator to be more sensible. We enter the frame whether we enter or resume the generator, and we call exit frame whenever we yield or exit the generator.

This fixes #2099.

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/Interpreter.java
+++ b/rhino/src/main/java/org/mozilla/javascript/Interpreter.java
@@ -1659,6 +1659,7 @@ public final class Interpreter extends Icode implements Evaluator {
                         interpreterResultDbl = frame.resultDbl;
                         break StateLoop;
                     } else if (result instanceof YieldResult) {
+                        exitFrame(cx, frame, null);
                         return ((YieldResult) result).yielding;
                     } else if (result instanceof ThrowableResult) {
                         ThrowableResult tr = (ThrowableResult) result;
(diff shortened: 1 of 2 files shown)


gbrail:
This seems straightforward to me, thanks!

PR #2430, opened 2026-06-13
Move `stackTop` from state to frame.

aardvark179:
This change moves `stackTop` tracking out of the `InterpreterState` object and into the frame itself, removing the need for a saved stack top in the frame. This is the next bit of setup for #2426, I moved it


rbri (approved):

gbrail:
Thanks!

PR #2431, opened 2026-06-14
Implement WeakRef

gbrail:
Implement WeakRef. We have another PR for this but I wanted to get this done
one feature at a time.

When this is done I will use the work already done in 
https://github.com/mozilla/rhino/pull/2058
to implement FinalizationRegistry, although I would like to adapt that a bit since
I think that the cleanup callbacks need to run in a different way.


gbrail:
And to be clear about that previous PR, I plan to start with the commit by @anivar so that he also gets credit in the commit log.

andreabergia (commented):
Looks nice and simple!

gbrail:
Yes, there is not too much to this one. Thanks!

PR #2432, opened 2026-06-15
Complete FinalizationRegistry

gbrail:
This builds on the work already done in https://github.com/mozilla/rhino/pull/2058
and https://github.com/mozilla/rhino/pull/2431 to complete the implementation
of the FinalizationRegistry.

The all-important reference queue where finalization callbacks are called is
invoked as part of microtask processing by the Context. Applications that have
many Contexts in use at once will see finalizations happening in the one where
the registry was originally created. It uses PhantomReference and a 
reference queue to ensure that finalizations happen when the JVM needs them
to happen.

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/Context.java
+++ b/rhino/src/main/java/org/mozilla/javascript/Context.java
@@ -16,6 +16,8 @@ import java.io.PrintWriter;
 import java.io.Reader;
 import java.io.StringWriter;
 import java.io.Writer;
+import java.lang.ref.Reference;
+import java.lang.ref.ReferenceQueue;
 import java.lang.reflect.Constructor;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
@@ -508,7 +510,7 @@ public class Context implements Closeable {
         }
         if (cx.enterCount < 1) Kit.codeBug();
         if (--cx.enterCount == 0) {
-            releaseContext(cx);
+            cx.releaseContext();
         }
     }
 
@@ -518,15 +520,17 @@ public class Context implements Closeable {
         if (--enterCount == 0) {
             assert (currentContext.get() == this)
                     : "currentContext: " + currentContext.get() + ", this: " + this;
-            releaseContext(this);
+            releaseContext();
         }
     }
 
-    private static void releaseContext(Context cx) {
+    private void releaseContext() {
+        // Drain the reference queue to prevent memory leaks
+        cleanUpReferences();
         // do not use contextLocal.remove() here, as this might be much slower, when the same thread
         // creates a new context. See ContextThreadLocalBenchmark.
         currentContext.set(null);
-        cx.factory.onContextReleased(cx);
+        factory.onContextReleased(this);
     }
 
     /**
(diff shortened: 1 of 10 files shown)


aardvark179:
I'll review this properly later, I need to make sure I underhand the graph of references and what that implies about lifetimes. Specifically I'm worried that we're keeping `FinalisationRegistry` objects alive when they should not remain so.

gbrail:
It would be cool to do a formal analysis. On that end, it probably makes
sense, for example, for the item on the referencequeue to have a weak
reference to the finalization registry in case the microtasks don't run
after the registry goes out of scope.

gbrail:
I have a bunch more ideas on this still, working...

gbrail:
I think this is ready for others to look at now. By default, the API is enabled but objects get garbage collected normally. "cx.setFinalizationEnabled" must be called before anything will be actually enqueued to be finalized. Finalization callbacks happen in the "processMicrotasks" method (which is called at the exit of every top-level function and script) but if someone is using Rhino in an environment where that never gets called, registered objects will pile up and possibly cause memory problems unless the queue is periodically cleaned.

rbri:
@gbrail will have a look during the next days...

rhino/src/main/java/org/mozilla/javascript/Context.java
```diff
@@ -523,6 +525,8 @@ public void close() {
     }
 
     private static void releaseContext(Context cx) {
+        // Drain the reference queue to prevent memory leaks
+        cx.cleanUpReferences(cx.referenceQueue.poll());
```
rbri: this looks strange, calling a method on cx that gets something i pulled out from the very same cx - seems to externally do internal behaviour

tests/testsrc/test262.properties
```diff
@@ -745,7 +745,10 @@ built-ins/Error 3/58 (5.17%)
     cause_abrupt.js
     proto-from-ctor-realm.js
 
-~built-ins/FinalizationRegistry 47/47 (100.0%)
```
rbri, line 748: great progress

rhino/src/main/resources/org/mozilla/javascript/resources/Messages.properties
```diff
@@ -1105,5 +1105,25 @@ msg.dataview.offset.range =\
 msg.dataview.length.range =\
   DataView length is out of range
 
+# WeakRef
+msg.weakref.invalid.target =\
```
rbri: this is never used in the provided code, maybe mixed in from another pr?

rbri (approved):

rhino/src/main/java/org/mozilla/javascript/Context.java
gbrail: Thanks -- I'm not sure if you mean polling and then passing in the result or having a static method that takes "this" as a parameter? I actually changed both to simplify all this.

rhino/src/main/resources/org/mozilla/javascript/resources/Messages.properties
gbrail: That was an artifact of splitting one PR into two -- I'm fixing that.

gbrail:
I cleaned up the weird stuff in Context, in which I think I was over-optimizing the microtask loop. I'm pretty happy with this now, I'd like to merge it soon and move on to other things.

PR #2433, opened 2026-06-16
Move `CallFrame` out of interpreter and tidy up some field names.

aardvark179:
Next chunk of code tidy up and reorg. This is the next part of prep for #2426. The next two PRs will be the extraction of abstract classes for interpreter, call frame, and interpreter data. I'm throught about combining the `ACallFrame` extraction with this change, but it results in more churn than is nice if we don't introduce some of the types at the same time.


rbri (approved):
I like the new names

PR #2434, opened 2026-06-18
Fix some warnings and make them appear in Bazel

gbrail:
Make Errorprone checks more consistent between Gradle and Bazel builds

Clean up some errors in code and in Javadoc that we let creep in recently that
Errorprone has been trying to tell us about.

Making them more likely to show up in Bazel also makes it more likely that I 
will notice them during code review!


rbri (approved):
looks like some unused code is gone - i like

maybe you can have a look at this two jdoc warnings also...

Image

rbri:
> maybe you can have a look at this two jdoc warnings also...

Thanks

aardvark179:
This looks good. I had something similar lined up in my queue but hadn't got to it yet.

PR #2435, opened 2026-06-18
Extract abstract frame and interpreted function data ancestors.

aardvark179:
This moves a bunch of fields and functionality that will be common between the two interpreters to abstract parent classes. These classes are parameterised to ensure the correct relationship between the various elements. This is the next part of #2426.


gbrail:
This looks good, thanks!

PR #2437, opened 2026-06-21
Final prep for interpreter v2 - extract an abstract super class of interpreter

aardvark179:
This pulls a lot of functionality needed by both interpreters up into an abstract super class where it can be shared.


gbrail:
Looks good, thanks!

PR #2438, opened 2026-06-22
Introduce interpreter V2 infrastructure.

aardvark179:
This is the first chunk of Interpreter V2 work. It introduces the call frame, interpreter data, instructions, operand and other associated classes, but I cut the actual interpreter out of this PR to make it a more manageable size, that will come in the next chunk.

The `InstructionSimplification` and `KnownType` classes allow the compiler to replace general instructions such as `Add` with specialised versions when the typee are known.

I've extracted this from #2426 with some additional tidy up.

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/interpreterv2/CompilerData.java
+++ b/rhino/src/main/java/org/mozilla/javascript/interpreterv2/CompilerData.java
@@ -113,7 +113,7 @@ public class CompilerData<T extends ScriptOrFn<T>> extends ACompilerData<T, Comp
         return LineNumberTable.getFirstLineNumber(lineNumberTable);
     }
 
-    public short getPcFirstLineNumber() {
+    public int getPcFirstLineNumber() {
         return LineNumberTable.getPcFirstLineNumber(lineNumberTable);
     }
 
--- a/rhino/src/main/java/org/mozilla/javascript/interpreterv2/CompilerData.java
+++ b/rhino/src/main/java/org/mozilla/javascript/interpreterv2/CompilerData.java
@@ -1,0 +1,212 @@
+package org.mozilla.javascript.interpreterv2;
+
+import java.io.PrintStream;
+import java.util.List;
+import org.mozilla.javascript.ACompilerData;
+import org.mozilla.javascript.Context;
+import org.mozilla.javascript.JSCode;
+import org.mozilla.javascript.JSDescriptor;
+import org.mozilla.javascript.ScriptOrFn;
+import org.mozilla.javascript.VarScope;
+import org.mozilla.javascript.config.RhinoConfig;
+import org.mozilla.javascript.interpreterv2.instruction.Instruction;
+
+public class CompilerData<T extends ScriptOrFn<T>> extends ACompilerData<T, CompilerData<?>> {
+
+    static boolean shouldDumpInstructions = RhinoConfig.get("rhino.printICodeV2", false);
+
+    public static final int EXCEPTION_TRY_START_SLOT = 0;
+    public static final int EXCEPTION_TRY_END_SLOT = 1;
+    public static final int EXCEPTION_HANDLER_SLOT = 2;
+    public static final int EXCEPTION_TYPE_SLOT = 3;
+    public static final int EXCEPTION_LOCAL_SLOT = 4;
+    public static final int EXCEPTION_SCOPE_SLOT = 5;
+    // SLOT_SIZE: space for try start/end, handler, start, handler type,
+    //            exception local and scope local
+    public static final int EXCEPTION_SLOT_SIZE = 6;
+
+    public final Instruction[] instructions;
+
+    /**
+     * @see LineNumberTable
+     */
+    private final LineNumberTable lineNumberTable;
+
+    private CompilerData(Builder<T> b) {
+        super(b.maxVars, b.maxLocals, b.maxStack, b.maxFrameSize, b.exceptionTable);
+        this.instructions = b.instructions;
+        this.lineNumberTable = b.builtLineNumberTable;
+    }
+
+    public Instruction[] getInstructions() {
+        if (instructions == null) {
+            throw new IllegalStateException("Instructions not set");
+        }
+        return instructions;
+    }
+
+    public int[] getLineNumbers() {
+        return LineNumberTable.getLineNumbers(lineNumberTable);
+    }
+
+    public int getInstructionCount() {
+        if (instructions == null) {
+            return 0;
+        }
+        return instructions.length;
+    }
+
+    public void dumpInstructions(PrintStream out, JSDescriptor.Builder<?> desc) {
+        if (!shouldDumpInstructions) {
+            return;
+        }
+
+        // Function header with metadata
+        String functionName = desc.name != null ? desc.name : "<anonymous>";
+        out.printf(
+                "function %s [maxStack=%d, locals=%d, instructions=%d]:%n",
+                functionName, maxStack, maxLocals, getInstructionCount());
+        out.println("  line number table: " + getLineNumberTableForDebug());
+
+        // Instructions with virtual register numbering
+        if (instructions != null) {
+            for (int pc = 0; pc < instructions.length; pc++) {
+                out.printf("%%%-3d: %s%n", pc, instructions[pc].toDebugString());
+            }
+        }
+
+        // Exception handlers (if any)
+        if (hasExceptionHandlers()) {
+            out.println();
+            out.println("exception handlers:");
+            dumpExceptionHandlers(out);
+        }
+
+        out.println();
+        out.flush();
+    }
+
+    private boolean hasExceptionHandlers() {
+        return exceptionTable != null && exceptionTable.length > 0;
+    }
+
+    private void dumpExceptionHandlers(PrintStream out) {
+        if (exceptionTable == null) {
+            return;
+        }
+
+        for (int i = 0; i < exceptionTable.length; i += EXCEPTION_SLOT_SIZE) {
+            int tryStart = exceptionTable[i + EXCEPTION_TRY_START_SLOT];
+            int tryEnd = exceptionTable[i + EXCEPTION_TRY_END_SLOT];
+            int handler = exceptionTable[i + EXCEPTION_HANDLER_SLOT];
+            int type = exceptionTable[i + EXCEPTION_TYPE_SLOT];
+            int local = exceptionTable[i + EXCEPTION_LOCAL_SLOT];
+            int scope = exceptionTable[i + EXCEPTION_SCOPE_SLOT];
+
+            out.printf(
+                    "  try %%%-3d..%%%-3d -> %%%-3d (type=%d, local=%d, scope=%d)%n",
+                    tryStart, tryEnd, handler, type, local, scope);
+        }
+    }
+
+    public int getFirstLineNumber() {
+        return LineNumberTable.getFirstLineNumber(lineNumberTable);
+    }
+
+    public short getPcFirstLineNumber() {
+        return LineNumberTable.getPcFirstLineNumber(lineNumberTable);
+    }
+
+    /**
+     * Get the set of line numbers associated with a given PC.
+     *
+     * @param pc The program counter
+     * @return List of line numbers, or null if PC has no associated lines
+     */
+    public List<Integer> getLineSetFromPc(int pc) {
+        return LineNumberTable.getLineSetFromPc(lineNumberTable, pc);
+    }
+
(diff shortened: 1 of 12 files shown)


rhino/src/main/java/org/mozilla/javascript/interpreterv2/instruction/Instruction.java
```diff
@@ -0,0 +1,24 @@
+package org.mozilla.javascript.interpreterv2.instruction;
+
+import org.mozilla.javascript.CallFrameV2;
+import org.mozilla.javascript.Context;
+import org.mozilla.javascript.interpreterv2.InstructionSimplification;
+import org.mozilla.javascript.interpreterv2.KnownType;
+
+public abstract class Instruction {
```
gbrail, line 8: Is there a good reason to make this an abstract class rather than an interface with some default methods? I'm not sure there's a big difference but an interface might unlock some future flexibility.

rhino/src/main/java/org/mozilla/javascript/interpreterv2/operand/Operand.java
```diff
@@ -0,0 +1,97 @@
...
+
+import java.math.BigInteger;
+import org.mozilla.javascript.CallFrameV2;
+import org.mozilla.javascript.Context;
+import org.mozilla.javascript.ScriptRuntime;
+import org.mozilla.javascript.Undefined;
+import org.mozilla.javascript.interpreterv2.InstructionSimplification;
+import org.mozilla.javascript.interpreterv2.KnownType;
+
+public abstract class Operand {
```
gbrail, line 9: In general I'd love a little bit of Javadoc on some of these generic abstractions that are going to serve us for years in the future. (Which is funny because I've noticed that AI loves to put comments on everything when I DON'T want them!) If people are going to use them to implement new things in the future they or their AI helpers might need some guidance.

I get why this one is abstract and not an interface.

rhino/src/main/java/org/mozilla/javascript/interpreterv2/operand/Operand.java
```diff
@@ -0,0 +1,97 @@
...
+    public Number retrieveNumber(Context cx, CallFrameV2 frame) {
+        if (isDouble(frame)) {
+            return retrieveDouble(frame);
+        } else {
+            var obj = retrieve(cx, frame);
+            return ScriptRuntime.toNumeric(obj);
+        }
+    }
+
+    public boolean coerceToBoolean(Context cx, CallFrameV2 frame) {
```
gbrail, line 52: Sure this is great and efficient but I feel like between Boolean.valueOf and other Boolean class methods and ScriptRuntime.toNumber, is this really a unique thing we need here?

rhino/src/main/java/org/mozilla/javascript/interpreterv2/CompilerData.java
```diff
@@ -0,0 +1,220 @@
...
+    }
+
+    public void dumpInstructions(PrintStream out, JSDescriptor.Builder<?> desc) {
+        if (!shouldDumpInstructions) {
+            return;
+        }
+
+        // Function header with metadata
+        String functionName = desc.name != null ? desc.name : "<anonymous>";
+        out.printf(
```
gbrail, line 66: I think I see what's going on here. But in general do we want to introduce java.util.Logging now or stickwith our existing "System.out" logging approach? I feel like we should pick one or the other.

rhino/src/main/java/org/mozilla/javascript/interpreterv2/CompilerData.java
```diff
@@ -0,0 +1,220 @@
...
+import org.mozilla.javascript.VarScope;
+import org.mozilla.javascript.config.RhinoConfig;
+import org.mozilla.javascript.debug.DebuggableScript;
+import org.mozilla.javascript.interpreterv2.instruction.Instruction;
+
+public class CompilerData<T extends ScriptOrFn<T>> extends ACompilerData<T, CompilerData<?>> {
+
+    static boolean shouldDumpInstructions = RhinoConfig.get("rhino.printICodeV2", false);
+
+    public static final int EXCEPTION_TRY_START_SLOT = 0;
```
gbrail, line 18: Should this be an enum?

Also we've had trouble with a few of these numeric constants that grow forever and we have started to use a convention like:

int FOO = 1;
int BAR = FOO + 1;
int BAZ = BAR + 1;

The reason is that when we add constants and merge conflicting PRs all the time then the renumbering gets out of hand. Not a big deal here but something to keep in mind for future PRs.

rhino/src/main/java/org/mozilla/javascript/interpreterv2/LineNumberTable.java
```diff
@@ -0,0 +1,275 @@
...
+ * lines = [5, 6, 7, 8]
+ * </pre>
+ *
+ * <p>Lookup uses binary search on pcs[] with O(log n) performance. Memory usage is approximately 12
+ * bytes per PC + 4 bytes per line number (no object overhead).
+ */
+public class LineNumberTable {
+    private static final Logger LOG = Logger.getLogger(LineNumberTable.class.getName());
+
+    private final short[] pcs;
```
gbrail: Are we sure that each set of offsets will never grow beyond 64k? (And I can help find a few corpuses of JavaScript to test if you'd like.) Or what happens if they do? 

If this thing is one per interpreter I'd suggest that there's no advantage to using short here versus int.

rhino/src/main/java/org/mozilla/javascript/interpreterv2/LineNumberTable.java
```diff
@@ -0,0 +1,275 @@
...
+
+            int totalLines = 0;
+            for (Set<Integer> lineSet : linesList) {
+                totalLines += lineSet.size();
+            }
+            short[] lines = new short[totalLines];
+
+            int lineIdx = 0;
+            for (int i = 0; i < numPcs; i++) {
+                pcs[i] = (short) (pcsList.get(i) & 0xFFFF);
```
gbrail, line 392: Are we trying to avoid overflow here or something with all the bitmasking?

rhino/src/main/java/org/mozilla/javascript/interpreterv2/LineNumberTable.java
```diff
@@ -0,0 +1,275 @@
+package org.mozilla.javascript.interpreterv2;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Set;
+import java.util.logging.Logger;
```
gbrail: This will be the first use of java.util.logging in Rhino. Is it worth it? Or should we stick with System.out like in other parts of the codebase? 

Either way, I don't recall in this PR how we're enabling or disabling logging but elsewhere we have used out existing "config" mechanism for this and I think we should use that here too.

gbrail (commented):
Thanks -- I have a bunch of questions but most are stylistic. No way to tell whether this is correct yet ;-)

I COULD suggest a few of the abstractions here like the line number table are great candidates for unit tests and if I were working on this I would insist on it (and let my AI get me started)

Finally, I think we should either bite the bullet and start using something like SLF4J in here like every other Java project on earth, or stick with "System.out" logging like other parts of Rhino.

aardvark179:
There should definitely be a unit test for the line number table, and there's one on our fork, but it looks like it's way later on my branch. Let me see about pulling that back to here.

Please do bring up any stylistic stuff. I've been doing some tidy as I've gone, but there's definitely room for more cleanup. Some things I want to tidy up, but want to wait to do properly, e.g. exception handling can definitely be unified and made simpler, but I'd like to do that after we've got v2 in because that will affect v1, v2, and class compilation.

rhino/src/main/java/org/mozilla/javascript/interpreterv2/instruction/Instruction.java
aardvark179: Yes. Invoke virtual has a measurable performance difference from invoke interface because even if the receiver is not known the offset of the method is. This makes for an almost 10% improvement in some benchmarks.

rhino/src/main/java/org/mozilla/javascript/interpreterv2/operand/Operand.java
aardvark179: Totally reasonable, I'll add some java docs.

rhino/src/main/java/org/mozilla/javascript/interpreterv2/CompilerData.java
aardvark179: I agree that this should simply be the same style we use everywhere else. I think this was done here to capture stuff to our logs during development.

rhino/src/main/java/org/mozilla/javascript/interpreterv2/CompilerData.java
aardvark179: There's quite a few things I'd like to turn into enums. I'd thought I'd do this as part of an overall change to clean up our exception handling, but since this is really a duplicate of similar constants in `Interpreter`, it might be worth trying to do some of that now.

aardvark179:
I've done some more clean up on this, and on #2426, still need to put some more tests on there at the right points.

aardvark179:
@gbrail I've sent you a message on Slack.

aardvark179:
I've pushed a large line number table implementation that can be used if we see anything that doesn't fit in a short during the build process. @andreabergia I'd appreciate your review on this as well.

rhino/src/main/java/org/mozilla/javascript/interpreterv2/operand/Operand.java
aardvark179: I think I agree that this shouldn't have a duplicate implementation. It's worth handling the double case to avoid a round of boxing, but after that it should fall back to `ScriptRuntime`.

The specialised operands for constants etc. can reasonable override this, but that doesn't appear to have been done. Need to do a quick comparison benchmark to see if that has a useful effect?

rhino/src/main/java/org/mozilla/javascript/interpreterv2/CompilerData.java
aardvark179: Okay, specialising this in the operands makes a small positive difference. I'll get those changes rolled into the appropriate commits.

aardvark179: Done.

rhino/src/main/java/org/mozilla/javascript/interpreterv2/LineNumberTable.java
aardvark179: Large version and test introduced.

rhino/src/main/java/org/mozilla/javascript/interpreterv2/LineNumberTable.java
aardvark179: Removed,

rhino/src/main/java/org/mozilla/javascript/interpreterv2/LineNumberTable.java
```diff
@@ -0,0 +1,489 @@
...
+ * pcs = [10, 15]
+ * lineOffsets = [0, 3]
+ * lineCounts = [3, 1]
+ * lines = [5, 6, 7, 8]
+ * </pre>
+ *
+ * <p>Lookup uses binary search on pcs[] with O(log n) performance. Memory usage is approximately 12
+ * bytes per PC + 4 bytes per line number (no object overhead).
+ */
+public abstract class LineNumberTable {
```
andreabergia, line 33: All of this looks reasonable to me. There's a bit of duplication, but I can't think of any way to solve this better sadly.

rhino/src/test/java/org/mozilla/javascript/interpreterv2/LineNumberTableTest.java
```diff
@@ -0,0 +1,262 @@
+package org.mozilla.javascript.interpreterv2;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+import java.util.List;
+import org.junit.jupiter.api.Test;
+
+class LineNumberTableTest {
```
andreabergia, line 8: This needs to be evolved to support both the short and int line number table versions.

gbrail:
This looks OK to me now -- I'd still like to make sure we're clear going forward on enums versus integers and on Javadoc, but I'm also OK doing that later.

gbrail:
Let's merge this now and keep going. Thanks!

PR #2439, opened 2026-06-30
Introduce main V2 interpreter loop.

aardvark179:
This PR is the next part #2426 and introduces the main interpreter v2 interpreter loop, error handling code, and utilitiy functions used by multiple instructions. This will be followed by PRs for the operands and instructions, and finally by the compiler itself.

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/InterpreterV2.java
+++ b/rhino/src/main/java/org/mozilla/javascript/InterpreterV2.java
@@ -1,0 +1,598 @@
+package org.mozilla.javascript;
+
+import static org.mozilla.javascript.UniqueTag.DOUBLE_MARK;
+
+import java.math.BigInteger;
+import java.util.List;
+import java.util.Objects;
+import org.mozilla.javascript.ast.ScriptNode;
+import org.mozilla.javascript.debug.DebuggableScript;
+import org.mozilla.javascript.interpreterv2.CompilerData;
+import org.mozilla.javascript.interpreterv2.GeneratorState;
+import org.mozilla.javascript.interpreterv2.instruction.JumpInstruction;
+import org.mozilla.javascript.interpreterv2.operand.Operand;
+
+public class InterpreterV2 extends AInterpreter<CallFrameV2, CompilerData<?>> {
+    private static final int EX_CATCH_STATE = 2; // Can execute JS catch
+    private static final int EX_FINALLY_STATE = 1; // Can execute JS finally
+    private static final int EX_NO_JS_STATE = 0; // Terminate JS execution
+
+    public static final int INVOCATION_COST = 100;
+    // arbitrary exception cost for instruction counting
+    private static final int EXCEPTION_COST = 100;
+
+    static {
+        RhinoException.registerInterpreterMethod(InterpreterV2.class, "interpretInner");
+    }
+
+    public InterpreterV2() {}
+
+    public static Object interpret(Context cx, CallFrameV2 frame, Object throwable) {
+        var oldFrame = cx.lastInterpreterFrame;
+        try {
+            return interpretInner(cx, frame, throwable);
+        } finally {
+            cx.lastInterpreterFrame = oldFrame;
+        }
+    }
+
+    private static Object interpretInner(Context cx, CallFrameV2 frame, Object throwable) {
+        cx.lastInterpreterFrame = frame;
+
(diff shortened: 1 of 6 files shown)


rhino/src/main/java/org/mozilla/javascript/interpreterv2/instruction/JumpInstruction.java
```diff
@@ -0,0 +1,30 @@
+package org.mozilla.javascript.interpreterv2.instruction;
```
rbri, line 1: do we have to always add the common header?

rhino/src/main/java/org/mozilla/javascript/interpreterv2/instruction/JumpInstruction.java
```diff
@@ -0,0 +1,30 @@
...
+
+import java.util.Set;
+import org.mozilla.javascript.interpreterv2.InstructionFormatter;
+
+public abstract class JumpInstruction extends Instruction {
+    protected int offset;
```
rbri: does it make sense to have a protected field and a getter/setter for this internal class

aardvark179: Good point. I made it private and have changed the upcoming commits to remove direct references.

rhino/src/main/java/org/mozilla/javascript/interpreterv2/instruction/JumpInstruction.java
aardvark179: Good question. If we should add it to every file then we've done a bad job of that for a while, even `Slot.java` doesn't have a licence header.

gbrail: We actually are, I got lazy about that, I think that there must be a tool that can fix this for us.

gbrail (commented):
I may be an old fuddy-duddy but I see lots of opportunities for unit tests in here, simple stuff like the "formatter" but even the actual bits of the interpreter too. I will not lose sleep if they are written by AI as long as they have decent coverage. It would make me more confident if I were writing the code myself (or with a tool) to have those additional tests as well.

rhino/src/main/java/org/mozilla/javascript/interpreterv2/instruction/JumpInstruction.java
aardvark179: I believe spotless can enforce this, and I can certainly produce a PR to fill in the missing ones.

aardvark179:
That's reasonable, I'll see what I can do.

aardvark179:
I've added a unit test for instruction formatting. I looked at adding tests for some of the other bits of infrastructure around instruction counting and error handling, but I had to add a few too many of the real instructions to make it worthwhile until the rest of the instructions are in.

I do agree that having some good unit tests round both interpreters infrastructure would be good, and I'll try and generate some to add later in this series.

rhino/src/main/java/org/mozilla/javascript/InterpreterV2.java
```diff
@@ -0,0 +1,610 @@
...
+            if (frame.frozen) {
+                frame = frame.cloneFrozen();
+            }
+            // TODO(jimmy)
+            // Need to figure out when this should happen
+            // or if the way we do things makes this unnecessary
+            // setCallResult(frame, interpreterResult, interpreterResultDbl);
+        }
+
+        return frame.result == DOUBLE_MARK
```
gbrail, line 136: My AI friend is worried about this return statement, because if we changed frame to frame.parentFrame above, we're now returning the result of the parent invocation, not the current invocation, which is certainly not what we want. I'm sure that this will turn up later, so it's up to you if you'd like to fix it now or merge and fix it later.!

aardvark179: So, interpreter V2 does not currently support multiple frames in a single interpreter call, so a non null `parentFrame` should be an illegal state. I've changed cases where we're looking for parent frames to throw exceptions.

I also realised that I had completely refactored out uses of the separate V2 `ContinuationJump` class, so I've removed it completely.

gbrail:
OK, this one is good to go, thanks for all the work so far!

PR #2445, opened 2026-07-13
Reduce heap space used by built in slots.

aardvark179:
In production we're seen some cases where we have an extremely high number of function objects. The move away from `IdScriptableObject`s causes a significant increase in memory footprint as every function normally has 3 or 4 slots ('length', 'name', 'arity', and 'arguments'). The `BuiltInSlot` objects each had a shallow size of 56, and had to be help in (in our case) a `ThreadSafeEmbeddedSlotMap` which added the cost of that object, an array, and a stamped lock to the cost.

This PR separates our slot hierarchy into standard slots and compact slots, and introduces small immutable slot maps that can be used for functions. Compact slots have descriptors which can be shared between many slots. We change all the uses of built in slots to use these descriptors, and the creation of the normal properties in `BaseFunction` to create a small immutable slot map.

This gets the overhead for function properties down to 160 bytes which is a significant improvement. I believe we can improve this in future work by only generating slot objects when required by our APIs and get the footprint down to at most 32 bytes per function, and extend this immutable slot map concept to the majority of built in prototypes as well.

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/BaseFunction.java
+++ b/rhino/src/main/java/org/mozilla/javascript/BaseFunction.java
@@ -167,19 +167,21 @@ public class BaseFunction extends ScriptableObject implements Function {
     }
 
     protected void createProperties() {
-        ScriptableObject.defineBuiltInProperty(this, DONTENUM | READONLY, LENGTH_DESCRIPTOR);
-        ScriptableObject.defineBuiltInProperty(this, DONTENUM | READONLY, NAME_DESCRIPTOR);
+        ASlot<Scriptable> lengthSlot = LENGTH_DESCRIPTOR.createSlot(this, DONTENUM | READONLY);
+        ASlot<Scriptable> nameSlot = NAME_DESCRIPTOR.createSlot(this, DONTENUM | READONLY);
+        ASlot<Scriptable> aritySlot = null;
+        ASlot<Scriptable> argsSlot = null;
 
         Context cx = Context.getCurrentContext();
         if (cx == null || !cx.isStrictMode()) {
-            ScriptableObject.defineBuiltInProperty(
-                    this, PERMANENT | DONTENUM | READONLY, ARITY_DESCRIPTOR);
+            aritySlot = ARITY_DESCRIPTOR.createSlot(this, PERMANENT | DONTENUM | READONLY);
 
             if (cx == null || cx.getLanguageVersion() < Context.VERSION_ES6) {
-                ScriptableObject.defineBuiltInProperty(
-                        this, PERMANENT | DONTENUM, ARGUMENTS_DESCRIPTOR);
+                argsSlot = ARGUMENTS_DESCRIPTOR.createSlot(this, PERMANENT | DONTENUM);
             }
         }
+
+        setMap(new ImmutableSmallSlotMap<>(lengthSlot, nameSlot, aritySlot, argsSlot));
     }
 
     private static Object lengthGetter(BaseFunction function, Scriptable start) {
--- a/rhino/src/main/java/org/mozilla/javascript/BaseFunction.java
+++ b/rhino/src/main/java/org/mozilla/javascript/BaseFunction.java
@@ -51,24 +51,19 @@ public class BaseFunction extends ScriptableObject implements Function {
     private static final BuiltInSlot.Descriptor<BaseFunction> ARGUMENTS_DESCRIPTOR =
             new BuiltInSlot.Descriptor<>(
                     "arguments", BaseFunction::argumentsGetter, BaseFunction::argumentsSetter);
-    private static final SlotMapDescriptor<Scriptable, BaseFunction> BASIC_MAP =
-            new SlotMapDescriptor.Builder<Scriptable, BaseFunction>()
+    private static final SlotMapDescriptor<BaseFunction> BASIC_MAP =
+            new SlotMapDescriptor.Builder<BaseFunction>()
                     .withSlot(LENGTH_DESCRIPTOR, DONTENUM | READONLY)
                     .withSlot(NAME_DESCRIPTOR, DONTENUM | READONLY)
                     .build();
 
-    private static final SlotMapDescriptor<Scriptable, BaseFunction> ARITY_MAP =
-            new SlotMapDescriptor.Builder<Scriptable, BaseFunction>()
-                    .withSlot(LENGTH_DESCRIPTOR, DONTENUM | READONLY)
-                    .withSlot(NAME_DESCRIPTOR, DONTENUM | READONLY)
+    private static final SlotMapDescriptor<BaseFunction> ARITY_MAP =
+            SlotMapDescriptor.Builder.extending(BASIC_MAP)
                     .withSlot(ARITY_DESCRIPTOR, PERMANENT | DONTENUM | READONLY)
                     .build();
 
-    private static final SlotMapDescriptor<Scriptable, BaseFunction> ARGUMENTS_MAP =
-            new SlotMapDescriptor.Builder<Scriptable, BaseFunction>()
-                    .withSlot(LENGTH_DESCRIPTOR, DONTENUM | READONLY)
-                    .withSlot(NAME_DESCRIPTOR, DONTENUM | READONLY)
-                    .withSlot(ARITY_DESCRIPTOR, PERMANENT | DONTENUM | READONLY)
+    private static final SlotMapDescriptor<BaseFunction> ARGUMENTS_MAP =
+            SlotMapDescriptor.Builder.extending(ARITY_MAP)
                     .withSlot(ARGUMENTS_DESCRIPTOR, PERMANENT | DONTENUM)
                     .build();
 
@@ -188,7 +183,7 @@ public class BaseFunction extends ScriptableObject implements Function {
     }
 
     protected void createProperties() {
-        SlotMapDescriptor<Scriptable, BaseFunction> desc;
+        SlotMapDescriptor<BaseFunction> desc;
 
         Context cx = Context.getCurrentContext();
         if (cx == null || !cx.isStrictMode()) {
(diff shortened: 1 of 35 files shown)


rhino/src/main/java/org/mozilla/javascript/BaseFunction.java
```diff
@@ -149,29 +167,21 @@ public BaseFunction(VarScope scope, Scriptable prototype) {
...
     protected void createProperties() {
-        ScriptableObject.defineBuiltInProperty(
-                this, "length", DONTENUM | READONLY, BaseFunction::lengthGetter);
-        ScriptableObject.defineBuiltInProperty(
-                this,
-                "name",
-                DONTENUM | READONLY,
-                BaseFunction::nameGetter,
-                BaseFunction::nameSetter);
+        Slot<Scriptable> lengthSlot = LENGTH_DESCRIPTOR.createSlot(this, DONTENUM | READONLY);
```
gbrail: Does it make sense to have the flags (DONTENUM | READONLY etc) in the descriptor rather than in the "createSlot" method itself? I imagine in most cases where we do this we will never change those flags, so you have less code.

rhino/src/main/java/org/mozilla/javascript/BaseFunction.java
```diff
@@ -149,29 +167,21 @@ public BaseFunction(VarScope scope, Scriptable prototype) {
...
-                        this,
-                        "arguments",
-                        PERMANENT | DONTENUM,
-                        BaseFunction::argumentsGetter,
-                        BaseFunction::argumentsSetter);
+                argsSlot = ARGUMENTS_DESCRIPTOR.createSlot(this, PERMANENT | DONTENUM);
             }
         }
+
+        setMap(new ImmutableSmallSlotMap<>(lengthSlot, nameSlot, aritySlot, argsSlot));
```
gbrail: What are the implications of making this an "immutable" map? Unless I missed something it sounds like it's kind of like making instances of this class and it's subclasses non-extendable, although by throwing a Java rather than a JavaScript exception. JavaScript code loves to muck with the properties of just about everything -- is that really going to work out?

rhino/src/main/java/org/mozilla/javascript/BaseFunction.java
```diff
@@ -149,29 +167,21 @@ public BaseFunction(VarScope scope, Scriptable prototype) {
...
-        ScriptableObject.defineBuiltInProperty(
-                this,
-                "name",
-                DONTENUM | READONLY,
-                BaseFunction::nameGetter,
-                BaseFunction::nameSetter);
+        Slot<Scriptable> lengthSlot = LENGTH_DESCRIPTOR.createSlot(this, DONTENUM | READONLY);
+        Slot<Scriptable> nameSlot = NAME_DESCRIPTOR.createSlot(this, DONTENUM | READONLY);
+        Slot<Scriptable> aritySlot = null;
+        Slot<Scriptable> argsSlot = null;
```
gbrail: Is there a way to do this with some sort of a Builder pattern instead of creating Slot instances directly? I tend to think of Slot as an implementation detail of ScriptableObject although maybe we're past that point already!

gbrail (commented):
Just a few small questions, I see where this is going -- thanks!

rhino/src/main/java/org/mozilla/javascript/BaseFunction.java
aardvark179: It's a good question. Flags are mutable without replacing a slot in the same way a `value` can be, which is why I've kept them out of the descriptor half for now. I think it's going to take a few rounds and some more experimentation to work out what the right split is going to end up being.

rhino/src/main/java/org/mozilla/javascript/BaseFunction.java
aardvark179: Making this immutable means that any change to the slots causes the whole slot map to be replaced. This allows for thread safe versions to skip having a lock as they can do a CAS of the map.

rhino/src/main/java/org/mozilla/javascript/BaseFunction.java
aardvark179: I think a builder is a good idea, but I'm tempted to go one stage further and say we should build a map descriptor and make the map from that. I think it could then be extended to be used by class descriptors.

aardvark179:
Okay, created a descriptor API and a builder for that. Although attributes aren't yet part of slot descriptors themselves I have put them into the map descriptor. I'll experiment with how this looks if I try expanding it for our class descriptors as that will force passing values in as well just the owner.

gbrail:
I'm OK with this but we have a real failure on the Android tests from a new method unfortunately.

rhino/src/main/java/org/mozilla/javascript/SlotMapDescriptor.java
```diff
@@ -0,0 +1,61 @@
...
+
+    public static class Builder<O extends ScriptableObject> {
+        List<CompactSlot.Descriptor<?, Scriptable, O>> slots = new ArrayList<>();
+        List<Integer> attributes = new ArrayList<>();
+
+        public Builder() {}
+
+        private Builder(SlotMapDescriptor<O> old) {
+            slots = new ArrayList<>(old.slots);
+            attributes = new ArrayList<>(Arrays.stream(old.attributes).mapToObj(v -> v).toList());
```
tonygermano: To mitigate Android test failure.
```suggestion
            attributes = new ArrayList<>(Arrays.stream(old.attributes).mapToObj(v -> v).collect(Collectors.toList()));
```

rhino/src/main/java/org/mozilla/javascript/SlotMapDescriptor.java
```diff
@@ -0,0 +1,61 @@
...
+        }
+
+        public static <O extends ScriptableObject> Builder<O> extending(
+                SlotMapDescriptor<O> start) {
+            return new Builder<>(start);
+        }
+
+        public Builder<O> withSlot(
+                CompactSlot.Descriptor<?, Scriptable, O> descriptor, int attributes) {
+            if (slots.size() > 4) {
```
tonygermano: If it's > 4, does that mean you've already added a 5th slot?
```suggestion
            if (slots.size() >= 4) {
```

tonygermano (commented):
Not a full review, but a couple of questions/suggestions.

rhino/src/main/java/org/mozilla/javascript/SlotMapDescriptor.java
aardvark179: You're right, now fixed.

rhino/src/main/java/org/mozilla/javascript/SlotMapDescriptor.java
aardvark179: Yeah, now resolved.

rhino/src/main/java/org/mozilla/javascript/StandardSlot.java
```diff
@@ -0,0 +1,125 @@
...
+ * A Slot is the base class for all properties stored in the ScriptableObject class. There are a
+ * number of different types of slots. This base class represents an "ordinary" property such as a
+ * primitive type or another object. Separate classes are used to represent properties that have
+ * various types of getter and setter methods.
+ */
+public class StandardSlot<T extends PropHolder<T>> extends Slot<T> {
+    @Serial private static final long serialVersionUID = -6090581677123995491L;
+    private final Object name;
+    private int indexOrHash;
+    private short attributes;
```
gbrail, line 18: I didn't find this but an AI did, but "attributes" shadows the field with the same name in Slot.

rhino/src/main/java/org/mozilla/javascript/StandardSlot.java
```diff
@@ -0,0 +1,125 @@
...
+
+    StandardSlot(Object name, int index, int attributes) {
+        super(attributes);
```
gbrail: This is setting attributes on the parent

gbrail (commented):
One thing that popped up here is that StandardSlot has an "attributes" field that doesn't seem to be used, but if it were it would not be consistent when we're using "attributes" on the parent Slot object and when on StandardSlot.

rhino/src/main/java/org/mozilla/javascript/StandardSlot.java
aardvark179: Ha! I spotted the same problem and fixed it a few commits later on my prototype branch. Let me see if I can pull the fix in here without dragging too much else in with it.

rhino/src/main/java/org/mozilla/javascript/ImmutableSmallSlotMap.java
```diff
@@ -0,0 +1,119 @@
...
+            CompoundOperationMap<T> mutableMap,
+            Object key,
+            int index,
+            SlotComputer<S, T> compute) {
+        return newMap(owner, 0).compute(owner, mutableMap, key, index, compute);
+    }
+
+    @Override
+    public boolean isEmpty() {
+        return false;
```
gbrail: I wish I could take credit as a human for spotting this, but we don't want this to always return false, do we? I imagine returning size() == 0 would make more sense.

gbrail (commented):
There's a really minor problem below -- otherwise I think that this is getting ready to go.

rhino/src/main/java/org/mozilla/javascript/ImmutableSmallSlotMap.java
aardvark179: We only ever create these maps with at least one entry, which is why this always returns false, but I allowed for all the slots being null in other places so I should do that here.

gbrail:
I think this is good now -- thanks!

rbri:
for the records: works fine with HtmlUnit

rhino/src/main/java/org/mozilla/javascript/BaseFunction.java
aardvark179: I have been doing some more prototyping towards compact maps and I am having to move attributes to the descriptors. I had thought they could go into a `short[]` alongside an `Object[]` for values. The problem is that slots don't always need a single value. Accessors have 3 (okay, wasteful but fine) but built ins just require the owner, don't require value storage, and so don't leave
Me with any matching attributes storage.

Oh well, time to move things around on the API and see how things shake out.

PR #2446, opened 2026-07-14
interpreterv2 operands.

aardvark179:
This PR is the next part #2426 and introduces the operands used by the interpreter V2 instructions.

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/interpreterv2/operand/GetVarOperand.java
+++ b/rhino/src/main/java/org/mozilla/javascript/interpreterv2/operand/GetVarOperand.java
@@ -1,0 +1,57 @@
+package org.mozilla.javascript.interpreterv2.operand;
+
+import org.mozilla.javascript.CallFrameV2;
+import org.mozilla.javascript.Context;
+
+public final class GetVarOperand extends Operand {
+    private final int index;
+
+    private static final GetVarOperand[] CACHE = new GetVarOperand[16];
+
+    static {
+        for (int i = 0; i < CACHE.length; i++) {
+            CACHE[i] = new GetVarOperand(i);
+        }
+    }
+
+    public static GetVarOperand createOperand(int index) {
+        if (index >= 0 && index < CACHE.length) {
+            return CACHE[index];
+        }
+        return new GetVarOperand(index);
+    }
+
+    private GetVarOperand(int index) {
+        this.index = index;
+    }
+
+    @Override
+    public Object retrieve(Context cx, CallFrameV2 frame) {
+        return frame.getVar(index);
+    }
+
+    @Override
+    public Object retrieveAndWrap(Context cx, CallFrameV2 frame) {
+        return frame.getVarAndWrap(index);
+    }
+
+    @Override
+    public double retrieveDouble(CallFrameV2 frame) {
+        return frame.getVarDouble(index);
+    }
+
+    @Override
+    public boolean isDouble(CallFrameV2 frame) {
+        return frame.isVarDouble(index);
+    }
+
+    @Override
+    public Object viewValue(Context cx, CallFrameV2 frame, int offset) {
+        return frame.getVarAndWrap(index);
+    }
+
+    @Override
+    public void appendDebugString(StringBuilder sb) {
+        sb.append("var.").append(index);
+    }
+}
--- a/rhino/src/main/java/org/mozilla/javascript/interpreterv2/operand/IntOperand.java
+++ b/rhino/src/main/java/org/mozilla/javascript/interpreterv2/operand/IntOperand.java
@@ -1,0 +1,50 @@
+package org.mozilla.javascript.interpreterv2.operand;
+
+import org.mozilla.javascript.CallFrameV2;
+import org.mozilla.javascript.Context;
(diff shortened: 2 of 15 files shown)


aardvark179:
@gbrail ir @rbri, could you kick the failed job off again? I think it's a transient infrastructure failure.

gbrail:
Just stylistic questions on this: 

First, is there a good reason for all these classes to be "final"? I know that sometimes it affects performance but generally we don't make classes final unless we're really sure we need that and we're really sure that it helps with performance. (I do love to make private variables final but that's a different use of final IMO.)

Second, I see that some of the methods, like "retrieve," return boxed types like Double and Integer and other methods return double and int. Obviously we want to avoid boxing and unboxing until we no choice, and Java is pretty good these days at autoboxing in the compiler. Do we need it there? (Or is it a way to specify an "optional" result?")

rhino/src/main/java/org/mozilla/javascript/interpreterv2/operand/IntOperand.java
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    @Override
+    public double retrieveDouble(CallFrameV2 frame) {
+        return value;
+    }
+
+    @Override
+    public boolean isDouble(CallFrameV2 frame) {
+        return value == -1.0;
```
gbrail: I don't understand what this is supposed to do. Is an int ever a double?

rhino/src/main/java/org/mozilla/javascript/interpreterv2/operand/ShortOperand.java
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    @Override
+    public double retrieveDouble(CallFrameV2 frame) {
+        return value;
+    }
+
+    @Override
+    public boolean isDouble(CallFrameV2 frame) {
+        return value == -1.0;
```
gbrail: Same question here as for IntOperand.

rhino/src/main/java/org/mozilla/javascript/interpreterv2/operand/GetVarOperand.java
```diff
@@ -0,0 +1,85 @@
...
+            return false;
+        }
+    }
+
+    @Override
+    public Object viewValue(Context cx, CallFrameV2 frame, int offset) {
+        if (!frame.useActivation) {
+            return frame.getVarAndWrap(index);
+        } else {
+            String name = frame.fnOrScript.getDescriptor().getParamOrVarName(offset);
```
gbrail: Other methods here use "index" as the parameter not "offset". Is this right? It confused by AI...

gbrail (commented):
Just a few more small things, thanks for reading!

rhino/src/main/java/org/mozilla/javascript/interpreterv2/operand/IntOperand.java
aardvark179: I don't understand this either. Yes, a double value (not a `Double`) can be compared to an int and will cause type promotion of the `int` side, but I'm not sure what it's intended to achieve here. @camnwalter, do you know what was intended? If you can't remember I'll change this to `false`.

camnwalter: I believe this is from something we check internally in our fork (for reasons I don't know), and when making interpreter v2, we wanted to copy the existing interpreter as closely as possible. I'm fairly sure we don't use it anywhere in this repo, so it should be fine to change it.

rhino/src/main/java/org/mozilla/javascript/interpreterv2/operand/GetVarOperand.java
aardvark179: You're right. That was code that was only taken in the debugger when we artificially added activation frames. Since we no longer do that I've removed those paths entirely.

rhino/src/main/java/org/mozilla/javascript/interpreterv2/operand/IntOperand.java
aardvark179: Okay, I've changed this to return `false` always.

aardvark179:
> Just stylistic questions on this:
> 
> First, is there a good reason for all these classes to be "final"? I know that sometimes it affects performance but generally we don't make classes final unless we're really sure we need that and we're really sure that it helps with performance. (I do love to make private variables final but that's a different use of final IMO.)

I think this was done because it sometimes helps the JIT, and I think `Operand` were originally an interface where it helped more.

> Second, I see that some of the methods, like "retrieve," return boxed types like Double and Integer and other methods return double and int. Obviously we want to avoid boxing and unboxing until we no choice, and Java is pretty good these days at autoboxing in the compiler. Do we need it there? (Or is it a way to specify an "optional" result?")

So, like the original interpreter we have a stack/local variable space that holds either `double` or object, and so operands have been designed to fit that model. I'm not convinced it's a great model—we'd likely do better with a type lattice that included ints as they are so commonly used for array offsets and similar—but changing it is a job for another day.

gbrail:
OK -- let's keep moving forward on this. Thanks!

PR #2447, opened 2026-07-16
Handle strict mode better in direct eval.

aardvark179:
I'm trying to reduce the differences between our fork and here to make merging down easier. This one came up because it was originally implemented as a special case in the interpreter, but seemed better suited to being in `evalSpecial` itself.


rbri:
@aardvark179 spotlessApply

rbri (approved):

rbri:
smoke tested with HtmlUnit - fine ❤️

gbrail:
Looks good, thanks!

PR #2448, opened 2026-07-17
Reduce metaspace usage by moving to method handles instead of auxiliary classes.

aardvark179:
When we introduced function descriptors in #2091 we generated additional auxiliary classes to hook up the underlying functions as JSCode objects. This was a relatively simple solution that could work for all supported platforms, but it did produce a relatively large number of additional classes. I have recently had reason to examine the total metaspace usage of Rhino, and noticed that although the solution introduced in #2091 did reduce metaspace usage a reasonable amount on large compiled scripts (mostly due to the removal of the large utility methods), we were still using a significant amount due to the additional classes.

Before #2091 our metaspace usage looked like this:
```
  80: CLD 0x000000087b865180: instance of org.mozilla.javascript.DefiningClassLoader
      Loaded classes:
         1:    org.mozilla.javascript.gen._Users_duncan_macgregor_git_mozilla_rhino_moment_js_2
      -total-: 1 class
  Non-Class:   53 chunks,      2.34 MB capacity,    2.34 MB (100%) committed,     2.33 MB (>99%) used,    14.98 KB ( <1%) free,     8 bytes ( <1%) waste , deallocated: 27 blocks with 5.02 KB
      Class:    1 chunk,       2.00 KB capacity,    2.00 KB (100%) committed,     1.99 KB (>99%) used,     8 bytes ( <1%) free,     0 bytes (  0%) waste , deallocated: 0 blocks with 0 bytes
       Both:   54 chunks,      2.35 MB capacity,    2.35 MB (100%) committed,     2.33 MB (>99%) used,    14.99 KB ( <1%) free,     8 bytes ( <1%) waste , deallocated: 27 blocks with 5.02 KB
```
and afterward it looked like this:
```
  72: CLD 0x000000081a65e760: instance of org.mozilla.javascript.DefiningClassLoader
      Loaded classes:
         1:    org.mozilla.javascript.gen._Users_duncan_macgregor_git_mozilla_rhino_moment_js_2ojsc414
         2:    org.mozilla.javascript.gen._Users_duncan_macgregor_git_mozilla_rhino_moment_js_2ojsc413
       ...
       415:    org.mozilla.javascript.gen._Users_duncan_macgregor_git_mozilla_rhino_moment_js_2ojsc0
       416:    org.mozilla.javascript.gen._Users_duncan_macgregor_git_mozilla_rhino_moment_js_2
      -total-: 416 classes
  Non-Class:   28 chunks,      1.06 MB capacity,    1.06 MB (100%) committed,     1.06 MB (>99%) used,     5.79 KB ( <1%) free,     0 bytes (  0%) waste , deallocated: 5 blocks with 14.27 KB
      Class:   11 chunks,    263.00 KB capacity,  263.00 KB (100%) committed,   251.78 KB ( 96%) used,    11.22 KB (  4%) free,     0 bytes (  0%) waste , deallocated: 0 blocks with 0 bytes
       Both:   39 chunks,      1.32 MB capacity,    1.32 MB (100%) committed,     1.30 MB ( 99%) used,    17.01 KB (  1%) free,     0 bytes (  0%) waste , deallocated: 5 blocks with 14.27 KB
```
That represents an excellent reduction, but moving to method handles gets us to
```
  79: CLD 0x0000000abb577340: instance of org.mozilla.javascript.DefiningClassLoader
      Loaded classes:
         1:    org.mozilla.javascript.gen._Users_duncan_macgregor_git_mozilla_rhino_moment_js_3
      -total-: 1 class
  Non-Class:   12 chunks,    592.00 KB capacity,  592.00 KB (100%) committed,   583.46 KB ( 99%) used,     8.54 KB (  1%) free,     0 bytes (  0%) waste , deallocated: 5 blocks with 7.65 KB
      Class:    1 chunk,       2.00 KB capacity,    2.00 KB (100%) committed,   528 bytes ( 26%) used,     1.48 KB ( 74%) free,     0 bytes (  0%) waste , deallocated: 0 blocks with 0 bytes
       Both:   13 chunks,    594.00 KB capacity,  594.00 KB (100%) committed,   583.98 KB ( 98%) used,    10.02 KB (  2%) free,     0 bytes (  0%) waste , deallocated: 5 blocks with 7.65 KB
```
We are paying a small extra cost in main heap usage, and in metaspace usage from MH `LambdaForm` classes. I believe I have balanced out the formed by de-duplicating some of our calcite machinery, while the latter seems to be at most 15K in additional hidden classes which is substantially less than the 700KB reduction in our class loader.


aardvark179:
@gbrail and @rbri, what are your feelings on our test for generated class names. I'n guessing the intention is to check that an exception or other stack trace contains reasonable class names, or maybe reasonable source locations? Should we test for that instead of trying to fish the class out of the compilation processes? I've disabled the test for now.

gbrail:
"Metaspace" and not "meatspace" -- that makes more sense now!

Anyway, I am not particularly attached to tests for the exact name of things in generated code, so unless @rbri has a use case for that I will not lose any sleep if we change or remove them.

aardvark179:
Curse you, autocorrect!

benchmarks/build.gradle
```diff
@@ -32,7 +32,7 @@ jmh {
             }
         }
     }
-    jvmArgs = ['-XX:+BackgroundCompilation']
+    jvmArgs = ['-XX:-BackgroundCompilation']
```
andreabergia, line 35: Since you keep trying to sneak this in unrelated PRs... 🤣 here's an actual one: https://github.com/mozilla/rhino/pull/2451

rhino/src/main/java/org/mozilla/javascript/optimizer/Bootstrapper.java
```diff
@@ -70,20 +73,99 @@ public class Bootstrapper {
...
+                new ChainedCallSite(dedupDesc(new CallSiteDescriptor(LOOKUP, op, mType))));
+    }
+
+    private static CallSiteDescriptor dedupDesc(CallSiteDescriptor desc) {
+        synchronized (CALLSITE_DESC_CACHE) {
+            return CALLSITE_DESC_CACHE.computeIfAbsent(desc, k -> new WeakReference<>(desc)).get();
+        }
     }
 
+    private static NamespaceOperation GET_PROOPERTY =
```
andreabergia: Typo: `PROOPERTY`

benchmarks/build.gradle
aardvark179: I didn't even intend to sneak that in here, it was pure incompetence.

rhino/src/main/java/org/mozilla/javascript/optimizer/Bootstrapper.java
aardvark179: Fixed.

aardvark179:
> "Metaspace" and not "meatspace" -- that makes more sense now!
> 
> Anyway, I am not particularly attached to tests for the exact name of things in generated code, so unless @rbri has a use case for that I will not lose any sleep if we change or remove them.

I've heard no objections so have removed that test.

rhino/src/main/java/org/mozilla/javascript/optimizer/Bootstrapper.java
```diff
@@ -70,20 +73,99 @@ public class Bootstrapper {
         linker = factory.createLinker();
     }
 
+    private static final MethodHandles.Lookup LOOKUP = MethodHandles.lookup();
+    private static final WeakHashMap<CallSiteDescriptor, WeakReference<CallSiteDescriptor>>
```
gbrail, line 83: I will read further but that's interesting -- what's the advantage of putting a WeakReference inside a WeakHashMap? Do we want to clean up in case either the key or the value is GCed?

rhino/src/main/java/org/mozilla/javascript/optimizer/Bootstrapper.java
```diff
@@ -70,20 +73,99 @@ public class Bootstrapper {
...
             throws NoSuchMethodException {
         Operation op = parseOperation(name);
+
         // ChainedCallSite lets a call site have a few options for complex situations.
         // It caches up to eight invocations, so that we can quickly select the best
         // implementation in situations where the same call site is invoked in different
         // contexts.
-        return linker.link(new ChainedCallSite(new CallSiteDescriptor(lookup, op, mType)));
+        return linker.link(
+                new ChainedCallSite(dedupDesc(new CallSiteDescriptor(LOOKUP, op, mType))));
```
gbrail, line 99: By passing a hard-coded LOOKUP here instead of the object passed to the function parameter, are we going to end up with weird problems accessing various fields?

rhino/src/main/java/org/mozilla/javascript/optimizer/MHJSScriptCode.java
```diff
@@ -0,0 +1,117 @@
...
+                        throw e;
+                    } catch (Throwable e) {
+                        throw (RuntimeException) e;
+                    }
+                };
+        resume =
+                resumeMH != null
+                        ? (cx, xobj, state, scope, operation, value) -> {
+                            try {
+                                return execMH.invokeExact(cx, xobj, state, scope, operation, value);
```
gbrail: Do you mean to use "resumeMH" here?

gbrail (commented):
This overall makes sense, but I have a few specific questions. Thanks!

rhino/src/main/java/org/mozilla/javascript/optimizer/Bootstrapper.java
aardvark179: Really I'm using it as a weak set. I guess it could be a simple marker object, but the APIs for getting that stored key, aren't ideal, I'll try iterating on that a bit.

rhino/src/main/java/org/mozilla/javascript/optimizer/Bootstrapper.java
aardvark179: I had a check, and we don't really use the lookup. Normally when writing call sites I would drop the lookup on the floor almost immediately, but we can't quite do that here since dynalink bakes it into the descriptor.

rhino/src/main/java/org/mozilla/javascript/optimizer/MHJSScriptCode.java
aardvark179: Actually, scripts are never generators, yet, so I'll strip that code out entirely. At some point we will need to support top level async, which means script would need to compile to generators under the hood, but we're not there yet.

rhino/src/main/java/org/mozilla/javascript/optimizer/Bootstrapper.java
aardvark179: I couldn't find a better way to do this, so added a comment explaining why we can't simply use something like `compute()` in combination with marker objects.

gbrail:
This looks OK to me now. Thanks!

PR #2449, opened 2026-07-18
Reflect.construct() fix

rbri:
Fix: Reflect.construct used ctorBaseFunction.call() to invoke the target, but LambdaConstructor targets can be non-callable as they only support .construct() which would result in a TypeError when called.

This supports a common workaround for subclassing built-in objects (like Set) without using ES6 class/extends syntax, by using Reflect.construct() with a newTarget argument.


rbri:
HtmlUnit tests passing...

gbrail:
This is nice and fixes a bunch of stuff. Thanks!

PR #2450, opened 2026-07-21
Add lazy source code provision.

aardvark179:
In many cases where we are loading large quantities of JavaScript we would like to have source code available, but do not want to pay the resource cost of having those strings on the heap all the time if they can be retrieved from a database, files, or other sources. This PR introduces a pluggable system that can lazily resolve source code requests only when required. If the feature is not used then we pay a cost of 16 bytes per source for the `EagerSourceCodeProvider`.

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/CompileSpec.java
+++ b/rhino/src/main/java/org/mozilla/javascript/CompileSpec.java
@@ -1,0 +1,154 @@
+package org.mozilla.javascript;
+
+import java.io.Reader;
+import java.util.function.Consumer;
+import org.mozilla.javascript.ast.ScriptNode;
+import org.mozilla.javascript.sourcemap.SourceMapper;
+
+/**
+ * Parameters for compiling a JavaScript script (i.e. top-level program source). Build instances via
+ * {@link #fromSource(String)} or {@link #fromReader(Reader)} and pass them to {@link
+ * Context#compileScript(ScriptCompileSpec)} or {@link Context#evaluateScript(ScriptCompileSpec,
+ * VarScope)}.
+ *
+ * @see FunctionCompileSpec
+ */
+public abstract class CompileSpec<T extends ScriptOrFn<T>> {
+    private final String source;
+    private final String sourceName;
+    private final int lineno;
+    private final Object securityDomain;
+    private final Evaluator compiler;
+    private final ErrorReporter compilationErrorReporter;
+    private final Consumer<CompilerEnvirons> compilerEnvironsProcessor;
+    private final SourceMapper sourceMapper;
+    private final SourceCodeSupplier sourceCodeSupplier;
+
+    public CompileSpec(Builder<T, ?> builder) {
+        this.source = builder.source;
+        this.sourceName = builder.sourceName;
+        this.lineno = Math.max(builder.lineno, 0);
+        this.securityDomain = builder.securityDomain;
+        this.compiler = builder.compiler;
+        this.compilationErrorReporter = builder.compilationErrorReporter;
+        this.compilerEnvironsProcessor = builder.compilerEnvironsProcessor;
+        this.sourceMapper = builder.sourceMapper;
+        this.sourceCodeSupplier = builder.sourceCodeSupplier;
+    }
+
+    public String getSource() {
+        return source;
+    }
+
+    public VarScope getScope() {
+        return null;
+    }
+
+    public String getSourceName() {
+        return sourceName;
+    }
+
+    public int getLineno() {
+        return lineno;
+    }
+
+    public Object getSecurityDomain() {
+        return securityDomain;
+    }
+
+    public Evaluator getCompiler() {
+        return compiler;
+    }
+
+    public ErrorReporter getCompilationErrorReporter() {
+        return compilationErrorReporter;
+    }
+
+    public Consumer<CompilerEnvirons> getCompilerEnvironsProcessor() {
+        return compilerEnvironsProcessor;
+    }
+
+    public SourceMapper getSourceMapper() {
+        return sourceMapper;
+    }
+
+    public SourceCodeSupplier getSourceCodeSupplier() {
+        return sourceCodeSupplier;
+    }
+
+    abstract boolean returnFunction();
+
+    abstract CompileFn<T> compilationFunction();
+
+    @FunctionalInterface
+    interface CompileFn<T extends ScriptOrFn<T>> {
+        CompilationResult<T> compile(
+                Evaluator evaluator, CompilerEnvirons env, ScriptNode tree, String rawSource);
+    }
+
+    public abstract static class Builder<T extends ScriptOrFn<T>, U extends Builder<T, U>> {
+        protected final String source;
+        protected String sourceName;
+        protected int lineno = 0;
+        protected Object securityDomain;
+        protected Evaluator compiler;
+        protected ErrorReporter compilationErrorReporter;
+        protected Consumer<CompilerEnvirons> compilerEnvironsProcessor;
+        protected SourceMapper sourceMapper;
+        protected SourceCodeSupplier sourceCodeSupplier;
+
+        protected Builder(String source) {
+            this.source = source;
+        }
+
+        public U sourceName(String sourceName) {
+            this.sourceName = sourceName;
+            return getThis();
+        }
+
+        public U lineno(int lineno) {
+            this.lineno = lineno;
+            return getThis();
+        }
+
+        public U securityDomain(Object securityDomain) {
+            this.securityDomain = securityDomain;
+            return getThis();
+        }
+
+        public U compiler(Evaluator compiler) {
+            this.compiler = compiler;
+            return getThis();
+        }
+
+        public U compilationErrorReporter(ErrorReporter compilationErrorReporter) {
+            this.compilationErrorReporter = compilationErrorReporter;
+            return getThis();
+        }
+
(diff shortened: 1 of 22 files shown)


aardvark179:
As part of porting this I did quite a bit of tidying up in our compiler interface, which might be worth splitting out into a separate PR. I'm also not sure I like the `LazySourceCodeProvider`'s use of a weak reference, it feels like it could produce a lot of churn, but I'm also not convinced Rhino should be responsible for any sort of cache here.

If anybody has a good suggestion on how to slice this I'd be interested.

rhino/src/main/java/org/mozilla/javascript/NullSourceCodeProvider.java
```diff
@@ -0,0 +1,12 @@
+package org.mozilla.javascript;
```
gbrail, line 1: Looks like a typo in the class name here

rhino/src/main/java/org/mozilla/javascript/optimizer/ClassCompiler.java
```diff
@@ -307,9 +310,23 @@ private void buildDescriptor(
         cfw.add(builder.isEvalFunction ? ByteCode.ICONST_1 : ByteCode.ICONST_0);
         cfw.add(builder.hasRestArg ? ByteCode.ICONST_1 : ByteCode.ICONST_0);
         cfw.addLoadConstant(builder.sourceFile);
-        cfw.addLoadConstant(builder.rawSource);
-        cfw.addLoadConstant(builder.rawSourceStart);
-        cfw.addLoadConstant(builder.rawSourceEnd);
+        if (compilerEnv.isGeneratingSource()) {
```
gbrail, line 315: Is this intended? At the moment we have two branches that do the same thing -- Is it possible that we want something different here?

rhino/src/main/java/org/mozilla/javascript/LazySourceCodeProvider.java
```diff
@@ -0,0 +1,55 @@
...
+    public LazySourceCodeProvider(SourceCodeSupplier sourceSupplier) {
+        Objects.requireNonNull(sourceSupplier);
+        this.sourceSupplier = sourceSupplier;
+    }
+
+    @Override
+    public String getSource(JSDescriptor<?> descriptor, int start, int end) {
+        String source;
+        if (sourceCodeRef != null && (source = sourceCodeRef.get()) != null) {
+            var funcionSrc = source.substring(start, end);
```
gbrail: Looks like a typo in that variable name

rhino/src/main/java/org/mozilla/javascript/LazySourceCodeProvider.java
```diff
@@ -0,0 +1,55 @@
...
+            descriptor.replaceSourceProvider(new ResolvedSourceProvider(funcionSrc));
```
gbrail: If we share Class objects between scripts or threads, and presumably they also share LazySourceCodeProvider instances, do we have a thread-safety problem? It may not matter, but it's probably worth tracing through the effect of various race conditions. If the worst case scenario is that we create extra garbage, that's OK, but it's worth just thinking it through.

gbrail (commented):
This seems like a fine idea and while it's not high on my list of Rhino problems I can see why it's on yours! I'd like to look at some of these small problems here first.

rhino/src/main/java/org/mozilla/javascript/LazySourceCodeProvider.java
andreabergia: Yes, but I think it's wholly benign. At worst, the supplier is invoked twice, and there will be a race to select which version to use. But consider how unlikely that is, it is fine.

The supplier should be repeatable, not something that can be consumed just once. I'm not sure if we made that clear in the javadoc, but that is definitely the idea behind it.

gbrail: I'm fine with that -- nice building in Java rather than Rust sometimes ;-)

rhino/src/main/java/org/mozilla/javascript/NullSourceCodeProvider.java
aardvark179: Fixed.

rhino/src/main/java/org/mozilla/javascript/optimizer/ClassCompiler.java
aardvark179: Sorry, I'd commented that part out while I worked out what to do in that area. It's an example of a problem I think we need to deal with more generally for other stuff, which is how we handle constants for some of our internal objects. I'd like to try adding [candy](https://docs.oracle.com/javase/specs/jvms/se25/html/jvms-4.html#jvms-4.4.10) support to the class file writer for things like this, symbols, regexes, and template literals, but that should be a separate change.

rhino/src/main/java/org/mozilla/javascript/LazySourceCodeProvider.java
aardvark179: Fixed.

rhino/src/main/java/org/mozilla/javascript/LazySourceCodeProvider.java
```diff
@@ -0,0 +1,65 @@
...
+        String source;
+        if (sourceCodeRef != null && (source = sourceCodeRef.get()) != null) {
+            var functionSrc = source.substring(start, end);
+            descriptor.replaceSourceProvider(new ResolvedSourceProvider(functionSrc));
+            return functionSrc;
+        } else {
+            source = sourceSupplier.get();
+            if (source != null) {
+                sourceCodeRef = new WeakReference<>(source);
+                var funcionSrc = source.substring(start, end);
```
gbrail: Typo in "funcionSrc"...

rhino/src/main/java/org/mozilla/javascript/CompileSpec.java
```diff
@@ -0,0 +1,154 @@
+package org.mozilla.javascript;
+
+import java.io.Reader;
+import java.util.function.Consumer;
+import org.mozilla.javascript.ast.ScriptNode;
+import org.mozilla.javascript.sourcemap.SourceMapper;
+
+/**
+ * Parameters for compiling a JavaScript script (i.e. top-level program source). Build instances via
+ * {@link #fromSource(String)} or {@link #fromReader(Reader)} and pass them to {@link
```
gbrail: This is producing a warning from Errorprone.

rhino/src/main/java/org/mozilla/javascript/EagerSourceCodeProvider.java
```diff
@@ -0,0 +1,25 @@
...
+        this.rawSource = rawSource;
+    }
+
+    @Override
+    public String getSource(JSDescriptor<?> desc, int start, int end) {
+        return rawSource.substring(start, end);
+    }
+
+    // Escape hatch for class compilation.
+    public String getRawSource() {
```
gbrail, line 22: This producing an ErrorProne warning because it's missing @Override

rhino/src/main/java/org/mozilla/javascript/JSDescriptor.java
```diff
@@ -39,9 +39,9 @@ public final class JSDescriptor<T extends ScriptOrFn<T>> implements Serializable
     private final boolean[] paramIsConst;
     private final int flags;
     private final String sourceFile;
-    private final String rawSource;
-    private final int rawSourceStart;
-    private final int rawSourceEnd;
+    private SourceCodeProvider sourceCodeProvider;
```
gbrail: I'm OK with the unsynchronized access to this because the worst case is that we have a benign race condition. However, I think this should be volatile for two reasons. First, it's theoretically possible that without that some threads may never see the result of the update, and more importantly (to me) to remind us all that this variable is mutated without a lock.

gbrail (commented):
Thanks! I know that I'm being pedantic here but I have a few more things.

In general -- can you run "./gradlew compileJava" and try to fix the warnings from Errorprone? We were completely clean for a long time and now various warnings are starting to creep back in. Thanks!

rhino/src/main/java/org/mozilla/javascript/JSDescriptor.java
aardvark179: That's reasonable, fixed.

rhino/src/main/java/org/mozilla/javascript/EagerSourceCodeProvider.java
aardvark179: Fixed.

rhino/src/main/java/org/mozilla/javascript/CompileSpec.java
aardvark179: Fixed.

rhino/src/main/java/org/mozilla/javascript/LazySourceCodeProvider.java
aardvark179: Fixed.

aardvark179:
I've fixed all the warnings from error prone except the one about an unused method in interpreter v2 and the ambiguous method name on top level. I took the opportunity to clean up bunch of casts in `ClassCompiler`.

rbri:
have done a test round with htmlunit, works fine

rbri:
have restarted the test, looks better now

gbrail:
This looks good to me now. Thanks!

PR #2452, opened 2026-07-28
Simplify modular java related checks

ZZZank:
- The "isModularJava" bool field moved to `ReflectUtils`
- `.isExportedClass(Class)` rewritten to call methods directly instead of via reflection. Main logic is gated after `ReflectUtils.IS_MODULAR_JAVA`, for Android compatibility


gbrail:
Sorry this got caught behind -- this works and looks, fine, thanks!

PR #2453, opened 2026-07-28
Remove `fieldsAndMethods` map from JavaMembers

ZZZank:
since `fieldsAndMethods` and `staticFieldsAndMethods` is never mutated now, they will always be an empty map, basically invalidating related code logic. This PR removed these two now unused map, and restored FieldAndMethods handling by rebinding `javaObject` in `JavaMembers#get(...)`


gbrail:
I understand this and I understand some, but not all of it. I really think we need a test for this specific functionality -- looking at test coverage, it doesn't look like some of the functionality changed here is being hit by the tests.

aardvark179:
I'd really appreciate @andreabergia's review of this, I think they are more familiar with this area than me.

PR #2454, opened 2026-07-29
PoC: Replace `NativeJavaMethod` with descriptors that can be cached across contexts.

aardvark179:
We cache a large number of `NativeJavaMethod`s between contexts in production instances to avoid repeatedly reflecting on the same things, but this is tricky as they must be detached from their original prototypes and if anything is wrong we end up leaking resources. This change tries to refactor things to fit native java methods into the same descriptor framework we use elsewhere, giving us something we can more easily cache and reuse.

Making a draft for PR for now to gather feedback, I feel like this area should be possible to simplify further with a bit more work.

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/JavaMembers.java
+++ b/rhino/src/main/java/org/mozilla/javascript/JavaMembers.java
@@ -26,6 +26,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
+import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.stream.Collectors;
 import org.mozilla.javascript.lc.ReflectUtils;
 import org.mozilla.javascript.lc.member.ExecutableBox;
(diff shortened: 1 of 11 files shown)


rhino/src/main/java/org/mozilla/javascript/JavaMembers.java
```diff
@@ -88,6 +89,37 @@ boolean has(String name, boolean isStatic) {
         return findExplicitFunction(name, isStatic) != null;
     }
 
+    static JSDescriptor<JSFunction> buildNativeJavaMethodDescriptor(
+            ExecutableBox[] methods, String name) {
+        int arity = methods.length > 0 ? methods[0].getArgTypes().size() : 0;
```
ZZZank, line 94: Not sure if it affects anything, but will how about using biggest arity instead of a random select
```suggestion
        int arity = Arrays.stream(methods).mapToInt(m -> m.getArgTypes().size()).max().orElse(0);
```

rhino/src/main/java/org/mozilla/javascript/JavaMembers.java
```diff
@@ -895,7 +931,12 @@ RuntimeException reportMemberNotFound(String memberName) {
     private final Map<String, Object> staticMembers;
 
     private final Map<String, ExecutableOverload.WithField> staticFieldAndMethods;
-    NativeJavaMethod ctors; // we use NativeJavaMethod for ctor overload resolution
+
+    /** Overloaded constructors, used for constructor overload resolution. */
+    ExecutableBox[] ctors;
+
+    /** Per-class cache of resolved constructor overloads, keyed by argument runtime types. */
+    final CopyOnWriteArrayList<ResolvedOverload> ctorOverloadCache = new CopyOnWriteArrayList<>();
```
ZZZank, line 939: replace executables+cache with `JavaAccessor`?

rhino/src/main/java/org/mozilla/javascript/JavaMembers.java
```diff
@@ -904,18 +945,44 @@ final class BeanProperty {
...
+}
+
+/**
+ * A callable bean getter or setter, backed by one or more overloaded Java methods. Overload
+ * resolution and invocation are delegated to {@link NativeJavaOverloadResolver}.
+ */
+final class JavaAccessor {
+    final String name;
+    final ExecutableBox[] methods;
+    final CopyOnWriteArrayList<ResolvedOverload> overloadCache = new CopyOnWriteArrayList<>();
```
ZZZank, line 959: Since `JavaAccessor` is already "mutable" (overloadCache), we can also make it act as a cache for descriptors.

Also, for me `JavaAccessor` looks like yet another `ExecutableOverload`, but with overload cache. I think merging JavaAccessor into ExecutableOverload can clarify the overall structure a bit.

TLDR: add resolved overload cache & descriptor cache to ExecutableOverload, and migrate JavaAccessor to ExecutableOverload

rhino/src/main/java/org/mozilla/javascript/JavaMembers.java
```diff
@@ -904,18 +945,44 @@ final class BeanProperty {
...
 
-class FieldAndMethods extends NativeJavaMethod {
+class FieldAndMethods extends JSFunction {
     @Serial private static final long serialVersionUID = -9222428244284796755L;
 
     FieldAndMethods(VarScope scope, ExecutableOverload.WithField withField) {
-        super(withField.methods, withField.name);
+        super(
+                scope,
+                JavaMembers.buildNativeJavaMethodDescriptor(withField.methods, withField.name),
```
ZZZank, line 982: one side effect of adding descriptor cache to ExecutableOverload (see review on JavaAccessor) is cached descriptor can be easily reached here

rhino/src/main/java/org/mozilla/javascript/NativeJavaCode.java
```diff
@@ -0,0 +1,86 @@
...
+ * JSCode implementation for Java methods and constructors exposed to JavaScript.
+ *
+ * <p>Handles reflective method invocation with overload resolution and caching.
+ *
+ * @see NativeJavaOverloadResolver for overload resolution logic
+ */
+public class NativeJavaCode extends JSCode<JSFunction> {
+
+    private final ExecutableBox[] methods;
+    private final CopyOnWriteArrayList<ResolvedOverload> overloadCache =
```
ZZZank, line 22: Is the cache here necessary? There seem to be more than one overload cache

PR #2458, opened 2026-08-02
Literals and simple keyword instructions.

aardvark179:
First round of instructions for interpreter v2. This covers simple literals and keyword operations.

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/interpreterv2/instruction/VoidInstruction.java
+++ b/rhino/src/main/java/org/mozilla/javascript/interpreterv2/instruction/VoidInstruction.java
@@ -1,0 +1,42 @@
+package org.mozilla.javascript.interpreterv2.instruction;
+
+import org.mozilla.javascript.CallFrameV2;
+import org.mozilla.javascript.Context;
+import org.mozilla.javascript.interpreterv2.InstructionFormatter;
+import org.mozilla.javascript.interpreterv2.operand.Operand;
+import org.mozilla.javascript.interpreterv2.operand.PopOperand;
+
+public class VoidInstruction extends Instruction {
+    private static final VoidInstruction voidInstructionPopOperand =
+            new VoidInstruction(PopOperand.instance);
+
+    public static VoidInstruction ofOperand(Operand obj) {
+        if (obj == PopOperand.instance) {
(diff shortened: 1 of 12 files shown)


aardvark179:
There was on commit that came before this that introduced a stats reporting feature, that I think was left behind after development, and so I have dropped it. I might introduce the later behind a system property.

gbrail:
This seems fine and doesn't cause any regressions yet anyway! I'm curious about your comments in the "Int" and "ShortNumber" classes about the comparision to -1.0. Do you want to leave it like this, or fix it now? I'm OK with adding a TODO if we must.

aardvark179:
> This seems fine and doesn't cause any regressions yet anyway! I'm curious about your comments in the "Int" and "ShortNumber" classes about the comparision to -1.0. Do you want to leave it like this, or fix it now? I'm OK with adding a TODO if we must.

I've fixed it. There was quite a bit of cleanup I ended up having to do on this stuff, but I'm sure there'll be something else down the line that is still wrong.

rhino/src/main/java/org/mozilla/javascript/interpreterv2/instruction/VoidInstruction.java
```diff
@@ -0,0 +1,42 @@
...
+
+    private final Operand obj;
+
+    private VoidInstruction(Operand obj) {
+        this.obj = obj;
+    }
+
+    @Override
+    public void interpret(Context cx, CallFrameV2 frame) {
+        frame.pc += 1;
```
gbrail: Every other one of these implementations increments pc after doing the thing that it does. This means that PC will be in the wrong place if "obj.retrieve" throws.

gbrail (commented):
Found something you might want to look at, but since this is all new code that isn't executed anywhere we can certainly move forward -- let me know what you think.

rhino/src/main/java/org/mozilla/javascript/interpreterv2/instruction/VoidInstruction.java
aardvark179: Fixed. I'll need to go through the other instructions and make sure they are consistent.

gbrail:
This looks good now -- thanks!

PR #744, opened 2020-07-06
Added tests for pending issues

luisggpina:
Added tests for issues #662 #663 #645 as recommended by @rbri

Issue #663, opened 2020-05-07
IllegalArgumentException in Label.setName


jon-bell:
We ran into an unexpected IllegalArgumentException when attempting to compile the following string using Rhino 1.7.12: `\u000a:S<6`

Stack trace:
```Exception in thread "main" java.lang.IllegalArgumentException: invalid label name
	at org.mozilla.javascript.ast.Label.setName(Label.java:57)
	at org.mozilla.javascript.Parser.name(Parser.java:3236)
	at org.mozilla.javascript.Parser.primaryExpr(Parser.java:3125)
	at org.mozilla.javascript.Parser.memberExpr(Parser.java:2732)
	at org.mozilla.javascript.Parser.unaryExpr(Parser.java:2626)
	at org.mozilla.javascript.Parser.mulExpr(Parser.java:2547)
	at org.mozilla.javascript.Parser.addExpr(Parser.java:2531)
	at org.mozilla.javascript.Parser.shiftExpr(Parser.java:2512)
	at org.mozilla.javascript.Parser.relExpr(Parser.java:2487)
	at org.mozilla.javascript.Parser.eqExpr(Parser.java:2459)
	at org.mozilla.javascript.Parser.bitAndExpr(Parser.java:2448)
	at org.mozilla.javascript.Parser.bitXorExpr(Parser.java:2437)
	at org.mozilla.javascript.Parser.bitOrExpr(Parser.java:2426)
	at org.mozilla.javascript.Parser.andExpr(Parser.java:2415)
	at org.mozilla.javascript.Parser.orExpr(Parser.java:2404)
	at org.mozilla.javascript.Parser.condExpr(Parser.java:2368)
	at org.mozilla.javascript.Parser.assignExpr(Parser.java:2324)
	at org.mozilla.javascript.Parser.expr(Parser.java:2303)
	at org.mozilla.javascript.Parser.nameOrLabel(Parser.java:2052)
	at org.mozilla.javascript.Parser.statementHelper(Parser.java:1252)
	at org.mozilla.javascript.Parser.statement(Parser.java:1116)
	at org.mozilla.javascript.Parser.parse(Parser.java:653)
	at org.mozilla.javascript.Parser.parse(Parser.java:583)
	at org.mozilla.javascript.Context.parse(Context.java:2572)
	at org.mozilla.javascript.Context.compileImpl(Context.java:2521)
	at org.mozilla.javascript.Context.compileString(Context.java:1525)
[...]

PR #2463, opened 2026-08-09
Add a test verifying that issue #663 is fixed in ES6 mode

rbri:
With this finally added to the test suite, we can close
* pr #744 and
* issue #663


gbrail:
Thanks for following up!

PR #2465, opened 2026-08-14
PoC: Make `Scriptable` objects work with more ES property methods.

aardvark179:
Very rough quick prototype built by my robot friend. Open questions:
1. Should we do this on `Scriptable` or a subclass?
2. Should we do this at all?
3. Are there other APIs that should also be expanded to support `Scriptable`s? Should we delay that until we can maybe refactor things to be closer to ES primitives in terms of return results? Does any of us even have a strategy for how to make that transition?


aardvark179:
Our robot friends can do all the hard bits, but then they forget to add the new files to git.

Should probably update AGENTS.md to fix that.

PR #2466, opened 2026-08-18
use the provided context instead of looking out for the current to avoid the thread local lookup

rbri:


rbri:
another minor cleanup step - hopefully

Not sure if we should still offer a version of getArrayElements() without cx parameter for backward compatibility. Maybe someone else has an opportunity on this.

gbrail:
This looks good, and simple -- thanks!

PR #2467, opened 2026-08-18
BigIntTypedArrays: Fix with() value parameter not allowing a bigint

camnwalter:
Previously, when passing in a biginteger to a BigInt64TypedArray or BigUint64TypedArray's with() method, a type error would occur. This resolves that problem.

See https://262.ecma-international.org/#sec-%typedarray%.prototype.with for the spec


camnwalter:
I'm not sure why the check is failing. Doesn't seem related to my changes?

rbri:
@camnwalter have restarted the task, now everything works fine

camnwalter:
Thanks!

tests/testsrc/jstests/es2023/typedArray-with.js
```diff
@@ -58,6 +53,49 @@ var types = [Int8Array, Uint8Array, Int16Array, Uint16Array,
 	}
 })();
 
+(function withBigIntValue() {
```
rbri, line 56: maybe i'm wrong but you change is done in NativeTypedArrayView; i think in a way that it works for all array types. If yes we have to test it for the other types also...

camnwalter: They can have big ints passed in, but they type error according to the `withNonBigIntTypedArrayRejectsBigIntValue` test at the end of the file (which is what should happen)

camnwalter: I could combine the 2 into 1 test if that'd make more sense

rbri: Omg i'm soo blind... sorry

gbrail:
Thanks! There are more changes coming in this area, nice to have this one!

PR #2469, opened 2026-08-22
Fix numeric type handling in ScriptRuntime.same

gbrail:
This fixes how Object.is works as well. This bug did not show up in pure
JavaScript, but only when some of the values were generated by native code
and therefore of different numeric types.

Also, clean up same(), sameZero(), eq(), and shallowEq() to use new
instanceof syntax and fix a few other uglinesses.

Add a comprehensive AI-generated test suite for all of the above.


rhino/src/main/java/org/mozilla/javascript/ScriptRuntime.java
```diff
@@ -4371,49 +4371,49 @@ public static Object toPrimitive(Object input, Class<?> preferredType) {
...
         } else if (x == y) {
             return true;
         } else if (x instanceof CharSequence) {
             return eqString((CharSequence) x, y);
-        } else if (x instanceof Boolean) {
-            boolean b = ((Boolean) x).booleanValue();
-            if (y instanceof Boolean) {
-                return b == ((Boolean) y).booleanValue();
+        } else if (x instanceof Boolean bx) {
+            boolean b = bx;
```
rbri, line 4396: I guess we can use bx in the code afterwards and remove this assignment and the var

rbri (commented):
I think both implementations are hard to read in the sense to mentally check if all cases are handled the way they should. Having that if x is this, else if x ist that pattern at the beginning looks understanable but then it seems to switch to something that has more complex conditions. Not related to the code changes but maybe we can try to make this more readable

gbrail:
I did some AI-driven DRY on this and the result is a bit cleaner. It's still super complex given the spec, Java, and things we've added over the years like the "equivalentValues" code which is actually only used by our XML implementation. But please take a look and see if it's any better!

gbrail:
FWIW, if I ever had to give a talk about "why JavaScript is complicated" I would just show them this ;-)

rhino/src/main/java/org/mozilla/javascript/ScriptRuntime.java
gbrail: Actually, no, that doesn't compile. We could just use .equals() but I think this is clean so I'm going to leave it.!

PR #2470, opened 2026-08-24
Add better infrastructure for checking test262

gbrail:
This is strictly for the Bazel build. Add a way to use Bazel to verify that
no test262 tests marked as failing actually pass. The test, which is disabled
by default, will fail in that case, and we'll still need Gradle to fix
test262.properties.


gbrail:
I expect this will be uncontroversial for folks so I am going to merge it and move on to other stuff.

PR #2471, opened 2026-08-24
WIP support for timeouts, intervals, and other async tasks.

aardvark179:
@gbrail this is a very rough port of our internal support for asynchronous tasks such as timeouts, intervals, and the fetch API. This isn't ready for proper review, but I thought was worth pushing as you said you were considering rearchitecting micro task stuff. This is missing tests, I haven't updated the timeout or interval classes to use descriptors, there's a couple more bits of refactoring I think we could do to improve these APIs, and this stuff isn't exposed yet in the REPL or other contexts.

### DIFF
--- a/rhino/src/main/java/org/mozilla/javascript/Context.java
+++ b/rhino/src/main/java/org/mozilla/javascript/Context.java
@@ -21,14 +21,29 @@ import java.lang.ref.ReferenceQueue;
 import java.lang.reflect.Constructor;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
+import java.time.Clock;
+import java.time.Instant;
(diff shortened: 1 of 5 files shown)


gbrail:
Thanks -- I think that I can work with this, although I do have some different ideas as well.

Here's a question: It makes sense for, somehow, Rhino to have capabilities that aren't built in to V8 but usually part of the browser and Node.js, namely:

* setTimeout
* setImmediate
* Workers and postMessage
* structuredClone

Do you think that we should:

A) Build these things into rhino-core
B) Build them in to Rhino but as an optional plug in or something you have to wire in manually
C) Not have them in Rhino at all?

Either way, we need to have a timer capability and a thread-safe task queue, but I would like to get a sense of how people would like to handle these higher-level features

PR #2473, opened 2026-08-24
JFR instrumentation.

aardvark179:
Add JFR events for internal Rhino processes and an option to emit JFR events from tests.

This adapts a pattern we have used internally to avoid unnecessary creation of J FR event objects, and instead uses the technique to make the event objects opaque. This technique will scale fine for small set of event types, but will need some rework to be really performant if the number grows.

I included the `TEST_EVENT` as a useful way to temporarily instrument areas of the code.

I have also added support for emitting standard junit events.

### DIFF
--- a/README.md
+++ b/README.md
@@ -192,6 +192,17 @@ When running the benchmarks you may find a couple of environment variables usefu
   argument. This allows for things like running JFR as the profiler to
   collect information on lock contention or other events.
 
+### JFR events
+
+Rhino will produces JFR events for a number of internal operations
+such as initialising global objects, parsing, compiling, and script
+execution These, along with standard junit events, can be enabled for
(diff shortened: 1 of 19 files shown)


aardvark179:
I've been using the test instrumentation to understand where we are spending time during testing, especially in test 262. I'll do a follow up PR once that work is tidied up.

gbrail:
Although adds a dependency on another Java module it seems OK (I mean, it doesn't break the Android build apparently). I don't use JFR much beyond what the profiler tool gives me. What's the purpose of the new "tests.jfc" file? Is that just so that we can automatically run tests with JFR and get some useful output?

aardvark179:
That jfc file contains a configuration that enables our events and sets a fairly low limit for recording monitor blocking, without having to make them enabled for everyone by default.

gbrail:
I don't mind merging this then!

PR #2475, opened 2026-08-26
Comparison and object operations.

aardvark179:
Next set of instructions, comparison, equality, and types and instance of operations.


rhino/src/main/java/org/mozilla/javascript/interpreterv2/instruction/Comparison.java
```diff
@@ -0,0 +1,95 @@
...
+        } else {
+            lhs = this.lhs.retrieve(cx, frame);
+        }
+
+        boolean valBln;
+        object_compare:
+        {
+            number_compare:
+            {
+                Number rNum, lNum;
```
gbrail, line 59: if DOUBLE_MARK is true for both lhs and rhs we are missing a big optimization opportunity to do the comparison without any boxing. This would require adding a "compare" function in ScriptRuntime that takes two doubles as arguments, but it would prevent us from undoing the optimizations that we got. OTOH I don't know how much the interpreter uses the ability to avoid boxing of doubles, since the compiler does that by generating different bytecode.

rhino/src/main/java/org/mozilla/javascript/interpreterv2/instruction/TypeofName.java
```diff
@@ -0,0 +1,30 @@
...
+public class TypeofName extends Instruction {
+    private final String name;
+
+    public TypeofName(String name) {
+        this.name = name;
+    }
+
+    @Override
+    public void interpret(Context cx, CallFrameV2 frame) {
+        frame.push(ScriptRuntime.typeofName(frame.scope, name));
```
gbrail: ScriptRuntime.typeofName calls Context.getContext(). Since you have the context already you can save a thread-local variable lookup if you use "cx".

gbrail (commented):
Both of these small things were found by AI but I verified them. Both affect performance but after all that's what we're trying to accomplish here. If you'd rather take care of them some other time (or not at all) that's OK too.

rhino/src/main/java/org/mozilla/javascript/interpreterv2/instruction/Comparison.java
aardvark179: Fixed.

rhino/src/main/java/org/mozilla/javascript/interpreterv2/instruction/TypeofName.java
aardvark179: Fixed.

aardvark179:
Thanks for spotting those @gbrail, while fixing them I also moved a couple of switch statements over to switch expressions, since I was touching those areas of the code anyway.

aardvark179:
BTW @gbrail I got my robot friend to apply the review comments from this and the previous instruction PR to all of the other instruction commits. I've put everything on https://github.com/aardvark179/rhino/tree/aardvark179-interpreterv2 (the branch for #2426). I don't see the test failure on that PR locally, but I have found various ways to push similar scripts over the edge and cause a stack overflow in node transformation—I can cause the same issues with V1 and class compilation. I'm going to try some refactoring to reduce the problem in node transformation and then see if I can provoke the issue in code gen as well.

gbrail:
Thanks, looking good!

Issue #1235, opened 2022-06-07
Promise in new Java Thread bug


fizly:
At present, I use the method of babel+babelify to use rhino with typescript, and it is very convenient. But now encounter difficulties, Promise is no problem in the main js thread, but after creating Java Thread, or using Promise in the execution body of Java thread callback, it will block and not return.

// @ts-ignore
export let Thread = java.lang.Thread
// @ts-ignore
export let Runnable = java.lang.Runnable

async function test1() {
    console.log("just test async func")
    return 1234
}

async function test2() {
    Thread.sleep(1000)
    console.log("test sleep async func")
    return 888
}

async function failedFunction() {
    let javaThread = new Thread(new Runnable({
        run: () => {
            (async () => {
                await test1();// here await the function permanently blocked
            })()

            test1().then(ret => {
                console.log("here is unreachable")
            })

            Promise.resolve(555)
                .then(ret => {
                    console.log("here is also unreachable")
                })
        }
    }))

    javaThread.start()
}

async function successFun() {
    console.log("in success function")
    await test1()
    test2() //test2 blocked 1000ms ,because Thread.sleep(1000)
    return 4567
}

successFun().then(ret => {
    console.log(`normal async/await is available:${ret}`)
})
failedFunction()

p-bakker:
To make it easier for someone to look at this, please provide a simple, runnable example. Besides the code example above being an image, it's neither the compiled JS code that can be run in Rhino, nor is it a simple example that demonstrates the issue with the simplest code possible

fizly:
> To make it easier for someone to look at this, please provide a simple, runnable example. Besides the code example above being an image, it's neither the compiled JS code that can be run in Rhino, nor is it a simple example that demonstrates the issue with the simplest code possible

new java.lang.Thread(new JavaAdapter(java.lang.Runnable, {
    run: function run() {
        Promise.resolve(123).then(ret => {
            print("unreachable")
        })
    }
})).start()

p-bakker:
@grob Tnx for that pointer.

So here's some more generic implementation example that will handle Promises in threads (also including a NativeConsole implemention):
```
    ContextFactory.getGlobal().addListener(new Listener() {

        @Override
        public void contextCreated(Context ctx) {
            ctx.setLanguageVersion(200);
            ctx.setTrackUnhandledPromiseRejections(true);
        }

        @Override
        public void contextReleased(Context ctx) {
            ctx.processMicrotasks();

[...]

Issue #1109, opened 2021-11-30
Async debugging support


p-bakker:
With the[ introduction of Promises](#160) in v1.7.14, Rhino now supports async programming out of the box.

With that, we should look into the support for debugging in async scenarios, at least one thing being async callstacks

V8 has [this](https://docs.google.com/document/d/13Sy_kBIJGP0XT34V1CV3nkWya4TwYx9L3Yv45LdGB6Q/edit), but it works only for [async/await and Promise.all/any](https://v8.dev/docs/stack-trace-api#async-stack-traces)

Maybe this case only becomes relevant when we get to implementing [async/await](#395)

Issue #1108, opened 2021-11-30
Java Interop for Promises


p-bakker:
In v1.7.14 support for JavaScript Promises [landed](#160)

However, no specific logic is inplemented yet to deal with JavaScript Promises from the Java side.

This issue is to figure out and implement the interoperability between Java and JavaScript for Promises.

As for the direction to take, maybe this gives some inspiration: the Promises section at [GraalVM JavaInteroperability](https://docs.oracle.com/en/graalvm/enterprise/20/docs/reference-manual/js/JavaInteroperability/) (also see [Asynchronous polyglot programming with Java and JavaScript on GraalVM](https://medium.com/graalvm/asynchronous-polyglot-programming-in-graalvm-javascript-and-java-2c62eb02acf0)) or at a higher level: [Java Promise](https://github.com/riversun/java-promise)

rbri:
I did this already during the last days for HtmlUnit and the good news - it is possible even with the current state of the impl.
You have to solve two problems
* make sure the async processing is done (cx.processMicrotasks() has to be called)
* be able to construct promises if some other js function impl e.g. returns one

For this i use code like this

        final Scriptable scope = ScriptableObject.getTopLevelScope(this);
        final LambdaConstructor ctor = (LambdaConstructor) getProperty(scope, "Promise");
        final LambdaFunction resolve = (LambdaFunction) getProperty(ctor, "resolve");
        return resolve.call(Context.getCurrentContext(), this, ctor, new Object[] {""});

p-bakker:
@rbri Great that it's already possible, added the `docs`tag to document this once we have docs again

In your opinion though: is what you describe above the best way to go about this and thus, should this case be closed? Or is this just something that works, but we need a better/easier approach?

rbri:
> In your opinion though: is what you describe above the best way to go about this and thus, should this case be closed? Or is this just something that works, but we need a better/easier approach?

We need a better approach to make it simpler to construct a promise from java code. At the moment all the methods are more or less unreachable for java classes.

p-bakker:
Regarding instantiating a Promise from Java code: haven't tried it yet, but context.newObject(scope, 'Promise, args) should do the trick I think

[...]

PR #2477, opened 2026-08-27
Add convenience methods for manipulating Promises from Java

gbrail:
Add methods to NativePromise so that native Java code can easily create
a new Promise, resolve or reject it, and be notified when it resolves or is
fulfilled.

This should support use cases like:

* Creating a Promise from Java and passing it to JavaScript code
* Fulfilling or rejecting a promise from Java code
* Being notified in Java when a promise is resolved or rejected

These are convenience methods for any code that embeds Rhino and needs to
manipulate promises.

For example, we are going to need this to implement Atomics.waitAsync, which
requires us to return a Promise and resolve it when the condition is fulfilled.


rhino/src/main/java/org/mozilla/javascript/NativePromise.java
```diff
@@ -613,67 +753,39 @@ private static class ResolvingFunctions {
                     new LambdaFunction(
                             topScope,
                             1,
-                            (Context cx, VarScope scope, Object thisObj, Object[] args) ->
-                                    resolve(
-                                            cx,
-                                            scope,
-                                            promise,
-                                            (args.length > 0 ? args[0] : Undefined.instance)));
+                            (Context cx, VarScope scope, Object thisObj, Object[] args) -> {
```
aardvark179, line 756: These lambdas close over `promise` (which is an argument we aren't storing as a field), and `this` (because they are looking at the `alreadyResolved` field, and it's hard to tell if that was intentional. I think it might be worth putting promise in a final field, moving these lambdas to private methods and then just passing a method reference like `this::resolvePromise`.

aardvark179 (commented):
Looks pretty good, my only questions are around a couple of the lambdas and what they close over.

gbrail:
I cleaned up a few of the lambdas to use method references where they make sense. There are a few lambdas that actually need to capture some state so I left those alone.

p-bakker:
Any existing issues resolved by this PR? #1108, #1109, #1235 maybe?

rbri:
Will try the new stuff when back...

PR #2482, opened 2026-09-06
Switch ArrayBuffer from raw byte[] to ByteBuffer

gbrail:
This replaces the implementation of ArrayBuffer from using a raw byte array
to a ByteBuffer. This is the first step in implementing shared ArrayBuffers,
since we will need direct ByteBuffers in that case if we want to get atomic
operations across Java threads.

This also does away with the old "ByteIo" class and moves conversion logic for
"float16" types into a new class, with new unit tests.


aardvark179:
Do we need direct byte buffers for atomic operations? VarHandles should be able to do any memory fencing you need on arrays.

gbrail:
Yes, we need direct ByteBuffers if we want efficient atomic operations.

VarHandle only supports atomic operations on direct ByteBuffers, and only for int and long (and float and double which JS doesn't care so much about). This has changed between releases, and older versions of Java actually supported more than current versions.

rhino/src/main/java/org/mozilla/javascript/typedarrays/NativeArrayBuffer.java
```diff
@@ -150,7 +182,10 @@ public NativeArrayBuffer slice(double s, double e) {
         int len = end - start;
 
         NativeArrayBuffer newBuf = new NativeArrayBuffer(len);
-        System.arraycopy(buffer, start, newBuf.buffer, 0, len);
+        buffer.position(start);
+        newBuf.buffer.put(buffer);
```
aardvark179: I think you need to use the `ByteBuffer.put(ByteBuffer, int, int)` method here, otherwise you can overflow the new buffer if `end` is less than the original buffer end, and you'll get an exception.

Alternatively you could create a `tmp` buffer like you do elsewhere and set the limit on that.

rhino/src/main/java/org/mozilla/javascript/typedarrays/NativeBigInt64Array.java
```diff
@@ -90,11 +90,7 @@ protected Object js_get(int index) {
         if (checkIndex(index)) {
             return Undefined.instance;
         }
-        long base =
-                ByteIo.readUint64Primitive(
-                        arrayBuffer.buffer,
-                        (index * BYTES_PER_ELEMENT) + offset,
-                        useLittleEndian());
```
aardvark179, line 97: Are there any remaining uses left of `useLittleEndian()`?

aardvark179 (changes requested):
My robot friend found a couple of issues with this, most were not real issues (it doesn't understand that you can fit an unsigned short into a short as long as you know it's an unsigned short), but it did spot a couple of problems that do look genuine.

rhino/src/main/java/org/mozilla/javascript/typedarrays/NativeArrayBuffer.java
gbrail: That method is probably not failing because it is unused. I don't recall why I put it in in the first place, but since in 2.0 we're changing other public methods, I think we should just delete that method entirely.

rhino/src/main/java/org/mozilla/javascript/typedarrays/NativeBigInt64Array.java
gbrail: Nope, there are no more usages!

gbrail:
I removed the unused methods (the public "split" and the unused byte order one.

aardvark179 (approved):

PR #2483, opened 2026-09-07
Add the additional Uint8Array methods

camnwalter:
This adds support for the Uint8Array additional methods.

The only 4 failing tests are 2 depending on classes, and 2 depending on immutable array buffers, which we haven't implemented yet.


rhino/src/main/java/org/mozilla/javascript/typedarrays/NativeUint8Array.java
```diff
@@ -82,6 +109,457 @@ private static Object js_constructor(
...
+            }
+            return index;
+        }
+
+        private static Result fromBase64(String string, String alphabet, String lastChunkHandling)
+                throws IOException {
+            return fromBase64(string, alphabet, lastChunkHandling, NativeNumber.MAX_SAFE_INTEGER);
+        }
+
+        private static Result fromBase64(
```
aardvark179: Is there a reason you aren't using the base64 decoder in the standard library? I believe Android supports it too.

rhino/src/main/resources/org/mozilla/javascript/resources/Messages.properties
```diff
@@ -1137,7 +1137,10 @@ msg.bad.lastchunkhandling =\
   The "lastChunkHandling" option must be either "loose", "strict", or "stop-before-partial"
 
 msg.invalid.base64 =\
-  Invalid base64
+  String is not valid base64
+
+msg.invalid.hex =\
+  String is not valid hexadecimal
```
rbri: Why not having the wrong input as part of te error msg? What do Browsers do here?

tests/testsrc/test262.properties
```diff
@@ -2765,69 +2765,11 @@ built-ins/TypedArrayConstructors 140/738 (18.97%)
     internals/Set/tonumber-value-throws.js
     of/custom-ctor-returns-immutable-arraybuffer.js
 
-built-ins/Uint8Array 62/70 (88.57%)
-    fromBase64/alphabet.js
-    fromBase64/descriptor.js
+built-ins/Uint8Array 4/70 (5.71%)
```
rbri, line 2768: 😎

rhino/src/main/resources/org/mozilla/javascript/resources/Messages.properties
camnwalter: Oh yeah I'll definitely add that.

aardvark179 (commented):
These changes look good, but I'm not sure why we're writing our own base64 and hex encoders and decoders when the JDK already includes them.
https://docs.oracle.com/en/java/javase/25/docs/api/java.base/java/util/HexFormat.html
https://docs.oracle.com/en/java/javase/25/docs/api/java.base/java/util/Base64.html

camnwalter:
I looked into HexFormat, and from how I was using it, if it reaches an illegal character it throws an exception (looking at `HexFormat.of().parseHex()`), whereas the spec wants to report back the partial progress it made. I didn't know if it's worth trying to catch an exception each iteration of the loop instead of manually checking if it's a valid character.

For the Base64 case, with how there are different flags for last chunk handling, I didn't even try to see if it would work well to map 1:1 with the spec. I could look more into that.

gbrail:
I will check this out soon. I of course prefer to use what's built in to the JDK but I am not surprised at all if it turns out that the assumptions of JS and Java are different for these libraries. We already encountered this for things like dates, iterators on Map and Set, sorting behavior, and other things.

gbrail:
Thanks -- this is a good addition, although I'd like to consider a few things.

First, since this PR includes its own base64 decoder (and it has to because the Java implementation and ECMAscript differ), I would really prefer if that were broken out into a separate class with its own unit tests. Same goes for hex encoding and decoding, although that is a lot simpler. We are relying entirely on the test262 tests to verify whether encoding and decoding is correct and I have no idea how comprehensive those are (or whether they just test conformance to the many arcane rules of the spec). Humans still read and modify the code on this project and probably will for another few weeks at least ;-)

Second, I (and my AI friends) see a lot of opportunities for better performance, which I believe that moving those bits to a separate class would help make more clear. For example, it's entirely possible to calculate the output length of hex or base64 encoding or decoding, making it unnecessary to use a self-growing output stream. High-performance encoders and decoders in both of these cases in many languages tend to use lookup tables to assign values, and things like that, and it seems like a faster implementation should be possible. There's lots of open source to learn from in this area and your favorite AI model probably learned it already.

Thanks!

camnwalter:
Sounds good! Will do.

PR #2484, opened 2026-09-09
ignore me

gausie:
Claude made this PR while I was diagnosing a Rhino dependent without informing me


gbrail:
If Claude finds a bug and you understand what it did (or what it fixes) we're happy to see it anyway as long as we can talk to you (and not just Claude) about it!

gausie:
I was just looking into when we fall back to interpreter mode instead of compilation/codegen. I am happy to submit work from claude i just would never do so without extensive human review from me!

aardvark179:
This is a genuine bug but also hard to truly fix. You can split the methods, but ultimately hit constant pool limits, and you can also hit those with dynamic constants, but you can arrange your resolvers to be sparing in their constant pool usage, and ultimately you can start splitting into multiple classes.

I'm in the process of moving regexps to dynamic constants which will solve half of this problem, but I need to fix a problem in `jsc`. Sorry, I have been failing to finish that so that #2461 can be merged.

Template literals are a little harder to represent as dynamic constants, and I think have a bug involving multiple contexts, but I think those problems can be resolved, and then we can get rid of the init methods entirely.

rbri:
@aardvark179 sounds promising, great

gausie:
Well I am glad to know that my accidental PR has triggered some good conversation. Sounds like you have the right approach, @aardvark179.

PR #2485, opened 2026-09-11
Make console output consistent when JLine is used

gbrail:
* Make sure that we flush so that output appears when it is redirected
* Make sure that we write to stderr the way we did before JLine


gbrail:
I think that this is uncontroversial and will merge it soon...

PR #2486, opened 2026-09-11
Use fixed paralellism and remove monitor contention on trackers.

aardvark179:
Adding the JFR event support in #2473 helped identify some areas of lock contention during test execution, and strong indications that the number of threads picked by JUnit was actually slowing things down. On my laptop it appears that anything greater than four threads causes an increase in the time required to run the suite and a reduction in thread utilisation observed in JFR events. 

This change configures JUnit to use a fixed pool of four threads, and refactors the test trackers to remove one of the major points of thread contention. We still see contention in the harness script cache, but trying to populate that more eagerly shows we have some tests that are failing because we cannot load all the harness scripts, so I'll leave that one for now.

@gbrail and @rbri, it would be interesting to see if you also see performance differences with test262 and this change, or if we should choose the configuration based on more information.

### DIFF
--- a/tests/src/test/java/org/mozilla/javascript/tests/Test262SuiteTest.java
+++ b/tests/src/test/java/org/mozilla/javascript/tests/Test262SuiteTest.java
@@ -31,10 +31,12 @@ import java.util.Map.Entry;
 import java.util.Scanner;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.Consumer;
 import java.util.function.Function;
 import java.util.regex.Matcher;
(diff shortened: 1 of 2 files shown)


gbrail:
At home I mostly build with Bazel now and I have some spare Linux boxes to distribute the build so I don't often run the whole test suite at once. But running "./gradlew check --rerun" with different settings for fixed parallelism didn't really help -- in fact, the whole thing took about 15:40 with this change, and 14:51 with the change to junit-platform.properties commented out, so if anything I think it made it a bit slower.

Not sure what to suggest on the configuration, I'd be inclined to leave it alone, but I think that the changes to reduce contention in Test262SuiteTest make sense.

aardvark179:
Okay, I might just refactor the junior settings so they are easy to override for individuals.

PR #2487, opened 2026-09-11
Report columns in stack traces

gausie:
Stack traces currently carry a file and a line. This adds the column, and makes both backends agree on where a failure is attributed. I have tested this against some very large compiled scripts and while there are some more steps to go for full coverage (for chains of sourcemaps or for indexed maps), this branch stands on its own to get us there.

## What changes

`ScriptStackElement` gains a `columnNumber`, `Error` objects gain a `columnNumber` property, and `CallSite.getColumnNumber()` stops returning `undefined`. V8-style rendering prints a real `file:line:column` instead of a hardcoded `:0`.

Positions are recorded at each property read, element read and call as well as at each statement, so a failure is blamed on the operation rather than on the start of its statement: for `foo.bar.baz()` the column is that of `baz`, which is what V8 reports for the same code. AST nodes keep their existing convention of being positioned at the start of the whole expression; error reporting takes the position of the part being read instead.

Under a source map a frame names the original file, line and column, including for a bundle that maps back to several original files, where each frame can name a different one.

## How

**Interpreter.** Source positions move out of the icode into a table beside it, keyed by the bytecode offset they take effect at (`PositionTable`), so nothing is spent maintaining them as the code runs; only error reporting and the debugger read them. The operands carried by the `LINE`, `THROW`, `CALLSPECIAL`, `YIELD` and generator icodes are gone. `LINE` itself stays, now without an operand, so the debugger can still step by line. The table is a varint delta stream, the form V8's source position table and HotSpot's line number table take: about three and a half bytes per position on a real bundle, where fixed words would cost eight and outweigh the icode itself.

**JVM bytecode backend.** A compiled frame exposes one datum that varies from position to position: the `line_number` of a `LineNumberTable` entry, sixteen bits per JVMS 4.7.12. A position whose line is not yet spoken for emits its real line, which is what a Java debugger expects; a second position on a claimed line emits a marker from the top of the range instead, which `CompiledPositions` maps back. Markers are handed out per method, since a lookup is keyed by method as well. Classes written out ahead of time by `ClassCompiler` are loaded without the step that attaches that table, so they emit real line numbers only and report no column where a line holds several positions.

`ClassFileWriter.addLineNumberEntry(short)` is deprecated in favour of an `int` overload: the field is unsigned, so a signed short could not express half of it, and a line number above `Short.MAX_VALUE` sign-extended into the `start_pc` half of the packed entry.

## Compatibility

The existing `ScriptStackElement` and `JavaScriptException` constructors are kept, and `Evaluator.getSourcePosition` has a default that reports the line only, so an evaluator that does not track columns needs no change.

`docs/source-maps.md` gains a section describing both schemes.


aardvark179:
It is awesome to see this progress, but would it be possible to present it in a more staged fashion which would make it a little easier to review? The fact your adding a position to the frame objects makes me slightly worried that you're going to be updating it a lot, and this is the sort of thing it's good to avoid for performance.

Maybe something like:
1. Moving the interpreter to line maps instead of the line instruction.
2. Making the changes to the byte code compiler
3. Making the changes to how stack traces are generated?

We're also part way through landing interpreter v2, and doing this at the same time is going to make that even more complicated.

In general for a complex change like this I recommend breaking things down into changes of at most +500/-500 lines of change that each make one change.

I'm also very nervous of messing with the line number map in compiled code. Not all exceptions originate in JS code, and you won't always be able to rewrite the stack trace the user sees.

gausie:
Yeah I wanted to stage it more but this was the first gate I could get to where I can really be sure it works and introduce a meaningful usable change. I can definitely work to those steps though, it’s a good idea.

I will test against non JS errors - good point.

-------- Original Message --------

gausie:
> The fact your adding a position to the frame objects makes me slightly worried that you're going to be updating it a lot, and this is the sort of thing it's good to avoid for performance.

Wait, I'm not am I? It previously did that (sort of) and I now track it in a side table. In any case, moving ahead with your ideas, I'm going to open two initial smaller PRs today and will sort the rest another time.

aardvark179:
> Wait, I'm not am I? It previously did that (sort of) and I now track it in a side table. In any case, moving ahead with your ideas, I'm going to open two initial smaller PRs today and will sort the rest another time.

I could have sworn I saw a position field ended to `ACallFrame` but apparently the evening jettag was really kicking in. 🫢

PR #2488, opened 2026-09-11
Write a line number to the whole field a class file gives it

gausie:
A `LineNumberTable` entry's `line_number` is an unsigned sixteen-bit value, per JVMS 4.7.12. `ClassFileWriter` packed the entry by addition and took the line as a signed `short`, so a line above `Short.MAX_VALUE` sign-extended and borrowed into the `start_pc` half of the packed word:

```java
itsLineNumberTable[N] = (itsCodeBufferTop << 16) + lineNumber;
```

What comes out is an entry the JVM rejects outright, with `ClassFormatError: Invalid pc in LineNumberTable`.

Mask instead of adding, and take the line as an `int`. `addLineNumberEntry(short)` stays, deprecated, since `ClassFileWriter` is used outside Rhino; nothing in the tree calls it any more.

Reachable today for a script compiled past line 32767, since `BodyCodegen` casts to `short` on the way in. The test fails without the fix.

Split out of #2487, where it was groundwork. It stands on its own and depends on nothing else in that branch.


hamzahamidi (approved):
The unsigned 16-bit masking matches the LineNumberTable representation, and the regression test covers the prior sign-extension failure.

hamzahamidi (commented):
No actionable issue found. The unsigned 16-bit masking and the high-line regression test correctly address the sign-extension failure.

gbrail:
Yes, this is fine, thanks!

PR #2489, opened 2026-09-11
Keep source positions as a stream of deltas

gausie:
Groundwork split out of #2487. Nothing calls it yet; the interpreter is next.

Both backends want the same thing: a source position looked up by a key that never decreases -- the bytecode offset a position takes effect at in the interpreter, or the line number a compiled frame reports -- and read only when an error is reported or a debugger asks.

Stored as fixed words such a table outweighs the code it describes, so `PositionTable` stores it the way V8's source position table and HotSpot's own line number table do: as a stream of variable-length deltas, about three and a half bytes per position on a real bundle where fixed words would cost eight. A table big enough to be worth it builds an index the first time it is read, since the few that are read are read many times over; a smaller one is walked instead and allocates nothing. The index is packed too: one record per entry in a single byte array, each column an offset from its own smallest value, a byte wide wherever the span allows.

An entry records whether it is a statement's, because only those mark lines for the debugger.

Separated from the interpreter change that follows so that change reads as a straight substitution rather than a new data structure and its adoption at once.


hamzahamidi (changes requested):
When an expression replaces a statement at the same key, `statements[at]` preserves the statement flag but `lines[at]` is overwritten with the expression line. If they are on different lines, `lines()` reports the expression line rather than the statement line. The test only uses the same line for both positions. Please preserve the statement line separately, or document and enforce a same-key, same-line invariant.

rhino/src/main/java/org/mozilla/javascript/PositionTable.java
```diff
@@ -0,0 +1,227 @@
...
+
+        /** Appends an entry, or replaces the last one when it has the same key. */
+        void add(int key, int line, int column, int source, boolean statement) {
+            boolean replacing = size > 0 && keys[size - 1] == key;
+            int at = replacing ? size - 1 : size++;
+            if (at == keys.length) {
+                keys = Arrays.copyOf(keys, size * 2);
+                lines = Arrays.copyOf(lines, size * 2);
+                columns = Arrays.copyOf(columns, size * 2);
+                sources = Arrays.copyOf(sources, size * 2);
```
hamzahamidi: When this replaces a statement entry with an expression from another line, `statements[at]` remains true but this line becomes the expression line. `lines()` then marks the wrong debugger line. Preserve the original statement line separately, or enforce that both entries are on the same line.

rhino/src/main/java/org/mozilla/javascript/PositionTable.java
```diff
@@ -0,0 +1,227 @@
...
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import org.mozilla.javascript.sourcemap.Position;
+
+/**
+ * Source positions keyed by an integer that never decreases: the bytecode offset a position takes
```
aardvark179: This description feels like it could be clearer. It took several re-readings for me to understand that it is only the bytecode index that is guaranteed to increase.

Maybe something like

> Maps bytecodes to source positions. The default format for the map is as a list of variable length deltas encoded as a byte array. Each entry consisting of an unsigned delta for bytecode position followed by…

Or something like that.

rhino/src/main/java/org/mozilla/javascript/PositionTable.java
```diff
@@ -0,0 +1,227 @@
...
+            for (Cursor c = new Cursor(); c.advance(); ) {
+                e.add(c.key, c.line, c.column, c.source, c.statement);
+            }
+            entries = e;
+        }
+        return e;
+    }
+
+    /** Entries as parallel arrays sorted by key: the form they are built in and searched in. */
+    private static final class Entries {
```
aardvark179, line 212: The minimum size of a decoded map with this structure is about 400 bytes (40 bytes for the basic object, 80 bytes for each 16 entry int array, and 32 bytes for the Boolean array). In production I would expect a lot of these maps to be instantiated for various reasons so this would end up adding a lot to our memory footprint.

I think it would be worth optimising for small maps and scale up as needed. Those small maps can like avoid the arrays entirely.

aardvark179 (changes requested):
Only a partial review, I haven't taken a really detailed look at structures for different script compilations.

I see what you're doing, but I think this is going to need more work. I'd work under the assumption that every source map will be instantiated eventually and try to ensure that these maps are as light weight as we can make them.

There are a couple of things you might be able to exploit to achieve that. In most cases files with significant numbers of lines will not have long lines, and minified files (where column numbers become more important) often only have a single line. Also many functions will only be small even though they may start on a high line or column numbers. Representing things as an offset to a starting position might be useful in a lot of cases.

gausie:
Thanks both for the feedback. I will apply most of it now.

I see this issue with statements vs lines. I'll write a test and probably just make it so a statement keeps the key, and an expression with the same one is dropped.

> Representing things as an offset to a starting position might be useful in a lot of cases.

The stream is already delta-coded so every entry after the first is an offset from its predecessor!

aardvark179:
> The stream is already delta-coded so every entry after the first is an offset from its predecessor!

The byte stream is already delta encoded, but the decoded entries aren't. Most functions are small enough that you could use byte arrays for the lines and columns if they are offsets from a starting line and column.

Issue #798, opened 2020-12-08
NullPointerException when  exec const in ScriptEngine


zfyq124:
exec const in  ScriptEngine, a NullPointerException throw

#### version
1.7.13

#### command
no

#### testcase
```java 
    public void testRhinoExec() throws ScriptException {
        ScriptEngineManager manager = new ScriptEngineManager();
        ScriptEngine scriptEngine = manager.getEngineByName("rhino");
        Bindings bindings = scriptEngine.createBindings();
        CompiledScript compiledScript = ((Compilable) scriptEngine).compile("const a = 10;");
        Object o = compiledScript.eval(bindings);
        System.out.println(o);
    }
```
#### output

```java
java.lang.NullPointerException
	at com.perfma.xocean.jmeter.JMeterRunner.testRhinoExec(JMeterRunner.java:50)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:56)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)

```

#### expected output
```java
exec success
```

zfyq124:
the reason is for const variable set call org.mozilla.javascript.ScriptableObject#putConst, but get call org.mozilla.javascript.engine.BindingsObject#get can not get the set value

p-bakker:
Could you provide the stacktrace that includes the trace writing the Rhino codebase?

tonygermano:
@p-bakker is this what you're looking for?
It doesn't throw the exception when the script is `var a = 10`, but returns undefined instead. I ran this against the latest snapshot, but haven't looked at the source yet.

```
Caused by: java.lang.NullPointerException
	at org.mozilla.javascript.ScriptableObject.getBase(ScriptableObject.java:2342)
[...]

PR #2490, opened 2026-09-12
Fix Java object conversion in ScriptEngine bindings

Hanabi9248:
After `engine.put("file", file)` and `engine.eval("var copy = file")`, `engine.get("copy")` returns a `NativeJavaObject` rather than the original Java object. `BindingsObject` currently applies the conversions in the opposite directions.

Convert Java values to JavaScript on reads and unwrap JavaScript values on writes. The five regression cases cover direct and compiled evaluation in both execution modes, plus global bindings. They fail on the original implementation; all 39 engine tests pass with the fix.

The existing [Rhino CI workflow](https://github.com/Hanabi9248/rhino/actions/runs/34712331805) passed all five jobs on this branch: Java 21, Java 25, debug, multithreaded, and Test262 verification.

The reversed conversions were also noted in #798; its separate const-scope issue remains outside this change. Please squash the two commits when merging.


gbrail:
Agreed -- I guess this can't be used all that much...

The first one is clear -- get should do the conversion in the other direction.

However, put is triggered when the JavaScript code sets a global that came from the context. But by calling jsToJava with Object.class as the target type, we end up with a converted Java object, which we might not always want. For example, imagine if we wanted to return the original JavaScript object, such as Undefined? This code would actually convert it to something else. I find the Java specs to be unhelpful on this, but do we want a conversion, or perhaps we want a different conversion?

Hanabi9248:
You’re right: `jsToJava(value, Object.class)` is too broad here. With `var missing = undefined`, the preceding revision stores the string `"undefined"`, so reading the binding back changes its JavaScript semantics.

I changed `put` to unwrap only `Wrapper` values and leave everything else unchanged. This keeps the original Java object for `var copy = file` without coercing native JavaScript values. Four added cases cover direct/compiled evaluation in both execution modes, checking `Undefined` identity and subsequent use of a date, object, array and function. All four fail on the preceding revision; all 43 engine tests pass with this change.

The updated head is `39d05f2`. The [full Linux workflow](https://github.com/Hanabi9248/rhino/actions/runs/35861716196) is running; its result is still pending.

PR #2492, opened 2026-09-16
fix: an accessor property's value is not constant

### LINKED ISSUE #1834: Regression: a property getter inside a function closure gets called only once
> [!NOTE]
> This works correctly in 1.7.5 (see below).

When I define a getter on a global scoped object using `Object.defineProperty`, and then access the property inside of a function using dot notation, the getter is accessed only once.

The same **does not** happen if the property is accessed using index notation (see demo).

The same **does not** happen if the property is defined beforehand in the object literal (as value or getter).

The same happens regardless of whether the function is a named function or an arrow function.

### Example code
```js

let _currentCounter = 0;

const counter = {
  // count: undefined // if this line is enabled, it works as expected.
};

Object.defineProperty(counter, "count", {
  get() {
    print("Reading value");
    return ++_currentCounter;
  },
});

function dotNamed() {
  return counter.count;
}

const dotArrow = () => counter.count;

function index() {
  return counter["count"];
}

print("Direct access via dot notation:")
print(counter.count);
print(counter.count);
print(counter.count);

print("");
print("Access in named function closure via dot notation:")
print(dotNamed());
print(dotNamed());
print(dotNamed());

print("");
print("Access in arrow function closure via dot notation:")
print(dotArrow());
print(dotArrow());
print(dotArrow());

print("");
print("Access in function closure via index notation:")
print(index());
print(index());
print(index());
```

rhino-1.7.5.jar works correctly (and identical to node)
```
Direct access via dot notation:
Reading value
1
Reading value
2
Reading value
3

Access in named function closure via dot notation:
Reading value
4
Reading value
5
Reading value
6

Access in arrow function closure via dot notation:
Reading value
7
Reading value
8
Reading value
9

Access in function closure via index notation:
Reading value
10
Reading value
11
Reading value
12
```

Output with rhino-all-1.8.1-20250215.122236-7.jar
```
Direct access via dot notation:
Reading value
1
Reading value
2
Reading value
3

Access in named function closure via dot notation:
Reading value
4
4
4

Access in arrow function closure via dot notation:
Reading value
5
5
5

Access in function closure via index notation:
Reading value
6
Reading value
7
Reading value
8
```

### Versions tested

- 1.7.15 **(not affected)**
- 1.7.16-20241221.122243-17
- 1.8.0
- 1.8.1-20250215.122236-7

midgleyc:
Fixes #1834.

Uses the code from that as a regression test.

The const aware linker caches properties which looked constant. Unfortunately, this included accessor getters, which aren't necessarily constant. Omit these.

Not sure about the test placement or name, let me know if you want them moved or changed.


aardvark179:
This looks good to me as a change, but also looks bad in that we're missing a bunch of stuff in this linker.

Not using this linker doesn't break anything, so when we split scopes off from objects we didn't see any failures, but we also didn't expand this linker to handle `ScopeObject`s as well as `ScriptableObject`s. Testing that this linker is working, _and that it is being used_, feels like it is something we'd really need to do from Java and at a somewhat lower level. I think what you'd need is
1. Some custom slot types for testing which record their accesses.
2. Some test utilities to use those custom slot types in scopes and objects.
3. Compiled test scripts / functions which can be run either with our custom scopes as parents or passing in our custom objects.

I was also worried there might be more edge cases (such as the `length` property of an array), but they generally seem to obey all the rules of not changing value if they are marked as read only. I think there may be an issue with things like `NativeProxy` which overrides `has` but not `getAttributes` so could throw an unexpected error at runtime. Maybe we should change the code to just do the `has` get a slot if it exists, and then base the remaining decisions on that slot.

gbrail:
Yes, I agree that there is a bunch more to do including more tests. This PR fixes a real bug, though, so I think we should take it and work on more.

aardvark179:
Fine with me.

PR #2493, opened 2026-09-17
Move LiveConnect and other reflection support to a separate module

gbrail:
This moves all of the functionality for invoking Java code from JavaScript
directly, including the top-level "packages", "import", "Java", and other
packages, as well as the ability to construct a JavaScript class from an
arbitrary Java object using reflection, from the main Rhino module to a
new module called "rhino-reflect".

Without "rhino-reflect" loaded, These top-level packages are not available
and direct access to Java object methods and properties is not allowed,
including the Rhino-specific "javaException" attribute on the Error object.

If "rhino-reflect" is in the classpath, then all that functionality is
automatically loaded and Rhino behaves as it has always behaved.

For applications that use Rhino to run untrusted code, this eliminates the
possibility that malicious code could use Java reflection to access unexposed
parts of the Java API and thus cause additional risks.


aardvark179:
This seems to be missing `FunctionObject` and `MemberBox`, but that might mean we also need to extract some stuff around `AccessorSlot.Getter/Setter`  but I wonder if we should be trying to do something like #2454 first to see if we can avoid needing to leak those details outside of the main rhino module.

gbrail:
No, I don't think FunctionObject needs to move, actually, nor does the code in ScriptableObject (I think) that constructs a class out of FunctionObject, or the code that looks for annotations and does the same.

What we're trying to move out here is the built-in functionality that lets any arbitrary Java object be automatically transformed into a JavaScript object with all its methods and properties callable from JavaScript. That makes it too easy to access unintended internal Java code and classes by accessing the classloader and other powerful things.

FunctionObject and its friends, OTOH, only expose, via reflection, the Java methods and properties that the embedder implicitly decided were OK to expose to the script. IMO the fact that reflection isn't used is not the problem, but the automatic wiring of lots of Java stuff to JavaScript.

If it helps, we can pick another name for "rhino-reflect," like "rhino-liveconnect". I just wasn't sure that anyone knows what "liveconnect" means any more!

rhino/src/main/java/org/mozilla/javascript/WrapProcessor.java
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.javascript;
+
+import org.mozilla.javascript.lc.type.TypeInfo;
+
+public interface WrapProcessor {
```
ZZZank: Can we merge this into `LiveConnect`? or this is separated to allow 3rd party implementations?

rhino/src/main/java/org/mozilla/javascript/WrapProcessor.java
```diff
@@ -0,0 +1,24 @@
...
+    Object wrap(
+            Context cx, VarScope scope, Object obj, Class<?> staticType, boolean wrapPrimitives);
```
ZZZank: I think this `.wrap()` method, which basically duplicates the one with `TypeInfo` param, can be eliminated. `WrapFactory` can do the conversion itself.

rhino/src/main/java/org/mozilla/javascript/WrapFactory.java
```diff
@@ -25,11 +21,13 @@
...
-     */
-    public final Object wrap(Context cx, VarScope scope, Object obj, Class<?> staticType) {
-        return wrap(cx, scope, obj, TypeInfoFactory.GLOBAL.create(staticType));
+    private final WrapProcessor processor;
+
+    private boolean javaPrimitiveWrap = true;
+
+    public WrapFactory() {
+        // Set the processor to null if reflection is not available
+        processor = LiveConnectSupport.get().getWrapProcessor();
```
ZZZank: This is a bit confusing, we can make the `processor` field static if it should only be loaded by Rhino, or make it a constructor param with default value if 3rd party implementation is allowed.

ZZZank (changes requested):
I think the name `rhino-reflect` is good enough. LiveConnect is not that famous now, and Rhino implementation does not strictly follows LiveConnect anyway (e.g. conversion weight comparison when choosing method overload)

PR #2495, opened 2026-09-19
Arithmetic and bitwise instructions.

aardvark179:
I'm grouping these two commits together partly because they feel fairly closely related in that they all operator instructions, and because we've already gone round various review aspects with the first two commits so I'm hoping we can accelerate landing these a little.

I got my robot friend to apply all the review comments from the first two PRs to these instructions.


rhino/src/main/java/org/mozilla/javascript/interpreterv2/instruction/StringConcat.java
```diff
@@ -0,0 +1,37 @@
...
+    private final Operand rhs;
+
+    public StringConcat(Operand lhs, Operand rhs) {
+        this.lhs = lhs;
+        this.rhs = rhs;
+    }
+
+    @Override
+    public void interpret(Context cx, CallFrameV2 frame) {
+        Object rObj = rhs.retrieve(cx, frame);
```
gbrail: This doesn't check isDouble so if that optimization is in place then DOUBLE_MARK will appear instead of the numeric value.

rhino/src/main/java/org/mozilla/javascript/interpreterv2/instruction/Add.java
```diff
@@ -0,0 +1,161 @@
...
+                throw ScriptRuntime.typeErrorById("msg.cant.convert.to.number", "BigInt");
+            } else {
+                frame.push(lNum.doubleValue() + resultDouble);
+            }
+        }
+    }
+
+    @Override
+    public Instruction simplify(InstructionSimplification simplifier) {
+        KnownType rhsType = rhs.getKnownType(simplifier);
```
gbrail, line 132: AI claims that a ClassCastException can be caused by this bit of code with an expression like:

    ("a" - 2) + 3

rhino/src/main/java/org/mozilla/javascript/interpreterv2/instruction/UnsignedRightShift.java
```diff
@@ -0,0 +1,61 @@
...
+        }
+
+        if (shouldCoerceLeft) {
+            lValue = ScriptRuntime.toNumber(lObj);
+        }
+        if (shouldCoerceRight) {
+            rValue = ScriptRuntime.toInt32(rObj);
+        }
+
+        long value = ScriptRuntime.toUint32(lValue) >>> ((int) rValue & 0x1F);
```
gbrail: AI comment makes sense here so I'll paste it:

 UnsignedRightShift wrong shift count for large doubles:

UnsignedRightShift.java:46
Uses a raw (int) rValue & 0x1F instead of ScriptRuntime.toInt32(rValue). Java's narrowing cast saturates instead of wrapping mod 2³², so: 1 >>> 2**40 → 0 (v1 and spec: 1); 1 >>> Infinity → 0 (v1 and spec: 1). v1 uses stack_int32 (Interpreter.java:1877) which goes through toInt32. The non-double path (line 43) is correct; only the double path is.

gbrail (commented):
Happy to move this through more quickly. There do seem to be some inconsistencies though -- let me know how you want to address these until we get test coverage up for this feature.

rhino/src/main/java/org/mozilla/javascript/interpreterv2/instruction/StringConcat.java
aardvark179: Good catch. It required a little jumping through hoops to get a test case that wasn't reduced to a constant earlier., But `function f(x) { let y = typeof x == 'string' ? 'bob' : 0.5; return `hello ${y}!`; }` triggered this.

rhino/src/main/java/org/mozilla/javascript/interpreterv2/instruction/Add.java
aardvark179: Just tested this (including added debug output to check stuff is getting triggered) and it works fine. I'll generate a bunch of unit tests round this known type stuff and simplification that we can push later once the compiler and everything is in place.

rhino/src/main/java/org/mozilla/javascript/interpreterv2/instruction/UnsignedRightShift.java
aardvark179: Oh, yeah. I see the potential problem there. It _seems_ to work, but it also feels like it should be clear it works.

gbrail:
This looks good, I'll merge it soon, thanks!

PR #2496, opened 2026-09-19
Fix backslash handling in jline.

aardvark179:
Hilarious as it is to see Claude go into a complete tail spin when `print('Line 1.\n; Line 2.');` outputs `Line 1.n; Line 2.` it's not really helpful for users of the REPL.


gbrail:
Thanks!

PR #2498, opened 2026-09-20
FIx constant pool construction and throw correct exception on overflow.

aardvark179:
@gbrail this should deal with the problems you saw related to #2461  and I was seeing on my const and loop fix branch. I've put it as a separate PR because I don't know what order you want to make these changes in.


rhino/src/main/java/org/mozilla/classfile/ConstantPool.java
```diff
@@ -348,6 +353,9 @@ Object getConstantData(int index) {
     }
 
     void setConstantData(int index, Object data) {
+        if ((index & 0xffff) != index) {
```
gbrail: My AI friend's is convinced that 65535 is not a valid index even though it would pass this test because of the way that we often compute or store itsTopIndex + 1.  (And FWIW we seem to +2 it a lot as well.) That would imply a different check here. A unit test to verify the duplication or non-duplication here would probably answer for sure.

rhino/src/main/java/org/mozilla/classfile/ConstantPool.java
```diff
@@ -281,23 +281,28 @@ short addMethodRef(String className, String methodName, String methodType) {
...
+            short ntIndex = addNameAndType(methodName, methodType);
+            short classIndex = addClass(className);
+            ensure(5);
+            itsPool[itsTop++] = CONSTANT_InterfaceMethodref;
+            itsTop = ClassFileWriter.putInt16(classIndex, itsPool, itsTop);
+            itsTop = ClassFileWriter.putInt16(ntIndex, itsPool, itsTop);
+            setConstantData(itsTopIndex, r);
+            itsPoolTypes.put(itsTopIndex, CONSTANT_InterfaceMethodref);
+            index = itsTopIndex++;
+        }
```
gbrail, line 331: Is the goal here to put the calculated value of "index" back in the "itsConstantHash" table? It doesn't appear that we do.

rhino/src/main/java/org/mozilla/classfile/ConstantPool.java
gbrail: It's also a fair point that we make this overflow check in setConstantData but not in other places that allocate handles, including addConstant and others.

gbrail (commented):
I'm happy to bring this in before the larger fix, but there are opportunities to make this change a bit more robust if we can.

aardvark179:
Okay, I got my robot friend to generate a bunch of unit tests and then do  a more extensive rewrite. I moved all the indices to be `ints` to avoid the types of conversion errors we have previously seen. I think there is much more comprehensive refactoring and API change lurking in there which might help significantly improve class file compilation speed (profiling suggests we spend a large amount of the time repeatedly building and finding things like `TypeInfo` structures), but that's not a change to do right now.

aardvark179:
I've added some simple class splitting to handle large scripts and emit a size exception if we emit too many `invokeDynamic` instructions in a single class (which seems to be the cause of the `three.js` problems. It works on the spot level of scripts, and cannot split large single nodes, or individual large methods, so compilation can still fall back to the interpreter for some scripts. `three.js` now successfully compiles to class files in the benchmarks.

